PR #2, opened 2014-04-02
Create parent dirs for useradd -m

thorstenb:
Code from pwdutils 3.2.2 with slight adaptations, under BSD-3-clause project license of course. Fixes the 'XXX - create missing parent directories.  --marekm' problem.

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -1894,24 +1894,54 @@ static void usr_update (void)
 static void create_home (void)
 {
 	if (access (user_home, F_OK) != 0) {
+		char path[strlen (user_home) + 2];
+		char *bhome, *cp;
+
+		path[0] = '\0';
+		bhome = strdup (user_home);
+		++bhome;
+
 #ifdef WITH_SELINUX
 		if (set_selinux_file_context (user_home) != 0) {
 			fail_exit (E_HOMEDIR);
 		}
 #endif
-		/* XXX - create missing parent directories.  --marekm */
-		if (mkdir (user_home, 0) != 0) {
-			fprintf (stderr,
-			         _("%s: cannot create directory %s\n"),
-			         Prog, user_home);
+
+		/* Check for every part of the path, if the directory
+		   exists. If not, create it with permissions 755 and
+		   owner root:root.
+		 */
+		cp = strtok (bhome, "/");
+		while (cp) {
+			strcat (path, "/");
+			strcat (path, cp);
+			if (access (path, F_OK) != 0) {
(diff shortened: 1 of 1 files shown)


thorstenb:
A polite ping - is this PR still of interest?

src/useradd.c
```diff
@@ -1894,24 +1894,54 @@ static void usr_update (void)
 static void create_home (void)
 {
 	if (access (user_home, F_OK) != 0) {
+		char path[strlen (user_home) + 2];
+		char *bhome, *cp;
+
+		path[0] = '\0';
+		bhome = strdup (user_home);
+		++bhome;
```
hallyn: Need to check for strdup failure before incrementing here.

src/useradd.c
```diff
@@ -1894,24 +1894,54 @@ static void usr_update (void)
...
 #endif
-			fail_exit (E_HOMEDIR);
+					fail_exit (E_HOMEDIR);
+				}
+				if (chown (path, 0, 0) < 0) {
+					fprintf (stderr,
+					         _("%s: warning: chown on `%s' failed: %m\n"),
+					         Prog, path);
+				}
+				if (chmod (path, 0777) < 0) {
```
hallyn: You're making the directories root owned and 777?

hallyn:
Thanks for doing this patch, sorry for the delay.  Indeed it seems worth doing, although the alternative (for which there is an open debian bug) is to simply drop this functionality from shadow and have everyone switch to other implemenetations.

hallyn:
Hi - one last ping;  If no reply (or if you ask me to) I'll update the patch myself and apply.

Thanks.

src/useradd.c
```diff
@@ -1894,24 +1894,54 @@ static void usr_update (void)
...
 	if (access (user_home, F_OK) != 0) {
+		char path[strlen (user_home) + 2];
+		char *bhome, *cp;
+
+		path[0] = '\0';
+		bhome = strdup (user_home);
+		++bhome;
+
 #ifdef WITH_SELINUX
 		if (set_selinux_file_context (user_home) != 0) {
```
hallyn: One more change - this set_selinux_file_context() must happen after the parent directories have been created, else it will apply to the parent directories we create.

thorstenb: Ah yes please, I've somewhat moved on from that code. Given my response latency, perhaps best if you roll your proposed changes in.

jubalh:
What's the current state of this PR?

Is there still interest in adding this feature?

PR #3, opened 2015-02-14
man: useradd(8): fix typo in German translation

vapier:
When referring to USERGROUPS_ENAB, the German mentions /etc/default/useradd
when it should be /etc/login.defs (like the original English does).

Reported-by: Stefan Kiesler <heavymetal@gmx.de>
Signed-off-by: Mike Frysinger <vapier@gentoo.org>


hallyn:
Thank you.

PR #4, opened 2015-03-20
Make shadow more robust in hostile environments

xnox:
I'm patching shadow to be more robust when operating in a-typical environments, but these improvements are general enough, that I believe warrant inclusion upstream by default.

Specifically there are deployments that use nss-altfiles / nss-extrausers and thus ship alternative group/passwd/shadow/gshadow files elsewhere on the filesystem (e.g. /var/lib/passwd). In such configurations admin modifiable files /etc/passwd, /etc/group and so on may not exist. Furthermore if one is bootstrapping a new distribution from scratch, it would be nice to point shadow utilities at an empty /etc and start creating default system accounts with useradd/usermod/groupadd/etc utilities without writing initial files by hand. Hence these changes:

* create all databases, with correct (?! used typical permissions here, please correct if different defaults are desired) permissions, even if they are missing / are empty.

* next /etc/shadow file existence is used as a flag file, whether or not shadow passwords should be used. I think this is very odd, hence I add FORCE_SHADOW option to make sure shadow/gshadow are used, even if those files are non-existent.

* during testing I have noticed that when shadow is compiled with PAM support, settings that are not-applicable any more, but present in the stock /etc/login.defs are being complained about. So I made a change for shadow to not complain about well-known settings, which are not in effect when compiled with PAM support. If this is undesirable, maybe instead we would want to pre-process login.defs at compile time to make sure they do not contain any unknown settings in a given configuration (with/without pam, with/without subuids, etc.)

* Lastly, I noticed that login command would bail out and exit, if /etc/login.defs file is not present on disk. I find that very odd - given that there are compiled in defaults for every single value for all shadow utils to operate normally. Thus I made it non fatal for login to operate without /etc/login.defs.

Overall my goal is to have fully usable system with empty /etc and with these initial patches this is achievable. At the moment I'm also working on adding full usermod support, when operating with nss-altfiles.

Please review and consider including these patches.

### DIFF
--- a/etc/login.defs
+++ b/etc/login.defs
@@ -393,3 +393,8 @@ USERGROUPS_ENAB yes
 #
 #CREATE_HOME     yes
 
+#
+# Force use shadow, even if shadow passwd & shadow group files are
+# missing.
+#
+#FORCE_SHADOW    yes
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -968,11 +968,10 @@ int commonio_close (struct commonio_db *db)
 	} else {
 		/*
 		 * Default permissions for new [g]shadow files.
-		 * (passwd and group always exist...)
 		 */
-		sb.st_mode = 0400;
-		sb.st_uid = 0;
-		sb.st_gid = 0;
+		sb.st_mode = db->st_mode;
+		sb.st_uid = db->st_uid;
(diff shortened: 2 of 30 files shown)


hallyn:
Sorry, I lost track of this, I thought you were discussing it with someone else.  Should we merge this at this point?

PR #6, opened 2015-07-11
Clear passwords on __gr_dup/__pw_dup errors and fix memory leak

stoeckmann:
The functions __gr_dup and __pw_dup do not explicitly zero the
memory which hold the passwords after free. The gr_free and pw_free
functions do this explicitly.

To guarantee same behaviour, it's possible to call these *_free
functions directly from __*_dup, because the memory is initialized
with zeros at the beginning. Calling free(NULL) has no negative
effect and can be considered safe these days.

Also prevent a memory leak in commonio on one of its error paths.


hallyn:
Thanks, this appears to be a nice cleanup and memleak fix.

PR #13, opened 2016-03-03
Add ability to clear or set lastlog record for user via lastlog command

t8m:
This functionality is useful because there is now a feature
of Linux-PAM's pam_lastlog module to block expired users (users
which did not login recently enough) from login. This commit
complements it so the sysadmin is able to unblock such expired user.

Signed-off-by: Tomáš Mráz <tmraz@fedoraproject.org>


hallyn:
Thanks, this looks good to me.

PR #21, opened 2016-05-16
Add --prefix argument

fariouche:
This is my proposal to add an alternative to the chroot option.
This --prefix option will prefix all files needed to create/modify/delete a group/user inside a directory, even if there is no valid rootfs inside. (/etc/{group,passwd,gshadow,shadow,logins.defs and subuid/gid)
The goal was to add a cross-compilation compatible option that is easy to use.
Currently, the --prefix option is available to user{add,mod,del} and group{add,mod,del}
The limitations are the followings:
- No selinux. if the -Z option is used with --prefix, you will get an error.
- No NIS/LDAP. For now NIS is explicitly disabled if --prefix is used
- PAM authenticate the user using the host credentials (this is the intended behavior as PAM is only used to verify that the user has the rights to add/delete a user/group in the system)

The implementation of this option touch quite a few files, but the majority of the modifications are in the libmisc/prefix_flag.c
This way, I kept the code path exactly the same as before if the --prefix option is not specified. Risks of regressions are minimal. At least I didn't encoutered any issue when testing.

I believe that NIS support is possible with minimal efforts, but I do not have any way to test so I prefer to not implement something blindly. If needed, the only needed file to modify will be most likely only prefix_flag.c

If you have any remarks, please let me know :)


fariouche:
Hi,
Just to let you know that I finished installing a full new gentoo cross compiled target system, and everything is OK.

hallyn:
Hi,

could you please do a rebase -i to squash commits like 'add one forgotten...'?  That'll make it easier to debug and in the future more bisectable.

I'll try to merge this on Friday.

glensc:
@hallyn you can enable "Squash Merge" in shadow project settings in GitHub. Then you can merge PR's like this into single squashed commit preserving commit author, and you can even modify commit message that GitHub proposes if you click the little arrow down instead big Merge button :)

https://github.com/shadow-maint/shadow/settings

@fariouche avoid making Pull Requests from `master` branch, as any change you push to your `master` gets added to this PR until the PR is still open. You may add mistakenly commits that you did not intend to include with PR.

hallyn:
Quoting Elan Ruusamäe (notifications@github.com):
> @hallyn you can enable "Squash Merge" in shadow project settings in GitHub. Then you can merge PR's like this into single squashed commit preserving commit author, and you can even modify commit message that GitHub proposes if you click the little arrow down instead big Merge button :)

I don't want to, I prefer to have the submitter squash in rebase - I assume
there should in fact be >1 commits, but not all the ones that are there (which
include merges from upstream)

thanks,
-serge

fariouche:
@glensc , I just discovered that last time when I pushed my modifications. I will create a specific branch next time.

@hallyn , indeed, some commits are just upstream alignments. I will try to squash some commits to makes things easier to read.

fariouche:
Ok, I think I've redone it properly (rebase instead of merge). I'm still new to git. I will push again shortly

hallyn:
On Thu, Dec 22, 2016 at 07:40:41AM -0800, fariouche wrote:
> Ok, I think I've redone it properly (rebase instead of merge). I'm still new to git. I will push again shortly

awesome, thanks.

hallyn:
Hm, this doesn't seem to work for me.  applied the patch, built, then ran './src/useradd --prefix ~/outout serge', but got:

useradd: Permission denied.
useradd: cannot lock /home/ubuntu/outout//etc/passwd; try again later.

while strace shows that in fact it's trying to lock /etc/passwd.lock:

19585 open("/etc/.pwd.lock", O_WRONLY|O_CREAT|O_CLOEXEC, 0600) = -1 EACCES (Permission denied)

Can't quite see why.

hallyn:
Is there any case where '-R' implemented with chroot() would be better than
-R implemented as --prefix wants to be?

vapier:
i cannot think of a case where `chroot()` would be better.  it disallows non-root usage even on paths that you own.  for example, if i'm building (as non-root) my own distro in $HOME/, i can't use the -R flag to set up passwords.  conversely, in the cases where you are running as root because you're operating on a path owned as root (like a backup distro on another drive), -R built on top of prefixed-paths instead of chroot would still function the same.

hallyn:
Right I couldn't think of one either.  If there is none, then I agree -R behavior
should be changed.  Thanks!

fariouche:
@hallyn , There are two lock mechanism, in commonio.c: one implemented by shadow (and is using the correct lock file) and one using lckpwdf from the libc (the default if the libc provides this function). THe later uses an hardcoded path.
I've re-implemented some of the libc functions like getpwd etc, and called them prefix_getpwd etc.
I can do the same for this function too, or we remove this code from shadow library (why use lckpwdf if already implemented in the shadow lib?)

Also, it seems like the lock file is opened first, and if failed, is checking the euid to see if we are root.
The passwd etc files should be only writable by root. So we'll need root access to modify the passwd file in the sysroot for example. What should we do?


About chroot, there are two differences not implemented by prefix option: selinux and auditing
I do not know if it does make sense to implement them here. And for selinux, I do not know if it is possible to use the selinux from the host (for the case of a sysroot from an other architecture)

fariouche:
Hi,

just a quick update of the patch so that it works with latets updates

fariouche:
Hello,

Any news on this subject? Do you need me to do some more work?

Thanks

hallyn:
Is being able to run as non-root actually the only advantage here?

It seems like we could get the same result with a far less invasive patch if we added a --map-root argument.  Uid 1000 could call useradd -R ~/mychroot --map-root $(id -u) joe.  useradd would then create a new user namespace mapping host 1000 to container uid 0, pivot or chroot to ~/mychroot, and act there.

fariouche:
except when the chroot is a cross target not using the same arch (the main reason of my patch was for using it with gentoo to cross compile my ARM rootfs)

heroxbd:
> It seems like we could get the same result with a far less invasive patch if we added a --map-root argument. Uid 1000 could call useradd -R ~/mychroot --map-root $(id -u) joe. useradd would then create a new user namespace mapping host 1000 to container uid 0, pivot or chroot to ~/mychroot, and act there.

Why do we need a user namespace or a container here?  It is more complex and does not work on older kernels.  On the other hand, adding a directory prefix so that a normal user can use the shadow CLI to update user/group files is more straightforward.

joakim-tjernlund:
Is this PR still on track? I am too would really like a a non chroot option for cross builds

fariouche:
I think yes. At least I'm keeping track with the upstream modifications until the maintainers have tested it and merge

lib/prototypes.h
```diff
@@ -271,6 +271,21 @@ extern void do_pam_passwd (const char *user, bool silent, bool change_expired);
...
+extern struct passwd *prefix_getpwnam(const char* name);
+extern struct spwd *prefix_getspnam(const char* name);
+extern struct group *prefix_getgr_nam_gid(const char *grname);
+extern void prefix_setpwent();
+extern struct passwd* prefix_getpwent();
+extern void prefix_endpwent();
+extern void prefix_setgrent();
+extern struct group* prefix_getgrent();
+extern void prefix_endgrent();
+
```
joakim-tjernlund: (void) vs. () ?

joakim-tjernlund:
upstream, what is keeping this from moving forward?

heroxbd:
> upstream, what is keeping this from moving forward?

I am afraid it is the laziness and incompetence of @hallyn.  He made a random comment without understanding the use case.  After 1 year, he come back again, forgets about what has been discussed and asks the same question.  This goes forever. 

Sorry @hallyn, I respect you stepping up to maintain this central but orphaned piece of software.  My comment is only on this issue.   If you need help co-maintain this feature in shadow, speak up. If you reject it, speak up and force people fork.

hallyn:
> I am afraid it is the laziness and incompetence of @hallyn.

wowza

lib/commonio.c
```diff
@@ -375,21 +375,30 @@ bool commonio_present (const struct commonio_db *db)
...
 	if (db->locked) {
 		return 1;
 	}
-
-	snprintf (file, sizeof file, "%s.%lu",
+	file_len = strlen(db->filename) + 11;/* %lu max size */
+	lock_file_len = strlen(db->filename) + 6; /* sizeof ".lock" */
+	file = (char*)malloc(file_len);
+	lock = (char*)malloc(lock_file_len);
+	snprintf (file, file_len, "%s.%lu",
```
hallyn: Ensure the mallocs succeeded before writing to them

libmisc/prefix_flag.c
```diff
@@ -0,0 +1,342 @@
...
+		fg = fopen(group_db_file, "rt");
+		if(!fg)
+			return NULL;
+		while(grp = fgetgrent(fg)) {
+			if(!strcmp(name, grp->gr_name))
+				break;
+		}
+		fclose(fg);
+		return grp;
+	}
```
hallyn: the } else is not needed here, since the if block always return.  (same in the ones below).

src/userdel.c
```diff
@@ -827,7 +832,19 @@ static int remove_mailbox (void)
 	if (NULL == maildir) {
 		return 0;
 	}
-	snprintf (mailfile, sizeof mailfile, "%s/%s", maildir, user_name);
+
+	len = strlen (prefix) + strlen (maildir) + strlen (user_name) + 2;
```
hallyn: It feels like a very long prefix could risk overflowing the stack in the alloca() below.  Do you think it's worth a sanity check?

libmisc/prefix_flag.c
fariouche: ok. I will add some checks.

src/userdel.c
fariouche: The max path length is limited in most systems. But indeed, I will add either a check on the max path or change the alloca() call to malloc()/free() (if possible) (but I think this is not specific to this part of the code. I do remember having seen alloca() in a few places)

fariouche:
Hi,

I've done the requested changes.
Also changed the lock function in commonio.c that in some case uses  lckpwdf() if available in the libc. However, this function use an hardcoded file path /etc/.pwd.lock.
So I'm forcing it to the fallback implementation in all cases.
An other possible implementation is to use the fallback only if a prefix is used.

hallyn:
Thanks, applied.

heroxbd:
Thank you so much @hallyn and @fariouche for your work.

Please forgive my rudeness. I owe you one.

joakim-tjernlund:
Great! How far off is a release with --prefix in it?

fariouche:
Great! Thank you!
I can now revive the Gentoo thread about this feature :-)

hallyn:
Quoting joakim-tjernlund (notifications@github.com):
> Great! How far off is a release with --prefix in it?

It shouldn't be far, a release is really overdue.

joakim-tjernlund:
Like now? :)

hallyn:
Hoping next week.  (HW key token should arrive this weekend making it easier to sign the release)

hallyn:
Hi, so while looking at a debian merge, i noticed something I hadn't before:

int commonio_lock (struct commonio_db *db)
{
/*#ifdef HAVE_LCKPWDF*/ /* not compatible with prefix option*/ 
#if 0

Does this mean that distros using HAVE_LOCKPWDF are not going to be able to run sanely at all?

fariouche:
Yes, this is my comment on March 28...
If needed, I can implement it only if a prefix is used

hallyn:
There are distros which definately will be broken without it.

fariouche:
ok. Then I propose the following patch, to use lckpwd only if no prefix is used
[lckpwdf_only_if_no_prefix.patch.txt](https://github.com/shadow-maint/shadow/files/1994882/lckpwdf_only_if_no_prefix.patch.txt)

PR #26, opened 2016-07-18
Fix sentence in usermod manual

jubalh:
Should have been: '[...] but only checkS [...]'.
So there was a missing 's'. Architectures isn't the right word either.
I decided to write the whole sentence new.


hallyn:
Looks better, thanks.

PR #29, opened 2016-08-04
Reduces syslog priority of common usage events

qua-bla:
- Log INFO instead of ERR on `su missing-user`
- Log NOTICE/WARN instead of ERR on pam_authenticate failure (wrong password for example)

*This PR is part of an ongoing effort to make syslog entries more expressive in Debian. Similar issues in important components like OpenSSH and PAM [were addressed upstream](https://github.com/hemio-ev/expressive-syslog/blob/master/README.md#sshd).*

### DIFF
--- a/src/su.c
+++ b/src/su.c
@@ -422,7 +422,7 @@ static void check_perms_pam (const struct passwd *pw)
 	int ret;
 	ret = pam_authenticate (pamh, 0);
 	if (PAM_SUCCESS != ret) {
-		SYSLOG ((LOG_ERR, "pam_authenticate: %s",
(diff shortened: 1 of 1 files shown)


hallyn:
Seems ok, but do you think there might be people grepping logs specifically for those levels expecting to see for instance wrong password messages?

qua-bla:
Are you asking why this change is useful? The reason is that we believe that level ERR should be reserved for messages that can be reported to system administrators automatically. There is little reason that the `su` errors imply that there is any action needed. (Usually it's just a misspelling by a admin anyway). I am addressing those specific problems because they became annoying on our systems.

If you are asking for backward compatibility: Log levels seem to have been used rarely before journald. Because of that, levels have been assigned carelessly in the past. There will be many other level changes in stretch.

PR #36, opened 2016-09-15
Add missing size args to snprintf calls.

atsampson:
This code was added in dd50014055a36969153e83cf3675d9559fa6a200,
but both calls to snprintf were missing their size arguments, which is
obviously wrong (and fails to compile with GCC 6).

Rebased on top of the fix added later.

### DIFF
--- a/src/su.c
+++ b/src/su.c
@@ -373,8 +373,8 @@ static void prepare_pam_close_session (void)
 		              stderr);
 		(void) kill (-pid_child, caught);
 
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks - sorry, I didn't see this when you first posted it, and during a feverish packaging attempt pushed my own version . I prefer your version with sizeof, please feel free to update your PR against current git HEAD and I'll merge it.

thanks.

atsampson:
Done!

hallyn:
Thanks!

PR #41, opened 2016-10-19
Remove non-POSIX option in chmod(1) used for src/Makefile.am

selkfoster:


vapier:
lgtm.  thanks!

hallyn:
Thanks.  My guess would be that the -f was added either out of habit at the time, or to work around a shortcoming in the testframe.  If the latter we can deal with it if/when it breaks again.

PR #43, opened 2016-11-05
buildsys: fix suidubins assignments

Blub:
These assignments were pasted as is into the Makefile and
ended up as part of a rule. (Usually the .PRECIOUS rule
which is why the build system never attempted to execute it
as commands, hiding the problem.)

Signed-off-by: Wolfgang Bumiller <wry.git@bumiller.com>
Reported-by: Rahel A <ra00177@surrey.ac.uk>


hallyn:
Thanks - I was worried about the fact that this seemed to be working anyway in Debian, but I see that seems to be done explicitly in debian/rules.

PR #45, opened 2016-11-10
Fixed suid bit for subuid tools

anti-social:
We detect that our containerization tool stopped working with `newuidmap` and `newgidmap` utilities from `shadow` 4.4: https://github.com/tailhook/vagga/issues/362

Also the same typo is present for the account tools: https://github.com/shadow-maint/shadow/blob/master/src/Makefile.am#L57
But possibly to fix it there should be changed default value for the `--enable-account-tools-setuid` flag. Because account utilities never have had suid bit due to the typo.


anti-social:
Duplicate of #43

PR #47, opened 2016-11-15
Make the group and user allocation more effective.

t8m:
Previously, the allocation was optimized for an outdated
deployment style (that of /etc/group alongside nss_db). The issue
here is that this results in extremely poor performance when using
SSSD, Winbind or nss_ldap.

There were actually two serious bugs here that have been addressed:

1) Running getgrent() loops won't work in most SSSD or Winbind
environments, as full group enumeration is disabled by default.
This could easily result in auto-allocating a group that was
already in use. (This might result in a security issue as well, if
the shared GID is a privileged group).

2) For system groups, the loop was always iterating through the
complete SYS_GID_MIN->SYS_GID_MAX range. On SSSD and Winbind, this
means hundreds of round-trips to LDAP (unless the GIDs were
specifically configured to be ignored by the SSSD or winbindd).
To a user with a slow connection to their LDAP server, this would
appear as if groupadd -r was hung. (Though it would eventually
complete).

This patch changes the algorithm to be more favorable for LDAP
environments, at the expense of some performance when using nss_db.
Given that the DB is a local service, this should have a negligible
effect from a user's perspective.

With the new algorithm, we simply first iterate through all entries
in the local database with gr_next(), recording the IDs that are in
use. We then start from the highest presumed-available entry and
call getgrgid() to see if it is available. We continue this until
we come to the first unused GID. We then select that and return it.

If we make it through all the remaining IDs without finding a free
one, we start over from the beginning of the range and try to find
room in one of the gaps in the range.

The patch was originally written by Stephen Gallagher and applied
identically also to the user allocation by Tomáš Mráz.

### DIFF
--- a/libmisc/find_new_gid.c
+++ b/libmisc/find_new_gid.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 1991 - 1994, Julianne Frances Haugh
  * Copyright (c) 2008 - 2011, Nicolas François
+ * Copyright (c) 2014, Red Hat, Inc.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


hallyn:
Hi - is this being used inthe current fedora packages?

t8m:
Yes, it is, for quite a long time.

hallyn:
Thanks, applied.

PR #48, opened 2016-11-15
Four simple patches from the Fedora package to merge

t8m:
Here are four simple patches from the Fedora shadow package that should be mergeable upstream.


hallyn:
Thanks.

PR #50, opened 2016-11-24
update Polish translation

m4sk1n:

PR #55, opened 2016-12-05
enable travis integration

glensc:
kickstart travis integration

currently fails because:
- #50 added invalid `po/pl.po` with fatal errors - it should be reverted (#50) or fixed (#56) (example build with reverted po: https://travis-ci.org/glensc/shadow/builds/181503106)

also removes `false` calls from `$(MAN_mans)` targets because it was not possible to disable man generation, it always invoked those targets regardless of `--disable-man` option

does not enable all features:
```
shadow will be compiled with the following features:
	auditing support:		no
	CrackLib support:		no
	PAM support:			no
	SELinux support:		no
	ACL support:			no
	Extended Attributes support:	no
	tcb support (incomplete):	no
	shadow group support:		yes
	S/Key support:			no
	SHA passwords encryption:	yes
	nscd support:			yes
	subordinate IDs support:	yes
```

appropriate debian dev packages need to be figured out, but that can be done in subsequent PRs

also `make distcheck` should be added to travis, but it's currently broken, should be fixed first.

after being merged, repo owner needs to enable travis integration from https://travis-ci.org/shadow-maint/shadow

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+sudo: false
+
+language: c
+
+compiler:
+  - gcc
+  - clang
+
+addons:
+  apt:
+    packages:
+      - autopoint
+      - xsltproc
+
+script:
+ - ./autogen.sh --without-selinux --disable-man
+ - grep ENABLE_ config.status
+ - make
+
+# vim:et:ts=2:sw=2
--- a/man/Makefile.am
+++ b/man/Makefile.am
@@ -215,5 +215,4 @@ $(man_MANS):
 	@echo "Error: you need to run configure with '--enable-man'"
 	@echo "       in order to regenerate man pages."
 	@echo ""
-	@false
 endif
--- a/man/generate_mans.mak
+++ b/man/generate_mans.mak
@@ -47,7 +47,6 @@ clean-local:
 else
 $(man_MANS):
 	@echo you need to run configure with --enable-man to generate man pages
-	@false
 endif
 
 man8/grpconv.8 man8/grpunconv.8 man8/pwunconv.8: man8/pwconv.8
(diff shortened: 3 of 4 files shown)


vapier:
imo it's a common mistake to try and stick all the build steps into `.travis.yml`.  instead, all the travis logic should live in a script and/or subdir of scripts.  this makes it easy to reproduce the behavior of travis locally (at least, in terms of build steps).

i don't see an existing dir that looks appropriate, so maybe just create a top level `travis/` dir and stick the script in there ?  name it like `travis/main.sh` and have `.travis.yml` simply do:

```
script: ./travis/main.sh
```

glensc:
i don't see that as mistake, it gives the better view (colorized) of each command being run.

other than that i'd propose `ci` subdir, for possible future ci systems (gitlab-ci is nice)

or just add more `make` targets and use these from travis. `make ci`?

vapier:
it is a mistake: growing beyond one or two commands gets unwieldy quickly, especially if you want to do any sort of dynamic logic.

it's easy to get the script output highlighted/colored the same as travis does.  it's simply scraping the output looking for specific sequences.  that means you can do the same in the script.

```
whitebg=$(tput setab 7)
blackfg=$(tput setaf 0)
normal=$(tput sgr0)
...
echo "${whitebg}${blackfg}fooooooooo${normal}"
```

adding it to the makefile wouldn't be that useful imo -- you'd still need the autoreconf+configure step before you could try and run `make`.

hallyn:
Let's go ahead and give it a shot, and we can break it out in later pull requests.

glensc:
Seems working!

https://travis-ci.org/shadow-maint/shadow

I hope it serves the benefit of preventing totally broken PR's like #50 getting merged (no offense intended).

Issue #49, opened 2016-11-22
message is unclear


crayxt:
Hello,

In kk.po file I see new message
```%s: failed to add uid range %lu-%lu from '%s'\```
is the phrase```from '%s'``` correct here? Shouldn't it be ```to '%s'```?

Thanks,

crayxt:
Lines
https://github.com/shadow-maint/shadow/blob/master/src/usermod.c#L2151
https://github.com/shadow-maint/shadow/blob/master/src/usermod.c#L2177

hallyn:
Hi - good catch, I think indeed those should be 'add to', and were
presumably wrongly copied from 'remove from'.

crayxt:
@hallyn ok, I am waithing for this and then updating the Kazakh translation, thanks

PR #57, opened 2016-12-07
Fix messages.

crayxt:
PR to fix https://github.com/shadow-maint/shadow/issues/49

Needs an update of PO files after merging.


hallyn:
D'oh, I missed htis and hand-applied it while look at the issue.  Sorry about that.

crayxt:
@hallyn that's ok, please update the po files.

PR #58, opened 2016-12-13
shadow: Add auditing support to su

juiceme:
This patch extends the auditing feature used in login to su.
Signed-off-by: Jussi Ohenoja <jussi.ohenoja@nokia.com>

On implementing audit traces to our LFS distro I noticed that the "su" binary from shdow package does not create any audit traces. This patch adds 2 audit logging points to su; one for a failure to authenticate and one for a successiful authentication.


hallyn:
Thanks, applied.

PR #62, opened 2017-01-24
zh_TW.po: Remove duplicated Plural-Forms

jwilk:
Found using [i18nspector](http://jwilk.net/software/i18nspector).


hallyn:
Thanks, applied.

PR #63, opened 2017-01-26
Don't crash on bogus keys in login.defs if PAM is enabled

berolinux:
Without this patch, PAM enabled builds crash when encountering an
invalid key in login.defs or key overrides because of array overflows

To reproduce, simply
	useradd -K Windows=broken

Signed-off-by: Bernhard Rosenkränzer <bero@lindev.ch>

### DIFF
--- a/lib/getdef.c
+++ b/lib/getdef.c
@@ -148,6 +148,7 @@ static struct itemdef knowndef_table[] = {
 #ifdef USE_PAM
 	PAMDEFS
 #endif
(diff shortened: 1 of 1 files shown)


vapier:
pretty sure it'll crash regardless of PAM settings if you put unknown values into the config file.

this type of crash is typically described as a "out of bounds read" rather than "array overflow".

hallyn:
Thanks guys, applied.

PR #65, opened 2017-02-09
useradd: Read defaults after changing root directories

dm0-:
There was an apparently unintentional change in behavior with 65c261714027d6d2bcf10cd103c5209ae9f7617f.  Previously, using the `--root` option with `useradd` would cause defaults to be loaded from the target root directory, but that commit caused defaults to be loaded before changing roots.  This was initially reported in coreos/bugs#1787.


hallyn:
Thanks, applied.

PR #66, opened 2017-02-10
Deleted a misplaced semicolon.

jmoellers:
There is a misplaced semicolon which causes an error message to be generated despite the fact that the clearing of the error log was successful.


hallyn:
Thanks, applied.

PR #70, opened 2017-03-31
Fix buffer overflow if NULL line is present in db.

t8m:
If ptr->line == NULL for an entry, the first cycle will exit,
but the second one will happily write past entries buffer.
We actually do not want to exit the first cycle prematurely
on ptr->line == NULL.
Signed-off-by: Tomas Mraz <tmraz@fedoraproject.org>


hallyn:
Thanks.

PR #71, opened 2017-04-10
Make the sp_lstchg shadow field reproducible.

lamby:
The third field in the /etc/shadow file (sp_lstchg) contains the date of
the last password change expressed as the number of days since Jan 1, 1970.
As this is a relative time, creating a user today will result in:

    username:17238:0:99999:7:::

whilst creating the same user tomorrow will result in:

    username:17239:0:99999:7:::

This has an impact for the Reproducible Builds[0] project where we aim to
be independent of as many elements the build environment as possible,
including the current date.

This patch changes the behaviour to use the SOURCE_DATE_EPOCH[1]
environment variable (instead of Jan 1, 1970) if valid.

 [0] https://reproducible-builds.org/
 [1] https://reproducible-builds.org/specs/source-date-epoch/

Signed-off-by: Chris Lamb <lamby@debian.org>

### DIFF
--- a/lib/prototypes.h
+++ b/lib/prototypes.h
@@ -179,6 +179,9 @@ extern int getrange (char *range,
                      unsigned long *min, bool *has_min,
                      unsigned long *max, bool *has_max);
 
+/* gettime.c */
+extern time_t gettime ();
+
 /* get_uid.c */
 extern int get_uid (const char *uidstr, uid_t *uid);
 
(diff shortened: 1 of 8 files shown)


hallyn:
Thanks!

Do you think there is any value in checking that the time passed in the environment variable is at least in the past?  (To avoid users trying to make their own passwords expire later than they should)

Is there any danger in an unprivileged user setting a password that's older than it really is?  I think the worst they can do is lock themselves out early...

lamby:
Good call; updated :)

hallyn:
Thanks!

PR #72, opened 2017-05-14
Reset pid_child only if waitpid was successful.

stoeckmann:
Do not reset the pid_child to 0 if the child process is still
running. This else-condition can be reached with pid being -1,
therefore explicitly test this condition.

This is a regression fix for CVE-2017-2616. If su receives a
signal like SIGTERM, it is not propagated to the child.

Reported-by: Radu Duta <raduduta@gmail.com>


hallyn:
Thanks very much!

PR #74, opened 2017-05-22
support dynamically added users via pam_group

AdamMajer:
Dynamically added users via pam_group are not listed in groups
databases but are still valid.


hallyn:
Thanks, looks good.  Can you describe how we could test this?

PR #82, opened 2017-08-14
newgrp: avoid unnecessary group lookups

t8m:
In case a system uses remote identity server (LDAP) the group lookup
can be very slow. We avoid it when we already know the user has the
group membership.


hallyn:
Thanks.  Looks good to me.

PR #86, opened 2017-09-19
Make language more inclusive

WheresAlice:
This started by making the language in usermod more inclusive by removing the reference to binary genders since I was getting annoyed by this.  However, it quickly became apparent it's a bigger problem than that one single occurrence.

It is not necessary to refer to the gender of the user at all, and so I have gone through removing all gendered language, replacing it with gender neutral terms.

It is possible that I have not caught everything, and there is almost certainly some work to be done in the translations, but this makes things a lot better in terms of inclusive language.


vapier:
lgtm, if only because "he/she" is an abomination of a construction

glensc:
probably should strip comments at all for test data files, for future maintainability.

brauner:
Also could be a script that calls `sed` so that users can cherry-pick the commit and apply it to their docs locally.

hallyn:
> lgtm, if only because "he/she" is an abomination of a construction

Yes, that is exactly what I was thinking :)

I've been holding off on hitting merge because it is 6.5k loc.  I was
hoping someone would say "I've done a test build and it works" but failing
that I will take a close look when I get some time.

vapier:
the travis-ci passed ... is that not sufficient coverage ?

WheresAlice:
Given this is such a large commit and it is help and documentation rather than code I feel like tests aren't necessarily going to catch any problems with this. I can provide a shell script which makes these changes if it helps show what I've done better.

WheresAlice:
```bash
sed -i 's/his\/her/their/g;s/him\/her/the user/g' contrib/groupmems.shar
sed -i 's/his\/her/their/g;s/him\/her/the user/g' etc/login.defs
find man/ -type f | xargs sed -i 's/his\/her/their/g;s/him\/her/the user/g'
find po/  -type f | xargs sed -i 's/his\/her/their/g;s/him\/her/the user/g'
sed -i 's/his\/her/their/g;s/him\/her/the user/g' src/usermod.c
find tests -type f -name login.defs | xargs sed -i 's/his\/her/their/g;s/him\/her/the user/g'
find tests -type f -name usage.out | xargs sed -i 's/his\/her/their/g;s/him\/her/the user/g'
```

hallyn:
Thanks, I'll take another look very soon.

hallyn:
Thank you!

PR #88, opened 2017-10-22
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

Most of these changes shouldn't be controversial.

`pam_pass_non_interractive` and friends are almost certainly an api change.
Since each thing is a distinct commit, dropping `interactive` is pretty easy.

If you want things split, I can probably do that w/in reason. Note that this isn't a priority project for me (I'm currently working through chromium which will take a few months).


jsoref:
Oh, as a general note looking forward, I'd encourage you not to intentionally include misspellings in your changelog files. They make tasks like this much more complicated.

(I did try not to correct those instances and the names of misspelled files from Debian patches)

hallyn:
Thanks.  The spelling changes are not controversial.  Changing the Changlog is.  (Which is why I've been postponing acting on this for so long, sorry)

Note that most of those come from previous maintainers, and they're a part of history.  Maybe I'm being silly, but I feel wrong changing them.  If someone else (with a history in the project) wants to +1 accepting these fixes, then I'll accept it.

jsoref:
Sorry, I'll try to get around to dropping the relevant bits eventually. I didn't lose it, it's just not high on my tasks...

Issue #89, opened 2017-11-02
userdel -R option does not work as expected


t8m:
When removing users from chroot the userdel still checks their presence with getpwnam(). However this call is expected to fail if the chroot does not contain the required nsswitch libraries. The userdel should not call getpwnam() if operating in chroot.

hallyn:
Thanks for reporting that, indeed that shouldn't check against the host database.

Would you be able to provide a patch to fix it?

PR #90, opened 2017-12-21
Make userdel to work with -R.

t8m:
The userdel checks for users with getpwnam() which might not work
properly in chroot. Check for the user's presence in local files only.
Fixes issue #89

### DIFF
--- a/src/userdel.c
+++ b/src/userdel.c
@@ -96,6 +96,7 @@ static char *user_home;
 static bool fflg = false;
 static bool rflg = false;
 static bool Zflg = false;
+static bool Rflg = false;
 
 static bool is_shadow_pwd;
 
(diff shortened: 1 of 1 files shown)


src/userdel.c
```diff
@@ -1150,7 +1156,7 @@ int main (int argc, char **argv)
 	 * Note: This is a best effort basis. The user may log in between,
 	 * a cron job may be started on her behalf, etc.
 	 */
-	if (user_busy (user_name, user_id) != 0) {
+	if (!Rflg && user_busy (user_name, user_id) != 0) {
```
hallyn: Hi - I'm fine merging as-is, but just to make sure - do you want to check whether the user_name is busy in the chroot?

t8m: No, I don't. It would be basically impossible to do.

PR #91, opened 2018-01-03
Add note to passwd(1) that --maxdays -1 disables the setting.

kloeri:
This note already exists in chage(1).


hallyn:
Thanks for the patch.

PR #92, opened 2018-01-17
Fixes mispelling of MAX_DAYS help text

IronicBadger:
Tiny, tiny change but one I noticed during a training course.

Maximum was misspelled as maximim in two places.


hallyn:
Thanks.

PR #97, opened 2018-02-15
newgidmap: enforce setgroups=deny if self-mapping a group

cyphar:
This is necessary to match the kernel-side policy of "self-mapping in a
user namespace is fine, but you cannot drop groups" -- a policy that was
created in order to stop user namespaces from allowing trivial privilege
escalation by dropping supplementary groups that were "blacklisted" from
certain paths.

This is the simplest fix for the underlying issue, and effectively makes
it so that unless a user has a valid mapping set in /etc/subgid (which
only administrators can modify) -- and they are currently trying to use
that mapping -- then /proc/$pid/setgroups will be set to deny. This
workaround is only partial, because ideally it should be possible to set
an "allow_setgroups" or "deny_setgroups" flag in /etc/subgid to allow
administrators to further restrict newgidmap(1).

Ref: https://bugs.launchpad.net/ubuntu/+source/shadow/+bug/1729357
Fixes: CVE-2018-7169
Reported-by: Craig Furman <craig.furman89@gmail.com>
Signed-off-by: Aleksa Sarai <asarai@suse.de>

### DIFF
--- a/src/newgidmap.c
+++ b/src/newgidmap.c
@@ -46,32 +46,37 @@
  */
 const char *Prog;
 
-static bool verify_range(struct passwd *pw, struct map_range *range)
+
+static bool verify_range(struct passwd *pw, struct map_range *range, bool *allow_setgroups)
 {
 	/* An empty range is invalid */
 	if (range->count == 0)
 		return false;
 
-	/* Test /etc/subgid */
-	if (have_sub_gids(pw->pw_name, range->lower, range->count))
+	/* Test /etc/subgid. If the mapping is valid then we allow setgroups. */
+	if (have_sub_gids(pw->pw_name, range->lower, range->count)) {
+		*allow_setgroups = true;
 		return true;
+	}
 
-	/* Allow a process to map its own gid */
-	if ((range->count == 1) && (pw->pw_gid == range->lower))
+	/* Allow a process to map its own gid. */
+	if ((range->count == 1) && (pw->pw_gid == range->lower)) {
+		/* noop -- if setgroups is enabled already we won't disable it. */
 		return true;
+	}
 
 	return false;
 }
 
 static void verify_ranges(struct passwd *pw, int ranges,
-	struct map_range *mappings)
+	struct map_range *mappings, bool *allow_setgroups)
 {
 	struct map_range *mapping;
 	int idx;
 
 	mapping = mappings;
 	for (idx = 0; idx < ranges; idx++, mapping++) {
-		if (!verify_range(pw, mapping)) {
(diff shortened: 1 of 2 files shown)


src/newgidmap.c
```diff
@@ -46,32 +46,41 @@
...
 {
 	/* An empty range is invalid */
 	if (range->count == 0)
 		return false;
 
-	/* Test /etc/subgid */
-	if (have_sub_gids(pw->pw_name, range->lower, range->count))
+	/* Test /etc/subgid. If the mapping is valid then we allow setgroups. */
+	if (have_sub_gids(pw->pw_name, range->lower, range->count)) {
+		*allow_setgroups |= true;
```
brauner: Why are you `|=` this instead of simply assigning if it is a bool?

brauner: I might just be dense.

cyphar: Ah, that's an artifact of my old patchset (which used a bit-mask to specify the different states rather than setting the policy in `verify_mapping`). I'll swap it to just assignment.

cyphar: Fixed, thanks!

src/newgidmap.c
```diff
@@ -89,6 +94,36 @@ static void usage(void)
...
+	if (setgroups_fd < 0) {
+		/*
+		 * If it's an ENOENT then we are on too old a kernel for the setgroups
+		 * code to exist. Emit a warning and bail on this.
+		 */
+		if (ENOENT == errno) {
+			fprintf(stderr, _("%s: kernel doesn't support setgroups restrictions\n"), Prog);
+			return;
+		}
+		fprintf(stderr, _("%s: couldn't open process setgroups: %m\n"), Prog);
```
AkihiroSuda: I think it is better to avoid GNU extension %m

src/newgidmap.c
```diff
@@ -89,6 +94,36 @@ static void usage(void)
 	exit(EXIT_FAILURE);
 }
 
+void write_setgroups(int proc_dir_fd, unsigned int allow_setgroups)
+{
+	int setgroups_fd = openat(proc_dir_fd, "setgroups", O_WRONLY);
```
vapier: should add O_CLOEXEC, if only for "good hygiene"

hallyn:
Thanks, everyone!  I like this.

hallyn:
I suppose this means we should do a release.

cyphar:
I will work on the whole `allow_setgroups` patchset we discussed over the weekend (the complicated part is that `subordinate_range` is not accessible from `src/newgidmap.c`).

PR #99, opened 2018-02-18
newsubgid: add deny_setgroups option to /etc/subgid

cyphar:
Add a new deny_setgroups (and corresponding allow_setgroups) flag to
/etc/subgid. The purpose of this flag is to extend the security
protections against CVE-2018-7169, so that even group mapping configured
in /etc/subgid by an administrator can still disable setgroups.

However, rather than the fairly lenient semantics for self-mapping, the
semantics of /etc/subgid are stronger. If a mapping is encountered where
"deny_setgroups" is set, then no other mapping can "undo" this
restriction. The reason for this is that "deny_setgroups" indicates that
(according to the administrator) the mapping is unsafe to allow setgroups
in, and adding more mappings will not change this fact. "allow_setgroups"
is the default, and setting it is a noop. The logic used when applying
setgroups policies is unchanged (only denies are written, and we don't
write anything if it's already denied).

Signed-off-by: Aleksa Sarai <asarai@suse.de>

### DIFF
--- a/lib/subordinateio.c
+++ b/lib/subordinateio.c
@@ -18,9 +18,11 @@ struct subordinate_range {
 	const char *owner;
 	unsigned long start;
 	unsigned long count;
+	char **options;
 };
 
-#define NFIELDS 3
+#define NFIELDS_MIN 3
+#define NFIELDS_MAX 4
 
 /*
  * subordinate_dup: create a duplicate range
@@ -46,6 +48,7 @@ static /*@null@*/ /*@only@*/void *subordinate_dup (const void *ent)
 	}
 	range->start = rangeent->start;
 	range->count = rangeent->count;
+	range->options = dup_list(rangeent->options);
 
 	return range;
 }
@@ -58,8 +61,9 @@ static /*@null@*/ /*@only@*/void *subordinate_dup (const void *ent)
 static void subordinate_free (/*@out@*/ /*@only@*/void *ent)
 {
 	struct subordinate_range *rangeent = ent;
-	
+
 	free ((void *)(rangeent->owner));
+	free_list(rangeent->options);
 	free (rangeent);
 }
 
@@ -76,9 +80,9 @@ static void *subordinate_parse (const char *line)
 {
 	static struct subordinate_range range;
 	static char rangebuf[1024];
-	int i;
+	int i, j;
 	char *cp;
-	char *fields[NFIELDS];
+	char *fields[NFIELDS_MAX];
 
 	/*
 	 * Copy the string to a temporary buffer so the substrings can
@@ -93,7 +97,7 @@ static void *subordinate_parse (const char *line)
 	 * field.  The fields are converted into NUL terminated strings.
 	 */
 
-	for (cp = rangebuf, i = 0; (i < NFIELDS) && (NULL != cp); i++) {
+	for (cp = rangebuf, i = 0; (i < NFIELDS_MAX) && (NULL != cp); i++) {
 		fields[i] = cp;
 		while (('\0' != *cp) && (':' != *cp)) {
 			cp++;
@@ -108,16 +112,23 @@ static void *subordinate_parse (const char *line)
 	}
 
 	/*
-	 * There must be exactly NFIELDS colon separated fields or
-	 * the entry is invalid.  Also, fields must be non-blank.
+	 * There must be at least NFIELDS_MIN and at most NFIELDS_MAX colon
+	 * separated fields or the entry is invalid. The first NFIELDS_MIN fields
+	 * must be non-blank.
 	 */
-	if (i != NFIELDS || *fields[0] == '\0' || *fields[1] == '\0' || *fields[2] == '\0')
+	if (i < NFIELDS_MIN || i > NFIELDS_MAX)
 		return NULL;
+	for (j = 0; j < NFIELDS_MIN; j++)
+		if (*fields[j] == '\0')
+			return NULL;
+
 	range.owner = fields[0];
 	if (getulong (fields[1], &range.start) == 0)
 		return NULL;
 	if (getulong (fields[2], &range.count) == 0)
 		return NULL;
+	if (i >= 4)
+		range.options = comma_to_list(fields[3]);
 
 	return &range;
 }
@@ -133,11 +144,15 @@ static void *subordinate_parse (const char *line)
 static int subordinate_put (const void *ent, FILE * file)
 {
 	const struct subordinate_range *range = ent;
+	char *options = comma_from_list(range->options);
 
-	return fprintf(file, "%s:%lu:%lu\n",
-			       range->owner,
-			       range->start,
-			       range->count) < 0 ? -1  : 0;
+	int n = fprintf(file, "%s:%lu:%lu:%s\n",
+			      range->owner,
+			      range->start,
+			      range->count,
+			      options ?: "");
+	free(options);
+	return n < 0 ? -1 : 0;
 }
 
 static struct commonio_ops subordinate_ops = {
@@ -285,38 +300,41 @@ static const struct subordinate_range *find_range(struct commonio_db *db,
 }
 
 /*
- * have_range: check whether @owner is authorized to use the range
- *             (@start .. @start+@count-1).
+ * get_range: check whether @owner is authorized to use the range
+ *             (@start .. @start+@count-1) and return the first matching range.
  * @db: database to check
  * @owner: owning uid being queried
  * @start: start of range
  * @count: number of uids in range
  *
- * Returns true if @owner is authorized to use the range, false otherwise.
+ * Returns the corresponding subordinate_range if @owner is authorized to use
+ * the range, NULL otherwise.
  */
-static bool have_range(struct commonio_db *db,
-		       const char *owner, unsigned long start, unsigned long count)
+static const struct subordinate_range *get_range(struct commonio_db *db,
+						 const char *owner,
+						 unsigned long start,
+						 unsigned long count)
 {
 	const struct subordinate_range *range;
 	unsigned long end;
 
 	if (count == 0)
-		return false;
+		return NULL;
 
 	end = start + count - 1;
 	range = find_range (db, owner, start);
 	while (range) {
-		unsigned long last; 
+		unsigned long last;
 
 		last = range->start + range->count - 1;
 		if (last >= (start + count - 1))
-			return true;
+			return range;
 
 		count = end - last;
 		start = last + 1;
 		range = find_range(db, owner, start);
 	}
-	return false;
+	return NULL;
 }
 
 /*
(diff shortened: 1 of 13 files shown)


cyphar:
This is currently still work-in-progress (even though it all works fine now). The issue is that I cannot get `lib/` to use `libmisc/list.c` because automake really doesn't like me trying to add an `LDADD` for `libmisc.la` to `lib_la_LDADD`. So currently I just copied `list.c` to `lib/` (which is ... bad).

src/newgidmap.c
```diff
@@ -46,16 +46,83 @@
  */
 const char *Prog;
 
+enum quadopt_t {
```
AkihiroSuda: how is this "quad"?

libmisc/list.c
```diff
@@ -169,6 +170,23 @@
 	return tmp;
 }
 
+/*
+ * Free a list.
+ * The output list isn't modified, but the memory is freed.
+ */
+void free_list (char *const *list)
```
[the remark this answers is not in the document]
cyphar: I did this to match the signature of `dup_list`, but I don't really mind either way.

src/newuidmap.c
```diff
@@ -46,15 +46,40 @@
...
+
+	if (NULL == flags)
+		return;
+
+	for (i = 0; NULL != flags[i]; i++) {
+		char *flag = flags[i];
+
+		if (strlen(flag) < 1)
+			continue;
+
```
AkihiroSuda: Please add a comment about that no uid flag is valid atm

cyphar:
@AkihiroSuda Working on it, I'm trying to figure out how to fix AutoMake first.

src/newgidmap.c
cyphar: I'm using the term `quadoption` from Mutt, but you're right that this might not be the best name.

AkihiroSuda: Actually this looks `triopt`, as it has three values.

In github.com/moby/buildkit we call this `BoolOrAuto`.

cyphar:
Alright the AutoMake issue has been resolved, so I've removed `lib/list.c`. I've also now included manpage updates for this change as well as #97 (to clarify when and how `newgidmap` will write `deny`), as well as changed `quadoption` to `autobool` (and added helpers for `autobool_t`).

There aren't any `newgidmap` tests, but I have a feeling we should probably add some (though the tests aren't run as part of Travis -- restricting their usefulness).

hallyn:
> There aren't any `newgidmap` tests, but I have a feeling we should probably add some (though the tests aren't run as part of Travis -- restricting their usefulness).

We can fix that

cyphar:
Oh, I also haven't added any way for people to set `deny_setgroups` (or other options in the future) within `usermod` or `useradd` -- I couldn't figure out what the interface should look like. Any suggestions, or should we leave that feature for later?

man/subgid.5.xml
```diff
@@ -86,6 +89,42 @@
...
+      Options are comma-separated. Empty options are ignored, and if the
+      field is missing entirely it is treated as an empty string. Attempting
+      to use an unknown option will cause <command>newgidmap</command> to emit
+      an error. Setting options from the same option-set multiple times in a
+      single entry will result in the last option specified taking precedence.
+    </para>
+
+    <variablelist>
+      <varlistentry>
+        <term><option>allow_setgroups</option> (default),
```
hallyn: Hm, is this the right place to put this kind of thing though?  Maybe it is, but:

The only case I know of for deny-setgroups is when an administrator has placed a user in a group with a negative acl, to prevent that user from accessing some resource.

In that case, does it make more sense to place a sort of "group_locked" option on the group itself in /etc/group?  Then when newgidmap runs, it checks whether the target pid is in any locked group, and if so sets deny_setgroup.

Placing the deny/allow_setgroup variable in the gid mapping seems a step removed, and might cause an admin to not notice that "hey, the user being allowd to grab those gids should have been locked into group nos3critdocs.

cyphar: Hmm, I hadn't though of this point.

You're quite right that most users of negative ACLs are done with groups (and I think it does make intuitive sense to be able to say "if you are in a bad group then you cannot use `setgroups`"), but I have a couple of questions about edge cases:

* Should an administrator be able to override this decision for particular users or groups? Should that also be in `/etc/groups` (or `/etc/passwd`)?

* What if it's a numeric group (so not one registered in `/etc/groups`)? Is the solution to just tell people "tough luck, you should register your group"?

* This is a bit of an edge condition (and is an edge condition that actually isn't really easy to handle with the current implementation either), but how will this affect containers on the system that are also using group blacklisting? While `/etc/sub*` are global, if you have containers using allocated portions of the UID/GID space then you would expect that a user shouldn't be able to use `newgidmap` to drop the negative ACLs for a different group set than the one you are joining? Now, this is quite a weird case because it would require you to have a quite strange ACL setup -- so maybe this is a use case we don't care about?

hallyn: > Should an administrator be able to override this decision for particular users or groups? Should that also be in /etc/groups (or /etc/passwd)?

I don't really think so, as you either want to deny the group certain access or you don't, but my gut says there are more subtle cases.  So let's say yes.

> What if it's a numeric group (so not one registered in /etc/groups)? Is the solution to just tell people "tough luck, you should register your group"?

I think requiring a name makes sense.  Do you?

> how will this affect containers on the system that are also using group blacklisting? 

Hmm - that applies to your approach too right?

It really seems like this requires a new kernel functionality to make the most sense - a new per-process 'lockedgrp' from which groups can never be removed.  Otherwise all we can really do in the parent container is say "never let the child container do a setgroup" (right?  or have i confused myself on this topic once again?)

cyphar:
Alright, I'm going to work on the `/etc/group` configuration method in a separate branch.

hallyn:
Thanks @cyphar - should I close this one then?

cyphar:
Yup, I'll close it.

hallyn:
Hey @cyphar did this topic continue elsewhere?

cyphar:
No, I started working on it and ran into some roadbump and then moved on to something else. I can take a look at reviving this when I get back from vacation in ~2 weeks.

cyphar:
No, I started working on it and ran into some roadbump and then moved on to something else. I can take a look at reviving this when I get back from vacation in ~2 weeks.

hallyn:
> No, I started working on it and ran into some roadbump and then moved on to something else. I can take a look at reviving this when I get back from vacation in ~2 weeks.

ping? :)

cyphar:
Okay, I've started looking at this again, and I've remembered the roadblock I ran into last time.

@hallyn How should I handle the fact that `struct group` comes from libc and thus we can't add fields to it? Adding an argument to pass back the flags would be a little invasive, should I just wrap `struct group` into `struct group_ext` (or similar)?

hallyn:
> Okay, I've started rewriting at this again, and I've remembered the roadblock I ran into last time.
> 
> @hallyn How should I handle the fact that `struct group` comes from libc and thus we can't add fields to it? Adding an argument to pass back the flags would be a little invasive, should I just wrap `struct group` into `struct group_ext` (or similar)? Are there users of our internal libraries, or can we change stuff like this without issue?

D'oh, sorry, I never replied to this...

There are no outside users of our internal libraries that I know of.  Only libsubid.  Introducing struct group_ext should be fine.  I can't visualize right now where all you need this, so not quite sure whether there would be a better way.  Could you post a link to your current branch so I can take a look?

hallyn:
Hi @cyphar - no pressure, just wondering - have you given this any more thought? :)

alejandro-colomar:
This would need rebasing.

cyphar:
It needs a complete rework to be honest (rather than marking the mapping as denying `setgroups`, there should be a marking for a group to indicate it is "locked" and thus must not be dropped). Last I looked, I was worried about changing the structures we use for groups but @hallyn said it was fine, so it might be okay to just do it in the most straightforward way.

Unfortunately, this seems to keep falling off my plate...

lib/subordinateio.c
```diff
@@ -133,11 +144,15 @@ static void *subordinate_parse (const char *line)
...
 
-	return fprintf(file, "%s:%lu:%lu\n",
-			       range->owner,
-			       range->start,
-			       range->count) < 0 ? -1  : 0;
+	int n = fprintf(file, "%s:%lu:%lu:%s\n",
+			      range->owner,
+			      range->start,
+			      range->count,
+			      options ?: "");
```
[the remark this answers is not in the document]
alejandro-colomar: We have recently embraced this GNU extension.

libmisc/list.c
```diff
@@ -269,3 +287,30 @@ bool is_on_list (char *const *list, const char *member)
...
+	char *comma;
+	int i, commalen = 0;
+
+	if (NULL == list)
+		return NULL;
+
+	for (i = 0; NULL != list[i]; i++)
+		commalen += strlen(list[i]) + 1;
+
+	comma = xmalloc(commalen + 1);
```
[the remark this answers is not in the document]
alejandro-colomar: Interestingly, the loops are simpler if we leave a trailing comma, so that a CSV would look like `a,b,c,\0`.  It is technically incorrect as a CSV --as you have a trailing empty element--, but we do support such CSVs elsewhere, so we could keep the +1.

libmisc/list.c
```diff
@@ -269,3 +287,30 @@ bool is_on_list (char *const *list, const char *member)
...
+	memset(comma, '\0', commalen + 1);
+
+	for (i = 0; NULL != list[i]; i++) {
+		int j = strlen(comma);
```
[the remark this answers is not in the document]
alejandro-colomar: We now have stpecpy(), which is designed precisely with this in mind.

libmisc/list.c
```diff
@@ -269,3 +287,30 @@ bool is_on_list (char *const *list, const char *member)
...
+		if (j > 0)
+			comma[j++] = ',';
+		strncat(comma, list[i], commalen - j);
+	}
+
+	return comma;
+}
```
alejandro-colomar, line 316: Please use stpecpy() instead.  We have forbidden strncat() with string input; we only allow it with `[[gnu::nonstring]]` input --that is, utmp(5) members--.

With stpecpy() --which is defined in `lib/string/`--, it would be:

```c
char  *p, *e;

[...]

p = comma;
e = comma + commalen + 1;
for (int i = 0; NULL != list[i]; i++) {
	p = stpecpy(p, e, list[i]);
	p = stpecpy(p, e, ",");
}

if (p == NULL) {
	free(comma);
	return NULL;
}

return comma;
```

alejandro-colomar: Or, since we know there won't be truncation, we can just use stpcpy(3) from libc.

```c
char  *p;

[...]

p = comma;
for (int i = 0; NULL != list[i]; i++) {
	p = stpecpy(p, list[i]);
	p = stpecpy(p, ",");
}

return comma;
```

PR #100, opened 2018-03-12
Complete translation to Norwegian bokmål

akrosikam:
Translate remaining strings to Norwegian bokmål (nb). Also, cure previous translation of excessive anglicism and apply a more consistent use of actual Norwegian syntax.


hallyn:
Thanks!

PR #112, opened 2018-05-15
Create parent dirs for useradd -m

jubalh:
Equivalent of `mkdir -p`. It will create all parent directories.
Example: `useradd -d /home2/testu1 -m testu1`

Based on https://github.com/shadow-maint/shadow/pull/2 by Thorsten Kukuk
and Thorsten Behrens which was Code from pwdutils 3.2.2 with slight adaptations.

Adapted to so it applies to current code.

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2018,6 +2018,19 @@ static void usr_update (void)
 static void create_home (void)
 {
 	if (access (prefix_user_home, F_OK) != 0) {
+		char path[strlen (prefix_user_home) + 2];
+		char *bhome, *cp;
+
+		path[0] = '\0';
+		bhome = strdup (prefix_user_home);
+		if (!bhome) {
+			fprintf (stderr,
+							_("%s: error while duplicating string %s\n"),
+							Prog, user_home);
+			fail_exit (E_HOMEDIR);
+		}
+		++bhome;
+
 #ifdef WITH_SELINUX
 		if (set_selinux_file_context (prefix_user_home) != 0) {
 			fprintf (stderr,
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks!  Will take a look.

seiferteric:
FYI, there is an issue if the username contains a `/` which unfortunately debian at least allows.

```
$ sudo useradd -m 'abc/def'
useradd: warning: chown on `/home/abc/def' failed: No such file or directory
$ ls -l /home/
total 8
drwxr-xr-x  3 root     root     4096 May 23 20:22 abc
$ ls -l /home/abc/
total 4
drwxr-xr-x 2 abc/def abc/def 4096 May 23 20:22 def
$ grep abc /etc/passwd
abc/def:x:1001:1001::/home/abc/def:/bin/sh

```

ikerexxe:
> FYI, there is an issue if the username contains a `/` which unfortunately debian at least allows.
> 
> ```
> $ sudo useradd -m 'abc/def'
> useradd: warning: chown on `/home/abc/def' failed: No such file or directory
> $ ls -l /home/
> total 8
> drwxr-xr-x  3 root     root     4096 May 23 20:22 abc
> $ ls -l /home/abc/
> total 4
> drwxr-xr-x 2 abc/def abc/def 4096 May 23 20:22 def
> $ grep abc /etc/passwd
> abc/def:x:1001:1001::/home/abc/def:/bin/sh
> ```

This seems like a Debian related bug as I'm unable to reproduce it in Fedora:
```
# useradd -m 'abc/def'
useradd: invalid user name 'abc/def': use --badname to ignore
# useradd -m abc/def
useradd: invalid user name 'abc/def': use --badname to ignore
```

I'd recommend you to open a bug in their tracking system.

jubalh:
> FYI, there is an issue if the username contains a / which unfortunately debian at least allows.

Someone tested it for me now on Debian Bookworm and Bullseye and reported that it works in both cases.

seiferteric:
Yes it has to do with Debian patching chkname.c to allow almost all characters in usernames by default. That and this change to create parent directories will create the issue. My issue is that when creating the directories it will do strtok("/") on the home path, but it does not distinguish between a "/" in the path and one in the username itself. I beelive that if the "/" is in the username itself, it should not attempt to create it as a parent directory since it is not, but rather part of the username. So that is a bug I think here.

seiferteric:
@jubalh in Bullseye and Bookworm, I just tested in docker, and I don't se the "chown" error, but behavior is the same. The other issue is due to the ambiguity of "/" in username and home path, it doesn't remove correctly when removing the user with "userdel -r":
```
eseifert@bilbo:~$ docker run -it debian:bullseye
root@ba14121520fb:/# 
root@ba14121520fb:/# 
root@ba14121520fb:/# useradd -m "abc/def"
root@ba14121520fb:/# ls -l /home/
total 4
drwxr-xr-x 3 root root 4096 May 25 14:41 abc
root@ba14121520fb:/# grep abc /etc/passwd
abc/def:x:1000:1000::/home/abc/def:/bin/sh
root@ba14121520fb:/# userdel -r "abc/def"
userdel: abc/def mail spool (/var/mail/abc/def) not found
root@ba14121520fb:/# ls -l /home/
total 4
drwxr-xr-x 2 root root 4096 May 25 14:42 abc
root@ba14121520fb:/# grep abc /etc/passwd
root@ba14121520fb:/# 
```

jubalh:
> Someone tested it for me now on Debian Bookworm and Bullseye and reported that it works in both cases.

Correction: they confused useradd and adduser ;)

PR #118, opened 2018-06-24
[WIP] Support systems that only have utmpx

awilfox:
This allows shadow-utils to build on systems like Adélie, which have no
<utmp.h> header or `struct utmp`.  We use a <utmpx.h>-based daemon,
[utmps], which uses `struct utmpx` only.

Tested both `login` and `logoutd` with utmps and both work correctly.

[utmps]: http://skarnet.org/software/utmps/

---

NOTE: This has not been tested on a glibc system yet (I don't have any outside of chroots; it is not possible for me to test `login` / `logoutd` functionality in a chroot.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -74,12 +74,6 @@ AC_CHECK_MEMBERS([struct utmp.ut_type,
                   struct utmp.ut_time,
                   struct utmp.ut_xtime,
                   struct utmp.ut_tv],,,[[#include <utmp.h>]])
-dnl There are dependencies:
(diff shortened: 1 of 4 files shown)


hallyn:
Hi,

you marked this [WIP] - what else are you planning on doing before you'd consider it ready to merge?

awilfox:
It needs to be tested on a glibc system to ensure no regressions.  I still don't have access to one at this moment.

hallyn:
Builds, installs and runs fine here on glibc, thanks.

PR #119, opened 2018-06-25
su.c: run pam_getenvlist() after setup_env

mvo5:
When "su -l" is used the behaviour is described as similar to
a direct login. However login.c is doing a setup_env(pw) and then a
pam_getenvlist() in this scenario. But su.c is doing it the other
way around. Which means that the value of PATH from /etc/environment
is overridden. I think this is a bug because the man-page claims that 
"-l": "provides an environment similar to what the user would expect had 
the user logged in directly." and login.c is using the PATH from /etc/environment.

This will fix:
https://bugs.launchpad.net/ubuntu/+source/shadow/+bug/984390

### DIFF
--- a/src/su.c
+++ b/src/su.c
@@ -913,27 +913,7 @@ static void set_environment (struct passwd *pw)
 		addenv ("IFS= \t\n", NULL);	/* ... instead, set a safe IFS */
 	}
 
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks for the patch.  Looks good.

PR #128, opened 2018-09-13
Flush sssd caches in addition to nscd caches

jhrozek:
Some distributions, notably Fedora, have the following order of nsswitch
modules by default:
    passwd: sss files
    group:  sss files

The advantage of serving local users through SSSD is that the nss_sss
module has a fast mmapped-cache that speeds up NSS lookups compared to
accessing the disk an opening the files on each NSS request.

Traditionally, this has been done with the help of nscd, but using nscd
in parallel with sssd is cumbersome, as both SSSD and nscd use their own
independent caching, so using nscd in setups where sssd is also serving
users from some remote domain (LDAP, AD, ...) can result in a bit of
unpredictability.

More details about why Fedora chose to use sss before files can be found
on e.g.:
    https://fedoraproject.org//wiki/Changes/SSSDCacheForLocalUsers
or:
    https://docs.pagure.org/SSSD.sssd/design_pages/files_provider.html

Now, even though sssd watches the passwd and group files with the help
of inotify, there can still be a small window where someone requests a
user or a group, finds that it doesn't exist, adds the entry and checks
again. Without some support in shadow-utils that would explicitly drop
the sssd caches, the inotify watch can fire a little late, so a
combination of commands like this:
    getent passwd user || useradd user; getent passwd user
can result in the second getent passwd not finding the newly added user
as the racy behaviour might still return the cached negative hit from
the first getent passwd.

This patch more or less copies the already existing support that
shadow-utils had for dropping nscd caches, except using the "sss_cache"
tool that sssd ships.

It is a minimal patch; I have more or less just copied the nscd module.
I'm not sure if a better approach would be to e.g. instead of adding
sssd_flush_cache() to where nscd_flush_cache() is called to maybe add a
generic flush_caches call that would call nscd or sssd functions as
appropriate.

I'm also not sure if I should add any tests. I thought that submitting a
PR and then discussing it further is probably a better idea.

Thank you for any comments.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -280,6 +280,9 @@ AC_ARG_WITH(sha-crypt,
 AC_ARG_WITH(nscd,
 	[AC_HELP_STRING([--with-nscd], [enable support for nscd @<:@default=yes@:>@])],
 	[with_nscd=$withval], [with_nscd=yes])
+AC_ARG_WITH(sssd,
+	[AC_HELP_STRING([--with-sssd], [enable support for flushing sssd caches @<:@default=yes@:>@])],
+	[with_sssd=$withval], [with_sssd=yes])
 AC_ARG_WITH(group-name-max-length,
 	[AC_HELP_STRING([--with-group-name-max-length], [set max group name length @<:@default=16@:>@])],
 	[with_group_name_max_length=$withval], [with_group_name_max_length=yes])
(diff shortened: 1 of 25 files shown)


hallyn:
Hi @jhrozek ,

the code itself looks fine to me.  Is this patch in use in Fedora today?

jhrozek:
Thank you for the reply. This patch itself is not used in Fedora. To be absolutely honest, I wanted to create this patch when we introduced the "sss passwd" order in Fedora, but then forgot -- and was only reminded when some of the SSSD tests for this feature started failing with race conditions on very slow test VMs.

I can open a bug in Fedora against shadow-utils so the patch can also be discussed with the maintainer.

jhrozek:
Hmm, I guess I can also @mention @t8m who maintains the Fedora package to see what he thinks.

t8m:
I have just one question - can the spawned cache flush command hang for a long time, or is that impossible or highly unlikely?

t8m:
So I have no problem with applying this downstream in Fedora to test it. I'll do that soon.

jhrozek:
@hallyn since @t8m agreed to merge this patch to fedora, can we merge it upstream? Or should it be the other way around? :-)

hallyn:
Hi,

I was asking about it being in Fedora just to know how well tested it was.

Let me get a test build going here, I'll merge barring any blockers.

PR #132, opened 2018-10-08
newuidmap/newgidmap: do not require CAP_SYS_ADMIN in the parent user namespace

giuseppe:
More details about the issue: https://github.com/genuinetools/img/issues/170


AkihiroSuda:
seems related to https://github.com/torvalds/linux/commit/41c21e351e79004dbb4efa4bc14a53a7e0af38c5

cc @amluto

rhatdan:
@giuseppe Great find.  Does this for CAP_SETPCAP to be added to newuidmap and newgidmap?  Could these utils be run with filecaps rather then having to be setuid?

rhatdan:
@t8m PTAL

rhatdan:
@vbatts FYI

vbatts:
Wait. I'm already having an updated shadow build that has a newxidmap. We just talked about this

giuseppe:
@rhatdan I've not tried but I think filecaps should work as well

vbatts:
oh oh good. I was seeing the forked shadow, but that is a narrow bypass for this CAP issue.

hallyn:
Hi,

can you give a concise example of a way to trigger this?

The patch is mostly fine, except build should absolutely not fail if libcap is not available.

hallyn:
Hey @brauner , just wondering, is this something you've already thought through?

brauner:
So, the I would really like to see reproducer for this scenario to better reason about it. :)

AkihiroSuda:
> misleading because the setuid is set on them.

setuid is set but CAP_SYS_ADMIN is dropped by default in Docker/Kubernetes.
Without enabling CAP_SYS_ADMIN explicitly, newuidmap fails with `newuidmap: write to uid_map failed: Operation not permitted` https://github.com/genuinetools/img/issues/170

hallyn:
Or - you're running these builds, unprivileged, inside a docker container?  Then to run unprivileged you need newuidmap to setup a userns for you?  But the dropped CAP_SYS_ADMIN is an external limitation, not something you are doing yourself?

AkihiroSuda:
Yes

hallyn:
Or - you're running these builds, unprivileged, inside a docker container?  Then to run unprivileged you need newuidmap to setup a userns for you?  But the dropped CAP_SYS_ADMIN is an external limitation, not something you are doing yourself?

If I'm reading this right, then would somehing like the following bit work instead? : 

```
#ifdef LIBCAP
   // define have_cap_sys_admin()

   if (!have_cap_sys_admin() && pw->pw_uid != geteuid()) {
      // perhaps do sanity check that only pw->pw_uid is being mapped?
      if (seteuid(pw->pw_uid) != 0) die();
   }
#endif 
```

(making sure to not fail the build in configure.ac if libcap is not available)

hallyn:
(Sorry for the repeat there - i spent too long tweaking the pseudocode in flaky web browser text editor :)

brauner:
> setuid is set but CAP_SYS_ADMIN is dropped by default in Docker/Kubernetes.
Without enabling CAP_SYS_ADMIN explicitly, newuidmap fails with newuidmap: write to uid_map failed: Operation not permitted genuinetools/img#170

This is confusing:
- Where is CAP_SYS_ADMIN dropped?
   - In the target user namespace `new*idmap` is supposed to write a mapping to?
   - In the user namespace `new*idmap` is called from

I think it is the latter in which case see above: The analysis is wrong. Not `CAP_SYS_ADMIN` is required in the userns `new*idmap` is called from. It never was. `CAP_SETUID` and `CAP_SETGID` are.

brauner:
That is in the generic case. I you only map your own uid onto itself or onto nsuid 0 you don't even need `CAP_SET*ID`.

brauner:
What permissions does your regular `new*idmap` binary have in this example aka does it have fcaps? Because if they have fcaps this wouldn't be an issue at all.

brauner:
Also, the commit messages are wrong. There is no CAP_SYS_ADMIN requirement in the parent userns. Your example even shows that you're adding CAP_SET*ID.

brauner:
Please show:
```
ls -al /usr/bin/newuidmap
getcap /usr/bin/newuidmap
```

giuseppe:
> What permissions does your regular `new*idmap` binary have in this example aka does it have fcaps? Because if they have fcaps this wouldn't be an issue at all.

it is a setuid binary owned by root.  The newuidmap process cannot gain CAP_SYS_ADMIN since it is not in the bounding set, so the filecap wouldn't help as well.

giuseppe:
> ```
> ls -al /usr/bin/newuidmap
> getcap /usr/bin/newuidmap
> ```

[root@cbdbfe4f741f /]# ls -al /usr/bin/newuidmap
-rwsrwsr-x. 1 root root 155752 Oct 19 10:56 /usr/bin/newuidmap
[root@cbdbfe4f741f /]# getcap /usr/bin/newuidmap
[root@cbdbfe4f741f /]#

brauner:
```
podman run -rm -ti --cap-add SETUID --cap-add SETGID -v /etc:/etc:ro --user $(id -u):$(id -g) fedora bash
```
In that container what does `/proc/self/uid_map` show and what capabilities are dropped and what are kept?

giuseppe:
> In that container what does `/proc/self/uid_map` show and what capabilities are dropped and what are kept?
```console
bash-4.4$ cat /proc/self/uid_map
         0          0 4294967295
bash-4.4$ getpcaps 1
Capabilities for `1': = cap_setgid,cap_setuid+eip
```endif

**Comment by giuseppe** [comment]
yes, file capabilities work as well

**Comment by brauner** [comment]
So, if we go this way, then instead of duplicating this code in two places wouldn't it be better if we did something like this? And I think a check for `sys/capability.h` was missing too:
```c
---
 configure.ac        |  2 +-
 libmisc/idmapping.c | 59 +++++++++++++++++++++++++++++++++++++++++++--
 libmisc/idmapping.h |  3 ++-
 src/newgidmap.c     |  3 ++-
 src/newuidmap.c     |  6 ++---
 5 files changed, 65 insertions(+), 8 deletions(-)

diff --git a/configure.ac b/configure.ac
index 10ad70cf..0d2fa6d9 100644
--- a/configure.ac
+++ b/configure.ac
@@ -34,7 +34,7 @@ AC_HEADER_STDBOOL
 
 AC_CHECK_HEADERS(errno.h fcntl.h limits.h unistd.h sys/time.h utmp.h \
 	utmpx.h termios.h termio.h sgtty.h sys/ioctl.h syslog.h paths.h \
-	utime.h ulimit.h sys/resource.h gshadow.h lastlog.h \
+	utime.h ulimit.h sys/resource.h sys/capability.h gshadow.h lastlog.h \
 	locale.h rpc/key_prot.h netdb.h acl/libacl.h attr/libattr.h \
 	attr/error_context.h)
 
diff --git a/libmisc/idmapping.c b/libmisc/idmapping.c
index 20f2d9c7..fdba0864 100644
--- a/libmisc/idmapping.c
+++ b/libmisc/idmapping.c
@@ -34,6 +34,10 @@
 #include <limits.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <sys/prctl.h>
+#if HAVE_SYS_CAPABILITY_H
+#include <sys/capability.h>
+#endif
 #include "prototypes.h"
 #include "idmapping.h"
 
@@ -119,15 +123,19 @@ struct map_range *get_map_ranges(int ranges, int argc, char **argv)
  */
 #define ULONG_DIGITS ((((sizeof(unsigned long) * CHAR_BIT) + 9)/10)*3)
 
-
 void write_mapping(int proc_dir_fd, int ranges, struct map_range *mappings,
-	const char *map_file)
+		   const char *map_file, uid_t uid, gid_t gid)
 {
 	int idx;
 	struct map_range *mapping;
 	size_t bufsize;
 	char *buf, *pos;
 	int fd;
+#if HAVE_SYS_CAPABILITY_H
+	bool uid_map;
+	struct __user_cap_header_struct hdr = {_LINUX_CAPABILITY_VERSION_3, 0};
+	struct __user_cap_data_struct data[2] = {{0}};
+#endif
 
 	bufsize = ranges * ((ULONG_DIGITS  + 1) * 3);
 	pos = buf = xmalloc(bufsize);
@@ -148,6 +156,53 @@ void write_mapping(int proc_dir_fd, int ranges, struct map_range *mappings,
 		pos += written;
 	}
 
+#if HAVE_SYS_CAPABILITY_H
+	if (capget(&hdr, data) < 0) {
+		fprintf(stderr, _("%s: Could not get capabilities\n"), Prog);
+		exit(EXIT_FAILURE);
+	}
+
+	if (strcmp(map_file, "uid_map") == 0) {
+		uid_map = true;
+	} else if (strcmp(map_file, "gid_map") == 0) {
+		uid_map = false;
+	} else {
+		fprintf(stderr, _("%s: Invalid map file %s specified\n"), Prog, map_file);
+		exit(EXIT_FAILURE);
+	}
+
+	if (!(data[0].effective & CAP_TO_MASK(CAP_SYS_ADMIN)) &&
+	    ((uid_map ? uid : gid) != (uid_map ? geteuid() : getegid()))) {
+		int ret;
+
+		memset(data, 0, sizeof(data));
+
+		if (prctl(PR_SET_KEEPCAPS, 1, 0, 0, 0) < 0) {
+			fprintf(stderr, _("%s: Could not prctl(PR_SET_KEEPCAPS)\n"), Prog);
+			exit(EXIT_FAILURE);
+		}
+
+		if (uid_map) {
+			data[0].permitted |= CAP_TO_MASK(CAP_SETUID);
+			ret = seteuid(uid);
+		} else {
+			data[0].permitted |= CAP_TO_MASK(CAP_SETGID);
+			ret = setegid(gid);
+		}
+		if (ret < 0) {
+			fprintf(stderr, _("%s: Could not seteid to %d\n"), Prog,
+				uid_map ? uid : gid);
+			exit(EXIT_FAILURE);
+		}
+
+		data[0].effective = data[0].permitted;
+		if (capset(&hdr, data) < 0) {
+			fprintf(stderr, _("%s: Could not set caps\n"), Prog);
+			exit(EXIT_FAILURE);
+		}
+	}
+#endif
+
 	/* Write the mapping to the mapping file */
 	fd = openat(proc_dir_fd, map_file, O_WRONLY);
 	if (fd < 0) {
diff --git a/libmisc/idmapping.h b/libmisc/idmapping.h
index 58d000f2..b7722437 100644
--- a/libmisc/idmapping.h
+++ b/libmisc/idmapping.h
@@ -38,7 +38,8 @@ struct map_range {
 
 extern struct map_range *get_map_ranges(int ranges, int argc, char **argv);
 extern void write_mapping(int proc_dir_fd, int ranges,
-	struct map_range *mappings, const char *map_file);
+			  struct map_range *mappings, const char *map_file,
+			  uid_t uid, gid_t gid);
 
 #endif /* _ID_MAPPING_H_ */
 
diff --git a/src/newgidmap.c b/src/newgidmap.c
index 59a2e75c..c5c71db7 100644
--- a/src/newgidmap.c
+++ b/src/newgidmap.c
@@ -250,7 +250,8 @@ int main(int argc, char **argv)
 	verify_ranges(pw, ranges, mappings, &allow_setgroups);
 
 	write_setgroups(proc_dir_fd, allow_setgroups);
-	write_mapping(proc_dir_fd, ranges, mappings, "gid_map");
+	write_mapping(proc_dir_fd, ranges, mappings, "gid_map", pw->pw_uid,
+		      pw->pw_gid);
 	sub_gid_close();
 
 	return EXIT_SUCCESS;
diff --git a/src/newuidmap.c b/src/newuidmap.c
index 1ba25e7a..5376cb3a 100644
--- a/src/newuidmap.c
+++ b/src/newuidmap.c
@@ -145,7 +145,7 @@ int main(int argc, char **argv)
 				(unsigned long) getuid ()));
 		return EXIT_FAILURE;
 	}
-	
+
 	/* Get the effective uid and effective gid of the target process */
 	if (fstat(proc_dir_fd, &st) < 0) {
 		fprintf(stderr, _("%s: Could not stat directory for target %u\n"),
@@ -178,8 +178,8 @@ int main(int argc, char **argv)
 		usage();
 
 	verify_ranges(pw, ranges, mappings);
-
-	write_mapping(proc_dir_fd, ranges, mappings, "uid_map");
+	write_mapping(proc_dir_fd, ranges, mappings, "uid_map", pw->pw_uid,
+		      pw->pw_gid);
 	sub_uid_close();
 
 	return EXIT_SUCCESS;
-- 
2.19.1

```

giuseppe:
thanks for the suggestion, I've included it in the new revision.  I've done only few things differently:

- I've added the check for sys/capability.h in addition to checking for `-l cap`.
- use `seteuid` also when writing `gid_map`.


I've tested it locally

hallyn:
Why are you setting the dumpable flag?

hallyn:
Why are you setting the dumpable flag?

hallyn:
Why are you setting the dumpable flag?

hallyn:
Why are you setting the dumpable flag?

hallyn:
Why are you setting the dumpable flag?

hallyn:
Why are you setting the dumpable flag?

I could see this being a problem in sites where they

hallyn:
Why are you setting the dumpable flag?

I could see this being a problem in sites where they try to hide allocations...

configure.ac
```diff
@@ -358,6 +358,15 @@ AC_SUBST(LIBCRYPT)
...
 	[AC_MSG_ERROR([crypt() not found])])
 
+AC_SUBST(LIBCAP)
+
+AC_CHECK_HEADERS(sys/capability.h,
+	[AC_CHECK_LIB(cap, capget, [
+		LIBCAP=-lcap
+		AC_DEFINE(HAVE_LIBCAP, 1, [Defined if you have libcap.])],
+		[LIBCAP=""])],
+	[LIBCAP=""])
```
brauner: Hm, why do you even check for libcap if all you use are the two raw syscalls `cap{g,s}et()`? There's no need to link `new{g,u}idmap` against `libcap` imho.

brauner: Hm, why do you even check for libcap if all you use are the two raw syscalls `cap{g,s}et()`? There's no need to link `new{g,u}idmap` against `libcap` imho.

hallyn:
Why are you setting the dumpable flag?

hallyn:
Gah - sorry about that.

github was wonky lastnight.

libmisc/idmapping.c
```diff
@@ -121,17 +125,56 @@ struct map_range *get_map_ranges(int ranges, int argc, char **argv)
...
+		if (prctl(PR_SET_KEEPCAPS, 1, 0, 0, 0) < 0) {
+			fprintf(stderr, _("%s: Could not prctl(PR_SET_KEEPCAPS)\n"), Prog);
+			exit(EXIT_FAILURE);
+		}
+		if (seteuid(uid) < 0) {
+			fprintf(stderr, _("%s: Could not seteuid to %d\n"), Prog, uid);
+			exit(EXIT_FAILURE);
+		}
+
+		data[0].permitted |= CAP_TO_MASK(uid_map ? CAP_SETUID : CAP_SETGID);
```
brauner: Please clear all other capabilities. You really only need `CAP_SET*ID` and set it in effective too otherwise this will not work. :)

ebiederm:
I am trying to undetstand what is being proposed and what is going on here.

- newuidmap is required when the caller does not have CAPSETID.  It limits the caller to just what is allowed in /etc/subuid.

- Making newuidmap setcap instead of setuid root is reasonable, and takes no code changes and should just work.  The downside is it is called as someone other than the creator of the user namespace it will fail.

- Why would you call newuidmap with a set of credentials different from how the namespace was created?

- If you are already changing your uid in the caller of newuidmap than the caller should be the one making the changes to uid.

- newuidmap should be as simple as absolutely possible because it is designed to be a setuid root executable.   It should have no bells and no whistles because every extra line of code is a possible bug which could be a possible security hole.

So I see a great argument for making newuidmap use file capabilities to gain SETUID and newgidmap to use file capabilities to gain SETGID.   Beyond that everything appears to be the callers problem.  

I don't see a justification for unsharing a user namespace as one euid and calling newuidmap with another euid.  As the caller is doing both then the caller can take care of whatever needs to be take care of.

rhatdan:
@giuseppe You are saying that this is needed for when you run podman inside of a container that does not have access to SYS_ADMIN privs,  But does have SETUID/SETGID.   Would this also happen if a user account dropped SYS_ADMIN from the bounding set and then attempted to run podman?

giuseppe:
@rhatdan without CAP_SYS_ADMIN podman won't work, both when running as root or as rootless with CAP_SYS_ADMIN gained in a new user namespace

ebiederm:
@giuseppe  Why does it make sense to change the code of newuidmap and newgidmap to support this environment where setuid does not grant CAP_SYS_ADMIN?  Why not change the spec file or Makefile to install newuidmap and newgidmap with file capabilities?

It seems like changing how permissions are granted would be the simpler and much less risky change.

giuseppe:
> @giuseppe Why does it make sense to change the code of newuidmap and newgidmap to support this environment where setuid does not grant CAP_SYS_ADMIN? Why not change the spec file or Makefile to install newuidmap and newgidmap with file capabilities?
> 
> It seems like changing how permissions are granted would be the simpler and much less risky change.

I am completely fine with using file capabilities.

I've used the seteuid approach as all other programs that need capabilities are installed as suid binaries and changing newuidmap/newgidmap to use file caps seemed to me like a bigger change than a slightly different code path that is used only when CAP_SYS_ADMIN is not available (so in an already "safe" environment).

I've opened another PR where file capabilities are used so we can more easily compare the two options: https://github.com/shadow-maint/shadow/pull/136

ebiederm:
If make the goal here to always make the code in newuidmap and newgidmap always run with the minimal set of capabilities and uids.  As if the utilities were exclusively using file capabilities I think there are security improvements to be had.  Something like this (untested):

[0001-newuidmap-newgidmap-Only-run-with-the-minimal-necess.txt](https://github.com/shadow-maint/shadow/files/2511375/0001-newuidmap-newgidmap-Only-run-with-the-minimal-necess.txt)

hallyn:
@ebiederm part of me feels like "the minimal amount of work we do, the better", but please feel free to post that as a PR.  note that it can't be literally what you posted, since we cannot assume that libcap is available, but I can do the autoconf tweaking if you prefer.

I'm closing this request, we'll go with one of the filecaps PRs instead.

Thanks, everyone.

AkihiroSuda:
As mentioned in https://github.com/shadow-maint/shadow/pull/136 , file capabilities does not work well with user namespaces for older kernels.

@ebiederm Could you open PR?

hallyn:
That may be, but remember all this hubub is abuot a pretty silly context - where you do not have cap_sys_admin but have cap_setuid.  So now for nsfscaps to come into play, we're talking about a case where you are in a user namespace, run docker in there, and now want to run unprivileged builds inside that?

ebiederm:
@hallyn @AkihiroSuda I am going to pass at opening a PR at this time.  I don't currently care enough to take the code I have posted any farther.

I just wanted to point out that it is possible and without conditionals, and with just the capset system call that glibc supports to lockdown newuidmap and newgidmap and in so doing happen to fix this issue.

giuseppe:
@hallyn since file caps are not going to be the default anyway for a while, could we please reconsider this patch for now?

hallyn:
What about verifying, along with

+       if (!(data[0].effective & CAP_TO_MASK(CAP_SYS_ADMIN)) &&
+           uid != geteuid()) {

that ruid is either 0 or uid ?  Otherwise if two users somehow share a uid delegation, I think this would let them mess with each other's namespaces.

giuseppe:
> Uhm... I was about to send a proper patch for this.

was any comment left unaddressed code wise?  I don't see any functional difference in the other PR, in facts the new code block seems to duplicate what is already done below it and doesn't build when `<sys/capability.h>` is present

brauner:
Forgot, to push over my missing rebase. The functional changes should be pretty obvious though:
- unconditionally drop all unneeded caps
- check whether we can safely seteuid()
- allocate memory after we have done all capability handling

Further simplifcations:
- do not use a bool use a cap and pass the cap directly instead of needing to check an intermediate bool variable
- move everything additional under one single ifdefine

PR #133, opened 2018-10-10
Fix some issues found in Coverity scan.

t8m:
This fixes some fairly trivial issues found in Coverity scan review.

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -343,7 +343,7 @@ static void fail_exit (int code)
 static void get_defaults (void)
 {
 	FILE *fp;
-	char* default_file = USER_DEFAULTS_FILE;
+	char *default_file = USER_DEFAULTS_FILE;
 	char buf[1024];
 	char *cp;
 
@@ -353,6 +353,8 @@ static void get_defaults (void)
 
 		len = strlen(prefix) + strlen(USER_DEFAULTS_FILE) + 2;
 		default_file = malloc(len);
+                if (default_file == NULL)
+                       return;
 		wlen = snprintf(default_file, len, "%s/%s", prefix, USER_DEFAULTS_FILE);
 		assert (wlen == (int) len -1);
 	}
@@ -363,7 +365,7 @@ static void get_defaults (void)
 
 	fp = fopen (default_file, "r");
 	if (NULL == fp) {
-		return;
+		goto getdef_err;
 	}
 
 	/*
@@ -474,7 +476,7 @@ static void get_defaults (void)
 		}
 	}
 	(void) fclose (fp);
-
+     getdef_err:
 	if(prefix[0]) {
 		free(default_file);
 	}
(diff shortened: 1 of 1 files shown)


lib/commonio.c
```diff
@@ -388,12 +388,10 @@ int commonio_lock_nowait (struct commonio_db *db, bool log)
 	lock_file_len = strlen(db->filename) + 6; /* sizeof ".lock" */
 	file = (char*)malloc(file_len);
 	if(file == NULL) {
-		err = ENOMEM;
```
hallyn: Why are you changing this to return 0?

lib/commonio.c
```diff
@@ -388,12 +388,10 @@ int commonio_lock_nowait (struct commonio_db *db, bool log)
...
 		goto cleanup_ENOMEM;
 	}
 	lock = (char*)malloc(lock_file_len);
 	if(lock == NULL) {
-		err = ENOMEM;
```
hallyn: same here.

lib/spawn.c
```diff
@@ -69,7 +69,7 @@ int run_command (const char *cmd, const char *argv[],
 	do {
 		wpid = waitpid (pid, status, 0);
 	} while (   ((pid_t)-1 == wpid && errno == EINTR)
-	         || (wpid != pid));
+	         || ((pid_t)-1 != wpid && wpid != pid));
```
hallyn: Since pid cannot be -1 here (we checked for that above), the extra check for wpid != -1 seems redundant.

libmisc/console.c
```diff
@@ -50,7 +50,7 @@ static bool is_listed (const char *cfgin, const char *tty, bool def);
 static bool is_listed (const char *cfgin, const char *tty, bool def)
 {
 	FILE *fp;
-	char buf[200], *s;
+	char buf[1024], *s;
```
hallyn: Why increase the size?

libmisc/console.c
```diff
@@ -70,7 +70,8 @@ static bool is_listed (const char *cfgin, const char *tty, bool def)
 
 	if (*cons != '/') {
 		char *pbuf;
-		strcpy (buf, cons);
+		strncpy (buf, cons, sizeof (buf));
```
hallyn: Agreed on principle, although I don't believe overflow can happen in practice just bc we always pass in CONSOLE.
So I definately want this, thanks.

lib/commonio.c
t8m: The caller (commonio_lock) expects zero return as error and non-zero as success. So this is a bug fix.

lib/commonio.c
t8m: As above

lib/spawn.c
t8m: You're right that according to the waitpid manpage the wpid == -1 && errno != EINTR condition cannot happen because the other possible errors are ECHILD and EINVAL which should not happen due to the parameters used and no SIGCHLD set to SIG_IGN. But if there is another undocumented permanent waitpid() error, without this added condition the loop will hang forever as wpid == -1 and pid !=-1 so wpid != pid is true and the loop continues forever.

libmisc/console.c
t8m: Well, the console path can be theoretically longer than 200 characters. I know I am replacing one arbitrary number with another but 200 characters seems to me to be quite small limit for paths.

libmisc/console.c
t8m: The overflow can happen. It is not a security issue because the login.defs file is controlled by sysadmin, but still possible crashes are uggly.

PR #136, opened 2018-10-24
newuidmap/newgidmap: install with file capabilities

giuseppe:
do not install newuidmap/newgidmap as suid binaries.  Running these
tools with the same euid as the owner of the user namespace to
configure requires only CAP_SETUID and CAP_SETGID instead of requiring
CAP_SYS_ADMIN when it is installed as a suid binary.

Alternative solution for https://github.com/shadow-maint/shadow/pull/132

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>


src/Makefile.am
```diff
@@ -138,3 +135,7 @@ if WITH_TCB
 		chmod $(sgidperms) $(DESTDIR)$(ubindir)/$$i; \
 	done
 endif
+if ENABLE_SUBIDS
```
brauner: Make sure that this is opt-in. So that the distros can decide whether they want these to be `suid` binaries or `fscaps` binaries. This is important when wanting to build `shadow` in user namespaces.

AkihiroSuda:
> Alternative solution for #132

I'm not sure this can be alternative, because setting file capabilities in user namespace is not supported for Kernel < 4.14

src/Makefile.am
giuseppe: I've added a `--with-fcaps` switch to configure

giuseppe:
> > [](https://github.com/shadow-maint/shadow/pull/132)
> 
> I'm not sure this can be alternative, because setting file capabilities in user namespace is not supported for Kernel < 4.14

thinking more of it: the two solutions are not really complementary to each other, they can be both used and they address the issue in a different way.

brauner:
> thinking more of it: the two solutions are not really complementary to each other, they can be both used and they address the issue in a different way.

When compiling it needs to be `suid` `xor` `fscaps`.

ebiederm:
@AkihiroSuda 
> > Alternative solution for #132
> 
> I'm not sure this can be alternative, because setting file capabilities in user namespace is not supported for Kernel < 4.14

Where does this matter?

I presume if you are playing with a container and you are dropping capabilities you are doing so because you have real root inside the container.  As such what you can do inside a user namespace should not be relevant.

Further file capabilities set outside of a user namespace continue to work inside the user namespaces on any kernel that supports user namespaces.  The binaries that use file capabilities are just then limited to having those file capabilities inside the user namespace in which they run.

hallyn:
I opened PR #137 before I saw this, sorry.

Please update configure.ac to ensure that setcap is installed before doing everything else.

Then assuming you've test-built this with and without filecaps flag, I'll merge this one.

AkihiroSuda:
>> I'm not sure this can be alternative, because setting file capabilities in user namespace is not supported for Kernel < 4.14
>
> Where does this matter?

It matters 
* when building a rootfs image (especially for containers but not limited to) with newuidmap/newgidmap binaries in a user namespace. e.g. building [an image of `img`](https://github.com/genuinetools/img/blob/master/Dockerfile) using `img` itself.
* when unpacking a rootfs archive with newuidmap/newgidmap binaries in a user namespace

> I presume if you are playing with a container and you are dropping capabilities you are doing so because you have real root inside the container. 

No

AkihiroSuda:
I'm also wondering there might be some container builder & runtime implementations that does not support filecaps?

cyphar:
> I'm also wondering there might be some container builder & runtime implementations that does not support filecaps?

In general they would be non-OCI compliant or just not a useful runtime (most distributions use fscaps for `/usr/bin/ping`). But AUFS would have some issues since it has "interesting" xattr support -- but AUFS has many other issues and I don't think it makes much sense to accommodate it given that very few people use it nowadays.

giuseppe:
@brauner @hallyn should we default to installing with file caps?  I am not really sure what the default should be :)

brauner:
@giuseppe, default to `setuid` for now and in a year when namespaced file caps have found even more adoption and the new LTS kernel is out we make an announcement and make fscaps the default. We can announce in this release that we plan on switching over `new*idmap` to fscaps in the next release.

giuseppe:
@brauner That makes sense, thanks for the clarification.  Could we reconsider #132 in the meanwhile since newuidmap/newgidmap will still be used as suid binaries for now?

PR #138, opened 2018-10-27
new{g,u}idmap: align setuid and fscaps behavior

brauner:
Commit 1ecca84 ("new[ug]idmap: not require CAP_SYS_ADMIN in the parent userNS")
does contain a wrong commit message, is lacking an explanation of the
issue, misses some simplifications and hardening features. This commit
tries to rectify this.

In (crazy) environment where all capabilities are dropped from the
capability bounding set apart from CAP_SET{G,U}ID setuid- and
fscaps-based new{g,u}idmap binaries behave differently when writing
complex mappings for an unprivileged user:

1. newuidmap is setuid

unshare -U sleep infinity &
newuidmap $? 0 100000 65536

First file_ns_capable(file, ns, CAP_SYS_ADMIN) is hit. This calls into
cap_capable() and hits the loop

for (;;) {
        /* Do we have the necessary capabilities? */
        if (ns == cred->user_ns)
                return cap_raised(cred->cap_effective, cap) ? 0 : -EPERM;

        /*
         * If we're already at a lower level than we're looking for,
         * we're done searching.
         */
        if (ns->level <= cred->user_ns->level)
                return -EPERM;

        /*
         * The owner of the user namespace in the parent of the
         * user namespace has all caps.
        */
        if ((ns->parent == cred->user_ns) && uid_eq(ns->owner, cred->euid))
                return 0;

        /*
         * If you have a capability in a parent user ns, then you have
         * it over all children user namespaces as well.
        */
        ns = ns->parent;
}

The first check fails and falls through to the end of the loop and
retrieves the parent user namespace and checks whether CAP_SYS_ADMIN is
available there which isn't.

2. newuidmap has CAP_SETUID as fscaps set

unshare -U sleep infinity &
newuidmap $? 0 100000 65536

The first file_ns_capable() check for CAP_SYS_ADMIN is passed since the
euid has not been changed:

if ((ns->parent == cred->user_ns) && uid_eq(ns->owner, cred->euid))
        return 0;

Now new_idmap_permitted() is hit which calls ns_capable(ns->parent,
CAP_SET{G,U}ID). This check passes since CAP_SET{G,U}ID is available in
the parent user namespace.
Now file_ns_capable(file, ns->parent, CAP_SETUID) is hit and the
cap_capable() loop (see above) is entered again. This passes

if (ns == cred->user_ns)
        return cap_raised(cred->cap_effective, cap) ? 0 : -EPERM;

since CAP_SET{G,U}ID is available in the parent user namespace. Now the
mapping can be written.

There is no need for this descrepancy between setuid and fscaps based
new{g,u}idmap binaries. The solution is to do a
seteuid() back to the unprivileged uid and PR_SET_KEEPCAPS to keep
CAP_SET{G,U}ID. The seteuid() will cause the
file_ns_capable(file, ns, CAP_SYS_ADMIN) check to pass and the
PR_SET_KEEPCAPS for CAP_SET{G,U}ID will cause the CAP_SET{G,U}ID to
pass.

Fixes: 1ecca84 ("new[ug]idmap: not require CAP_SYS_ADMIN in the parent userNS")
Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

### DIFF
--- a/libmisc/idmapping.c
+++ b/libmisc/idmapping.c
@@ -38,7 +38,7 @@
 #include "idmapping.h"
 #include <sys/prctl.h>
 #if HAVE_SYS_CAPABILITY_H
-# include <sys/capability.h>
+#include <sys/capability.h>
 #endif
 
 struct map_range *get_map_ranges(int ranges, int argc, char **argv)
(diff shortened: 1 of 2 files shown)


brauner:
That was the intended patch. Let me fix this since we merged a strange version of that patch.

brauner:
@hallyn, please review.

giuseppe:
FWIW, I've tested the patch both as suid binary and with fcaps and in both cases it works well in a container where CAP_SYS_ADMIN is not present

brauner:
Thanks, @giuseppe. Sorry, was in a hurry when writing my comments. They may sound more harsh than intended. :)

libmisc/idmapping.c
```diff
@@ -132,49 +132,53 @@ void write_mapping(int proc_dir_fd, int ranges, struct map_range *mappings,
...
-			uid_map = true;
-		} else if (strcmp(map_file, "gid_map") == 0) {
-			uid_map = false;
-		} else {
-			fprintf(stderr, _("%s: Invalid map file %s specified\n"), Prog, map_file);
-			exit(EXIT_FAILURE);
-		}
+
+	/* Align setuid- and fscaps-based new{g,u}idmap behavior. */
+	if (!(data[0].effective & CAP_TO_MASK(CAP_SYS_ADMIN)) && uid != 0 &&
```
hallyn: As long as we're doing all this, can we have a comment at top of fn explaining what exactly 'uid' is, or fix up the variable name (maybe caller_ruid, except that can be wrong when actually called by root on behalf of a user) ?

brauner: done

brauner: I added an extensive comment now too.

libmisc/idmapping.c
```diff
@@ -123,58 +123,76 @@ struct map_range *get_map_ranges(int ranges, int argc, char **argv)
...
-		data[0].effective = data[0].permitted = CAP_TO_MASK(uid_map ? CAP_SETUID : CAP_SETGID);
-		if (capset(&hdr, data) < 0) {
-			fprintf(stderr, _("%s: Could not set caps\n"), Prog);
+		if (seteuid(ruid) < 0) {
+			fprintf(stderr, _("%s: Could not seteuid to %d\n"), Prog, ruid);
 			exit(EXIT_FAILURE);
 		}
 	}
+
+	/* Lockdown new{g,u}idmap by dropping all unneeded capabilities. */
```
hallyn: Doing this unconditionally *should* be safer, but the sendmail capabilities bug has me paranoid that we're actually opening up a new hole by always dropping privs.

But, we'll never make progress if we don't try :)

hallyn:
Thanks!

giuseppe:
>  Sorry, was in a hurry when writing my comments. They may sound more harsh than intended. :)

no problem :-)  And thanks a lot for improving it!

AkihiroSuda:
Unfortunately this commit seems to have broke newuidmap binary with SUID bit: https://github.com/genuinetools/img/issues/191

PR #141, opened 2018-11-20
idmap: always seteuid to the owner of the namespace

giuseppe:
simplify the condition for setting the euid of the process.  Now it is
always set when we are running as root, the issue was introduced with
the commit 52c081b02c4ca4432330ee336a60f6f803431e63

Closes: https://github.com/genuinetools/img/issues/191

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/libmisc/idmapping.c
+++ b/libmisc/idmapping.c
@@ -161,14 +161,8 @@ void write_mapping(int proc_dir_fd, int ranges, struct map_range *mappings,
 		exit(EXIT_FAILURE);
 	}
 
-	if (capget(&hdr, data) < 0) {
-		fprintf(stderr, _("%s: Could not get capabilities\n"), Prog);
-		exit(EXIT_FAILURE);
-	}
-
 	/* Align setuid- and fscaps-based new{g,u}idmap behavior. */
-	if (!(data[0].effective & CAP_TO_MASK(CAP_SYS_ADMIN)) && ruid != 0 &&
-	    ruid == getuid() && ruid != geteuid()) {
+	if (geteuid() != ruid) {
 		if (prctl(PR_SET_KEEPCAPS, 1, 0, 0, 0) < 0) {
 			fprintf(stderr, _("%s: Could not prctl(PR_SET_KEEPCAPS)\n"), Prog);
 			exit(EXIT_FAILURE);


libmisc/idmapping.c
```diff
@@ -161,14 +161,8 @@ void write_mapping(int proc_dir_fd, int ranges, struct map_range *mappings,
...
 
-	if (capget(&hdr, data) < 0) {
-		fprintf(stderr, _("%s: Could not get capabilities\n"), Prog);
-		exit(EXIT_FAILURE);
-	}
-
 	/* Align setuid- and fscaps-based new{g,u}idmap behavior. */
-	if (!(data[0].effective & CAP_TO_MASK(CAP_SYS_ADMIN)) && ruid != 0 &&
-	    ruid == getuid() && ruid != geteuid()) {
+	if (geteuid() == 0) {
```
t8m: Perhaps ruid != geteuid() would be more correct?

giuseppe: yes that works as well.  I've pushed a new version

hallyn:
But and fix confirmed - but I want to think about this a bit more closely.  I may or may not have some questions before I apply.

hallyn:
let's at least keep

```
        if (geteuid() == 0 && geteuid() != ruid) {
```

my concern, which may be a long shot, is that an unpriv user copy /usr/bin/newuidmap to ~/runme and make it setuid to $user, and trick a root user into running it.  Ensuring that euid == 0 at least makes sure that the code was vetted by root at some point, and this still works (at least for setuid - i didn't test with fcaps).

My remaining concern is - why was the check for CAP_SYS_ADMIN added (by Giuseppe Scrivano) in the first place?  Was there a particular threat it was fighting?  I don't always have the most active imagination, so just bc I can't think of a problem situation doesn't mean i trust there isn't one.

Still this definately is broken now, so I will merge this - with my additional geteuid == 0 check - in a bit if i don't hear any complaints.

hallyn:
Merged by hand with that change, thanks.

giuseppe:
> My remaining concern is - why was the check for CAP_SYS_ADMIN added (by Giuseppe Scrivano) in the first place? Was there a particular threat it was fighting? I don't always have the most active imagination, so just bc I can't think of a problem situation doesn't mean i trust there isn't one.

my first patch was addressing only the special case when `CAP_SYS_ADMIN` was not present as I didn't want to change too much of how newuidmap and newgidmap work.  I think the refactored version is cleaner as we don't treat not having `CAP_SYS_ADMIN` as a special case.

hallyn:
Ok - thanks.

PR #142, opened 2018-11-28
login.defs: Add LASTLOG_UID_MAX variable to limit lastlog to small uids.

t8m:
As the large uids are usually provided by remote user identity and
authentication service, which also provide user login tracking,
there is no need to create a huge sparse file for them on every local
machine.

### DIFF
--- a/lib/getdef.c
+++ b/lib/getdef.c
@@ -92,6 +92,7 @@ static struct itemdef def_table[] = {
 	{"GID_MIN", NULL},
 	{"HUSHLOGIN_FILE", NULL},
 	{"KILLCHAR", NULL},
+	{"LASTLOG_UID_MAX", NULL},
 	{"LOGIN_RETRIES", NULL},
 	{"LOGIN_TIMEOUT", NULL},
 	{"LOG_OK_LOGINS", NULL},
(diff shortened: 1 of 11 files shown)


t8m:
If this is accepted, I will add the support also to pam_lastlog.

karelzak:
Cool, I'm going to support this feature i util-linux login.c too.

hallyn:
Seems reasonable to me, thanks.

hallyn:
Well, but let me ask - would it seem more conservative to make the default, if the variable is unspecified, be 4294967295, i.e. so noone's behavior changes by default, then distros/packages can be encouraged to ship 99999 as the default in their default login.defs?

t8m:
It works for me either way. If this conservative approach is preferred by you, please tell me and I will update the patch.

hallyn:
Thanks.

glensc:
this got now merged with `fixup!` commit. should had `git rebase --autosquash -i` or use the github merge as squash feature.

hallyn:
Oh, in general I prefer for the submitter to do the git rebase.  I don't like to mangle other people's submissions, though maybe I should...  It's only been a few hours, I can rebase and force-push if you far prefer.

t8m:
The fixup commits purpose is to be autosquashed, so please do.

PR #143, opened 2018-12-18
usermod: Guard against unsafe change of ownership of home contents

t8m:
In case the home directory is not a real home directory
(owned by the user) but things like / or /var or similar,
it is unsafe to change ownership of home directory content.

The test checks whether the home directory is owned by the
user him/herself, if not no ownership modification of contents
is performed.


hallyn:
Thanks, sorry this took so long.  Looks good.

PR #144, opened 2018-12-21
WIP: Introduce MAX_ID_HOLE_SIZE setting

t8m:
For allowing more efficient new UID/GID allocation we introduce
MAX_ID_HOLE_SIZE setting to allocate new IDs in unused ranges
larger than MAX_ID_HOLE_SIZE.


t8m:
This is WIP and item for discussion. I would like to modify the ID allocation algorithm to avoid skipping large holes to fix https://bugzilla.redhat.com/show_bug.cgi?id=958842. @hallyn what do you think about it?

hallyn:
I'm sorry, I'm having a hard time following.  Can you give an example of how the behavior would change?

t8m:
Basically if a hole between the allocated IDs would be encountered during the search and it would be larger than the MAX_ID_HOLE_SIZE, it would be freely used to allocate a new user/group ID.
Let's say for example we have the allowed system uid range 100-999. Then we have some hardcoded user with uid 111 and no other allocated uid in the system uid range. The current algorithm will allocate a new system user id with uid 110. With the default MAX_ID_HOLE_SIZE it would allocate uid 999 as expected just fine.
The question is whether the same algorithm should apply to regular (non-system) uids. But I do not see why it should not.

t8m:
@hallyn ping, should I proceed with the patch, or do you think this feature would not  be useful?

hallyn:
Going by the bugzilla, I gather the concern is a tradeoff between re-using just-freed uids, and skipping too many unused uids...

I don't really think the feature would be particularly useful, but I've been wrong about this sort of thing before.

Anyone else have any opinions?

hallyn:
If you write the patch, I will seriously consider it. But I'm closing this for now, since I'll feel bad if I ask you to write it and then decide against it :)

Issue #131, opened 2018-10-04
Discussion: subvolume capable filesystems, useradd homedir, hooks


jubalh:
With the rise of popularity of filesystems like btrfs, it's possible that users want their home directory to be on a (btrfs) subvolume.
If I create a new user via `useradd -m`, useradd will create the (standard) home directory for me.
However if I create the subvolume first, useradd detects that the directory already exists and won't adapt the permissions nor copy the skel files.

One could argue that on such a custom setup everything should be done customized by hand:
 * creating subvolume
 * call useradd
 * change permissions
 * copy skel files

Another possibility would be to add some kind of hooking mechanism into useradd. Then a distro could specify any behaviour for that hook. In this case calling btrfs to create a subvolume and then continuing with normal useradd behaviour.

Adding explicitly btrfs (subvolume) support into useradd probably is not a nice solution. The hooking is the one way I could think of to achieve some desired behaviour. Maybe someone else has a better idea?

hallyn:
I like the hooking idea - could also be used to e.g. create an lvm lv.

hallyn:
Well, I'm not sure.  It seems more accessible to just have useradd.c:create_home() check for $somexdgprefix/shadow/hooks/create_home, which it simply execs, passing along the needed info.  Though that info passing might be a source of cves, and might prove problematic when we want to extend it for things like selinux.  But then create_home could really just be a shell script.

jubalh:
https://github.com/shadow-maint/shadow/pull/149 got merged now.

PR #149, opened 2019-01-24
Btrfs subvolume

AdamMajer:
This adds support for using `btrfs` subvolumes as a user home instead of just a plain directory. These subvolumes allow for snapshots to be taken to allow users to rollback changes. A use-case is described at,

https://www.suse.com/c/menu-du-jour-vivaneau-vert-sur-lit-de-legumes-au-beurre-et-supremes-de-pamplemousse-2/

It also implements #131

I've only implemented this as a long option as this is kind of specialized situation. In default case, nothing changes for useradd. For `userdel`, it can now handle subvolumes on `btrfs` instead of just printing an error. `usermod` will also throw an error instead of doing a deep-copy of a subvolume for cross-device home moves, which is better than current attempt at deep-copy and failure.

subvolume creation and deletion is handled by `btrfs` program.

```
 configure.ac        | 18 ++++++++++++++
 lib/prototypes.h    |  8 ++++++
 libmisc/Makefile.am |  5 ++++
 libmisc/btrfs.c     | 94 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 src/useradd.c       | 41 ++++++++++++++++++++++++++++++
 src/userdel.c       | 17 +++++++++++++
 src/usermod.c       |  9 +++++++
 7 files changed, 192 insertions(+)
```

### DIFF
--- a/libmisc/btrfs.c
+++ b/libmisc/btrfs.c
@@ -1,0 +1,94 @@
+#include <linux/btrfs_tree.h>
+#include <linux/magic.h>
+#include <sys/statfs.h>
+
+#include "prototypes.h"
+
+
+static int run_btrfs_subvolume_cmd(const char *subcmd, const char *arg1, const char *arg2)
+{
+	int status = 0;
+	const char *cmd = "/sbin/btrfs";
+	const char *argv[] = {
+		strrchr(cmd, '/'),
+		"subvolume",
+		subcmd,
+		arg1,
+		arg2,
+		NULL
+	};
+
+	if (argv[0] == NULL)
+		argv[0] = cmd;
+	else
+		argv[0] = argv[0] + 1;
+
+	if (access(cmd, X_OK)) {
+		return 1;
+	}
+
+	if (run_command(cmd, argv, NULL, &status))
+		return -1;
+	return status;
+}
+
+
+int btrfs_create_subvolume(const char *path)
+{
+	return run_btrfs_subvolume_cmd("create", path, NULL);
+}
+
+
+int btrfs_remove_subvolume(const char *path)
+{
+	return run_btrfs_subvolume_cmd("delete", "-C", path);
+}
+
+
+/* Adapted from btrfsprogs */
+/*
+ * This intentionally duplicates btrfs_util_is_subvolume_fd() instead of opening
+ * a file descriptor and calling it, because fstat() and fstatfs() don't accept
+ * file descriptors opened with O_PATH on old kernels (before v3.6 and before
+ * v3.12, respectively), but stat() and statfs() can be called on a path that
+ * the user doesn't have read or write permissions to.
+ *
+ * returns:
+ *   1 - btrfs subvolume
+ *   0 - not btrfs subvolume
+ *  -1 - error
+ */
+int btrfs_is_subvolume(const char *path)
+{
+	struct stat st;
+	int ret;
+
+	ret = is_btrfs(path);
+	if (ret <= 0)
+		return ret;
+
+	ret = stat(path, &st);
+	if (ret == -1)
+		return -1;
+
+	if (st.st_ino != BTRFS_FIRST_FREE_OBJECTID || !S_ISDIR(st.st_mode)) {
+		return 0;
+	}
+
+	return 1;
+}
+
+
+/* Adapted from btrfsprogs */
+int is_btrfs(const char *path)
+{
+	struct statfs sfs;
+	int ret;
+
+	ret = statfs(path, &sfs);
+	if (ret == -1)
+		return -1;
+
+	return sfs.f_type == BTRFS_SUPER_MAGIC;
+}
+
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -823,7 +823,9 @@ static void usage (int status)
 	                Prog, Prog, Prog);
 	(void) fputs (_("  -b, --base-dir BASE_DIR       base directory for the home directory of the\n"
 	                "                                new account\n"), usageout);
+#ifdef WITH_BTRFS
 	(void) fputs (_("      --btrfs-subvolume-home    use BTRFS subvolume for home directory\n"), usageout);
+#endif
 	(void) fputs (_("  -c, --comment COMMENT         GECOS field of the new account\n"), usageout);
 	(void) fputs (_("  -d, --home-dir HOME_DIR       home directory of the new account\n"), usageout);
 	(void) fputs (_("  -D, --defaults                print or change default useradd configuration\n"), usageout);
@@ -1104,7 +1106,9 @@ static void process_flags (int argc, char **argv)
 		int c;
 		static struct option long_options[] = {
 			{"base-dir",       required_argument, NULL, 'b'},
+#ifdef WITH_BTRFS
 			{"btrfs-subvolume-home", no_argument, NULL, 200},
+#endif
 			{"comment",        required_argument, NULL, 'c'},
 			{"home-dir",       required_argument, NULL, 'd'},
 			{"defaults",       no_argument,       NULL, 'D'},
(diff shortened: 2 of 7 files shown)


AdamMajer:
One thing that is missing is autotools probably should detect `btrfs` executable location and use that instead of hardcoded `/sbin/btrfs` as it is on openSUSE distributions.

jubalh:
@hallyn @brauner any comment? :)

src/useradd.c
```diff
@@ -2073,6 +2083,37 @@ static void create_home (void)
 			strcat (path, "/");
 			strcat (path, cp);
 			if (access (path, F_OK) != 0) {
+				/* Check if parent directory is BTRFS, fail if requesting
+				   subvolume but no BTRFS. The paths cound be different by the
+				   trailing slash
+				 */
+#if WITH_BTRFS
+				if (subvolflg && (strlen(prefix_user_home) - (int)strlen(path)) <= 1) {
```
brauner: Why do you cast to `int` here?

src/useradd.c
```diff
@@ -2073,6 +2083,37 @@ static void create_home (void)
...
+					char *btrfs_check = strdup(path);
+
+					if (!btrfs_check) {
+						fprintf (stderr,
+						         _("%s: error while duplicating string in BTRFS check %s\n"),
+						         Prog, path);
+						fail_exit (E_HOMEDIR);
+					}
+					btrfs_check[strlen(path) - strlen(cp) - 1] = '\0';
```
brauner: You're calculating `strlen(path)` twice, no? Wouldn't it make more sense to store it in a variable?

libmisc/btrfs.c
```diff
@@ -0,0 +1,94 @@
...
+#include <linux/magic.h>
+#include <sys/statfs.h>
+
+#include "prototypes.h"
+
+
+static int run_btrfs_subvolume_cmd(const char *subcmd, const char *arg1, const char *arg2)
+{
+	int status = 0;
+	const char *cmd = "/sbin/btrfs";
```
brauner: Does that make sense unconditionally? Can't we just pass `btrfs` in case it's not located in `/sbin`.

libmisc/btrfs.c
```diff
@@ -0,0 +1,94 @@
...
+	const char *argv[] = {
+		strrchr(cmd, '/'),
```
brauner: That feels like it shouldn't be done unconditionally. That seems like a good candidate for later bugs. :)

src/useradd.c
jubalh: Probably done by the compiler, no?

libmisc/btrfs.c
AdamMajer: It should be either in /sbin or /usr/sbin. And if in latter, there should be a symlink from /sbin. It should not be in user's PATH. We also need to path for line 26.

Alternatively, the path should be moved to the autoconf scripts.

libmisc/btrfs.c
AdamMajer: This is not unconditional. The condition is on line 21 for "just in case" scenario. Maybe a little bit "fancy" and the condition should be moved up. It should need another variable then.

Let me know if you'd like me to refactor this or it's ok.

src/useradd.c
AdamMajer: This is just an attempt at avoiding unsigned undeflow. I didn't check if this is impossible, but just in case. :crossed_fingers:

src/useradd.c
AdamMajer: yes, but I think it is clearer to do this here instead of adding another variable. We are not going to be saving anything here.

But yes, this can be added if you insist.

src/useradd.c
```diff
@@ -2073,6 +2083,37 @@ static void create_home (void)
...
+					if (is_btrfs(btrfs_check) <= 0) {
+						fprintf (stderr,
+						         _("%s: home directory \"%s\" must be mounted on BTRFS\n"),
```
AdamMajer: I noticed that "home" should probably be removed from the string...

jubalh:
@brauner did you see that @AdamMajer replied to your requests? :)

brauner:
@jubalh, thanks. @hallyn you fine with this?

hallyn:
@brauner - yeah, looks fine at first glance.  I can take a closer look tomorrow, or, it looks like you've reviewed it pretty closely so if you want to merge it I'm happy with that.

libmisc/btrfs.c
```diff
@@ -0,0 +1,94 @@
...
+		"subvolume",
+		subcmd,
+		arg1,
+		arg2,
+		NULL
+	};
+
+	if (argv[0] == NULL)
```
hallyn: If this is NULL at this point then something has gone very wrong, so best to just exit.

libmisc/btrfs.c
```diff
@@ -0,0 +1,94 @@
...
+}
+
+
+int btrfs_remove_subvolume(const char *path)
+{
+	return run_btrfs_subvolume_cmd("delete", "-C", path);
+}
+
+
+/* Adapted from btrfsprogs */
```
hallyn: btrfsprogs is GPL, shadow is bsd style.  So I'm afraid we need a more precise definitino of "Adapted from' :(

It's so trivial it's difficult to imagine a problem, but...

hallyn: (If it's just "inspired by", or "informed by", i.e. it showed how to use the constnats, then that's probably ok)

src/useradd.c
```diff
@@ -2073,7 +2079,35 @@ static void create_home (void)
 			strcat (path, "/");
 			strcat (path, cp);
 			if (access (path, F_OK) != 0) {
-				if (mkdir (path, 0) != 0) {
+				/* Check if parent directory is BTRFS, fail if requesting
```
hallyn: I have three issues with this (besides the existing formatting ARGH)

1. if the homedirs for some reason are something like /home/group1/group2//user1, and all of it is btrfs, then you will create a new btrfs subvolume for each directory that didn't exist.
2. Like Christian said, it seems like btrfs_check handling is awkward.  It's bikeshedding to a degree, though, and what i was going to suggest ends up with more work in the !btrfs case...
3. This is now way too much code for this simple "create if doesn't exist" loop.  The btrfs creation should be in its own function, and for that matter in the existing code there should just be a "mkdir_or_die()" call.

hallyn:
Thanks, so silly as it may seem the licensing question is my biggest concern.  If you or Christian can pursuade me it's no problem, then I don't mind editing your commit myself for the changes I requested.

libmisc/btrfs.c
brauner: Good catch. Serge is raising a good question. *Sigh* licenses.

AdamMajer: This part comes from the libbtrfsutil/subvolume.c which is actually LGPL, but that doesn't really help. I think the most tell-tale copy/paste is the comment itself. The rest is just basically 2 or 3 lines here. Line 74 and line 92.

I'm not sure there really is any alternative way of writing those 2 lines. But I can remove that comment which is copied. Read access is not relevant anyway.

brauner: Hm, I'd like to hear what @stgraber thinks.

stgraber: It does look like something that's sufficiently trivial and pretty much has only a single way of getting implemented that copyright is unlikely to apply. If we want to be extra safe here, it should be reasonably easy to find another piece of code out there, under a suitable license, which implements the same logic.

hallyn: Thanks.  You're all right, there's basically no other way of doing it, so let's take it.

hallyn:
I've pushed a few changes to my branch (bc for some reason it didn't let me push on top of yours) at https://github.com/hallyn/shadow/pull/new/btrfs .  But I still have one more question which I'll ask inline.

hallyn:
Manually merged the branch and pushed, so closing this.  Thank you.

jubalh:
Thanks for merging @hallyn.
Just for the record. This was merged in c1d36a8acb1d5fbe42099a37c8ed6f33e8a648e6 (plus the following commits).

PR #151, opened 2019-01-31
Fix the default mentioned in man page for SUB_UID/GID_COUNT variables.

t8m:
Subject says it all. A trivial fix.


brauner:
Please sign-off your commits next time. I just realized this now. But since this is a trivial change it's ok.

t8m:
OK

PR #156, opened 2019-02-27
Do not flush nscd and sssd cache in read-only mode

### LINKED ISSUE #155: grpck & pwck: nscd / sssd cache are flushed even in read-only mode
In read-only mode ("-r" flag), there is no changes about group and password  files.
A condition is missing before flushing the cache:
* [grpck flush cache](https://github.com/shadow-maint/shadow/blob/d66a1fe069e6727b8098289b732e358461f76627/src/grpck.c#L873)
* [pwck flush cache](https://github.com/shadow-maint/shadow/blob/d66a1fe069e6727b8098289b732e358461f76627/src/pwck.c#L880)

cvuillemez:
Fix #155

signed-off-by: Charlie Vuillemez <cvuillemez@users.noreply.github.com>

### DIFF
--- a/src/grpck.c
+++ b/src/grpck.c
@@ -870,8 +870,10 @@ int main (int argc, char **argv)
 	/* Commit the change in the database if needed */
 	close_files (changed);
 
-	nscd_flush_cache ("group");
-	sssd_flush_cache (SSSD_DB_GROUP);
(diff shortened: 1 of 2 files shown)


hallyn:
Thanks, makes sense.

PR #157, opened 2019-03-25
Do not crash in commonio_close if database FILE not opened.

t8m:
The db->fp can be NULL if commonio_unlock() is called when the
shadow file is opened but did not exist before.

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -925,7 +925,6 @@ static int write_all (const struct commonio_db *db)
 
 
 int commonio_close (struct commonio_db *db)
-	/*@requires notnull db->fp@*/
 {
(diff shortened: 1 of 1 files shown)


lib/commonio.c
```diff
@@ -938,8 +937,10 @@ int commonio_close (struct commonio_db *db)
 	db->isopen = false;
 
 	if (!db->changed || db->readonly) {
-		(void) fclose (db->fp);
-		db->fp = NULL;
+		if (NULL != db->fp) {
```
brauner: Should this rather be `if (!dp->fp)` to be in line with the coding style?

brauner: Cool, thanks!

PR #161, opened 2019-04-25
lib/spawn.c run_command: don't loop forever if waitpid() is returning ECHILD

tabraham:
lib/spawn.c run_command: don't loop forever if waitpid() is returning ECHILD

If SIGCHILD is being ignored, waitpid() will forever error with ECHILD and
this loop with never end, so don't loop if it erros with ECHILD.

### DIFF
--- a/lib/spawn.c
+++ b/lib/spawn.c
@@ -68,6 +68,8 @@ int run_command (const char *cmd, const char *argv[],
 
 	do {
 		wpid = waitpid (pid, status, 0);
+		if ((pid_t)-1 == wpid && errno == ECHILD)
+			break;
 	} while (   ((pid_t)-1 == wpid && errno == EINTR)
 	         || ((pid_t)-1 != wpid && wpid != pid));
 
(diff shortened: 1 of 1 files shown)


hallyn:
Hi,

Thanks.  This does look like a real problem, but I'm not sure your new check is right.  In particular, it will only continue through the waitpid loop if wpid == -1.  If wpid is ever not -1, then it will exit the loop.

This is probably easier to reason about if you put your new check inside the loop, and break, i.e.

```
    do {
        wpid = waidpid(pid, status, 0);
        if (wpid == -1 && errno == ECHILD)
            break;
    } while (...);
```

Or just put the other check inside the do/while as well if you prefer.

tabraham:
You're correct, of course. I'm missing a '!' -- loop while  !((pid_t)-1 == wpid && errno != ECHILD) ...
I do like your  solution better as it's easier to read.

tabraham:
*sigh* No. The logic in that comment was still off. The commit is correct, though.

jubalh:
@hallyn I'm curious, what do you think about updated patch?

PR #163, opened 2019-04-30
build: Port to itstool

jtojnar:
gnome-doc-utils are basically abandoned.

I am not sure this is correct as I am not very familiar with autotools. For example, what does remove the `.xml-config` files? Currently it is failing on

```
make[3]: *** No rule to make target 'man1/chfn.1', needed by 'all-am'.  Stop.
```


hallyn:
Yeah, I can't quite tell why your patch is causing this.  I'm sure it'll be obvious once we find it..

Thanks for the patch.  Indeed looks like it'll be needed.

PR #164, opened 2019-05-02
Use lckpwdf() again if prefix is not set and fix a possible DoS in locking

t8m:
This PR fixes two issues related to the locking of the databases.


t8m:
Ping @hallyn ?

PR #167, opened 2019-06-17
configure.ac: fix configure error with dash

yizhao1:
A configure error occurs when /bin/sh -> dash:
  checking for is_selinux_enabled in -lselinux... yes
  checking for semanage_connect in -lsemanage... yes
  configure: 16322: test: yesyes: unexpected operator

Use "=" instead of "==" since dash doesn't support this operator.

Signed-off-by: Yi Zhao <yi.zhao@windriver.com>


hallyn:
Thank you.

PR #171, opened 2019-08-01
Honor --sbindir and --bindir for binary installation

falconindy:
Some distros don't care about the split between /bin, /sbin, /usr/bin,
and /usr/sbin, so let them easily stuff binaries wherever they want.

This also fixes a problem during installation where-in a loop of 'chmod
4755' calls will mostly fail (regardless of configuration). However, because
the last two succeed (newuidmap/newgidmap), make considers the command to be a
success. Somewhat not-amusingly, configuring shadow with --with-fcaps will
cause installation to fail because the final chmod call is now a failing one.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -52,13 +52,14 @@ usbin_PROGRAMS = \
 # id and groups are from gnu, sulogin from sysvinit
 noinst_PROGRAMS = id sulogin
 
+suidusbins     =
 suidbins       = su
 suidubins      = chage chfn chsh expiry gpasswd newgrp
 if !WITH_TCB
 suidubins += passwd
 endif
 if ACCT_TOOLS_SETUID
-suidubins += chgpasswd chpasswd groupadd groupdel groupmod newusers useradd userdel usermod
+suidusbins += chgpasswd chpasswd groupadd groupdel groupmod newusers useradd userdel usermod
 endif
 if ENABLE_SUBIDS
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks, most of the patch looks good, but I don't understand the reasoning behind the first hunk.  Distros which don't care about /bin vs /usr/bin usually have one as symlink to another, so using $prefix doesn't hurt there.  Distros which do care will I assume want the $prefix in there.  Does dropping the first hunk cause any problems in your distro?

falconindy:
It's true, Arch creates symlinks to /usr/bin, but our packaging still needs to place binaries in /usr/bin. Doing this through configure seems much more natural, and removing an element of surprise that these standard flags don't do anything.

The net effect is that I'm saving a [few lines](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/shadow#n146) in our packaging code by making --bindir and --sbindir work for shadow.

glensc:
@falconindy reminder that you should never link to branches https://news.ycombinator.com/item?id=8046710 especially if pointing to a line number.

hallyn:
This has the potential to change the 'make install' result for some distro, so I'd prefer to make the first hunk a separate patch.  That way if a distro complains we can revert just that patch.

falconindy:
Sounds fine. Updated the pull request.

hallyn:
Thanks

PR #175, opened 2019-08-21
sgetpwent.c/sgetgrent.c: check for additional data at end of line

### LINKED ISSUE #94: pwck does not moan about entries with wrong column count
The pwck tool checks the entries of /etc/passed and /etc/shadow for correctnes.
At least the pwck tool in version 4.5 of the shadow suite does not moan about entries with 8 columns.

Example: add a line like this in /etc/passed and run pwck:
`clamav:x:64:64:ClamAntiVirus:/dev/null:/bin/:/usr/bin/nologin`

pwck does not complain.

edneville:
Additional check to test for data at the end of passwd and group, fixes #94

### DIFF
--- a/lib/sgetgrent.c
+++ b/lib/sgetgrent.c
@@ -136,7 +136,7 @@ struct group *sgetgrent (const char *buf)
 			cp++;
 		}
 	}
-	if (i < (NFIELDS - 1) || *grpfields[2] == '\0') {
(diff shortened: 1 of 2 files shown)


hallyn:
Hi,

Thanks for the patches.  First patch had incorrect whitespace, so I fixed them up and pushed by hand.

PR #176, opened 2019-08-23
chkname.c, pwck.c, useradd.c, usermod.c, newusers.c: Allow names that…

### LINKED ISSUE #121: --force-badname
Hello!

After reading [this mailing list thread](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=42874) I'm not sure whether it was intentionally removed, but is there a runtime option or compiler flag that I can use to force [`is_valid_name()`](https://github.com/shadow-maint/shadow/blob/6fbc11ce2178205968c37853db752729359c9893/libmisc/chkname.c#L49-L71) to always `return true`? If not, would you be open to having it added?

Thanks for your work on this project!

edneville:
Allow names that do not conform to standards. Fixes #121

### DIFF
--- a/src/newusers.c
+++ b/src/newusers.c
@@ -117,6 +117,8 @@ static void check_perms (void);
 static void open_files (void);
 static void close_files (void);
 
+extern int allow_bad_names;
+
 /*
  * usage - display usage message and exit
  */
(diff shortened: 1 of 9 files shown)


glensc:
You should generally stick to the same indenting style used in the rest of the files.

edneville:
Thank you @glensc, modified the branch to match what is nearby.

src/newusers.c
```diff
@@ -610,9 +610,9 @@ static void process_flags (int argc, char **argv)
 #endif
 	                         long_options, NULL)) != -1) {
 		switch (c) {
-        case 'b':
-            allow_bad_names = true;
-            break;
+			case 'b':
```
tpimh: Extra tab here?

edneville: @tpimh, you're correct. I've pushed another update.

PR #177, opened 2019-08-25
pwck.c: only check home dirs if set and not a system user

### LINKED ISSUE #126: Conflicts between system users, useradd and pwck
When using `useradd` and adding a system user or simply a user with no home directory, with the `-r` or `-M` option, `pwck` will fail saying that the home directory doesn't exist.

The directory is supposed to not exist though.

I noticed this, because after creating a system user in Arch, I had `shadow.service` fail.

A workaround I found was editing `/etc/passwd` and change the home directory to either `/` or `/dev/null`, but I don't feel like this is a solution.

Also that's assuming the problem is with `useradd`, but in my opinion it's with `pwck` for not accepting users that don't have home directories.

edneville:
Only check home directory if the UID is not within the SYS user range and there is a home directory set. Fixes #126.

### DIFF
--- a/src/pwck.c
+++ b/src/pwck.c
@@ -382,6 +382,8 @@ static void check_pw_file (int *errors, bool *changed)
 	struct commonio_entry *pfe, *tpfe;
 	struct passwd *pwd;
 	struct spwd *spw;
+	uid_t min_sys_id = (uid_t) getdef_ulong ("SYS_UID_MIN", 101UL);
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks.

Could you please change the 'strlen' check to something like 'if (pw_dir && pw_dir[0])" ?

Also, you can add 'Closes #126' to the commit message to automatically close the issue when this is merged.

PR #180, opened 2019-09-20
Add support for a vendor directory and libeconf

thkukuk:
With this, it is possible for Linux distributors to store their
supplied default configuration files somewhere below /usr, while
/etc only contains the changes made by the user. The new option
--enable-vendordir defines where the shadow suite should additional
look for login.defs if this file is not in /etc.
libeconf is a key/value configuration file reading library, which
handles the split of configuration files in different locations
and merges them transparently for the application.


hallyn:
Thanks for the patch.

The patch to def_load() is not following indentation style of the rest of the project.  Would you care to update it, or would you like me to do it?

thkukuk:
I hope I fixed the indentation correctly. If not, I think it would be easier if you could fix it.

Issue #182, opened 2019-09-26
 shadow uses obsolete SELinux API


bigon:
Hello,

It seems that the shadow source package is using obsolete SELinux API,
for example:
```
In file included from passwd.c:46:
/usr/include/selinux/flask.h:5:2: warning: #warning "Please remove any #include's of this header in your source code." [-Wcpp]
 #warning "Please remove any #include's of this header in your source code."
  ^~~~~~~
/usr/include/selinux/flask.h:6:2: warning: #warning "Instead, use string_to_security_class() to map the class name to a value." [-Wcpp]
 #warning "Instead, use string_to_security_class() to map the class name to a value."
  ^~~~~~~
In file included from passwd.c:47:
/usr/include/selinux/av_permissions.h:1:2: warning: #warning "Please remove any #include of this header in your source code." [-Wcpp]
 #warning "Please remove any #include of this header in your source code."
  ^~~~~~~
/usr/include/selinux/av_permissions.h:2:2: warning: #warning "Instead, use string_to_av_perm() to map the permission name to a value." [-Wcpp]
 #warning "Instead, use string_to_av_perm() to map the permission name to a value."
```
That should be fixed as it could lead to security issues (or at least
weakened security) for people using SELinux

Kind regards,

Laurent Bigonville

PR #183, opened 2019-10-15
migrate to new SELinux api

cgzones:
Using hard-coded access vector ids is deprecated and can lead to issues with custom SELinux policies.
Switch to `selinux_check_access()`.

Also use the libselinux log callback and log if available to audit.
This makes it easier for users to catch SELinux denials.

Drop legacy shortcut logic for passwd, which avoided a SELinux check if uid 0 changes a password of a user which username equals the current SELinux user identifier.
Nowadays usernames rarely match SELinux user identifiers and the benefit of skipping a SELinux check is negligible.

Closes: #182


src/passwd.c
```diff
@@ -740,17 +736,7 @@ static int check_selinux_access (const char *changed_user,
 	if (changed_uid != 0 && strcmp(changed_user, user) == 0) {
```
cgzones: `if (changed_uid != 0 && strcmp(changed_user, user) == 0)`
Is this shortcut really necessary?
Also the comment above does not match.

cgzones:
I asked on the [SELinux mailing](https://lore.kernel.org/selinux/CAJ2a_Dd=T1sQh5zCqJLf8XTSQVU6Qgz78aFpT2oQ30VYVbQnWw@mail.gmail.com/) list for feedback

lib/selinux.c
```diff
@@ -98,6 +98,50 @@ int reset_selinux_file_context (void)
...
+
+	/* caller has checked for (is_selinux_enabled () > 0) already */
+
+	passwd_class_id = string_to_security_class ("passwd");
+	if (0 == passwd_class_id) {
+		fprintf (stderr,
+			    _("%s: can not find SELinux class 'passwd'\n"),
+			    Prog);
+		SYSLOG ((LOG_WARN, "can not find SELinux class 'passwd'"));
+		if (0 != security_getenforce ()) {
```
stephensmalley: Checking security_deny_unknown() would be more appropriate here, and is what libselinux selinux_check_access() does.

lib/selinux.c
```diff
@@ -98,6 +98,50 @@ int reset_selinux_file_context (void)
...
+		}
+	}
+
+	permission_id = string_to_av_perm (passwd_class_id, perm_name);
+	if (0 == permission_id) {
+		fprintf (stderr,
+			    _("%s: can not find '%s' SELinux access vector for class 'passwd'\n"),
+			    Prog, perm_name);
+		SYSLOG ((LOG_WARN, "can not find '%s' SELinux access vector for class 'passwd'", perm_name));
+		if (0 != security_getenforce ()) {
```
stephensmalley: Ditto

lib/selinux.c
```diff
@@ -98,6 +98,50 @@ int reset_selinux_file_context (void)
...
+			closelog ();
+			exit (E_SELINUX_ERR);
+		} else {
+			return 0;
+		}
+	}
+
+	return selinux_check_passwd_access (permission_id);
```
stephensmalley: Wondering if we ought to just convert to direct use of libselinux selinux_check_access().  That would require calling getprevcon() here and passing it as the source and target contexts, but avoids mapping the class twice.  If so, would need/want to also use selinux_set_callback() to set the logging callback and ideally wire that up to audit.

src/passwd.c
stephensmalley: This logic appears to be a legacy of original SELinux reference implementation where SELinux user identity (i.e. first field of the security context) was the same as the Linux username.  Then the logic tries to require that in order to change the password of a different user, one needs to have the "passwd" permission.  Then someone further amended it to also require the "passwd" permission if changing the root password regardless. Seems like this will almost always fail since the SELinux user identity is rarely (never?) the same as the Linux username and thus we will always execute the else clause.  Likely should be removed.

stephensmalley:
For new code, try to use selinux_check_access() and let libselinux internally handle the string_to_security_class, string_to_av_perm, security_deny_unknown, security_getenforce, etc calls and handling.  See dbus-broker for an example.

cgzones:
updated to use `selinux_check_access()` and enabled logging via `selinux_set_callback(SELINUX_CB_LOG, ...)` (also if available inform audit)

lib/selinux.c
cgzones: switched to `selinux_check_access()`

lib/selinux.c
cgzones: switched to `selinux_check_access()`

lib/selinux.c
```diff
@@ -98,6 +98,136 @@ int reset_selinux_file_context (void)
...
+	va_end (ap);
+
+	if (r < 0) {
+		free (buf);
+		return 0;
+	}
+
+	fprintf (stderr, _("%s: libselinux: %s\n"),
+			    Prog, buf);
+	SYSLOG ((LOG_WARN, "libselinux: %s", buf));
```
stephensmalley: Do we want to always do both (log + audit) or only fall back to SYSLOG if audit not enabled?

lib/selinux.c
```diff
@@ -98,6 +98,136 @@ int reset_selinux_file_context (void)
...
+		    Prog, strerror (errno));
+		SYSLOG ((LOG_WARN,
+		    "can not get previous SELinux process context: %s",
+		    strerror (errno)));
+		freecon (user_context_str);
+		return (security_getenforce () != 0);
+	}
+
+	if (changed_user) {
+		/* get the "user" portion of the context
```
stephensmalley: I am of the opinion that this entire if statement and its associated dependencies can just go away, i.e. no shortcut logic because SELinux user identity will generally never be the same as Linux username on any modern Linux distro. Worst case is that we'll trigger a passwd permission check that wasn't truly needed.

src/chfn.c
```diff
@@ -280,7 +276,7 @@ static void process_flags (int argc, char **argv)
 		{NULL, 0, NULL, '\0'}
 	};
 
-	/* 
+	/*
 	 * The remaining arguments will be processed one by one and executed
```
stephensmalley: Usually you don't make unrelated whitespace fixes in the same commit as a real change, i.e. if you do them at all, do them in a separate commit.

src/passwd.c
```diff
@@ -331,7 +320,7 @@ static int new_password (const struct passwd *pw)
 
 		/*
 		 * If enabled, warn about weak passwords even if you are
-		 * root (enter this password again to use it anyway). 
+		 * root (enter this password again to use it anyway).
```
stephensmalley: ditto

src/passwd.c
```diff
@@ -1034,22 +977,15 @@ int main (int argc, char **argv)
...
-		if (getprevcon (&user_context) == 0) {
-			user = user_context; /* FIXME: use context_user_get? */
-		}
+	current_uid = getuid() ;
+	if ((0 == current_uid) &&
+	    (check_selinux_permit_shortcut ("passwd", name, pw->pw_uid) != 0)) {
 		SYSLOG ((LOG_ALERT,
-		         "%s is not authorized to change the password of %s",
-		         user, name));
+		         "uid %d is not authorized by SELinuxto change the password of %s",
```
stephensmalley: Missing a space.

lib/selinux.c
```diff
@@ -98,6 +98,136 @@ int reset_selinux_file_context (void)
...
+	r = vasprintf (&buf, fmt, ap);
+	va_end (ap);
+
+	if (r < 0) {
+		free (buf);
+		return 0;
+	}
+
+	fprintf (stderr, _("%s: libselinux: %s\n"),
+			    Prog, buf);
```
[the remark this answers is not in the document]
cgzones: dropped the `stderr` logging

lib/selinux.c
cgzones: switched to try `audit`, use `syslog` only as fallback

lib/selinux.c
```diff
@@ -98,6 +98,136 @@ int reset_selinux_file_context (void)
 	return 0;
 }
 
+/*
+ *  libselinux uses this callback when access gets denied or other
+ *  events happen. If audit is turned on, messages will be reported
+ *  using audit netlink, otherwise they will be logged using the usual
+ *  channels.
+ *
+ *  Code adopted from systemd, which adopted from dbus.
```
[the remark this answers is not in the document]
cgzones: The code should be different enough to not cause problems, but I am not a copyright expert.
The systemd code can be found [here](https://github.com/systemd/systemd/blob/c8966bffdd9ad4ef7dcec89e0cc8d43529ef2aed/src/core/selinux-access.c#L96)

src/chfn.c
cgzones: dropped

src/chfn.c
```diff
@@ -391,7 +386,7 @@ static void check_perms (const struct passwd *pw)
 	/*
 	 * Non-privileged users are optionally authenticated (must enter the
 	 * password of the user whose information is being changed) before
-	 * any changes can be made. Idea from util-linux chfn/chsh. 
+	 * any changes can be made. Idea from util-linux chfn/chsh.
```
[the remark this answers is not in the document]
cgzones: dropped

src/passwd.c
cgzones: dropped

src/passwd.c
cgzones: fixed

lib/selinux.c
```diff
@@ -98,6 +98,136 @@ int reset_selinux_file_context (void)
...
+                                      uid_t changed_uid)
+{
+	char *user_context_str;
+	int r;
+
+	if (0 == is_selinux_enabled ()) {
+		return 0;
+	}
+
+	selinux_set_callback (SELINUX_CB_LOG, (union selinux_callback) log_callback);
```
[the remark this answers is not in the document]
cgzones: The running applications (passwd, chfn,...) should only check for SELinux permit once in their runtime, so the question is, is the overhead worth avoiding additionally pointer assignments?

lib/selinux.c
```diff
@@ -98,6 +99,146 @@ int reset_selinux_file_context (void)
...
+#ifdef WITH_AUDIT
+	static int selinux_audit_fd = -2;
+#endif
+
+	va_start (ap, fmt);
+	r = vasprintf (&buf, fmt, ap);
+	va_end (ap);
+
+	if (r < 0) {
+		free (buf);
```
stephensmalley: buf could be uninitialized here?

lib/selinux.c
```diff
@@ -98,6 +99,146 @@ int reset_selinux_file_context (void)
...
+		if (   (errno != EINVAL)
+		    && (errno != EPROTONOSUPPORT)
+		    && (errno != EAFNOSUPPORT)) {
+
+		    (void) fputs (_("Cannot open audit interface.\n"),
+		              stderr);
+		    SYSLOG ((LOG_WARN, "Cannot open audit interface."));
+		}
+	} else {
+		uid_t audit_uid = geteuid ();
```
stephensmalley: Is this the right uid to use here?  Should we be using the real UID, or the login UID, or just pass 0 like systemd?

lib/selinux.c
```diff
@@ -98,6 +99,146 @@ int reset_selinux_file_context (void)
...
+	selinux_set_callback (SELINUX_CB_LOG, (union selinux_callback) selinux_log_cb);
+
+	if (getprevcon (&user_context_str) != 0) {
+		fprintf (stderr,
+		    _("%s: can not get previous SELinux process context: %s\n"),
+		    Prog, strerror (errno));
+		SYSLOG ((LOG_WARN,
+		    "can not get previous SELinux process context: %s",
+		    strerror (errno)));
+		freecon (user_context_str);
```
stephensmalley: No need to free here; getprevcon() failed.  And user_context_str could be uninitialized.

lib/selinux.c
stephensmalley: Do you think we need this logic?

src/passwd.c
```diff
@@ -1034,22 +977,15 @@ int main (int argc, char **argv)
...
 	   changing a password without entering the old one */
-	if ((is_selinux_enabled() > 0) && (getuid() == 0) &&
-	    (check_selinux_access (name, pw->pw_uid, PASSWD__PASSWD) != 0)) {
-		security_context_t user_context = NULL;
-		const char *user = "Unknown user context";
-		if (getprevcon (&user_context) == 0) {
-			user = user_context; /* FIXME: use context_user_get? */
-		}
+	current_uid = getuid() ;
+	if ((0 == current_uid) &&
```
stephensmalley: We should probably be checking the "amroot" variable used elsewhere for consistency.  Otherwise, if the amroot logic changes later, we won't match.

lib/selinux.c
cgzones: fixed, thanks

lib/selinux.c
cgzones: fixed, thanks

lib/selinux.c
cgzones: I agree with removing this logic; I'd like to get an OK from upstream.
Should I announce this change on the mailing-list?

lib/selinux.c
cgzones: Good question; From an SELinux standpoint the uid probably does not matter at all (only the context).

src/passwd.c
cgzones: The result of `0 == current_uid` (where `current_uid = getuid()`) should be the same as the value `amroot`. So simplify the whole condition to (ignoring the shortcut logic for now)?
```
if (amroot && (check_selinux_permit_shortcut ("passwd", name, pw->pw_uid) != 0)) {
```

lib/selinux.c
stephensmalley: Looks like dbus-broker uses getuid() i.e. real UID.  For shadow utils, that would be the UID of the caller even if the util is setuid-root, so that seems preferable to using geteuid().  Or you could ask what uid to use on linux-audit mailing list.

lib/selinux.c
stephensmalley: Sure, you could ask there. May also want to compare with source code of passwd, which is a separate package in Fedora?

src/passwd.c
stephensmalley: Correct.  That way if someone later alters the amroot logic we'll pick up the change automatically, and keep in sync.  We want to do this check iff it was possible to change the password without entering the old one.

lib/selinux.c
stephensmalley: Still not sure it is safe/necessary to free(buf) here.  man page for vasprintf() says "If memory allocation wasn't  possible,  or  some other error occurs, these functions will return -1, and the contents of strp are undefined."  That means buf could be undefined after the call if it returns -1.

src/passwd.c
cgzones: changed

lib/selinux.c
cgzones: you're absolutely right, fixed; thanks

src/passwd.c
cgzones: Legacy shortcut dropped

lib/selinux.c
cgzones: Announced
(p.s.: fedora's passwd does not use something like this)

lib/selinux.c
cgzones: switched to hardcoded `0`, like systemd, dbus and fedora's passwd

hallyn:
Thanks everyone!

Issue #186, opened 2019-11-11
 Using  the command "useradd -s" 


whzhe51:
If the path is not exist or not executable, i think it should not be allow to create a user.

PR #187, opened 2019-11-11
useradd: check for valid shell argument

jubalh:
Check whether shell argument given with `-s` is actually present and executable.

Fix https://github.com/shadow-maint/shadow/issues/186

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -1094,6 +1094,7 @@ static void process_flags (int argc, char **argv)
 	const struct group *grp;
 	bool anyflag = false;
 	char *cp;
+	struct stat st;
 
 	{
 		/*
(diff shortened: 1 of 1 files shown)


src/useradd.c
```diff
@@ -1310,7 +1310,8 @@ static void process_flags (int argc, char **argv)
 				if (   ( !VALID (optarg) )
 				    || (   ('\0' != optarg[0])
 				        && ('/'  != optarg[0])
-				        && ('*'  != optarg[0]) )) {
+				        && ('*'  != optarg[0]) )
+				    || access(optarg, X_OK != 0)) {
```
brauner:
```
				    || access(optarg, X_OK != 0)) {
```
Doesn't that need to be
```
				    || access(optarg, X_OK) != 0) {
```
?

jubalh:
And I should use stat() to check whether it is a directory probably too, since a dir will also have the X bit set.

hallyn:
Thanks guys.

PR #188, opened 2019-11-11
Allow translation of new strings by adding new files to POTFILES.in

rbalint:


rbalint:
Originally reported at: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=759110

hallyn:
Thanks.

PR #194, opened 2019-12-01
Add myself to contributors

jubalh:


hallyn:
Thanks!  I *just* tagged 4.8, so this will go into 4.9.  Shout if you want me to finagle it into 4.8

jubalh:
> Shout if you want me to finagle it into 4.8

It would be cool but it's not that important of course :) Don't worry!

Issue #196, opened 2019-12-01
All binaries are now always installed `/bin` and `/sbin`


chutz:
Some distros still care about `/bin` vs `/usr/bin`. The change https://github.com/shadow-maint/shadow/commit/e293aa9cfca0619a63616af75532637dab60d49d makes it so all binaries are always installed to `/bin`/`/sbin`. The only way to restore the previous behaviour of installing some binaries to `/usr/bin`/`/usr/sbin` is to revert the patch.

hallyn:
Hm.  I'll have to look in a bit - I don't know why I didn't think it was doing that.

Please feel free to do a PR doing the revert.

chutz:
Maybe the "right" solution to get the behaviour I suspect you were aiming for is to have a `--binprefix` configure flag. Have it default to `prefix`, but allow it to be overridden.

hallyn:
Thanks, merged the revert, so closing this.

hallyn:
If you need a 4.8.1 release with this revert, please shout.

PR #197, opened 2019-12-01
Revert "Honor --sbindir and --bindir for binary installation"

chutz:
This reverts commit e293aa9cfca0619a63616af75532637dab60d49d.

See https://github.com/shadow-maint/shadow/issues/196

Some distros still care about `/bin` vs `/usr/bin`. This commit makes
it so all binaries are always installed to `/bin`/`/sbin`. The only way to
restore the previous behaviour of installing some binaries to
`/usr/bin`/`/usr/sbin` is to revert the patch.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -2,8 +2,8 @@
 EXTRA_DIST = \
 	.indent.pro
 
-ubindir = ${bindir}
-usbindir = ${sbindir}
+ubindir = ${prefix}/bin
+usbindir = ${prefix}/sbin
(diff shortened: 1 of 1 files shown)


glensc:
could you release 4.8.1 or something?

hallyn:
> could you release 4.8.1 or something?

Yup, if it helps, I'll do that.

glensc:
Yes please

hallyn:
I aim to do it today or tomorrow.

PR #199, opened 2019-12-16
Don't auto-enable ACCT_TOOLS_SETUID if PAM is detected

falconindy:
Here's a sad story:

* 70971457 is merged into shadow, allowing newgidmap/newuidmap to be
installed with file caps rather than setuid.
* https://bugs.archlinux.org/task/63248 is filed to take advantage of
this.
* The arch maintainer of the 'shadow' package notices that this doesn't
work, and submits a pull request to fix this in shadow.
* edf7547ad5 is merged, fixing the post install hooks.

The problem here is that distros have been building shadow with PAM for
O(years), but the install hooks have silently failed due to the
combination of the directory mismatch (suidubins vs suidsbins) and later
success with setuid'ing newgidmap/newuidmap.

With the install hooks fixed, those of us (Arch[1] and Gentoo[2] so far)
who never built shadow explicitly with --enable-account-tools-setuid are
now getting setuid account tools, and don't have PAM configuration
suitable for use with setuid account management tools.

It's entirely unclear to me why you'd want this, but I assume there's
some reason out there for it existing. Regardless, setuid binaries are
dangerous and shouldn't be enabled by default without good reason.

[1] https://bugs.archlinux.org/task/64836
[2] https://bugs.gentoo.org/702252

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -226,7 +226,7 @@ AC_ARG_ENABLE(account-tools-setuid,
 	   *) AC_MSG_ERROR(bad value ${enableval} for --enable-account-tools-setuid)
 	   ;;
 	 esac],
-	[enable_acct_tools_setuid="maybe"]
+	[enable_acct_tools_setuid="no"]
 )
 
 AC_ARG_ENABLE(utmpx,


awilfox:
So, my concern from the Gentoo bug is that this would actually disable PAM functionality entirely in the specified utilities.  Is that the case, or is that a misunderstanding on their part?

falconindy:
Yes, you're correct -- PAM functionality in the account management tools is dependent on `--with-libpam` *and* `--enable-account-tools-setuid`. However, without the tools being setuid, the PAM linkage doesn't really matter, right? What's the value in prompting for authentication if you aren't also escalating privileges?

Foxboron:
This issue has been assigned [CVE-2019-19882](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-19882)

brauner:
> This issue has been assigned [CVE-2019-19882](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-19882).

What is this CVE filed against shadow? Who has been notified of this CVE?

PR #204, opened 2019-12-18
Adding tty logging to the useradd command

### LINKED ISSUE #198: User add logging
On a multi-admin system it can be tricky to track down who added a user.

If a TTY exists when useradd is run, can that be logged in the syslog entry?

If this idea is acceptable, let me know and I'll raise a PR.

edneville:
This commit adds a from= field to the end of the useradd log entry.
Casting user_name to tallylog_reset to silence a compiler warning.

Closes #198.

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -223,7 +223,7 @@ static void open_files (void);
 static void open_shadow (void);
 static void faillog_reset (uid_t);
 static void lastlog_reset (uid_t);
-static void tallylog_reset (char *);
+static void tallylog_reset (const char *);
 static void usr_update (void);
 static void create_home (void);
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks.  The patch looks good, however please make two changes (or I can do it if you prefer):

1. squash the two commits
2. (if I'm not overlooking a reason why it's not possible) instead of casting user_name to char*, instead have tallylog_reset() take a const char *.

edneville:
Thanks for looking @hallyn. I've removed the cast and changed the function as suggested. I think you'll have to do a squash merge as I've pushed commits to the source branch.

hallyn:
Thanks!

PR #205, opened 2019-12-19
Makefile: bail out on error in for-loops

Blub:
`make` runs each line in a shell and bails out on error,
however, the shell is not started with `-e`, so commands in
`for` loops can fail without the error actually causing
`make` to bail out with a failure status.

For instance, the following make snippet will end
successfully, printing 'SUCCESS', despite the first `chmod`
failing:

    all:
        touch a b
        for i in a-missing-file a b; do \
            chmod 666 $$i; \
        done
        @echo SUCCESS

To prevent wrong paths in install scripts from remaining
unnoticed, let's activate `set -e` in the `for` loop
subshells.

Signed-off-by: Wolfgang Bumiller <w.bumiller@proxmox.com>


hallyn:
Thanks, I'm all for this.

Going to wait to merge a bit as (a) I've not tested and (b) I'd like to give others time to comment :)

PR #208, opened 2020-01-08
add new UMASK_HOME login.defs option

Duncaen:
This option can be used to set a separate mode for useradd(8) and
newusers(8) to create the home directories with.
If this option is not set, the current behavior of using UMASK
or the default umask is preserved.

There are many distributions that set UMASK to 077 by default just
to create home directories not readable by others and use things like
/etc/profile, bashrc or sudo configuration files to set a less restrictive
umask. This has always resulted in bug reports because it is hard
to follow as users tend to change files like bashrc and are not about
setting the umask to counteract the umask set in /etc/login.defs.

A recent change in sudo has also resulted in many bug reports about
this. sudo now tries to respect the umask set by pam modules and on
systems where pam does not set a umask, the login.defs UMASK value is
used.

### DIFF
--- a/etc/login.defs
+++ b/etc/login.defs
@@ -195,11 +195,16 @@ KILLCHAR	025
 # Default initial "umask" value used by login(1) on non-PAM enabled systems.
 # Default "umask" value for pam_umask(8) on PAM enabled systems.
 # UMASK is also used by useradd(8) and newusers(8) to set the mode for new
-# home directories.
+# home directories if HOME_MODE is not set.
 # 022 is the default value, but 027, or even 077, could be considered
 # for increased privacy. There is no One True Answer here: each sysadmin
 # must make up their mind.
 UMASK		022
+
+# HOME_MODE is used by useradd(8) and newusers(8) to set the mode for new
+# home directories.
+# If HOME_MODE is not set, the value of UMASK is used to create the mode.
+#HOME_MODE	0700
 
 #
 # Password aging controls:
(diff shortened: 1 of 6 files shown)


t8m:
The reversed approach would be also possible although that would require modification of pam_umask as well.
Instead of adding UMASK_HOME a variable UMASK_LOGIN would be added that would (if set) change behavior of login and pam_umask.

Duncaen:
Another alternative would be something like `HOME_MODE` instead of a umask and only use `UMASK` if `HOME_MODE` is not set.
Because if this is only used for this specific case, keeping it a umask doesn't make a lot of sense.

hallyn:
> Another alternative would be something like `HOME_MODE` instead of a umask and only use `UMASK` if `HOME_MODE` is not set.
> Because if this is only used for this specific case, keeping it a umask doesn't make a lot of sense.

I like that - would you mind updating the patch to do that?

Duncaen:
I updated the patch, to use `HOME_MODE`.

hallyn:
Thanks.

(This points out that newusers.c is not doing the extra work that adduser is doing.  I suppose this all ought to be in a helper.  One day...)

hallyn:
It looks like the login.defs.d/HOME_MODE.xml file is missing.  That breaks the build, so I reverted it for now.  Could you please add that file?

hallyn:
If you just want to upload the contents of the file somewhere (even pastebin or a comment here), I'll finagle it in.

jubalh:
For the record: reverted back in https://github.com/shadow-maint/shadow/commit/93f1f3512322222894911c34b8267891bae94a92

Issue #207, opened 2020-01-03
useradd shell check and nologin


jubalh:
https://github.com/shadow-maint/shadow/pull/187 has a small drawback:

Sometimes we might want to point to nologin for some users:
```
useradd -r -s /sbin/nologin -c "nobody" -g nobody -d /var/lib/nobody -u 65534 nobody
```

AFAIK `/sbin/nologin` is provided by util-linux only. On some systems during initial setup it might be that util-linux is not installed yet.

So I'm thinking whether we should
* skip the check introduced in #187 for system users (`-r`)
* allow nologin always and don't check for it
* leave the responsibility for the user/distribution and don't change anything
* more ideas?

hallyn:
https://github.com/shadow-maint/shadow/blob/master/src/nologin.c

hallyn:
I'm going to close this, as it seems to me the underlying assumption is wrong.  Please re-open if I'm misunderstanding.

PR #210, opened 2020-01-16
Make the check for non-executable shell only a warning.

t8m:
Although it is a good idea to check for an inadvertent typo
in the shell name it is possible that the shell might not be present
on the system yet when the user is added.

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -1328,14 +1328,21 @@ static void process_flags (int argc, char **argv)
 				if (   ( !VALID (optarg) )
 				    || (   ('\0' != optarg[0])
 				        && ('/'  != optarg[0])
-				        && ('*'  != optarg[0]) )
-				    || (stat(optarg, &st) != 0)
-				    || (S_ISDIR(st.st_mode))
-				    || (access(optarg, X_OK) != 0)) {
+				        && ('*'  != optarg[0]) )) {
 					fprintf (stderr,
 					         _("%s: invalid shell '%s'\n"),
 					         Prog, optarg);
 					exit (E_BAD_ARG);
+				}
+				if (    '\0' != optarg[0]
+				     && '*'  != optarg[0]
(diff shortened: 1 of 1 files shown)


jubalh:
Good idea @t8m! Sorry for not doing this in the first place.

This also helps with https://github.com/shadow-maint/shadow/issues/207

src/useradd.c
```diff
@@ -1328,15 +1328,22 @@ static void process_flags (int argc, char **argv)
...
 					fprintf (stderr,
 					         _("%s: invalid shell '%s'\n"),
 					         Prog, optarg);
 					exit (E_BAD_ARG);
 				}
+				if (    '\0' != optarg[0]
+				     && '*'  != optarg[0]
+				     && strcmp(optarg, "/sbin/nologin") != 0
+				     && (   stat(optarg, &st) != 0
+				         || S_ISDIR(st.st_mode)
```
hallyn: But if it's a directory then it is not likely to be fixed up later, right? :)

There are quite a few duplicated checks here which I'm worried will eventually (as code shifts) not be handled right.  Would it be better to have a single outer 'if' and then only exit (and maybe change the message) using a nested check for non-executable or missing shell?

t8m: Perhaps it could be fixed later. Why not? :D The point is, there is no big reason to hard disallow it. I can try to reorg the conditions a little bit.

t8m: Thinking about it more - I really have no idea how to reorg it in a way that would not be even more convoluted than the existing patch. The problem is that in the second if we just want to avoid trying to stat() empty or a '*...' shell as these are special values and the warning should not be provided for them. However just putting these as higher level if does not work because there is also the !VALID() check which in case of *.... shell should take a precedence.

hallyn: Ok.

hallyn:
Thanks.

PR #212, opened 2020-01-20
Do not mistake a regular user process for a namespaced one

t8m:
In case there is a regular user with a process running on a system
with uid falling into a namespaced uid range of another user.
The user with the colliding namespaced uid range will not be
allowed to be deleted without forcing the action with -f.

The user_busy() is adjusted to check whether the suspected process
is really a namespaced process of the user being deleted.

### DIFF
--- a/libmisc/user_busy.c
+++ b/libmisc/user_busy.c
@@ -39,6 +39,7 @@
 #include <sys/types.h>
 #include <dirent.h>
 #include <fcntl.h>
+#include <unistd.h>
 #include "defines.h"
 #include "prototypes.h"
 #ifdef ENABLE_SUBIDS
@@ -106,6 +107,31 @@ static int user_busy_utmp (const char *name)
 #endif				/* !__linux__ */
 
 #ifdef __linux__
+#ifdef ENABLE_SUBIDS
+#define in_parentuid_range(uid) ((uid) >= parentuid && (uid) < parentuid + range)
+static int different_namespace (const char *sname)
+{
+	/* 41: /proc/xxxxxxxxxx/task/xxxxxxxxxx/ns/user + \0 */
+	char path[41];
+	char buf[512], buf2[512];
+	ssize_t llen1, llen2;
+
+	snprintf (path, 41, "/proc/%s/ns/user", sname);
+
+	if ((llen1 = readlink (path, buf, sizeof(buf))) == -1)
+		return 0;
+
+	if ((llen2 = readlink ("/proc/self/ns/user", buf2, sizeof(buf2))) == -1)
+		return 0;
+
+	if (llen1 == llen2 && memcmp (buf, buf2, llen1) == 0)
+		return 0; /* same namespace */
+
+	return 1;
+}
+#endif                          /* ENABLE_SUBIDS */
+
+
 static int check_status (const char *name, const char *sname, uid_t uid)
 {
 	/* 40: /proc/xxxxxxxxxx/task/xxxxxxxxxx/status + \0 */
@@ -114,7 +140,6 @@ static int check_status (const char *name, const char *sname, uid_t uid)
 	FILE *sfile;
 
 	snprintf (status, 40, "/proc/%s/status", sname);
-	status[39] = '\0';
 
 	sfile = fopen (status, "r");
 	if (NULL == sfile) {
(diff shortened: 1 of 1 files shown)


libmisc/user_busy.c
```diff
@@ -123,26 +171,29 @@ static int check_status (const char *name, const char *sname, uid_t uid)
...
+				}
 #ifdef ENABLE_SUBIDS
-				    || have_sub_uids(name, ruid, 1)
-				    || have_sub_uids(name, euid, 1)
-				    || have_sub_uids(name, suid, 1)
-#endif				/* ENABLE_SUBIDS */
+				if (  (   have_sub_uids(name, ruid, 1)
+				       || have_sub_uids(name, euid, 1)
+				       || have_sub_uids(name, suid, 1))
+                                    && in_mapping(sname, ruid, euid, suid)
```
ansasaki: The indentation looks a bit off here. Maybe using tabs instead of spaces?

t8m: Fixed, thank you.

libmisc/user_busy.c
```diff
@@ -106,6 +107,54 @@ static int user_busy_utmp (const char *name)
 #endif				/* !__linux__ */
 
 #ifdef __linux__
+#ifdef ENABLE_SUBIDS
+#define in_parentuid_range(uid) ((uid) >= parentuid && (uid) < parentuid + range)
+static int in_mapping (const char *sname, unsigned long ruid, unsigned long euid, unsigned long suid)
+{
+	/* 41: /proc/xxxxxxxxxx/task/xxxxxxxxxx/uid_map + \0 */
```
hallyn: Not a critique of your patch as I see the existing case of this.  But I worry about the assumption that each pid is 9 characters.  One day the kernel might bump that number.  Then an old shadow might have trouble on newer kernel, and boy would that suck.

@brauner what do you think, is it worth bumping these values throughout the tree? (if they exist elsewhere)

t8m: Actually it is kind-of hard to imagine that the PID would be larger than UINT32_MAX (when today it is just 16 bit number) - which fits into 10 digits. The comment is not really precise here because we are not looking at threads. I'll adjust it.

t8m: Oh, disregard what I said about the comment not being precise. We actually are looking at threads. But anyway. The reserved space is 10 digits for for PID where currently it has just 5 digits.

hallyn: Yeah it seems unlikely.  But bumping the # to 100 would be so easy, and the pain if it *does* happen would be so ugly, it seems maybe worth it.

But I don't mean for that to happen in your patch, sorry for derailing this.

libmisc/user_busy.c
```diff
@@ -106,6 +107,54 @@ static int user_busy_utmp (const char *name)
...
+	snprintf (path, 41, "/proc/%s/ns/user", sname);
+
+	if ((llen1 = readlink (path, buf, sizeof(buf))) == -1)
+		return 0;
+
+	if ((llen2 = readlink ("/proc/self/ns/user", buf2, sizeof(buf2))) == -1)
+		return 0;
+
+	if (llen1 == llen2 && memcmp(buf, buf2, llen1) == 0)
+		return 0; /* same namespace */
```
hallyn: This part makes sense - if the process is in the same namespace then it can't be using one of the subuids.  But,

libmisc/user_busy.c
```diff
@@ -106,6 +107,54 @@ static int user_busy_utmp (const char *name)
...
+	}
+	if (fgets (buf, sizeof (buf), file) == buf) {
+		unsigned long nsuid, parentuid, range;
+
+		if (sscanf (buf,
+			    " %lu %lu %lu\n",
+			    &nsuid, &parentuid, &range) == 3) {
+			if (   in_parentuid_range(ruid)
+			    || in_parentuid_range(euid)
+			    || in_parentuid_range(suid) ) {
```
hallyn: ...  what is the case where you expect these not to be true?  The task would have to be running as a uid which not mapped into the task's own uid mapping?  Which is unlikely in the general container case since we've already established that one if its uids is in the victim's subuid range.  Can you explain what case this is meant to check?

t8m: The thing is I've originally implemented just this check and wanted to exempt just the case where all the uid range is mapped (i.e. as is for the regular non-namespaced tasks). But then I realized the logic would be incorrect for the case when the namespace is the same. So I've implemented the namespace comparison.

So you're right that this check actually does not make much sense now and I'm going to remove it.

libmisc/user_busy.c
```diff
@@ -114,7 +140,6 @@ static int check_status (const char *name, const char *sname, uid_t uid)
 	FILE *sfile;
 
 	snprintf (status, 40, "/proc/%s/status", sname);
-	status[39] = '\0';
```
hallyn: Why did you get rid of this?

libmisc/user_busy.c
```diff
@@ -106,6 +107,31 @@ static int user_busy_utmp (const char *name)
...
+#ifdef ENABLE_SUBIDS
+#define in_parentuid_range(uid) ((uid) >= parentuid && (uid) < parentuid + range)
+static int different_namespace (const char *sname)
+{
+	/* 41: /proc/xxxxxxxxxx/task/xxxxxxxxxx/ns/user + \0 */
+	char path[41];
+	char buf[512], buf2[512];
+	ssize_t llen1, llen2;
+
+	snprintf (path, 41, "/proc/%s/ns/user", sname);
```
hallyn: This should also get a just-in-case null termination.

libmisc/user_busy.c
t8m: Because snprintf always appends the NUL termination.

libmisc/user_busy.c
t8m: The just-in-case would have to be a buggy libc non-conforming to POSIX and C specs. Do we really care about such environments?

libmisc/user_busy.c
hallyn: Yes, you're right, and I guessthis step doesn't help with my real concern.  It's been bothering me that we don't check for if (ret < 0 || ret >= len).  All over the source.

But that's not your problem.  I think I'll change that in a separate PR.

Thanks.

PR #216, opened 2020-02-05
useradd: doesn't generate /var/spool/mail/$USER with the proper SELinux user identity

ikerexxe:
Explanation: use set_selinux_file_context() and reset_selinux_file_context() for create_mail() just as is done for create_home()

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1690527

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -190,6 +190,7 @@ static bool home_added = false;
 #define E_NAME_IN_USE	9	/* username already in use */
 #define E_GRP_UPDATE	10	/* can't update group file */
 #define E_HOMEDIR	12	/* can't create home directory */
+#define E_MAILBOXFILE	13	/* can't create mailbox file */
 #define E_SE_UPDATE	14	/* can't update SELinux user mapping */
 #ifdef ENABLE_SUBIDS
 #define E_SUB_UID_UPDATE 16	/* can't update the subordinate uid file */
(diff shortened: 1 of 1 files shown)


pbrezina:
I did not try it but code looks good to me.

src/useradd.c
```diff
@@ -2210,6 +2211,15 @@ static void create_mail (void)
 			sprintf (file, "%s/%s/%s", prefix, spool, user_name);
 		else
 			sprintf (file, "%s/%s", spool, user_name);
+
+#ifdef WITH_SELINUX
+		if (set_selinux_file_context (spool) != 0) {
```
cgzones: why are you using the path of `spool` and not `file`?

ikerexxe: I agree that this may not be a good idea from a security point of view. So, I'm changing the fix to change directly the file.

cgzones:
I do not think this is going to work: now you are just copying the context from `/var/spool/mail` to `/var/spool/mail/joe`.
These do not have the same context:
  - `/var/spool/mail` has usually the context: `system_u:object_r:mail_spool_t`
  - `/var/spool/mail/joe` has a user-identity modified context: `user_u:object_r:mail_spool_t`

I think the first version was quite good (just change `set_selinux_file_context (spool)` to `set_selinux_file_context (file)`).

ikerexxe:
But isn't this approach a little bit dangerous? It opens the gate for another process to create a file in this folder with that context.

ikerexxe:
What would happen if between the context set and the context reset another process creates another file in this same folder? Would this file have "unconfined_u" or "system_u" as the SELinux user identity? Would this pose a security risk?

ikerexxe:
I'm sorry but I got confused so please forget about my messages. I will proceed to fix this problem as you proposed.

bachradsusi:
Before you merge it, please consider a different commit message

> useradd: doesn't generate /var/spool/mail/$USER with the proper SELinux user identity

In fact the patch *does* generate /var/spool/mail/$USER with the proper SELinux label based on SELinux user identity

ikerexxe:
> Before you merge it, please consider a different commit message
> 
> > useradd: doesn't generate /var/spool/mail/$USER with the proper SELinux user identity
> 
> In fact the patch _does_ generate /var/spool/mail/$USER with the proper SELinux label based on SELinux user identity

Done!

hallyn:
Thanks, all.

PR #219, opened 2020-02-07
libmisc: Accept --root=path and --prefix=path option syntax

### LINKED ISSUE #218: chpasswd silently ignores --root=path
getopt_long accepts two styles of long options, `--arg param` and `--arg=param`.  However, chpasswd attempts to pre-extract the --root option, but the code to do this does not recognise the latter style.  An option specified by the user as `--root=path` is then silently ignored, with potentially harmful consequences.

See https://bugs.launchpad.net/bugs/1822501

The same applies to other executables in the project, and also the --prefix option.

I am preparing a PR.

infinnovation-dev:
Recognise --root=path in addition to --root path (and similarly for
--prefix) to match the syntax accepted by getopt_long.

Fixes #218

### DIFF
--- a/libmisc/prefix_flag.c
+++ b/libmisc/prefix_flag.c
@@ -74,10 +74,13 @@ extern const char* process_prefix_flag (const char* short_opt, int argc, char **
 	 * Parse the command line options.
 	 */
 	int i;
-	const char *prefix = NULL;
+	const char *prefix = NULL, *val;
 
 	for (i = 0; i < argc; i++) {
+		val = NULL;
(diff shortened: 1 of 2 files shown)


infinnovation-dev:
Note that this does not attempt to fix parsing of short option syntax.  Variants such as "-Rpath", "-eRpath" and "-eR path" will still be ignored.  Instead of trying to fix that in process_root_flag(), it might be better to refactor process_flags() and check_flags() so that only getopt_long() is used for argument parsing.  But that's for another day.

hallyn:
Thank you.  Indeed refactoring might be good.

PR #221, opened 2020-02-09
fix non-systemd/non-selinux compiling bug

sulincix:
most of distro compile shadow without systemd and selinux
And also some ditros does not have systemd
For example: alpine devuan sulinos artix

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -8,5 +8,5 @@ autoreconf -v -f --install || exit 1
 	--enable-maintainer-mode \
 	--disable-shared \
 	--without-libpam \
-	--with-selinux \
+	--without-selinux \
 	"$@"


sulincix:
I teted. non-systemd build works with systemd used gnu/linux. (debian 10 and archlinux)

autogen.sh
```diff
@@ -8,5 +8,5 @@ autoreconf -v -f --install || exit 1
 	--enable-maintainer-mode \
 	--disable-shared \
 	--without-libpam \
-	--with-selinux \
```
brauner: You talk about systemd but you disable selinux here so I'm confused what the issue is.

sulincix: 😂.

sulincix: selinux not needed.

hallyn: We're confused what you're actually trying to do.  Your commit message only and repeatedly mentioned systemd, not selinux, but you don't change anything to do with systemd.  Did a chunk of your patch get lost?

PR #226, opened 2020-03-04
useradd: clarify the useradd -d parameter behavior in man page

ikerexxe:
Explanation: clarify the useradd -d parameter as it does create directory HOME_DIR if it doesn't exit.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1677005


hallyn:
Thanks for the patch.  So long as we are touching that sentence I would like to change it a bit more to mamke it flow better.  I'll update your PR if that's ok with you.

ikerexxe:
Yes, please do so. If you want I can do it given that you tell me what to change.

PR #231, opened 2020-03-11
Removed hard-coded default mail spool in useradd

gjzkrug:
The useradd program should be consistent with userdel and usermod and use the
MAIL_SPOOL_DIR variable as the default spool, if it is defined. Otherwise,
don't create a new mailbox, because it won't be cleaned up by userdel when run
with the -r flag.

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2203,8 +2203,13 @@ static void create_mail (void)
 		mode_t mode;
 
 		spool = getdef_str ("MAIL_DIR");
+#ifdef MAIL_SPOOL_DIR
+		if ((NULL == spool) && (getdef_str ("MAIL_FILE") == NULL)) {
+			spool = MAIL_SPOOL_DIR;
+		}
+#endif /* MAIL_SPOOL_DIR */
 		if (NULL == spool) {
-			spool = "/var/mail";
+			return;
 		}
 		file = alloca (strlen (prefix) + strlen (spool) + strlen (user_name) + 2);
 		if(prefix[0])


hallyn:
Agreed they should be consistent, though changing something like this after all this time can be scary.

I don't like the fact that the MAIL_SPOOL_DIR value depends on whether /var/mail exists on the *build* host at configure time.

src/useradd.c
```diff
@@ -2203,8 +2203,13 @@ static void create_mail (void)
...
 
 		spool = getdef_str ("MAIL_DIR");
+#ifdef MAIL_SPOOL_DIR
+		if ((NULL == spool) && (getdef_str ("MAIL_FILE") == NULL)) {
+			spool = MAIL_SPOOL_DIR;
+		}
+#endif /* MAIL_SPOOL_DIR */
 		if (NULL == spool) {
-			spool = "/var/mail";
+			return;
```
hallyn, line 2212: MAIL_FILE, according to login.defs manpage,
    Defines the location of the users mail spool files relatively to their home directory.
So for userdel , returning here makes sense, as $HOME will be deleted anyway.
But if MAIL_FILE is defined, we should probably create it here,
not exit early.  Right?

lnussel: I'm not the author of the patch but found the same issue. So far useradd never did anything with MAIL_FILE. So the patch can't break that :-) The patch does fix an issue though. On systems that do not have /var/mail, useradd prints an error message at the moment, even when MAIL_SPOOL_DIR was intentionally either not set during build or detected to be a different location and /var/mail not a symlink pointing there.

hallyn: thanks.

PR #234, opened 2020-03-12
Adding run-parts for userdel

### LINKED ISSUE #79: Consider adding an option to remove user journals on userdel
Maybe systemd should be handling this itself but just in case,
would it be possible to have userdel have an option to remove /var/log/journal/<machine_id>/user-1001*.journal when the user with id 1001 is deleted?
This would prevent the next user with the same id from reading the old user's data.
For example, when doing userdel testuser2 and useradd testuser3 since they will have the same user id.
userdel already offers to remove mail spool data through an option.

edneville:
This should enable additional user tidy-up when running userdel.

Happy to repeat elsewhere if this is acceptable.

Closes #79.

### DIFF
--- a/lib/run_part.c
+++ b/lib/run_part.c
@@ -1,0 +1,101 @@
+#include <dirent.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+int run_part (char *script_path, char *name, char *action)
+{
+	int pid;
+	int wait_status;
+	int pid_status;
+	char *args[] = { script_path, NULL };
+
+	pid=fork();
+	if (pid==-1){
+		perror ("Could not fork");
+		return 1;
+	}
+	if (pid==0) {
+		setenv ("ACTION",action,1);
+		setenv ("SUBJECT",name,1);
+		execv (script_path,args);
+		perror ("execv");
+		exit(1);
+	}
+
+	pid_status = wait (&wait_status);
+	if (pid_status == pid) {
+		return (wait_status);
+	}
+
+	perror ("waitpid");
+	return (1);
+}
+
+int run_parts (char *directory, char *name, char *action)
+{
+	struct dirent **namelist;
+	int scanlist;
+	int n;
+	int execute_result;
+
+	scanlist = scandir (directory, &namelist, 0, alphasort);
+	if (scanlist<0) {
+		return (0);
+	}
+
+	for (n=0; n<scanlist; n++) {
+		int path_length;
+		struct stat sb;
+
+		path_length=strlen(directory) + strlen(namelist[n]->d_name) + 2;
+		char *s = (char*)malloc(path_length);
+		if (!s) {
+			printf ("could not allocate memory\n");
+			for (; n<scanlist; n++) {
+				free (namelist[n]);
+			}
+			free (namelist);
+			return (1);
+		}
+		snprintf (s, path_length, "%s/%s", directory, namelist[n]->d_name);
+
+		execute_result = 0;
+		if (stat (s, &sb) == -1) {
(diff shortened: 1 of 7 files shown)


hallyn:
Is /etc the right place for this?  Would /usr/share or something be better?  Is the intent for distro-specific packages to ship some hooks, or only for admins to be able to define them?  (/etc seems right in the latter case)  Do you have a specific use case for this?

edneville:
/etc feels like the right place for it, sort of akin to /etc/skel which is user setup time. It seems to me the sort of place that an admin would look.

At my day job we have some "cloud" computers, which unfortunately allow people to create users via the web portal. I'd like to capture as much of that action at useradd time, collect the tty, time, source and send that as a mail so the originator can be contacted. Clouds probably have specific end points that portal work comes from which would make life easier also.

Other ways may also exist to do the above, but if I can think of this, there's probably lots of things I can't think of also. Audit does seem to be a good starting point. Log aggregation may also benefit from this if particular strings are added in syslog that show were the additions and deletions come from in a way that is easy to manage.

lib/run_part.c
```diff
@@ -0,0 +1,86 @@
...
+
+		if (S_ISREG (sb.st_mode)) {
+			execute_result = run_part (s, name, system);
+		}
+
+		free (s);
+		free (namelist[n]);
+
+		if (execute_result!=0) {
+			break;
```
hallyn: In this case you'll leak namelist[n+1..scanlist]

edneville: Good point, this should be fixed now.

edneville:
@hallyn, I've updated the pull request with garbage collection. Thanks to spotting that.

hallyn:
Hi


Please squash these all into one commit.

The api for talking to the scripts feels a little odd to me.  Are you already using this?  In particular, I mean passing "USERDEL=username" to the script, instead of something like ACTION=userdel SUBJECT=username or something.  If you're already using it, we can go with how it is, and maybe my aesthetic sense is wrong here?

In the run_parts loop, you only execute if the file is S_ISREG, but what about a symlink?

Also, if the script fails, perhaps you should log which script fails?

Other than that, I'm happy to merge this.

edneville:
Hi @hallyn 

I can't squash the commits in my branch (as far as I know) without deleting the branch and creating a new pull request, which then might clutter the original issue ticket. I /think/ you can squash the merge though which should collapse the commits into one commit.

This isn't in production use yet, only test, so changing the environment variable names isn't a problem for me.

Thanks for the comments, looking forward to getting this one crossed off.

hallyn:
Hi,

try this:

git checkout 79_userdel # or whatever your current local branch name for this pr is
git checkout -b 79_userdel.rebase
git rebase -i HEAD~6
# in the editor, replace each 'pick' - except the first one\ - with 's' for squash.
git push --force edneville HEAD:79_userdel  # this is assuming your local name for your own remote is 'edneville'

I'll do it if you prefer, but you most certainly should be able to :)

edneville:
Hi Serge,

Thanks for that. I've updated my branch, I didn't think a pushed branch 
could be re-written, every day, something new!

Ed

edneville:
Hi @hallyn, is this waiting on anything from me or could this be merged now?

hallyn:
Oh, sorry - my implicit mental filter mistakenly thought there was more to discuss.

I'll take one last brief look and merge, thanks!

hallyn:
Just ONE thing - in some of the run_parts error cases you do not free namelist itself.  We'll be exiting immediately so it's hard to care too much, but one never knows how the code will evolve.

edneville:
@hallyn, yes, I see your point. I've updated the PR to free.

hallyn:
Great, thanks - there's only one other little detail - in the last hunk where you add 'perror', you do the perror after the frees.  That will change the errno value so perror won't report stat's errno.

hallyn:
I'll go ahead and make the change and merge.

PR #235, opened 2020-03-15
Increase default max UID

MorbidMind:
Increase max UID to include users created by systemd-homed.
Some applications, e.g. SDDM use login.defs to determine max UID of displayed users on login screen.

### DIFF
--- a/etc/login.defs
+++ b/etc/login.defs
@@ -236,7 +236,7 @@ CRACKLIB_DICTPATH	/var/cache/cracklib/cracklib_dict
 # Min/max values for automatic uid selection in useradd(8)
 #
 UID_MIN			 1000
-UID_MAX			60000
+UID_MAX			65000
 # System accounts
 SYS_UID_MIN		  101
 SYS_UID_MAX		  999


brauner:
Can you explain this a little more how systemd-homed comes into play here. @poettering is this needed?

MorbidMind:
I created accounts with homed and they got UIDs over 60000. This caused SDDM not to show them on the list of accounts to log into.

hallyn:
> I created accounts with homed and they got UIDs over 60000. This caused SDDM not to show them on the list of accounts to log into.

But why did it do that?  In particular, I could imagine that systemd-homed might specifically want to use ids above UID_MAX to not interfere with adduser.  Or, it might just be using a different configuration file to pick max uid to use, but then it's possible that file might specify a number even higher than 65k.

hallyn:
Ping?

I'm not sure that this will work the way you intend.  In particular, find_new_uid() says:

```
153 /*
154  * find_new_uid - Find a new unused UID.
155  *
156  * If successful, find_new_uid provides an unused user ID in the
157  * [UID_MIN:UID_MAX] range.
158  * This ID should be higher than all the used UID, but if not possible,
159  * the lowest unused ID in the range will be returned.
160  *
161  * Return 0 on success, -1 if no unused UIDs are available.
162  */
```
So once homed allocates 60k, the next uid allocated by shadow would be 60001, which is probably not what homed wanted.

poettering:
Sorry for not responding more timely. We picked the UID range we default to in homed precisely to avoid stepping needlessly into the territory adduser/useradd  allocate from. Hence if that default would change then our choice of defaults would be pointless.

Hence, please do not change the upstream shadow defaults, it would be against what we intend with homed.

hallyn:
thanks @poettering .

@MorbidMind , please work with upstream for your login manager to have them get the needed information from homed.

PR #237, opened 2020-03-30
Check only local groups when adding new supplementary groups to a user

ikerexxe:
usermod and useradd: check only local groups with -G option

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1727236

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -211,6 +211,7 @@ static void get_defaults (void);
 static void show_defaults (void);
 static int set_defaults (void);
 static int get_groups (char *);
+static struct group * get_local_group (char * grp_name);
 static void usage (int status);
 static void new_pwent (struct passwd *);
 
@@ -220,7 +221,10 @@ static void grp_update (void);
 
 static void process_flags (int argc, char **argv);
 static void close_files (void);
+static void close_group_files (void);
+static void unlock_group_files (void);
 static void open_files (void);
+static void open_group_files (void);
 static void open_shadow (void);
 static void faillog_reset (uid_t);
 static void lastlog_reset (uid_t);
@@ -732,6 +736,11 @@ static int get_groups (char *list)
 	}
 
 	/*
+	 * Open the group files
+	 */
+	open_group_files ();
+
+	/*
 	 * So long as there is some data to be converted, strip off
 	 * each name and look it up. A mix of numerical and string
 	 * values for group identifiers is permitted.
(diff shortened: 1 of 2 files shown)


src/usermod.c
```diff
@@ -328,12 +336,52 @@ static int get_groups (char *list)
 	 * Any errors in finding group names are fatal
 	 */
 	if (0 != errors) {
+		close_group_files ();
 		return -1;
 	}
 
+	close_group_files ();
```
t8m: Would it make sense to avoid the duplication here and move the close before the if()?

src/useradd.c
```diff
@@ -805,12 +814,54 @@ static int get_groups (char *list)
 	 * Any errors in finding group names are fatal
 	 */
 	if (0 != errors) {
+		close_group_files ();
```
t8m: And here as well I'd avoid the duplication and move the close/unlock before the if().

src/usermod.c
ikerexxe: As you say it makes more sense to close the files before the if(), in order to minimize future errors.

src/useradd.c
ikerexxe: As you say it makes more sense to close the files before the if(), in order to minimize future errors. I have just pushed the changes.

ikerexxe:
@hallyn could you please review this change?

hallyn:
Thanks, but it''s hard to review as is.  The only explanation of why is a pointer to https://bugzilla.redhat.com/show_bug.cgi?id=1727236 , but I'm not authorized to see that.

Please add a detailed rationale in the commit message.

t8m:
@hallyn I've opened up the RHEL bug report so you should be able to read it now.

hallyn:
> @hallyn I've opened up the RHEL bug report so you should be able to read it now.

Thanks.

hallyn:
It seems like these functions that are repeated in useradd and usermod could be done in lib/ or libmisc/ instead, right?

hallyn:
> It seems like these functions that are repeated in useradd and usermod could be done in lib/ or libmisc/ instead, right?

never mind, i see, i should probably restructure that set of files myself.

hallyn:
Thanks - IIUC this also fixes the FIXME in useradd.c following the old prefix_getgr_nam_gid call, right?

ikerexxe:
Yes, it's also fixed as this change only allows local groups to be added as supplementary groups.

PR #242, opened 2020-04-10
login.defs: warn about weak choices

topimiettinen:
According to crypt(5), MD5 and DES should not be used for new
hashes. Also the default number of SHA rounds chosen by libc is orders
of magnitude too low for modern hardware. Let's warn the users about
weak choices.

Signed-off-by: Topi Miettinen <toiwoton@gmail.com>


brauner:
Thanks!

PR #243, opened 2020-04-11
print a warning from useradd if -u is used with uid number outside ra…

### LINKED ISSUE #240: Stop the operation or print a warning from useradd if -u is used with uid number outside range.
eg:
     $ useradd -u 456 test456 
     $ useradd -r  -u 1050 test1050

    In the manual
     -r, --system 
     System users will be created with no aging information in /etc/shadow, and their numeric identifiers are chosen in the SYS_UID_MIN-SYS_UID_MAX range
     UID_MAX (number), UID_MIN (number)
           Range of user IDs used for the creation of regular users by useradd or newusers.
     

test456 is a regular users,but 456 outside of the UID_MIN and UID_MAX range.
1050 is a system user ,but 1050 outside of the SYS_UID_MIN and SYS_UID_MAX range. 
It is better to stop the operation or print a warning from useradd if -u is used with uid number outside range.

blueskycs2c:
fix #240

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -228,6 +228,7 @@ static void tallylog_reset (const char *);
 static void usr_update (void);
 static void create_home (void);
 static void create_mail (void);
+static void check_uid_range(int rflg, uid_t user_id);
 
 /*
  * fail_exit - undo as much as possible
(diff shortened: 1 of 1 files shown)


src/useradd.c
```diff
@@ -2257,6 +2258,27 @@ static void create_mail (void)
...
+static void check_uid_range(int rflg, uid_t user_id)
+{
+	uid_t uid_min ;
+	uid_t uid_max ;
+	if(rflg){
+		uid_min = (uid_t)getdef_ulong("SYS_UID_MIN",101UL);
+		uid_max = (uid_t)getdef_ulong("SYS_UID_MAX",getdef_ulong("UID_MIN",1000UL)-1);
+		if(uid_min <= uid_max){
+			if(user_id < uid_min || user_id >uid_max)
+				printf(_("warning: %s's uid %d outside of the SYS_UID_MIN %d and SYS_UID_MAX %d range.\n"),user_name, user_id, uid_min, uid_max);
```
hallyn, line 2270: I've already merged this, but now it occurred to me - should these go to stderr instead?

blueskycs2c: > 
> 
> I've already merged this, but now it occurred to me - should these go to stderr instead?

@hallyn 
I don't think so.  This is a warning other than a error.

t8m: No, warnings should go to stderr.

blueskycs2c: > 
> 
> No, warnings should go to stderr.

ok, I will prepare a patch for this.

PR #244, opened 2020-04-17
Fix hurd build

sthibaul:
Do not include <sys/prctl.h> we don't have <sys/capability.h>, we don't
need prctl in that case anyway.

Signed-off-by: Samuel Thibault <samuel.thibault@ens-lyon.org>


hallyn:
One day if we need prctl for something else this might confuse someone, but it's good for now :)

thanks.

PR #245, opened 2020-04-18
remove unused and misleading 'owner' argument from find_new_sub*

hallyn:
Signed-off-by: Serge Hallyn <shallyn@cisco.com>


brauner:
Huh, so we've never used them. Fair enough.

PR #247, opened 2020-04-25
Remove unused variables

jubalh:


brauner:
Thanks! We should really register this project with Coverity. Let me see if I can do that.

PR #248, opened 2020-04-25
travis: add more architectures + enable Coverity

brauner:
Signed-off-by: Christian Brauner <christian@brauner.io>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,38 @@
+dist: bionic
 sudo: false
 
 language: c
 
 compiler:
-  - gcc
-  - clang
+ - gcc
+ - clang
+
+arch:
+ - amd64
+ - arm64
+ - ppc64le
+ - s390x
+
+env:
+  global:
+   # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
+   #   via the "travis encrypt" command using the project repo's public key
+   - secure: "D3C+O5UjN8Oe0FROMk5yPRU0he4UNAaHdxwSVG2373ImzYIs3pIvre1kXzeQSajbXCRKiKE8p2U0TooIRZAo/JeaZDgb4dBhij4c4/Qv8MpFCTC1W7cgo6UeOQeZwdfqp7v9Zz5ubh6RW68bmSM2naqQim1zuD1vDSGy4Wir7z4="
+
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get -y install -qq automake autopoint xsltproc libselinux1-dev gettext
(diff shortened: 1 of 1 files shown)


jubalh:
AFAIK they only allow a certain number of builds (per day?) and for that reason they advise to have a dedicated branch and not run on master, if I remember correctly.

brauner:
> AFAIK they only allow a certain number of builds (per day?) and for that reason they advise to have a dedicated branch and not run on master, if I remember correctly.

True, but we don't have that many merges that we need to worry about that?

brauner:
> > AFAIK they only allow a certain number of builds (per day?) and for that reason they advise to have a dedicated branch and not run on master, if I remember correctly.
> 
> True, but we don't have that many merges that we need to worry about that?

Sincere question. :)

jubalh:
> True, but we don't have that many merges that we need to worry about that?

Probably :) I don't know the exact number of requests. But I thought if for some time you don't merge anything. And then have some free time slot and merge 20 requests on one day, maybe it already is too much.

So I don't know just wanted to mention it in case you were not aware of the limitation :)

.travis.yml
```diff
@@ -1,16 +1,38 @@
...
-  apt:
-    packages:
-      - autopoint
-      - xsltproc
+  coverity_scan:
+    project:
+      name: "shadow-maint/shadow"
+      description: "Upstream shadow utils tree"
+    notification_email: christian.brauner@ubuntu.com,serge@hallyn.com
+    build_command_prepend: "./autogen.sh --without-selinux --disable-man"
```
hallyn, line 33: Unrelated to this change, but I've beenw ondering whether we should not do --disable-man.

hallyn:
Thanks.

.travis.yml
brauner: We probably should...

.travis.yml
```diff
@@ -1,16 +1,38 @@
...
+
+env:
+  global:
+   # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
+   #   via the "travis encrypt" command using the project repo's public key
+   - secure: "D3C+O5UjN8Oe0FROMk5yPRU0he4UNAaHdxwSVG2373ImzYIs3pIvre1kXzeQSajbXCRKiKE8p2U0TooIRZAo/JeaZDgb4dBhij4c4/Qv8MpFCTC1W7cgo6UeOQeZwdfqp7v9Zz5ubh6RW68bmSM2naqQim1zuD1vDSGy4Wir7z4="
+
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get -y install -qq automake autopoint xsltproc libselinux1-dev gettext
```
hallyn, line 24: Oh, ...  switching from the yaml based list of packages to install seems like maybe a regression?  Arguable, I guess.

PR #250, opened 2020-04-25
[strawman] Implement libsubid

hallyn:
This implements a proposd new libsubid with an API to allow other daemons and programs to coordinate subuid and subgid manipulations:

```
struct subordinate_range **get_subuid_ranges(char *owner);
struct subordinate_range **get_subgid_ranges(char *owner);
void subid_free_ranges(struct subordinate_range **ranges);

int get_subuid_owners(uid_t uid, uid_t **owner);
int get_subgid_owners(uid_t uid, uid_t **owner);

/* range should be pre-allocated with owner and count filled in, start is
 * ignored, can be 0 */
bool grant_subuid_range(struct subordinate_range *range, bool reuse);
bool grant_subgid_range(struct subordinate_range *range, bool reuse);

bool free_subuid_range(struct subordinate_range *range);
bool free_subgid_range(struct subordinate_range *range);
```

It also gets travis automatically running the libsubid testcases.


hallyn:
ping, review - especially at the api level, of anyone who might want to use this, would be great.

Many container related users of course would use go bindings (and maybe I'll offer those here), so the api at that level may be slightly different anyway...

hallyn:
@ping - I will merge this soon if there are no comments.

rhatdan:
@giuseppe PTAL

rhatdan:
LGTM

giuseppe:
thanks, the API seems fine to me

Issue #239, opened 2020-04-06
Allow directory "/nonexistent" or the like


lifecrisis:
Greetings:

I use Debian, so robot users on my system have home directory set to `/nonexistent` in the `/etc/passwd` file.

Normally this is fine, but when my team builds custom Debian FAI images, we would like to be able to run `pwck` as part of the process.  This becomes problematic because users with `/nonexistent` home directories are flagged as having bad `/etc/passwd` entries (causing exit status 2).

I would like to submit a patch that allows for explicitly setting the home directory to `/nonexistent` when a user is not supposed to have a home.

Would such a patch be accepted?  I want to ask before doing the work...

Thanks!

vapier:
this doesn't sound like it has wider distro/*NIX usage, so adding it to the canonical package sounds inappropriate to me.

how about adding a configure flag or runtime config option for it instead ?

lifecrisis:
That sounds fine to me.  I think a runtime configuration option makes the most sense, so that no one has to re-build if they want to change it.

Would it be okay if I worked on a PR to add it?

vapier:
i think it sounds fine, but i'm not the one who would be merging the PR :)

hallyn:
The earlier part of your description makes sense, but I'm not sure I'm following this part:

```
I would like to submit a patch that allows for explicitly setting the home directory to /nonexistent when a user is not supposed to have a home.
```

Are you wanting for pwck to simply ignore a nonexisting homedir if it matches a particular path (which can be passed to ./configure), or for pwck to actually set the homedir to that path?  If the latter, how would pwck decide which accounts to do this for?

lifecrisis:
@hallyn:  Definitely the former.  If the home directory path matches a particular string set either at compile-time or in runtime config file, then `pwck` will not return an error when a user has this home directory and it doesn't exist.

hallyn:
I'm fine with this - you could even add it to login.defs?

lifecrisis:
@hallyn Great!

Sure, adding it to `login.defs` sounds like a good idea.  I'll get to work on a patch. :)

lifecrisis:
@hallyn -- Friendly update: See the PR that mentions this issue.

I think there is a lot that could be done here.  Of course, how much more _needs_ to be done is up for debate.  I went for the minimal solution in the PR that fixes the issue for Debian-based systems.

PR #251, opened 2020-05-08
Add "NONEXISTENT" to "login.defs"

lifecrisis:
Greetings:

This is my initial attempt at a patch that will help the `pwck` utility function more helpfully on systems that use a special string to indicate that a user account intentionally does not have a home directory on the system.

For example, on Debian, it would be normal to see an `/etc/passwd` entry such as the following:

```
lp:x:7:7:lp,,,,:/nonexistent:/usr/sbin/nologin
```

The special directory string `/nonexistent` does not refer to a path on the system.  Instead, it indicates that this account does not have a home directory on purpose.

However, as `pwck` is currently written, it will generate the following errors for such accounts.  For example, using `sudo pwck -r` to validate the `/etc/passwd` file, we see:

```
user 'lp': directory '/nonexistent' does not exist
user 'news': directory '/nonexistent' does not exist
user 'uucp': directory '/nonexistent' does not exist
...
user 'list': directory '/nonexistent' does not exist
user 'pulse': directory '/nonexistent' does not exist
user '_apt': directory '/nonexistent' does not exist
user 'nobody': directory '/nonexistent' does not exist
pwck: no changes
```

This is not desirable for situations in which we need to use `pwck` as part of an automated process.  In these scenarios, we would prefer to simply check the exit status of the command to verify our `/etc/passwd` file.  My team would like to use `pwck` in this way for custom FAI builds of Debian.

This PR will add a new setting to `login.defs` that will be used to indicate when a system account does not need a home directory.  This will prevent useless errors from `pwck`.

Currently, the setting acts as a last stage filter for checking this error condition.  I.e., if a system user has a home directory string that doesn't name a path on disk, then `pwck` will now compare that string to the `NONEXISTENT` setting.  If they match, no error will be printed.  The usage of this option could be made more all-encompassing, but I decided to fix the minimal case for now, keeping the PR small.

Let me know what you think!

Closes: #239.


hallyn:
This looks ok to me, thanks.

The 'Edit NONEXISTENT" patch should probably be folded into the patch defining NONEXISTENT.xml, if you're comfortable enough with git rebasing.

lifecrisis:
@hallyn: I rebased, as requested.  There were two commits with minor edits that have been squashed into ancestors. I also uploaded my public key to GitHub so that each commit is now properly marked as verified.

Let me know if anything else needs to be addressed!

hallyn:
Thanks, looks good.

PR #254, opened 2020-05-10
[RFC] Add --exec-command / -e to `su` for shell-bypass

vcaputo:
This adds the ability to run commands via `su` without an interpretive shell, enabling a thinner user-switching veneer over execve() mode of operation.


hallyn:
Thanks - this looks good to me.

I'm a bit skiddish about the possibility that, however correct it is, it might break someone's usage in some script somewhere from way back when.  So I'm going to take another look with fresh eyes soon, before I merge.

hallyn:
Actually, you marked this as 'draft' - are you intending to make more changes?

vcaputo:
@hallyn Well, I didn't update the docs in this PR, which I presumed would be necessary before merging.  So I figured leave it as an RFC draft in case we need to bikeshed a bit on the flag naming or anything else before updating the docs/man page.

But no, I didn't have any planned future modifications to the code.

My only apprehension with this PR is I fairly quickly did the getopt replacement in frustration and didn't make an exhaustive effort to verify I replicated exactly all of the error messages getopt might produce with bad flags.  I made *some* effort, but I didn't go crazy with it, like I didn't scrutinize the getopt code to run down all the error paths and ensure I replicated all of them.  I'm not sure how important that aspect is.

hallyn:
Hi,

in PR #256 I proposed an alternate fix for the 'user' handling in su.

While doing this I noticed that the argument handling for -c is wrong anyway - it appears if you do

```
su -c /bin/cat -- /somefile
```

It will do

```
$shell -c /bin/cat /somefile
```
but it would need to do

```
$shell -c "/bin/cat /somefile"
```
for the arguments to get sent to the command.  It seems an easy enough fix, but I didn't do it just now.

hallyn:
@vcaputo how would you feel about rebasing this patch set on top of PR#256 instead of patch 1/3 ?

hallyn:
ping?  I may go ahead and rebase it myself this weekend if you don't have time but don't object.

vcaputo:
See comments in #256

vcaputo:
BTW as demonstrated in #256, the way the POSIX shell handles subsequent arguments following `-c foo` makes the existing `su -c foo -- args to foo` *very* awkward and unintuitive to anyone actually attempting to make this work, even when there aren't `su` bugs interfering.

I don't know anyone who would naturally think to add explicit parameters suffixing the `-c foo` argument, but that's what's needed to actually plumb the arguments through the intermediate shell.  i.e. `su -c 'foo $0 $1' -- arg arg` etc...  At least not without some reading of the shell man page:
```
       -c        If the -c option is present, then commands are read from  the
                 first non-option argument command_string.  If there are argu‐
                 ments  after  the  command_string,  the  first  argument   is
                 assigned  to  $0  and any remaining arguments are assigned to
                 the positional parameters.  The assignment  to  $0  sets  the
                 name  of  the  shell, which is used in warning and error mes‐
                 sages.
```
predicated on the realization that there's another shell gating things in the middle there, so go look at its man page about *its* -c handling.

So, if some form of --exec lands, I'd like to see mentioning something about this in the man page - perhaps even promoting --exec use over --command or demoting --command in the manual as something you should only use if shell features are required.  The way argument passing behaves with --exec is much more sane IMHO.

HarmtH:
I don't think this PR is needed, as ```sh -s``` already does this?

vcaputo:
@HarmtH I presume you meant `su -s`.

It's not equivalent because that's `su` executing an arbitrary command as an *interactive* shell.  When `su --command` is used, the controlling tty is closed.  When `su --shell` without a command is used, the controlling tty remains.

And if you attempt to somehow contort this into working by mixing `--command` and `--shell`, you'll find there's always a `-c` separating what you passed to `--shell` and `--command`.  Because the argument of `--shell` is expected to be a POSIX-compliant shell accepting `-c command`.

`su` simply doesn't support executing commands without a shell.

HarmtH:
Yes, I meant ```su -s``` indeed 😃. Can you give me an example of the controlling tty being closed? I seem to be able to ^C programs with ether ```su -s``` or  ```su -c```.

vcaputo:
Here's the relevant code in su https://github.com/shadow-maint/shadow/blob/master/src/su.c#L1110

hallyn:
> BTW as demonstrated in #256, the way the POSIX shell handles subsequent arguments following `-c foo` makes the existing `su -c foo -- args to foo` _very_ awkward and unintuitive to anyone actually attempting to make use of `--`, even when there aren't `su` bugs interfering.

Yes, I thought I'd mentioned that in my pr but I guess I edited that out in a rebase.

I'm going to think just a bit more about this and then probably merge it and update the manpage.

vcaputo:
I'm in no hurry - it'll be forever before this percolates into my molasses-like distro anyways.  What started out as just wanting to add a shell bypass mode to `su` turned into finding a dusty skeleton in this old closet.

Honestly, I didn't exhaustively test this PR, because it was apparent that there'd appropriately be some friction, and likely discussion and iteration necessary if it would land at all.  I just made things presentable enough to try move things forward, get the ball rolling on upstream understanding what had been uncovered.

So definitely take care in sanity checking this stuff before merging.  I'll see if I can make some time to do more testing of this PR myself as well.

Let's not replace an existing argument handling bug with new ones out of haste, especially with a suid binary.

hallyn:
Yeah - I was planning on adding some tests based on the ones you showed above.

Thanks for this.

HarmtH:
> Here's the relevant code in su https://github.com/shadow-maint/shadow/blob/master/src/su.c#L1110

I meant an example which showes the necessity / use case of the controlling terminal being closed :sweat_smile: Maybe unnecessary to say, but having a controlling terminal and being an interactive shell are two entirely different concepts. You can have a non-interactive shell with a controlling terminal, and you can have an interactive shell without a controlling terminal (although you won't have job control then).

Btw, as soon as you give additional arguments to a (non-)shell passed to ```-e```, the controlling terminal also already gets closed, see https://github.com/shadow-maint/shadow/blob/master/src/su.c#L832

alejandro-colomar:
What's the status of this?

PR #256, opened 2020-05-17
[discussion] Fix -- handling in su

hallyn:
Su incorrectly tries to handle some flexible usage, including

su -c command
su -c command user
su -c command - user
su -c command -- args
su -c command user -- args

etc.  But it ignored the fact that if getopt_long runs into a
--, it will leave optind at the next word.  So

su -c command -- arg

tried to parse 'arg' as a user.  Fix that by checking for it.

Signed-off-by: Serge Hallyn <shallyn@cisco.com>

### DIFF
--- a/src/su.c
+++ b/src/su.c
@@ -760,6 +760,20 @@ static void save_caller_context (char **argv)
 	pw_free (pw);
 }
 
+static void choose_default_user(void) {
+	struct passwd *root_pw = getpwnam ("root");
+	if ((NULL != root_pw) && (0 == root_pw->pw_uid)) {
+		(void) strcpy (name, "root");
+	} else {
+		root_pw = getpwuid (0);
+		if (NULL == root_pw) {
+			SYSLOG ((LOG_CRIT, "There is no UID 0 user."));
+			su_failure (caller_tty, true);
+		}
+		(void) strcpy (name, root_pw->pw_name);
+	}
+}
+
 /*
  * process_flags - Process the command line arguments
  *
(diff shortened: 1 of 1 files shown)


vcaputo:
@hallyn I'm short on time for volunteer FOSS stuff currently...

I briefly went down this route before replacing getopt in #254.  From what I recall, there was ambiguity created by getopt moving the "--" around in the argv, but maybe I had an unrelated bug misleading me.

If you don't find any problems with using getopt then great!  I found it to initially work in testing --exec, but broken in more of the edge cases like missing vs. present user name.

Feel free to rebase or use whatever commits from my PR, it makes no difference to me, I'd just like the --exec/-e feature in upstream su.

vcaputo:
Ok, I took a few minutes to poke at `su` with this PR applied, and reproduced the malfunction which sent me down the "there's a reason I stopped using getopt decades ago" route in #254 .

Start with a simple script at `/tmp/foo` for instrumenting the euid and passed-through args:
```
#!/bin/bash

whoami
echo args: "$@"
```

Now invoking #256 `su` with this valid commandline for running `/tmp/foo` as `foouser`, it spuriously passes the switched-to user as $1, and runs as `root` requiring root's password.

```
./su -c '/tmp/foo $0 $1 $2' - foouser -- first second third
Password: 
root
args: - foouser first
```

Now, repeating this exercise with #254 instead:

```
/su -c '/tmp/foo $0 $1 $2' - foouser -- first second third
Password: 
foouser
args: first second third
```

I hope that helps clear up the nature of the problem.

Issue #255, opened 2020-05-16
lastlog `From` IPv6 gets cropped


Inrin:
# Steps to reproduce
1. Login via an IPv6 SSH session
2. Logout
3. Login again
4. lastlog -u "$USER"

# Output
```
Username         Port     From                     Latest
dummy            pts/0    2001:db8:aa:bb:: Sat May 16 21:53:31 +0200 2020
````

# Should be
```
Username         Port     From                         Latest
dummy            pts/0    2001:db8:aa:bb::c00:ff:ee    Sat May 16 21:53:31 +0200 2020
````
# Considerations
Please note, that in this constructed example `2001:db8:aa:bb::` is an valid IPv6-Address.
(Despite the documentation factor of course) so in a special case a mix-up might occur.

I think the source code to consider is:
https://github.com/shadow-maint/shadow/blob/320707fcb0b84ac9c816890328f1fea2aabd7b46/src/lastlog.c#L175

Change `%-16.16s` to `%-16.46s` or similar.
The size should contain at least one of:
1) `INET6_ADDRSTRLEN = 46` as an IPv4-Mapped-IPv6 written out gets lengthy
2) `39` for 1234:5678:90AB:CDEF:1234:5678:90AB:CDEF
3) `39+1+MAX-INTERFACES = ` if zone indices are supported like in `fe80::1%2`. I don't know how many are supported but I don't think it is above 100000000000000 :laughing:
4) `39+1+IFNAMSIZ = 56` if zone indices are supported like in `fe80::1%enp4s0`

Although 3) uses less space it is more common to see 4) on UNIX-Like systems.

Or do it like https://github.com/karelzak/util-linux with their `last` command and add a parameter to display the `From` at the end without bounding it.

# EDIT
While coding #259 I noticed, that 1-4) could be reduced to 42 characters, as `ll.ll_host` is always in minimized form.
I added a comment inside the PR which explains it in detail.

hallyn:
Hi,

Thanks for raising this issue.  I'm fine with 4, and also (in addition) having the option to move hostname to the end.

Perhaps we should do one loop calculating the maximum hostname length, so the columns can be aligned.

Are you willing to write a patch+PR for this?

Inrin:
Yes I can do it.
But only at the weekend.

hallyn:
Awesome, thanks.

Inrin:
I created two PRs considering maximum width and the mentioned parameter.
[...]

PR #258, opened 2020-05-24
Add -F/--fromlast Parameter to lastlog

Inrin:
As mentioned in #255 here is the solution to add a parameter to display the `From` column at last without bounding as for example done in the `last` command.

An manpage entry has also been added.

Of course the parameter name could be changed, if the current one doesn't suite you.


hallyn:
If you think people might still want this in addition to the other one, please re-open.

PR #259, opened 2020-05-24
Add maximum padding to fit IPv6-Addresses

Inrin:
As mentioned in #255 here is a padding, which suits IPv6 Addresses and Link-Local Addresses aswell. I tried to keep it as small as possible, without loosing information.

As the complete width exceeds a typical `80` oldschool terminal, this might be non practical to include. I have more on this topic in #255 and guess this PR gets denied in favor of a more dynamic one.

### DIFF
--- a/src/lastlog.c
+++ b/src/lastlog.c
@@ -42,6 +42,9 @@
 #include <sys/types.h>
 #include <time.h>
 #include <assert.h>
+#ifdef HAVE_LL_HOST
+#include <net/if.h>
+#endif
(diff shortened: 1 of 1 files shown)


hallyn:
I feel like this is the better approach.  I'm tempted to merge this one, rather than the -F one.

Rationale: anyone parsing the results should hopefully doing based on fields and not chars; anyone using ipv6 should see the full address;  80 char cols can wrap.

Objections?

PR #262, opened 2020-06-05
correct grammar in shadow(5)

andydna:


edneville:
Could be wrong, but means would seem more correct:

    An empty field and value 0 means that there is no minimum

andydna:
Sure.  Well, smoother at least.  "are" instead of "is" just hurts.

PR #263, opened 2020-06-13
Option to suppress group/gshadow inconsistencies

### LINKED ISSUE #261: grpck: questionable warning
Hi,

I have an intention to set up a system as follows:

* User `me` should not have `test-sg` in his list of auxiliary groups when logging in
* User `me` should be able to explicitly assume this group when needed, using `sg`, without password

This intention is satisfied by making `me` a member of `test-sg` in `/etc/gshadow`, but not in `/etc/group`. Even more, such group transitions are properly logged in `auth.log`.

The only problem is that this (valid) setup triggers a warning from grpck:

    'me' is a member of the 'test-sg' group in /etc/gshadow but not in /etc/group

As explained above, this is a setup that has a 100% valid use case. So please make this check optional, or completely remove it.

Note: in this issue, I express no firm opinion regarding the opposite warning. Let's assume it is still valid, until someone proves otherwise.

edneville:
'gshadow' man page suggests that "You should use the same list of users
as in /etc/group", but not must.

Closes #261

### DIFF
--- a/man/grpck.8.xml
+++ b/man/grpck.8.xml
@@ -34,7 +34,6 @@
 <!-- SHADOW-CONFIG-HERE -->
 ]>
 <refentry id='grpck.8'>
-  <!-- $Id$ -->
   <refentryinfo>
     <author>
       <firstname>Julianne Frances</firstname>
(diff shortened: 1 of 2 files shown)


hallyn:
@edneville I will change the option name later today and merge, unless you reply to say you want to do it.

edneville:
On 2020-08-11 09:48-0700, Serge Hallyn wrote:
> @edneville I will change the option name later today and merge, unless
> you reply to say you want to do it.

More than happy for you to update at merge time to avoid some 
back-and-forth :-)

Ed

PR #267, opened 2020-07-12
chage: Prevent signed integer overflows.

stoeckmann:
This is merely a stability fix, not a security fix.

As the root user, it is possible to set time values which later on
result in signed integer overflows.

For this to work, an sgetspent implementation must be used which
supports long values (glibc on amd64 only parses 32 bit, not 64).
Either use musl or simply call configure with following environment
variable:

$ ac_cv_func_sgetspent=no ./configure

Also it is recommended to compile with -fsanitize=undefined or
-ftrapv to see these issues easily.

Examples to trigger issues when calling "chage -l user":

$ chage -d 9223372036854775807 user

$ chage -d 106751991167300 user
$ chage -M 9999 user

$ chage -d 90000000000000 user
$ chage -I 90000000000000 user
$ chage -M 9999 user

$ chage -E 9223372036854775807 user

While at it, I fixed casting issues which could lead to signed integer
overflows on systems which still have a 32 bit time_t.

### DIFF
--- a/src/chage.c
+++ b/src/chage.c
@@ -203,10 +203,10 @@ static int new_fields (void)
 		return 0;
 	}
 
-	if (-1 == lstchgdate) {
+	if (-1 == lstchgdate || lstchgdate > LONG_MAX / SCALE) {
 		strcpy (buf, "-1");
 	} else {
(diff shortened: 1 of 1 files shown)


hallyn:
Thank you.

PR #268, opened 2020-07-12
chfn: Prevent buffer overflow.

stoeckmann:
This is a stability fix, not a security fix, because the affected -o
option can only be used by root and it takes a modified passwd file.

If a gecos field for a user has BUFSIZ characters without commas and an
equals sign (i.e. a huge slop/extra field) and chfn is called with -o,
then a buffer overflow occurs.

It is not possible to trigger this with shadow tools. Therefore, the
passwd file must be modified manually.

I have fixed this unlikely case the easiest and cleanest way possible.
Since chfn bails out if more than 80 characters excluding commas are
supposed to be written into gecos field, we can stop processing early on
if -o argument is too long.

### DIFF
--- a/src/chfn.c
+++ b/src/chfn.c
@@ -61,7 +61,7 @@ static char fullnm[BUFSIZ];
 static char roomno[BUFSIZ];
 static char workph[BUFSIZ];
 static char homeph[BUFSIZ];
-static char slop[BUFSIZ];
+static char slop[BUFSIZ + 1 + 80];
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks.  So we're confident that the old_gecos is limited to BUFSIZE by get_old_fields, and you're limiting the extra to 80 chars, so this looks sufficient.  (I'm recording that here because I spent too much time trying to verify that :)

PR #269, opened 2020-07-26
NOCONFIGURE=1 ./autogen.sh support added

sulincix:
most of build system need to this I think.

Most of build method look like this:
```bash
NOCONFIGURE=1 ./autogen.sh
./configure --prefix=/usr
make
make install DESTDIR="$pkgdir"

```
Autogen.sh script did not support noconfigude. I added it


hallyn:
No come on.  Then just run autoreconf.

PR #274, opened 2020-08-28
Fix su -- handling and add some tests

hallyn:
A few things stood out to me here

1. ubuntu is using su from util-linux.  Should we drop su from shadow?
2. some the existing su testcases assume a different understanding of the
   -- handling.  Given that, is it at all safe to make changes, or should
   we just document the current stat of things and move on?


hallyn:
(hm, why is travis not running)

hallyn:
Meh - I thought I had it, but "- -- f" is still not working.

Yeah, ok.  Getopt reorders the contents so that if you pass

```
./src/su - -c /tmp/foo -- f s t
```

Then after getopt loop you get

```
argc is 8 optind 4
argv[0] = ./src/su
argv[1] = -c
argv[2] = /tmp/foo
argv[3] = --
argv[4] = -
argv[5] = f
argv[6] = s
argv[7] = t
```

So - is now after --.

vcaputo was right - we just need to ditch getopt here.

PR #275, opened 2020-08-28
Add tests on top of #254

hallyn:
This is to get travis to test for me.


brauner:
Now I'm confused. Does this need to be reverted because it _depends_ on a prior patchset to be merged?

hallyn:
D'oh, sorry, I thought I'd marked it a draft, but that was the previous one.

Indeed - the previous patchset which this depends on appears to be bad, so it'll need to be reverted. I'll just do it by hand.

hallyn:
As for the tests commit itself, I'm still not quite sure whether it in fact represents the behavior we want, or not.  Based on some existing tests, it might not be, although I feel the existing tests are simply wrong.  But sometimes my opinions don't mean much.

PR #277, opened 2020-09-16
gpasswd.c: fix memory leak in sg_adm

whzhe51:
Signed-off-by: whzhe <wanghongzhe@huawei.com>

### DIFF
--- a/src/gpasswd.c
+++ b/src/gpasswd.c
@@ -1204,6 +1204,17 @@ int main (int argc, char **argv)
 	nscd_flush_cache ("group");
 	sssd_flush_cache (SSSD_DB_GROUP);
 
+#ifdef SHADOWGRP
+	if (sgent.sg_adm) {
+		xfree(sgent.sg_adm);
+	}
+	if (sgent.sg_mem) {
+		xfree(sgent.sg_mem);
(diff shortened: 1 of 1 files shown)


brauner:
This looks reasonable to me. I'll let @hallyn take a loo at it too.

brauner (approved):
Looks reasonable to me!

hallyn:
Hm, so why sg_adm, but not sg_mem, and why not the grent-> ones?

Note this patch as is will break compile if !SHADOWGRP.

But I'm also failing to see the point, since it's right before exit.

whzhe51:
> Hm, so why sg_adm, but not sg_mem, and why not the grent-> ones?
> 
> Note this patch as is will break compile if !SHADOWGRP.
> 
> But I'm also failing to see the point, since it's right before exit.

I think you are right，and i fix the commits.

whzhe51:
> Hm, so why sg_adm, but not sg_mem, and why not the grent-> ones?
> 
> Note this patch as is will break compile if !SHADOWGRP.
> 
> But I'm also failing to see the point, since it's right before exit.

I think you are right, and i will fix the mr.

PR #278, opened 2020-09-20
loop until waitpid returns pid_child or error

### LINKED ISSUE #104: su: check whether waitpid() return value is equal to pid_child
When su runs as pid 1 (happened to me in Docker), it can get adopted children.
If such child exits, it makes su exit prematurely with its retcode.
It would be great to simply loop here https://github.com/shadow-maint/shadow/blob/master/src/su.c#L357 if pid != -1 && pid != pid_child.

edneville:
closes #104

### DIFF
--- a/src/su.c
+++ b/src/su.c
@@ -354,7 +354,9 @@ static void prepare_pam_close_session (void)
 			pid_t pid;
 			stop = true;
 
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks - looks good.

PR #279, opened 2020-09-22
Fix incorrect translation (de)

juergenhoetzel:


jubalh:
"Eine Systemkonto erstellen" that sounds much worse :)

juergenhoetzel:
> Eine Systemkonto erstellen" that sounds much worse :
:fearful: amended

Issue #276, opened 2020-09-02
uknown login.defs options


karelzak:
Hi guys, 

as you know the file /etc/login.defs is shared between more packages. I'd like to ask how problematic for shadow-utils is new foreign options in the file? For example, if we add in util-linux something (a new option) to the file --  do you want to be informed about such change or your login.defs parser silently ignores unknown options? 

From lib/getdef.c:def_find() it seems that in some cases it prints 
     _("configuration error - unknown item '%s' (notify administrator)\n")

I can prepare PR with the missing options (if you want it/need it).

brauner:
Hm, good question. I think we should probably coordinate on adding new options so being informed about such a change is probably a good way to go. That's what my gut tells me.

hallyn:
PR would be appreciated, thanks.

karelzak:
What about to make shadow-utils login.defs a little bit more promiscuous and do not report unknown options at all? We use the same solution in util-linux.

karelzak:
OK, PR to add foreign items (util-linux in this case) created.

hallyn:
Thanks.

PR #286, opened 2020-10-08
getdefs: add foreign (non-shadow-utils) items

karelzak:
The login.defs is shared between more upstream projects (util-linux,
etc.). We need to improve compatibility between the projects do not
report valid, but foreign items.

Addresses: https://github.com/shadow-maint/shadow/issues/276


hallyn:
Thanks.

PR #288, opened 2020-10-13
xfree: move xfree() function to xmalloc.c

whzhe51:
Signed-off-by: whzhe <wanghongzhe@huawei.com>

### DIFF
--- a/lib/prototypes.h
+++ b/lib/prototypes.h
@@ -448,6 +448,7 @@ extern bool valid (const char *, const struct passwd *);
 extern /*@maynotreturn@*/ /*@only@*//*@out@*//*@notnull@*/char *xmalloc (size_t size)
   /*@ensures MaxSet(result) == (size - 1); @*/;
 extern /*@maynotreturn@*/ /*@only@*//*@notnull@*/char *xstrdup (const char *);
+extern void xfree(void *ap);
 
 /* xgetpwnam.c */
 extern /*@null@*/ /*@only@*/struct passwd *xgetpwnam (const char *);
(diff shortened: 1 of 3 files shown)


lib/prototypes.h
```diff
@@ -448,6 +448,7 @@ extern bool valid (const char *, const struct passwd *);
 extern /*@maynotreturn@*/ /*@only@*//*@out@*//*@notnull@*/char *xmalloc (size_t size)
   /*@ensures MaxSet(result) == (size - 1); @*/;
 extern /*@maynotreturn@*/ /*@only@*//*@notnull@*/char *xstrdup (const char *);
+extern /*@null@*/void xfree(void *);
```
hallyn, line 451: Should the @null@ be by the argument?

whzhe51: I fix it to ‘notnull’

hallyn: But the function does not return a value.  Instead of

```
+/*@notnull@*/void xfree(void *ap)
```

doesn't it need to be
```
+void xfree(/*@null@*/void *ap)
```

libmisc/xmalloc.c
```diff
@@ -66,3 +66,10 @@
 {
 	return strcpy (xmalloc (strlen (str) + 1), str);
 }
+
+void xfree(/*notnull@*/void *ap)
```
hallyn, line 70: But ...  in fact it *can* be null :)  Then we just do nothing.

whzhe51: i think so :).

hallyn (approved):

hallyn:
Thanks.

PR #290, opened 2020-10-17
add parsing support for advertised force long opt

jamincollins:
Existing help output advertises --force as a long opt.

  -f, --force                   delete group even if it is the primary group of a user

But errors when the long opt is used.

groupdel: unrecognized option '--force'

Signed-off-by: Jamin W. Collins <jamin.collins@gmail.com>

### DIFF
--- a/man/groupdel.8.xml
+++ b/man/groupdel.8.xml
@@ -92,6 +92,17 @@
     </para>
     <variablelist remap='IP'>
       <varlistentry>
+	<term>
+	  <option>-f</option>, <option>--force</option>
+	</term>
+	<listitem>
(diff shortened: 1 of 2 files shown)


ikerexxe:
I think that the change is correct but I'd also add another patch changing the man page and include force options, since it's not documented. The explanation of this option could be as follows:
```
This option forces the removal of the group, even if there's some user having the group as the primary one.
```

jamincollins:
I'll see about getting it added to the man page.

ikerexxe:
LGTM

hallyn:
Thanks

PR #293, opened 2020-11-02
commonio: force lock file sync

ikerexxe:
lib/commonio.c: after writing to the lock file, force a file sync to
the storage system.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1862056

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -157,7 +157,17 @@ static int do_lock_file (const char *file, const char *lock, bool log)
 	if (write (fd, buf, (size_t) len) != len) {
 		if (log) {
 			(void) fprintf (stderr,
-			                "%s: %s: %s\n",
+			                "%s: %s file write error: %s\n",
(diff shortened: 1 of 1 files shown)


lib/commonio.c
```diff
@@ -157,7 +157,17 @@ static int do_lock_file (const char *file, const char *lock, bool log)
...
 			(void) fprintf (stderr,
-			                "%s: %s: %s\n",
+			                "%s: %s file write error: %s\n",
+			                Prog, file, strerror (errno));
+		}
+		(void) close (fd);
+		unlink (file);
+		return 0;
+	}
+	if (fsync (fd) == -1) {
```
socketpair, line 167: `fdatasync()` is better (faster), since mtime is not meaningful for this file.

ikerexxe: Updated with `fdatasync()`

hallyn:
Thank you.

PR #303, opened 2020-12-27
Add yescrypt support

### LINKED ISSUE #300: Add yescrypt as ENCRYPT_METHOD
Could `yescrypt` (with prefix `$y$`) be added as a choice for /etc/login.defs ENCRYPT_METHOD?
(This is the default hash generated by `crypt()` nowadays, and is supported by linux-pam since 1.4.0.)

breard-r:
This PR should fix #300

Although I have not done intensive testing, I have been able to set yescrypt-hashed passwords using `chpasswd -c YESCRYPT`.

### DIFF
--- a/etc/login.defs
+++ b/etc/login.defs
@@ -326,6 +326,7 @@ CHFN_RESTRICT		rwh
 # If set to SHA256, SHA256-based algorithm will be used for encrypting password
 # If set to SHA512, SHA512-based algorithm will be used for encrypting password
 # If set to BCRYPT, BCRYPT-based algorithm will be used for encrypting password
+# If set to YESCRYPT, YESCRYPT-based algorithm will be used for encrypting password
 # If set to DES, DES-based algorithm will be used for encrypting password (default)
 # MD5 and DES should not be used for new hashes, see crypt(5) for recommendations.
 # Overrides the MD5_CRYPT_ENAB option
@@ -366,6 +367,19 @@ CHFN_RESTRICT		rwh
 #
 #BCRYPT_MIN_ROUNDS 13
 #BCRYPT_MAX_ROUNDS 13
+
+#
+# Only works if ENCRYPT_METHOD is set to YESCRYPT.
+#
+# Define the YESCRYPT cost factor.
+# With a higher cost factor, it is more difficult to brute-force the password.
+# However, more CPU time and more memory will be needed to authenticate users
+# if this value is increased.
+#
+# If not specified, a cost factor of 5 will be used.
+# The value must be within the 1-11 range.
+#
+#YESCRYPT_COST_FACTOR 5
 
 #
 # List of groups to add to the user's supplementary group set
--- a/libmisc/salt.c
+++ b/libmisc/salt.c
@@ -32,6 +32,10 @@ static /*@observer@*/const char *SHA_salt_rounds (/*@null@*/int *prefered_rounds
 static /*@observer@*/const char *gensalt_bcrypt (void);
 static /*@observer@*/const char *BCRYPT_salt_rounds (/*@null@*/int *prefered_rounds);
 #endif /* USE_BCRYPT */
+#ifdef USE_YESCRYPT
+static /*@observer@*/const char *gensalt_yescrypt (void);
+static /*@observer@*/const char *YESCRYPT_salt_cost (/*@null@*/long *prefered_rounds);
+#endif /* USE_YESCRYPT */
 
 #ifndef HAVE_L64A
 static /*@observer@*/char *l64a(long value)
@@ -263,6 +267,78 @@ static /*@observer@*/const char *gensalt_bcrypt (void)
 }
 #endif /* USE_BCRYPT */
 
+#ifdef USE_YESCRYPT
+/* Default cost if not explicitly specified.  */
+#define Y_COST_DEFAULT 5
+/* Minimum cost.  */
+#define Y_COST_MIN 1
+/* Maximum cost.  */
+#define Y_COST_MAX 11
+/*
+ * Return a salt prefix specifying the cost for the YESCRYPT method.
+ */
+static /*@observer@*/const char *YESCRYPT_salt_cost (/*@null@*/long *prefered_cost)
+{
+	static char cost_prefix[5];
+	long cost;
+
+	if (NULL == prefered_cost) {
+		cost = getdef_num ("YESCRYPT_COST_FACTOR", Y_COST_DEFAULT);
+	} else {
+		cost = *prefered_cost;
+	}
+
+	if (cost < Y_COST_MIN) {
+		cost = Y_COST_MIN;
+	}
+	if (cost > Y_COST_MAX) {
+		cost = Y_COST_MAX;
+	}
+
+	cost_prefix[0] = 'j';
+	if (cost < 3) {
+		cost_prefix[1] = 0x36 + cost;
+	} else if (cost < 6) {
+		cost_prefix[1] = 0x34 + cost;
+	} else {
+		cost_prefix[1] = 0x3b + cost;
+	}
+	cost_prefix[2] = cost >= 3 ? 'T' : '5';
+	cost_prefix[3] = '$';
+	cost_prefix[4] = 0;
+
+	return cost_prefix;
+}
+
+/*
+ * Default number of base64 characters used for the salt.
+ * 24 characters gives a 144 bits (18 bytes) salt. Unlike the more
+ * traditional 128 bits (16 bytes) salt, this 144 bits salt is always
+ * represented by the same number of base64 characters without padding
+ * issue, even with a non-standard base64 encoding scheme.
+ */
+#define YESCRYPT_SALT_SIZE 24
+/*
+ *  Generate a 22 character salt string for yescrypt.
+ */
+static /*@observer@*/const char *gensalt_yescrypt (void)
+{
+	static char salt[32];
+
+	salt[0] = '\0';
+
+	seedRNG ();
+	strcat (salt, l64a (random()));
+	do {
+		strcat (salt, l64a (random()));
+	} while (strlen (salt) < YESCRYPT_SALT_SIZE);
+
+	salt[YESCRYPT_SALT_SIZE] = '\0';
+
+	return salt;
+}
+#endif /* USE_YESCRYPT */
+
 /*
  *  Generate salt of size salt_size.
  */
@@ -302,6 +378,7 @@ static /*@observer@*/const char *gensalt (size_t salt_size)
  * If meth is specified, an additional parameter can be provided.
  *  * For the SHA256 and SHA512 method, this specifies the number of rounds
  *    (if not NULL).
+ *  * For the YESCRYPT method, this specifies the cost factor (if not NULL).
  */
 /*@observer@*/const char *crypt_make_salt (/*@null@*//*@observer@*/const char *meth, /*@null@*/void *arg)
 {
@@ -333,6 +410,11 @@ static /*@observer@*/const char *gensalt (size_t salt_size)
 		BCRYPTMAGNUM(result);
 		strcat(result, BCRYPT_salt_rounds((int *)arg));
 #endif /* USE_BCRYPT */
+#ifdef USE_YESCRYPT
+	} else if (0 == strcmp (method, "YESCRYPT")) {
+		MAGNUM(result, 'y');
+		strcat(result, YESCRYPT_salt_cost((int *)arg));
+#endif /* USE_YESCRYPT */
 #ifdef USE_SHA_CRYPT
 	} else if (0 == strcmp (method, "SHA256")) {
 		MAGNUM(result, '5');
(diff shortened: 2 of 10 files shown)


libmisc/salt.c
```diff
@@ -263,6 +267,78 @@ static /*@observer@*/const char *gensalt_bcrypt (void)
...
+static /*@observer@*/const char *YESCRYPT_salt_cost (/*@null@*/long *prefered_cost)
+{
+	static char cost_prefix[5];
+        long cost;
+
+        if (NULL == prefered_cost) {
+	    cost = getdef_num ("YESCRYPT_COST_FACTOR", Y_COST_DEFAULT);
+        } else {
+            cost = *prefered_cost;
+        }
```
solardiz, line 289: Inconsistent indentation.

libmisc/salt.c
```diff
@@ -263,6 +267,78 @@ static /*@observer@*/const char *gensalt_bcrypt (void)
...
+        if (cost < 3) {
+            cost_prefix[1] = 0x36 + cost;
+        } else if (cost < 6) {
+            cost_prefix[1] = 0x34 + cost;
+        } else {
+            cost_prefix[1] = 0x3b + cost;
+        }
+        cost_prefix[2] = cost >= 3 ? 'T' : '5';
+        cost_prefix[3] = '$';
+        cost_prefix[4] = 0;
```
solardiz, line 308: Oh, this is a rather compact way to mimic (I guess) libxcrypt's translation of a single cost parameter to yescrypt's multiple ones, which it'd then encode with `yescrypt_encode_params_r()`. I didn't verify it, but I assume the resulting strings were checked to match what libxcrypt produces? If so, that's fine.

libmisc/salt.c
```diff
@@ -263,6 +267,78 @@ static /*@observer@*/const char *gensalt_bcrypt (void)
...
+ *  Generate a 22 character salt string for yescrypt.
+ */
+static /*@observer@*/const char *gensalt_yescrypt (void)
+{
+	static char salt[32];
+
+	salt[0] = '\0';
+
+	seedRNG ();
+	strcat (salt, l64a (random()));
```
solardiz, line 331: This leading `strcat()` call looks unneeded as the loop contains the same (will just do one more iteration of this line is omitted).

solardiz (commented):
Thank you for your work on this.

Not a real review, but I skimmed and made some comments. (I'm not approving nor requesting changes as I'm not a maintainer.)

As I tweeted in https://twitter.com/solardiz/status/1343593126410268673, longer-term or additionally (when recent libxcrypt is available at compile-time) we could switch shadow to our `crypt_gensalt*()` APIs:

I guess you're trying to mimic libxcrypt's translation of a single cost parameter to several of yescrypt's parameters. This is just right for your use case. Maybe you could have invoked libxcrypt's `crypt_gensalt*()` to have it do this for you (using yescrypt APIs under the hood).

Historically, we introduced those `crypt_gensalt*()` APIs in an Owl glibc patch in 2001, and indeed our shadow suite patches on Owl use them. ALT Linux, too, now including for yescrypt. As libxcrypt finally adopted our APIs, maybe shadow should adopt our patches to use them, too?

This removes all of the per-hash-type complexity from shadow, and eliminates the need to update it to support future hash types (adding them to libxcrypt is enough to get them supported). https://cvsweb.openwall.com/cgi/cvsweb.cgi/Owl/packages/shadow-utils/shadow-4.0.4.1-owl-crypt_gensalt.diff
 https://packages.altlinux.org/en/sisyphus/srpms/shadow http://git.altlinux.org/people/sem/packages/shadow.git

That said, perhaps I should also document the kind of translation you came up with - from libxcrypt's yescrypt cost parameter directly to prefix strings. This avoids a dependency on APIs beyond crypt(), which might be a requirement in some other projects.

libmisc/salt.c
breard-r: Exactly. At first I tried to actually implement a proper parameter generation. Unfortunately, I found no documentation and very few comments in the code. I did not spent too much time on it, but I think it involves bit masks and non-standard base64 encoding. Instead of reading some foreign code for days and re-implementing it, I generated the 11 possible values using libxcrypt's `crypt_gensalt`.

libmisc/salt.c
breard-r: I copy/pasted this code from [`gensalt_bcrypt()`](https://github.com/shadow-maint/shadow/blob/c917ed7b76c31a9aa100477c6c4d281c7d835c93/libmisc/salt.c#L248-L264), which seems to be an adaptation from [`gensalt()`](https://github.com/shadow-maint/shadow/blob/c917ed7b76c31a9aa100477c6c4d281c7d835c93/libmisc/salt.c#L272-L289). Since I do not fully understand how this code works, I didn't touched it except for the salt's length.

libmisc/salt.c
breard-r: Thank you! I couldn't see it on my editor since it renders the two on the same number of columns.

libmisc/salt.c
solardiz: Makes sense to reuse the same (non-ideal) code for this PR - in fact, this is why I didn't complain about the uses of `random()`. I should have realized the redundant `strcat()` probably came from there as well.

I wouldn't increase salt size. Sure 144 bits gets encoded more optimally, but it's excessive and besides with `random()` we only have at most 32 bits for real: https://github.com/shadow-maint/shadow/blob/e84df9e163e133eb11a2728024ff3e3440592cf8/libmisc/salt.c#L72

The Owl patch I referred to above also switched to use of `/dev/urandom`, but of course this shouldn't be in this same PR.

BTW, it's weird the same code was used to generate salts for bcrypt, since bcrypt uses the crypt base64 alphabet in different order than all other crypts. This probably ended up relying on the bcrypt implementation truncating the unsupported bit values for the last character, probably leaving slightly fewer than 128 bits.

solardiz: > I wouldn't increase salt size. Sure 144 bits gets encoded more optimally

Oh, maybe the reason you did that is the salt strings truncated at 22 were not being accepted, because they contained extra 4 bits in the last char, and you didn't feel like implementing encoding of exactly 128 bits?

libmisc/salt.c
```diff
@@ -283,7 +283,7 @@ static /*@observer@*/const char *YESCRYPT_salt_cost (/*@null@*/long *prefered_co
         long cost;
```
solardiz, line 283: Also inconsistent indentation in the above two lines. (I didn't realize you wouldn't see all of them at once.)

libmisc/salt.c
breard-r: It's not really that I don't "feel like" implementing it. My reasoning was that 144 bits does not harm in any way (libxcrypt's `man 5 crypt` states that the salt can be up to 512 bits) and truncating to 128 bits, while not adding value, would require me to take time study the undocumented non-standard base64 encoding with the risk of introducing bugs in my implementation. Less code, less time spent, less risks and no loss in terms of security, all the odds were in favor of a 144 bits salt.

Obviously, if you recommend to use 128 bits anyway (I'm not sure what you imply by saying it's excessive), I can do that.

solardiz: > the undocumented non-standard base64 encoding

For salt and hash, yescrypt uses the exact same standard crypt base64 encoding that traditional Unix crypt(3) did since the 1970s. It's also the encoding that md5crypt, sha256crypt, and sha512crypt use. (bcrypt is an outlier here.)

However, the yescrypt implementation is strict in that it only accepts salts that are exactly the same as what would be in the resulting salt+hash string. This is different from bcrypt, which silently turns the last character of salt into how it actually sees it in binary, re-encoding it in the returned salt+hash string. This is also different from md5crypt, sha256crypt, and sha512crypt, which use the salt as a mostly opaque ASCII string, preserving the full 132 bits. So it's probably because of these differences that the same code producing 132-bit salts didn't work for yescrypt here, while it did for others. Not the different base64 encoding.

Sure 144 bits does not harm, it's just different. By "excessive", I was referring to 144 vs. 128, not to 128.

libmisc/salt.c
solardiz: I accept your criticism here: I should write a specification of the encoding, document the rationale behind the choices, and add some source code comments. The compact encoding of numeric parameters is actually specific to yescrypt, and needs comments.

libmisc/salt.c
```diff
@@ -263,6 +267,78 @@ static /*@observer@*/const char *gensalt_bcrypt (void)
 }
 #endif /* USE_BCRYPT */
 
+#ifdef USE_YESCRYPT
+/* Default cost if not explicitly specified.  */
+#define Y_COST_DEFAULT 8
```
solardiz, line 272: BTW, libxcrypt's default is 5, not 8. With 8, it's 128 MiB and ~100 ms - probably generally affordable these days when not under attack, but probably does make (inadvertent) DoS of some systems/services easier.

(This is not an objection. Just a comment.)

etc/login.defs
```diff
@@ -367,6 +368,19 @@ CHFN_RESTRICT		rwh
 #BCRYPT_MIN_ROUNDS 13
 #BCRYPT_MAX_ROUNDS 13
 
+#
+# Only works if ENCRYPT_METHOD is set to YESCRYPT.
+#
+# Define the YESCRYPT cost factor.
+# With a higher cost factor, it is more difficult to brute-force the password.
+# However, more CPU resources will be needed to authenticate users if
```
solardiz, line 376: It could be worth mentioning memory here, e.g. "more CPU time and more memory".

libmisc/salt.c
breard-r: You are right, 5 seams much more reasonable.

etc/login.defs
```diff
@@ -367,6 +368,19 @@ CHFN_RESTRICT		rwh
...
+#
+# Define the YESCRYPT cost factor.
+# With a higher cost factor, it is more difficult to brute-force the password.
+# However, more CPU time and more memory will be needed to authenticate users
+# if this value is increased.
+#
+# If not specified, a cost factor of 8 will be used.
+# The value must be within the 1-11 range.
+#
+#YESCRYPT_COST_FACTOR 8
```
solardiz, line 382: This still says 8 in two places.

src/chgpasswd.c
```diff
@@ -73,6 +73,9 @@ static long sha_rounds = 5000;
 #ifdef USE_BCRYPT
 static long bcrypt_rounds = 13;
 #endif
+#ifdef USE_YESCRYPT
+static long yescrypt_cost = 8;
```
solardiz, line 77: This still uses 8.

src/chpasswd.c
```diff
@@ -70,6 +70,9 @@ static long sha_rounds = 5000;
 #ifdef USE_BCRYPT
 static long bcrypt_rounds = 13;
 #endif
+#ifdef USE_YESCRYPT
+static long yescrypt_cost = 8;
```
solardiz, line 74: This still uses 8.

src/newusers.c
```diff
@@ -89,6 +89,9 @@ static long sha_rounds = 5000;
 #ifdef USE_BCRYPT
 static long bcrypt_rounds = 13;
 #endif				/* USE_BCRYPT */
+#ifdef USE_YESCRYPT
+static long yescrypt_cost = 8;
```
solardiz, line 93: This still uses 8.

solardiz (commented):
There are more instances of 8 to be changed to 5 now.

etc/login.defs
breard-r: I can't believe I forgot I defined this cost factor somewhere else >.<
Thank you.

solardiz:
@hallyn I think this is ready for your review, and should get in.

Longer-term, I'd like support for recent libxcrypt's `crypt_gensalt*()` to be added, which is more generic, simpler, and will use `/dev/urandom` for salts - but that's not a reason not to merge this PR.

hallyn:
> @hallyn I think this is ready for your review, and should get in.
> 
> Longer-term, I'd like support for recent libxcrypt's `crypt_gensalt*()` to be added, which is more generic, simpler, and will use `/dev/urandom` for salts - but that's not a reason not to merge this PR.

Thanks!

hallyn:
Could you please fix the formatting in libmisc/salt.c (and a few other files - tabs instead of spaces) ?  (Or if you prefer I can do it)

hallyn:
> Could you please fix the formatting in libmisc/salt.c (and a few other files - tabs instead of spaces) ? (Or if you prefer I can do it)

Oh - I see, that comes in later comits.

Can you squash those?  No reason to have those in history.

breard-r:
> Can you squash those? No reason to have those in history.

If I am not mistaken, the correct way to do so is during the merge using the `Squash and merge` option. If I try to squash on my side it would not be able to push it without forcing, which may cause troubles.

Squash and merge

If not available, it can be activated: https://docs.github.com/en/enterprise/2.14/user/articles/configuring-commit-squashing-for-pull-requests

solardiz:
@breard-r I don't know @hallyn's preference, but FWIW for projects that we manage at Openwall we currently prefer that the contributor squashes on their end and force-pushes - or avoids the multiple commits in the first place, force-pushing each change. Force-pushes work just fine, and GitHub UI allows to see diffs between them.

"Squash and merge" also works, but ends up with GitHub itself as the committer in git history (with the contributor as the author only, but not the committer), which feels unclean to me.

breard-r:
Thanx for the explanation, I understand. I'll try to squash & force-push.

solardiz:
Any reason this isn't merged yet?

hallyn:
No - thank you.

PR #307, opened 2021-01-02
usermod: check if shell is know

glysbaysb:
The man page for usermod says:
> -s, --shell SHELL
> The name of the user's new login shell. ...

This is wrong, usermod just set whatever was given as the new shell. So
if you pass `bash` instead of the path of bash a login afterwards might
not work.

The behaviour in chsh was different, it checked if the argument was on
of the known shells and only then set it[0]. To make it easier for
users, the behaviour should be similar in both programms.

To still allow for a user to run without a real shell (e.g. a daemon
user) and not change the old behaviour too much (while still helping
users) check if the new shell looks like a path and then just set it,
w/o further validation

[0] However root was allowed to set any value, no matter if it was a
known shell or not.

### DIFF
--- a/man/po/de.po
+++ b/man/po/de.po
@@ -1005,10 +1005,10 @@ msgstr ""
 
 #: usermod.8.xml:327(para) chsh.1.xml:123(para)
 msgid ""
-"The name of the user's new login shell. Setting this field to blank causes "
+"The path of the user's new login shell. Setting this field to blank causes "
 "the system to select the default login shell."
 msgstr ""
-"Der Name der neuen Anmelde-Shell des Benutzers. Falls dieses Feld leer "
+"Der Pfad der neuen Anmelde-Shell des Benutzers. Falls dieses Feld leer "
 "gelassen wird, verwendet das System die Standard-Anmelde-Shell."
 
 #: usermod.8.xml:334(term) useradd.8.xml:471(term)
--- a/man/usermod.8.xml
+++ b/man/usermod.8.xml
@@ -353,7 +353,7 @@
 	</term>
 	<listitem>
 	  <para>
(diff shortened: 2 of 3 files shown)


glysbaysb:
Before:
```sh
$ sudo usermod -s doesntexist shadow-test
$ su shadow-test
Password: 
su: failed to execute doesntexist: No such file or directory
```

After:
```sh
$ sudo src/usermod -s doesntexist shadow-test
doesntexist is an invalid shell
```

glysbaysb:
If the behaviour change is accepted I would be happy to also try to write some unit tests

jubalh:
Related https://github.com/shadow-maint/shadow/pull/187 and https://github.com/shadow-maint/shadow/commit/4ed08824e511fa5247006c71490b0440e1c17319#diff-9a7a2bfccabec64213bd054801b9efca8ad55636afbc49e0107714c0f8ffabbe

hallyn:
Thanks for this.  Sorry for the delay.

I'm ok to merge it, but am curious:  how about adding a check analogous to the one useradd, that access(shellpath, X_OK != 0), rather than just shellpath[0]=='/' ?  There's no reason to allow an absolute path which can't be executed, right?

glysbaysb:
No problem, I also usually do OSS on the weekend.

Well it looks like useradd is the third somewhat different implementation then. useradd just checks if either it's "/sbin/nologin" or if it's executable. usermod didn't check anything and chsh checks if it is a known shell.

Then maybe it is better to match useradd. e.g. I personally didn't know about /sbin/nologin and then always set /bin/true as the shell for users which weren't allowed to log in.

hallyn:
metoo on /bin/true :)

hallyn:
Thank you!

Issue #304, opened 2020-12-29
Why can gid mappings only be created when running under the user's primary group?


martijndegouw:
I'm trying to use rootless podman, but when I'm trying to run podman when I changed my primary group using either newgrp or runuser, newgidmap returns an error due to the following check:
https://github.com/shadow-maint/shadow/blob/e84df9e163e133eb11a2728024ff3e3440592cf8/src/newgidmap.c#L231

Now I'm wondering why this check is here and why newgidmap/newuidmap is only allowed when the passwd entry is verified.

If I disable this check and modify https://github.com/shadow-maint/shadow/blob/c917ed7b76c31a9aa100477c6c4d281c7d835c93/src/newgidmap.c#L63 from 
`if ((range->count == 1) && (pw->pw_gid == range->lower)) {`
to
`if ((range->count == 1) && (getgid() == range->lower)) {`
everything works as I expect.

Is this check here for security reasons (since we are running with setuid bit) or is it only a sanity check?

hallyn:
Arguably, when you are running a setgid application you may well not want it to have privilege over all your subgids.  There may be cases where that is ok, but if there are cases where it's not ok, then we have to limit it.

Can you say more about why you need to do this after changing your primary group?

martijndegouw:
In my case this we are using a 'service' account as user to the run build agents of our CI environment.

In our environment service accounts all have their primary group set to the service accounts group. This build service account is a member of several other projects (groups). We switch the primary group when running the build agents so files created during the builds have the group set correctly, instead of the service-accounts group.

Recently people requested the usage of docker/podman to create container images as well during their project builds. That's when we found out that the runuser/newgrp trick breaks the usage of rootless podman for us.

hallyn:
For the NFS restrictions, could they simply newgrp back to their original group before running builds?  The container images are presumably not sitting on NFS, right?

It's possible that relaxing this will be the best way forward, but I want to make sure we are not overlooking a simpler, safer solution.

martijndegouw:
I tested if newgrping back to the primary group is possible, once inside the project directory, but that only results in Permission denied errors. The container images are not stored on NFS whatsoever.

hallyn:
@brauner what do you think?

[...]

PR #308, opened 2021-01-07
newuidmap,newgidmap: Relax gid checking to allow running under alternative group ID

martijndegouw:
This patch allows running newuidmap/newgidmap running when newgrp/runuser is used. The discussion is in https://github.com/shadow-maint/shadow/issues/304


martijndegouw:
Additionally we could use group_member() to add a check similar to the pw_gid check?

src/newgidmap.c
```diff
@@ -60,7 +61,7 @@ static bool verify_range(struct passwd *pw, struct map_range *range, bool *allow
 	}
 
 	/* Allow a process to map its own gid. */
-	if ((range->count == 1) && (pw->pw_gid == range->lower)) {
+	if ((range->count == 1) && (getgid() == range->lower)) {
```
hallyn, line 64: Interesting.  If user is using non-primary id, should they still be allowed to
map pw->pw_gid?  Probably not, so this is probably right, and this may be a
bugfix.

src/newuidmap.c
```diff
@@ -158,9 +159,9 @@ int main(int argc, char **argv)
 	 * mappings we have been asked to set.
 	 */
 	if ((getuid() != pw->pw_uid) ||
-	    (getgid() != pw->pw_gid) ||
+	    (!getdef_bool("GRANT_AUX_GROUP_SUBIDS") && (getgid() != pw->pw_gid)) ||
```
hallyn, line 162: Sorry this should have occurred to me before.

You only need to support the user doing newgrp and then starting a container, right?

So can we (in both of these checks) still enforce that regid == egid? i.e.
    getresgid(&r, &e, &s) && r == e

martijndegouw: This test reflects my use-case pretty well, where the unshare -U and new{u,g}idmap is called by podman offcourse:
```
runuser foo -g bar -c "unshare -U sleep 10 & pid=\$!; \
        newgidmap \$pid 0 1001 1 1 1000000 1000; ret=\$?; \
	cat /proc/\$pid/gid_map >/tmp/test-gidmap/gid_map;
        kill \$pid; exit \$ret"
```

hallyn: Thanks.  I guess the only case I can imagine is:

user A used group Asharesdata to share data with user B.  Makes some
files owned by group Asharesdata which have group write perms, and
creates a setgid binary which allows user B to update it.

If the binary did a unshare, or could be made to do so, then userB
coudl run the setgid binary and quickly does a newuidmap+newgidmap
on it.  At this point B could map B's uid on the host to A's uid in
the namespace, potentially confusing it.

This seems a laughable scenario.  Let me think about it for one
more day and then I think I'll just take it as is.

PR #309, opened 2021-02-01
2021 01 31/newnss.1

hallyn:
I'll have to look this over with some fresh eyes later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ arch:
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get -y install -qq automake autopoint xsltproc libselinux1-dev gettext expect
+  - sudo apt-get -y install -qq byacc libtool
 script:
  - ./autogen.sh --without-selinux --disable-man
  - grep ENABLE_ config.status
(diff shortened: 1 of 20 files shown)


hallyn:
> What I meant with "name spaced" functions is `subid_$token_has_range` (or `shadow_subid_$token_has_range` if you prefer so).
> But the idea is to have $token in function names to avoid clashes between different plugins.
> 
> At least this is what glibc nss is doing including `module->name`: https://github.com/bminor/glibc/blob/master/nss/nss_module.c#L177 ,

Currently the code has, eg:

+                       subid_nss->has_range = dlsym(h, "shadow_subid_has_range");

Even if we open it up to support multiple concurrent nss modules, each one can define shadow_subid_has_range(); we'll h = dlopen() each one by one, and dlsym(h, "shadow_subid_has_range") from the opened handle, putting each into their own struct subid_nss_ops.

I'm not following what $token buys us.

I recognize that if we don't do it now, it'll be hard or impossible to support it later, so apologies if I'm missing something obvious.

hallyn:
(Will open a new PR soon with updated patch)

PR #310, opened 2021-02-02
Man clarifications

ikerexxe:
man/chage.1.xml: Indicate that -d option with a value of 0 forces the
user to change his password. Besides, set an example on how to use -E
option. Finally, add a general note to clarify that chage only takes
charge of local users and another note to indicate that it doesn't check
inconsistencies between shadow and passwd files.

man/groupadd.8.xml: clarify the exit values meaning.

man/shadow.5.xml: indicate the exact time and timezone for the dates.
Moreover, clarify that when the password expires the user won't be able
to login.

man/useradd.8.xml: specify the SELinux requirements for -m option.

man/usermod.8.xml: specify what happens when the current home directory
doesn't exist if using -d and -m options. Moreover, specify what happens
when the group ownership is changed and the uid's don't match in -u and
-g options.

### DIFF
--- a/man/chage.1.xml
+++ b/man/chage.1.xml
@@ -102,6 +102,9 @@
 	    Set the number of days since January 1st, 1970 when the password
 	    was last changed. The date may also be expressed in the format
 	    YYYY-MM-DD (or the format more commonly used in your area).
+	    If the <replaceable>LAST_DAY</replaceable> is set to
+	    <emphasis>0</emphasis> the user is forced to change his password
+	    on the next log on.
 	  </para>
 	</listitem>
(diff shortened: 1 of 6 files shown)


hubgit:
I was going to create a separate pull request for this, but maybe you could change `groupmod.8.xml` here as well?

```patch
- <para>E_GID_IN_USE: specified group doesn't exist</para>
+ <para>E_GID_IN_USE: group id already in use</para>
```

ikerexxe:
I've modified 4ea269d to include the improvement.

hallyn:
Thank you, these are great.

PR #312, opened 2021-02-23
getdate: improve year conversion

ikerexxe:
libmisc/getdate.y: improve year conversion

### DIFF
--- a/libmisc/getdate.y
+++ b/libmisc/getdate.y
@@ -152,6 +152,7 @@ static int	yyHaveDay;
 static int	yyHaveRel;
 static int	yyHaveTime;
 static int	yyHaveZone;
+static int	yyHaveYear;
 static int	yyTimezone;
 static int	yyDay;
 static int	yyHour;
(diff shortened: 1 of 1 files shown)


libmisc/getdate.y
```diff
@@ -395,7 +401,8 @@ relunit	: tUNUMBER tYEAR_UNIT {
 
 number	: tUNUMBER
           {
-	    if ((yyHaveTime != 0) && (yyHaveDate != 0) && (yyHaveRel == 0))
+	    if ((yyHaveTime != 0 || $1 >= 100) && !yyHaveYear
+		&& (yyHaveDate != 0) && (yyHaveRel == 0))
```
hallyn, line 405: I think this looks correct, but it's not exactly easy to parse.
What do you think about moving this baroque check into its own
helper function?

libmisc/getdate.y
```diff
@@ -802,7 +809,7 @@ yylex (void)
 	  return LookupWord (buff);
 	}
       if (c != '(')
-	return *yyInput++;
+	return (unsigned char)*yyInput++;
```
hallyn, line 812: The function returns int.  Why this cast to unsigned char?

hallyn (commented):
Thanks!  If you feel my comments are off please let me know and I'll merge.

ikerexxe:
I'll take those concerns into account and propose a new patch. Thanks for the feedback.

libmisc/getdate.y
ikerexxe: I'd like to do that but my understanding of yacc is low and I fear it'll take me a lot of time to do it. Do you mind if we keep the change as it is?

libmisc/getdate.y
ikerexxe: This didn't make any sense. Sorry for including it. I've restored the previous version.

hallyn:
The commit says "improve year conversion".  What change precisely is it intended to make?

(Would be reassuring if there were some testcases for getdate. :)

hallyn:
Ping - can you elaborate (in the commit message) on how this improves the conversion?

ikerexxe:
Sorry, I haven't been able to work on this topic lately and I don't think I'll be able to do it before the end of the year. I'll close the PR and open a new one when I have time to clarify it.

PR #313, opened 2021-03-12
man: include lastlog file caveat

ikerexxe:
man/lastlog.8.xml: add another point to the caveats section regarding
the handling of the lastlog file by external tools.


thalman:
Looks good to me. Thank you for the patch.

relax-pad (changes requested):
-

PR #314, opened 2021-03-12
passwd(5): Note the use of an ampersand in the comment field

sevan:


hallyn:
Not sure about this - near as I can tell that is an interpretation (and a fickle one) done by 'finger', not by anything in shadow utils, right?  It also is fiddly - only expanded in the first comma separated field in gecos.
Do we really want that in the passwd manpage?

sevan:
It's not just finger(1), I just rode on the back of that example already provided, another would be mail(1). You are right that it's not necessarily anything provided by shadow utils, but it is a [convention of passwd(5) used by other utilities](https://minnie.tuhs.org/cgi-bin/utree.pl?file=2.9BSD/usr/man/man5/passwd.5).

PR #315, opened 2021-03-14
login & su: Treat an empty passwd field as invalid

lanodan:
Otherwise it's treated like the “require no password” clause while it probably
should be treated like a normal su that can't validate anyway.

A similar change should be done for USE_PAM.

### DIFF
--- a/etc/login.defs
+++ b/etc/login.defs
@@ -444,3 +444,12 @@ USERGROUPS_ENAB yes
 # primary group.
 #
 #GRANT_AUX_GROUP_SUBIDS yes
+
+#
+# Prevents an empty password field to be interpreted as "no authentication
+# required".
+# Set to "yes" to prevent for all accounts
+# Set to "superuser" to prevent for UID 0 / root (default)
+# Set to "no" to not prevent for any account (dangerous, historical default)
+
+PREVENT_NO_AUTH superuser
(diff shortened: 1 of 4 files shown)


lanodan:
btw this is to avoid the possibility of accidents with logging to UID-0, I would be fine with allowing no-password for other users but  you should put a dumb password instead of nothing.  
And maybe there is multi-user scripts using this (non-documented?) feature but I don't think there is a good use case for this.

t8m:
This will certainly break existing legitimate use-cases even for the uid 0 account. I'd recommend to make it configurable via login.defs. Perhaps with the default to be as in the proposed patch but with possibility to revert.

lanodan:
Nice idea. I think I'll put the default to "allow for non-UID 0" to avoid a change that would be too destructive for not much gains.

relax-pad (changes requested):
Okimage-

relax-pad (changes requested):
Hey

etc/login.defs
```diff
@@ -444,3 +444,12 @@ USERGROUPS_ENAB yes
 # primary group.
 #
 #GRANT_AUX_GROUP_SUBIDS yes
+
+#
+# Prevents an empty password field to be interpreted as "no authentication
+# required".
+# Set to "yes" to prevent for all accounts
+# Set to "superuser" to prevent for UID 0 / root (default)
+# Set to "no" to not prevent for any account (dangerous, historical default)
```
hallyn, line 453: Conceptually sounds good, thanks.  I'll just need to do a closer code review.

PR #320, opened 2021-04-06
Add optional config directory /etc/login.defs.d/

OttoHollmann:
Add support for optional config directory `/etc/login.defs.d/`. Some programs may create files in this directory, so if this directory exists, load all files with `.defs` suffix. If it doesn't exists, no change and load only default config file `/etc/login.defs`.


Vogtinator:
That was actually implemented with https://github.com/shadow-maint/shadow/pull/180 already

OttoHollmann:
Ah, I didn't notice `libeconf`. So in this case we can close/reject this PR.

PR #321, opened 2021-04-09
Subids: support nsswitch

hallyn:
When starting any operation to do with subuid delegation, check
nsswitch for a module to use.  If none is specified, then use
the traditional /etc/subuid and /etc/subgid files.

Currently only one module is supported, and there is no fallback
to the files on errors.  Several possibilities could be considered:

1. in case of connection error, fall back to files
2. in case of unknown user, also fall back to files

etc...

When non-files nss module is used, functions to edit the range
are not supported.  It may make sense to support it, but it also
may make sense to require another tool to be used.

libsubordinateio also uses the nss_ helpers.  This is how for instance
lxc could easily be converted to supporting nsswitch.

Add a set of test cases, including a dummy libsubid_zzz module.  This
hardcodes values such that:

'ubuntu' gets 200000 - 300000
'user1' gets 100000 - 165536
'error' emulates an nss module error
'unknown' emulates a user unknown to the nss module
'conn' emulates a connection error ot the nss module


lib/nss.c
```diff
@@ -0,0 +1,150 @@
...
+		     token;
+		     token = strtok(NULL, " \n\t")) {
+			char libname[65];
+			void *h;
+			if (strcmp(token, "files") == 0) {
+				subid_nss = NULL;
+				goto done;
+			}
+			if (strlen(token) > 50) {
+				fprintf(stderr, "Subid NSS module name too long: %s\n", token);
```
AkihiroSuda, line 83: Can we print the max length in the error message?

lib/nss.c
```diff
@@ -0,0 +1,150 @@
...
+		if (strlen(line) < 8)
+			continue;
+		if (strncasecmp(line, "subid:", 6) != 0)
+			continue;
+		p = &line[6];
+		while ((*p) && isspace(*p))
+			p++;
+		if (!*p)
+			continue;
+		for (token = strtok(p, " \n\t");
```
giuseppe, line 73: could we use `strtok_r` here to avoid surprises if another thread is also using `strtok`?

Or do you suggest using `nss_init(NULL)` before there are multiple threads running?

libsubid/subid.h
```diff
@@ -13,5 +14,73 @@ enum subid_type {
...
+	 * nss_list_owner_ranges: list the subid ranges delegated to a user.
+	 *
+	 * @owner - string representing username being queried
+	 * @id_type - subuid or subgid
+	 * @ranges - * a NULL-terminated array of struct subordinate_range,
+	 * or NULL.  The returned array of struct subordinate_range must be
+	 * freed by the caller and it's members, if not NULL.
+	 * @ok - whether the result (even if 0) was successful from the nss module.
+
+	 * returns success if the module was able to determine an answer (true or false),
```
alexey-tikhonov, line 61: I guess this is a copy&paste

libsubid/subid.h
```diff
@@ -13,5 +14,73 @@ enum subid_type {
...
+	 * else an error status.
+	 */
+	enum subid_status (*list_owner_ranges)(const char *owner, enum subid_type id_type, struct subordinate_range ***ranges, bool *ok);
```
alexey-tikhonov, line 64: Why do we need `ok`?
Is there any case when `(return_code == SUBID_STATUS_SUCCESS) && (ok != true)` (or vice versa)?
I think "even if 0" case be indicated just via `SUBID_STATUS_SUCCESS && (*ranges == NULL)`?

lib/nss.c
alexey-tikhonov: Additionally, `nss_init()` is thread-unsafe in general, this might be an issue for `libsubid` users...

libsubid/subid.h
```diff
@@ -13,5 +14,73 @@ enum subid_type {
...
+	 * @uids - pointer to an array of uids which will be allocated by
+	 *         nss_find_subid_owners()
+	 * @id_type - subuid or subgid
+	 * @count - number of uids found
+	 * @ok - whether the result (even if 0) was successful from the nss module.
+	 *
+	 * returns success if the module was able to determine an answer (true or false),
+	 * else an error status.
+	 */
+	enum subid_status (*find_subid_owners)(unsigned long id, uid_t **uids, enum subid_type id_type, int *count, bool *ok);
```
alexey-tikhonov, line 79: Same as above about copy&paste and `ok` plus:
 - `id` isn't described
 - why do we use `count` here and NULL-terminated array in `list_owner_ranges()`? Wouldn't it make sense to have "symmetric" API?
 - could we have the same logical order of args - "subject - id_type - result" - as in `list_owner_ranges()`?

Something like:
`enum subid_status (*find_subid_owners)(unsigned long subid, enum subid_type id_type, uid_t **uids);`

lib/nss.c
hallyn: Yeah - I'll go ahead and add locking and use _r.

libsubid/subid.h
hallyn: re count - yeah it is inconsistent, it has kinda bothered me.  Which one do people prefer, a returned count or ending NULL?

libsubid is already out there using the ending null, and i really don't want to have to have libsubid massage what it gets back from nss.

But as far as I know noone is yet using libsubid, so maybe it's still ok to change so far.

alexey-tikhonov: > re count - yeah it is inconsistent, it has kinda bothered me. Which one do people prefer, a returned count or ending NULL?

Any option is fine for me. Just preferably to have it consistent.

lib/nss.c
```diff
@@ -0,0 +1,150 @@
...
+// then sssd will be consulted for subids.  Unlike normal NSS dbs,
+// only one db is supported at a time.  That's open to debate, but
+// the subids are a pretty limited resource, and local files seem
+// bound to step on any other allocations leading to insecure
+// conditions.
+static bool nss_initialized;
+
+static struct subid_nss_ops *subid_nss;
+
+struct subid_nss_ops *get_subid_nss_handle() {
```
alexey-tikhonov, line 25: Did you consider to call `nss_init()` here?
Looks like it could save a line of code in every `lib/subordinateio.c ` function without impacting readability...
(Just IMO)

lib/subordinateio.c
```diff
@@ -738,7 +795,19 @@ struct subordinate_range **list_owner_ranges(const char *owner, enum subid_type
...
+	struct subid_nss_ops *h;
+
+	nss_init(NULL);
+	h = get_subid_nss_handle();
+	if (h) {
+		status = h->list_owner_ranges(owner, id_type, &ranges, &ok);
+		if (status == SUBID_STATUS_SUCCESS && ok)
+			return ranges;
+		return NULL;
+	}
```
alexey-tikhonov, line 810: I think `ok` is excessive.
It could be just
```
    status = h->list_owner_ranges(owner, id_type, &ranges);
    if (status == SUBID_STATUS_SUCCESS)
        return ranges;
    return NULL;
```

libsubid/subid.h
```diff
@@ -13,5 +14,73 @@ enum subid_type {
...
 };
 
+enum subid_status {
+	SUBID_STATUS_SUCCESS = 0,
+	SUBID_STATUS_UNKNOWN_USER = 1,
+	SUBID_STATUS_ERROR_CONN = 2,
+	SUBID_STATUS_ERROR = 3,
+};
+
+struct subid_nss_ops {
```
alexey-tikhonov, line 24: What is the reason to keep this definition in `libsubid/*`?
It seems it is only used in `lib/nss.c` and `lib/subordinateio.c`.

lib/subordinateio.c
hallyn: Hm, well I was thinking for the SUBID_STATUS_UNKNOWN_USER case - but i guess what i really wanted is satisfied by empty list or count=0.  So I think I'll drop that, thanks.

alexey-tikhonov: Any `return code != SUBID_STATUS_SUCCESS` means caller shouldn't read out-args at all.

user not found => just `SUBID_STATUS_UNKNOWN_USER`
user found, but doesn't have any subid range assigned => SUBID_STATUS_SUCCESS && (*ranges == NULL)
user found and has ranges => SUBID_STATUS_SUCCESS && (*ranges != NULL)

imo that way make sense.

hallyn: Well, user unknown seemed special to me, but that's sensible.

hallyn:
I'll address your feedback with appended patches (which I will squash in before merging) soon, but probably not today.

lib/subordinateio.c
alexey-tikhonov: The same logic applies to all plugin functions.

alexey-tikhonov:
Hi @hallyn,

thank you for opening PR.

Please see few inline comments.

Besides that I think this PR should update `libsubid/api.c` as well: currently `get_sub?id_ranges()` and `get_sub?id_owners()` unconditionally call `sub_?id_open()`, `sub_?id_close()`, etc. I think this is at least redundant and might even cause issues (for example, if network source is configured and /etc/sub?id files are inaccessible for some reason).

It's even more prominent with `grant/free_sub?id_range()`: besides the same issue with unconditional execution of `sub_?id_open/lock/unlock/close()`, it operates on files directly, bypassing `sub_?id_add/remove()`. I think those function should return `EOPNOTSUPP` promptly in case nsswitch plugin configured.

Additionally, could you please clarify, if `const char *owner` means:
 - "username", 
 - text representation of "UID" 
 - or both [should be] supported
 in `subid_nss_ops` and in `libsubid/api.h`?

lib/nss.c
```diff
@@ -0,0 +1,150 @@
...
+
+// NSS plugin handling for subids
+// If nsswitch has a line like
+//    subid: sssd
+// then sssd will be consulted for subids.  Unlike normal NSS dbs,
+// only one db is supported at a time.  That's open to debate, but
+// the subids are a pretty limited resource, and local files seem
+// bound to step on any other allocations leading to insecure
+// conditions.
+static bool nss_initialized;
```
alexey-tikhonov, line 21: Btw, probably using `atomic_bool` could be an easy way to achieve thread safety.

Imo, this type and moving `nss_initialized = true;` right after `if (nss_initialized)` check in `nss_init()` could do the trick.

lib/nss.c
```diff
@@ -0,0 +1,150 @@
...
+				fprintf(stderr, "Using files\n");
+				subid_nss = NULL;
+				goto done;
+			}
+			subid_nss = malloc(sizeof(*subid_nss));
+			if (!subid_nss) {
+				dlclose(h);
+				goto done;
+			}
+			subid_nss->has_range = dlsym(h, "shadow_subid_has_range");
```
alexey-tikhonov, line 101: I still propose
```
_subid_$token_has_range
```
as a name template. I.e. for SSSD it would be `_subid_sss_has_range()`.

That would be consistent with glibc NSS conventions:
https://github.com/bminor/glibc/blob/master/nss/nss_module.c#L169

hallyn: It just seems like needless cve-inviting string manipulation for no benefit.

Except matching glibc.

Is there another advantage I'm overlooking?

@brauner what do you think?

lib/nss.c
hallyn: No I hadn't considered it.  That's a good idea.

libsubid/subid.h
hallyn: Seemed the thing to do at the time.

I've moved it to lib/prototypes.h.

lib/nss.c
hallyn: But then another thread may assume that nss_init() has completed because nss_initialized == true, while the first thread is still initializing.

Unless I'm missing something.

hallyn: two atomic_flags might work.

lib/nss.c
alexey-tikhonov: To avoid potential name clashes in multiple plugins.

lib/nss.c
alexey-tikhonov: You are right.

lib/nss.c
hallyn: But each plugin will come from its own dlopen'd handle.  So even if we start supporting multiple simultaneous plugins, we dlsym for the same name from each handle and get the right function pointer, which we'll keep in separate struct instances for each plugin.

Changing the name only seems to complicate things.

Still if you (sssd) as author of one of the modules think it makes things easier for you, I'm still considering changing it.

rhatdan:
Thanks @hallyn This looks good.

hallyn:
I probably need to have configure.ac look for yacc+libtool (per my .travis.yaml update), then depending on the open thread it should be ready.

lib/nss.c
alexey-tikhonov: > Still if you (sssd) as author of one of the modules think it makes things easier for you, I'm still considering changing it.

No, it doesn't. If you don't like this idea - let's skip.

lib/prototypes.h
```diff
@@ -299,14 +299,16 @@ struct subid_nss_ops {
 	 *
 	 * @owner - string representing username being queried
 	 * @id_type - subuid or subgid
-	 * @ranges - * a NULL-terminated array of struct subordinate_range,
-	 * or NULL.  The returned array of struct subordinate_range must be
-	 * freed by the caller and it's members, if not NULL.
+	 * @ranges - * an array of struct subordinate_range, or NULL.  The returned
```
alexey-tikhonov, line 302: This is actually a pointer to an array of **pointers** to struct.

This is somewhat inconsistent as compared to `find_subid_owners()`, where out_arg is really a pointer to an array of data.

Do we really need this to be an array of pointers..?

alexey-tikhonov: Ah... this seems to be deeply embedded into existing code...

lib/nss.c
```diff
@@ -18,16 +19,17 @@
 // the subids are a pretty limited resource, and local files seem
 // bound to step on any other allocations leading to insecure
 // conditions.
-static bool nss_initialized;
+static atomic_flag nss_init_started;
+bool nss_init_completed;
```
alexey-tikhonov: Maybe `atomic_bool nss_init_completed`?

lib/subordinateio.c
```diff
@@ -311,31 +311,31 @@ static bool append_range(struct subordinate_range ***ranges, const struct subord
 {
 	struct subordinate_range *tmp;
 	if (!*ranges) {
-		*ranges = malloc(2 * sizeof(struct subordinate_range **));
+		*ranges = malloc(sizeof(struct subordinate_range **));
```
alexey-tikhonov, line 314: Why is this `sizeof(struct subordinate_range **)` and not `sizeof(struct subordinate_range *)`?

Result is clearly the same. I'm just confused.

alexey-tikhonov:
Hi @hallyn,

thanks for an update. Good progress.

In the comments to
```
(*find_subid_owners)(unsigned long id, uid_t **uids, enum subid_type id_type, int *count);
```
'owner' is described instead of 'id'.
And I still think order of arguments is weird - don't you think it makes sense to move `id_type` before `uids`? Currently input arg fits between output arguments.

And what do you think about https://github.com/shadow-maint/shadow/pull/321#issuecomment-816759039 ?
Namely:
  - `libsubid/api.c` does some files operations unconditionally
  - what are the expectations towards meaning of "owner" (UID/username/both) in subid_nss_ops and in libsubid/api.h?

hallyn:
> In the comments to
> 
> ```
> (*find_subid_owners)(unsigned long id, uid_t **uids, enum subid_type id_type, int *count);
> ```
> 
> 'owner' is described instead of 'id'.
> And I still think order of arguments is weird - don't you think it makes sense to move `id_type` before `uids`? Currently input arg fits between output arguments.

Yes, id_type before the output value uids makes sense.  Sorry - if you've mentioned this before, I don't remember it.

lib/subordinateio.c
hallyn: As you say both pointer types are the same size , but technically because ranges is a ***, *range is a **.

lib/nss.c
hallyn: I did that first, but then realized that there is no api that i could find for just checking the value or even clearing it.  Only set-and-check.  That makes this unsuitable for the loop where I'm just waiting for another thread to set it.

And it doesn't need to be atomix - only nss_init_started does.

lib/prototypes.h
hallyn: Hm, I guess we don't.  It could just be a pointer to an array of structs...

hallyn:
>     * what are the expectations towards meaning of "owner" (UID/username/both) in subid_nss_ops and in libsubid/api.h?

owner is username, cannot (at the moment) be a userid.  That's from pre-existing subid code.

hallyn:
>     * `libsubid/api.c` does some files operations unconditionally

I'll think about that.

lib/subordinateio.c
hallyn: *boggle* i had that wrong.  But in any case I'm changing it.

lib/nss.c
alexey-tikhonov: > I did that first, but then realized that there is no api that i could find for just checking the value or even clearing it.

Actually there are `atomic_load()`/`atomic_store()`.

But if I understand correctly, in case you don't need to specify explicit memory order, then `atomic_bool` can be just assigned/read (keeping "atomic" properties).

> And it doesn't need to be atomix - only nss_init_started does.

I think both have to. Otherwise compiler is eligible for different optimizations (for example, to cache value of `nss_init_completed` on a register while code is in the loop). In the past some folks used `volatile` to prevent it but this is... wrong.

https://en.cppreference.com/w/c/language/atomic

hallyn: Oh, I see - i had looked through stdatomix.h by the bool and flag definitions, but it's done through #defines.  Thanks

libsubid/api.c
```diff
@@ -81,7 +81,7 @@ void subid_free_ranges(struct subordinate_range **ranges, int count)
 	return free_subordinate_ranges(ranges, count);
 }
 
-int get_subid_owner(unsigned long id, uid_t **owner, enum subid_type id_type)
+int get_subid_owner(unsigned long id, enum subid_type id_type, uid_t **owner)
```
alexey-tikhonov, line 56: This (and few other internal functions) could be `static`, I think? (as `get_subid_ranges()` is)

alexey-tikhonov:
> > ```
> > * what are the expectations towards meaning of "owner" (UID/username/both) in subid_nss_ops and in libsubid/api.h?
> > ```
> 
> owner is username, cannot (at the moment) be a userid. That's from pre-existing subid code.

Could you please add a short comment about this to `libsubid/api.h`?

lib/prototypes.h
```diff
@@ -299,7 +299,8 @@ struct subid_nss_ops {
 	 *
 	 * @owner - string representing username being queried
 	 * @id_type - subuid or subgid
-	 * @ranges - * an array of struct subordinate_range, or NULL.  The returned
+	 * @ranges - pointer to an array of struct subordinate_range, or NULL.  The
+	 *  returned
```
alexey-tikhonov, line 303: (sorry for being so pedantic) line of text is torn here.

lib/subordinateio.c
```diff
@@ -913,20 +977,71 @@ bool new_subid_range(struct subordinate_range *range, enum subid_type id_type, b
...
+		sub_gid_unlock();
+	}
+
+	return ret;
 }
 
 bool release_subid_range(struct subordinate_range *range, enum subid_type id_type)
 {
 	struct commonio_db *db;
-	if (id_type == ID_TYPE_UID)
```
alexey-tikhonov, line 868: I think the first thing that have to be done here and in `new_subid_range()` - check `get_subid_nss_handle()` and return false in case NSS module is used.

alexey-tikhonov:
(Strictly speaking, this is out of the scope of this PR, but...)
Btw, why `sub_uid_open()` isn't executed in `sub_uid_assigned()`->`range_exists()` and in `have_sub_uids()`->`have_range()`->`find_range()` (and in -gid- counterparts)?

lib/subordinateio.c
```diff
@@ -307,39 +307,23 @@ static bool have_range(struct commonio_db *db,
...
 		if (!new)
 			return false;
 		*ranges = new;
 	}
-	(*ranges)[n] = NULL;
-	tmp = subordinate_dup(new);
-	if (!tmp)
-		return false;
-	(*ranges)[n] = tmp;
+	(*ranges)[n] = *in;
```
alexey-tikhonov, line 324: This makes a copy of `subordinate_range::owner` pointer.
And it seems `commonio_next()` returns a pointer to internal structured (not copy).
So I guess dup of `subordinate_range::owner` is needed here?

alexey-tikhonov: Btw, `subordinate_dup()` seems to be unused after this change.

lib/subordinateio.c
```diff
@@ -307,39 +307,23 @@ static bool have_range(struct commonio_db *db,
...
 }
 
-void free_subordinate_ranges(struct subordinate_range **ranges, int count)
-{
-	int i;
-
-	if (!ranges)
-		return;
-	for (i = 0; i < count; i++)
-		subordinate_free(ranges[i]);
```
alexey-tikhonov, line 339: `subordinate_free()` seems to be unused anymore.

libsubid/api.h
```diff
@@ -1,9 +1,8 @@
 #include "subid.h"
 #include <stdbool.h>
 
-int get_subuid_ranges(const char *owner, struct subordinate_range ***ranges);
-int get_subgid_ranges(const char *owner, struct subordinate_range ***ranges);
-void subid_free_ranges(struct subordinate_range **ranges, int count);
```
alexey-tikhonov: Well, a helper still would be handy.
While we can free `ranges` in one shot (no need to free every `subordinate_range` element), we still need to iterate over every array item first to free `subordinate_range::owner`...

lib/subordinateio.c
```diff
@@ -812,14 +794,13 @@ int list_owner_ranges(const char *owner, enum subid_type id_type, struct subordi
 	commonio_rewind(db);
 	while ((range = commonio_next(db)) != NULL) {
 		if (0 == strcmp(range->owner, owner)) {
-			if (!append_range(&ranges, range, count++)) {
-				free_subordinate_ranges(ranges, count-1);
+			if (!append_range(ranges, range, count++)) {
+				free(*ranges);
```
alexey-tikhonov, line 811: See my comment to removal of `subid_free_ranges()`: we still need to iterate over every array item first to free subordinate_range::owner...

alexey-tikhonov:
Sorry for a bunch of new comments :)
But in general I think we are pretty close to a version that can be merged.
Public/module API looks good to me now, I'll try to draft SSSD plugin soon.

lib/subordinateio.c
hallyn: It's still used as a member of the commonio_ops struct.

libsubid/api.h
hallyn: Eh I should have left it as an array of poitners to subordinate_range.

hallyn:
> (Strictly speaking, this is out of the scope of this PR, but...)
> Btw, why `sub_uid_open()` isn't executed in `sub_uid_assigned()`->`range_exists()` and in `have_sub_uids()`->`have_range()`->`find_range()` (and in -gid- counterparts)?

The callers have to (and do) do it.  If these will be called from say libsubid, then it'll have to do *_open first.

lib/subordinateio.c
hallyn: yeah that's ugly.  I'd forgotten about the -> owner, and, given that, would prefer to revert  '(fold up) change list_owner_ranges to do **ranges, not ***ranges'  so that the code using subordinate_range can stay consistent.

lib/subordinateio.c
hallyn: Hm, right.  I missed those.  Thanks.

hallyn:
Unless there are any failures in the travis tests, or any more grievous problems which would require api changes to resolve, I intend to merge this tomorrow.

We can always fine-tune later.  I'm sure that as someone writes a real NSS module, we'll find some things lacking.

hallyn:
@alexey-tikhonov  to say a bit more about the ***ranges:  I find it particularly distasteful to return a pointer to an array of structs and require the caller to free a pointer in the return struct.  It just feels wrong.

lib/subordinateio.c
alexey-tikhonov: I don't object keeping this as an array of pointers to a struct, but still this should be `sizeof(struct subordinate_range *)` :)

lib/subordinateio.c
alexey-tikhonov: This member wasn't used in that version of PR. But doesn't matter already.

libsubid/api.h
alexey-tikhonov: That's what I meant when wrote "this seems to be deeply embedded into existing code..." :)

lib/subordinateio.c
alexey-tikhonov: Ok.

alexey-tikhonov:
> @alexey-tikhonov to say a bit more about the ***ranges: I find it particularly distasteful to return a pointer to an array of structs and require the caller to free a pointer in the return struct. It just feels wrong.

Both versions require the caller to free a pointer in the struct, so I don't see a big difference.
But I don't have a strong preference here. Initially I merely asked about rationale and I don't mind keeping it as `***ranges`.


> Unless there are any failures in the travis tests, or any more grievous problems which would require api changes to resolve,

In general, current version looks good to me codewise. Thank you for the patience while addressing my comments.

I would only ask to add a couple of comments to `libsubid/api.h` before merging (see https://github.com/shadow-maint/shadow/pull/321#issuecomment-818921578)


> I intend to merge this tomorrow. We can always fine-tune later. I'm sure that as someone writes a real NSS module, we'll find some things lacking.

JFYI, @ikerexxe is already working on packaging of `libsubid` and this PR in Fedora.
And I'm starting on SSSD plugin.
It's up to you if you prefer to wait for a feedback or proceed with merge and address it later (if there will be a need).

ikerexxe:
I have one concern but I don't know if this is the proper place to write. I'll write it here and if it doesn't fit I can open an issue or a PR.

Straight to the point, I fear that one of the names of the header for exposing the library is not correct. I see `api.h` and `subid.h` as the header names. `api.h` can be a sensible name for shadow-utils repository, but once you include the file in a distribution you don't know what you are handling as there are lots of packages providing APIs. Would you mind merging those two files into `subid.h`? Another option would be to rename both files, something like `subid_api.h` and `subid_data.h`. This is the approach that I have taken so far in Fedora.

I'm open to another way of naming them, what I'm trying to avoid is portability problems of applications between distributions because a header was named in a very generic way and then each distribution changed its name when packaging.

alexey-tikhonov:
> Straight to the point, I fear that one of the names of the header for exposing the library is not correct. I see `api.h` and `subid.h` as the header names. `api.h` can be a sensible name for shadow-utils repository, but once you include the file in a distribution you don't know what you are handling as there are lots of packages providing APIs.

Wouldn't you like to install those to $systempath/subid/ (i.e. /usr/include/subid/)?
Then include would be <subid/api.h> and that's ok, imo.

lib/subordinateio.c
hallyn: Right.

hallyn:
Yeah, I can rename those.  I also was going to actually comment the api.h functions :)  But maybe as a separate commit since technically that's libsubid, and nss support is a core shadow feature.

hallyn:
BTW I do notice that it's a bit inconsistent that find_subid_owners returns uids while the others take a char *username.

I'm loath to change that though - since we  can have more than one name per uid, it seems sensible

hallyn:
@ikerexxe 

> Straight to the point, I fear that one of the names of the header for exposing the library is not correct. I see `api.h` and `subid.h` as the header names. `api.h` can be a sensible name for shadow-utils repository, but once you include the file in a distribution you don't know what you are handling as there are lots of packages providing APIs. Would you mind merging those two files into `subid.h`? Another option would be to rename both files, something like `subid_api.h` and `subid_data.h`. This is the approach that I have taken so far in Fedora.

I've merged them into subid.h.  Since you're already massaging the files in your package, can you verify that the way I have it will work for you?

alexey-tikhonov:
Thank you, @hallyn.

hallyn:
@alexey-tikhonov @giuseppe @AkihiroSuda @ikerexxe @rhatdan 

Thanks for all your helpful comments.

ikerexxe:
@hallyn thanks to you for your work.

> I've merged them into subid.h. Since you're already massaging the files in your package, can you verify that the way I have it will work for you?

Yes, that's perfectly fine.

alexey-tikhonov:
Hi,

not sure if this is a best place, but I'd like to discuss one thing.

Currently `subuid.h::get_subuid_ranges()` -> `list_owner_ranges()` -> `get_subid_nss_handle()` -> `nss_init()` -> `fprintf(stderr, ...)`

That's ok for `new?idmap` to print to `stderr`, but I'm not so sure about `libsubid`...

@hallyn , what do you think?

giuseppe:
one potential issue I've found is that `subid.h` is not installed:

```shell
$ make install DESTDIR=$(pwd)/dest >/dev/null && find dest -name subid.h
libtool: warning: remember to run 'libtool --finish /lib'
$
```

alexey-tikhonov:
Hi @hallyn ,

looking at the usage of `has_any_range()`/`has_range()` in existing code, I have following questions:

1) newusers: sub_uid_assigned() -> nsswitch::has_any_range(), so if network source is configured in `nsswitch.conf` then `newusers` consults network source instead of /etc/subid, and this doesn't make much sense, imo
And this is the only use of `has_any_range()` I've found. Do we need this function in a plugin at all if only local user manipulation tools use it?

2) useradd: sub_uid_add() -> !files => -EOPNOTSUPP. Again, if network source is configured in `nsswitch.conf` then `useradd` will fail to update /etc/subuid

alexey-tikhonov:
Btw, if we suppose that a given subid can be used by multiple users (uids), then probably usage of `has_range()` in `user_busy.c` became wrong now (subrange doesn't identify owner anymore).

alexey-tikhonov:
Another one question about `get_sub?id_ranges()`/`list_owner_ranges()`: do we really need `owner` field in `struct subordinate_range`? Or do we expect those functions to populate this field?
Since those functions do a lookup for a specified `owner`, this field will clearly be the same in all ranges (and must match in arg). Feels like useless strcpy()'s...

hallyn:
So maybe we should have something like:

struct subordinate_id_range {
    unsigned long start;
    unsigned long count;
};

struct subordinate_delegated {
    const char *owner;
    struct subordinate_id_range range;
};

?  Although simply saying that it's ok to leave the subordinate_range->owner
NULL when those functions return would be ok with me too.  That might become an
issue for users who want to do several calls for different users, combine the
results, and then parse the collection.  But even they may not want to do all
that freeing later.

So if you like we can add a comment in these function descriptions saying
that owner may be returned as NULL.

hallyn:
> one potential issue I've found is that `subid.h` is not installed:
> 
> ```shell
> $ make install DESTDIR=$(pwd)/dest >/dev/null && find dest -name subid.h
> libtool: warning: remember to run 'libtool --finish /lib'
> $
> ```

Should be fixed with https://github.com/shadow-maint/shadow/pull/330

hallyn:
> Hi @hallyn ,
> 
> looking at the usage of `has_any_range()`/`has_range()` in existing code, I have following questions:
> 
>     1. newusers: sub_uid_assigned() -> nsswitch::has_any_range(), so if network source is configured in `nsswitch.conf` then `newusers` consults network source instead of /etc/subid, and this doesn't make much sense, imo
>        And this is the only use of `has_any_range()` I've found. Do we need this function in a plugin at all if only local user manipulation tools use it?
> 
>     2. useradd: sub_uid_add() -> !files => -EOPNOTSUPP. Again, if network source is configured in `nsswitch.conf` then `useradd` will fail to update /etc/subuid

Would you mind creating a new issue for these as well?  It'll be very hard to track conversation in this issue.

I don't mind dropping has_any_range(), I think I thought something like lxc-usernsexec could use it, but it can just as well just ask for the ranges and get back -ENOENT.

For the second, I think you're saying that useradd should ignore the failure to add subids?

alexey-tikhonov:
> > 1. newusers: sub_uid_assigned() -> nsswitch::has_any_range(), so if network source is configured in `nsswitch.conf` then `newusers` consults network source instead of /etc/subid, and this doesn't make much sense, imo
> >    And this is the only use of `has_any_range()` I've found. Do we need this function in a plugin at all if only local user manipulation tools use it?
> > 
> > 2. useradd: sub_uid_add() -> !files => -EOPNOTSUPP. Again, if network source is configured in `nsswitch.conf` then `useradd` will fail to update /etc/subuid
> 
> Would you mind creating a new issue for these as well? It'll be very hard to track conversation in this issue.

Ok. Just wasn't sure if you see this as an issue.

> I don't mind dropping has_any_range(), I think I thought something like lxc-usernsexec could use it, but it can just as well just ask for the ranges and get back -ENOENT.

`SUBID_STATUS_SUCCESS  && (*count == 0)`, to be precise.


> For the second, I think you're saying that useradd should ignore the failure to add subids?

No, I meant a little bit different thing.
What is our general approach to a case: "nsswitch has network source configured" and "local user manipulation are being done"?
I think local tools, that add/modify local users, should work with /etc/ even if "network source" is enabled in nsswitch?

alexey-tikhonov:
> struct subordinate_id_range {
> unsigned long start;
> unsigned long count;
> };
> 
> struct subordinate_delegated {
> const char *owner;
> struct subordinate_id_range range;
> };
> 
> ? Although simply saying that it's ok to leave the subordinate_range->owner
> NULL when those functions return would be ok with me too. That might become an
> issue for users who want to do several calls for different users, combine the
> results, and then parse the collection. But even they may not want to do all
> that freeing later.
> 
> So if you like we can add a comment in these function descriptions saying
> that owner may be returned as NULL.

I think my preference would be a solution with two structs (this is very similar to what I have in SSSD's plugin now).
This would allow `list_owner_ranges()` to work with `subordinate_id_range **` (instead of `subordinate_delegated ***`) - single malloc instead of (1 + count * 2) malloc's...

Issue #322, opened 2021-04-09
set_selinux_file_context in selinux.c is too broad


doverride:
currently `set_selinux_file_context` expects a single parameter: `dst_name` but the underlying `matchpathcon()` function also takes a `format bit mode` and mode matching is disabled in `set_selinux_file_context`. The manual page clearly states:

```
mode can be 0 to disable mode matching, but should be provided whenever possible, as it may affect the matching.
```
I've hit this issue where useradd tries to create the user home directory with a wrong context because the `matchpathcon()` result is not for directories. The set_selinux_file_context wrapper should instead probably allow one to specify the format bit mode. so that "dir" can be specified when creating user home dir and "file" can be specified when creating the user mail spool file

PR #323, opened 2021-04-09
SELinux modernizations

cgzones:
Closes: #322

### DIFF
--- a/lib/selinux.c
+++ b/lib/selinux.c
@@ -35,7 +35,7 @@
 #include "defines.h"
 
 #include <selinux/selinux.h>
-#include <selinux/context.h>
+#include <selinux/label.h>
 #include "prototypes.h"
 
 static bool selinux_checked = false;
@@ -53,8 +53,6 @@ static bool selinux_enabled;
  */
 int set_selinux_file_context (const char *dst_name)
 {
-	/*@null@*/char *scontext = NULL;
-
 	if (!selinux_checked) {
 		selinux_enabled = is_selinux_enabled () > 0;
 		selinux_checked = true;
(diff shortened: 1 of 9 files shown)


cgzones:
Send to SELinux mailing list for review: https://lore.kernel.org/selinux/20210413122508.24745-1-cgzones@googlemail.com/T/#t

src/vipw.c
```diff
@@ -243,13 +243,13 @@ vipwedit (const char *file, int (*file_lock) (void), int (*file_unlock) (void))
...
-		security_context_t passwd_context=NULL;
+		char *passwd_context_raw = NULL;
 		int ret = 0;
-		if (getfilecon (file, &passwd_context) < 0) {
+		if (getfilecon_raw (file, &passwd_context_raw) < 0) {
 			vipwexit (_("Couldn't get file context"), errno, 1);
 		}
-		ret = setfscreatecon (passwd_context);
-		freecon (passwd_context);
+		ret = setfscreatecon_raw (passwd_context_raw);
```
hallyn, line 251: Why do  _raw here?  The setfscreatecon manpage only says that it does "not perform context translation" but I don't know what that's referring to.

Note - I'm not doubting that it's the right thing, I just want to understand why it it.

lib/selinux.c
```diff
@@ -53,28 +53,40 @@ static bool selinux_enabled;
...
+		hnd = selabel_open(SELABEL_CTX_FILE, NULL, 0);
+		if (hnd == NULL) {
+			return security_getenforce () != 0;
+		}
+
+		r = selabel_lookup_raw(hnd, &fcontext_raw, dst_name, 0);
+		selabel_close(hnd);
+		if (r < 0) {
+			/* No context specified for the searched path */
+			if (errno == ENOENT) {
```
hallyn, line 87: Is selabel_close() implemented in a way that ensures that the errno from the failed selabel_lookup_raw() is not overwritten?

hallyn: I see - in the end it doesn't matter because you're removing that call.  Feel free to ignore my previous question about selabel_close().

src/vipw.c
cgzones: A SELinux context consists of 3 or 4 parts (separated by `:`), depending on whether the loaded policy enables MCS/MLS; for example `system_u:object_r:user_t` or `system_u:object_r:user_t:s0-s0:c0.c1023`.
The last part, in this case `s0-s0:c0.c1023`, is the sensitivity and/or category part of the context.

To improve usability it is possible to set an alias name for those `sX`/`cX` identifiers.
libselinux supplies the functions `selinux_trans_to_raw_context()` and `selinux_raw_to_trans_context()`, which are called by the non-raw functions behind the scenes, to convert raw contexts into human readable ones and vice versa.

It is useful to use the translated non-raw functions if the context is either being presented to the user, e.g. by ls(1) or ps(1), or the context is read from a user supplied input, e.g. pam_selinux(7).

shadow does neither of these two things, so translating the context is not necessary.

See also https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/5/html/deployment_guide/sec-mcs-config-categories, or [setrans.conf(5)](https://manpages.debian.org/unstable/mcstrans/setrans.conf.5.en.html)

lib/selinux.c
cgzones: Good catch.
Should I add errno-saving code to the commit "selinux.c: use modern selabel interface instead of deprecated matchpathcon" and or squash "selinux.c: use modern selabel interface instead of deprecated matchpathcon" and "selinux: only open selabel database once", so that there is no intermediate state with this issue?

hallyn: Adding the errno-saving code would be overkill here :)  You can squash them, or ignore it.  Thanks.

hallyn:
Hi - before I merge, did you want to add James Carter's Ack from the selinux list thread?

cgzones:
Added the Ack's by James Carter.

PR #326, opened 2021-04-29
Tests if shadow-utils are immune against bugs in 2006:0032

aborah-sudo:
1. Creating tmp directory
2. Allowing '.' in usernames
3. Added two patches to shadow-utils for
additional usermod flag
4. The user foo must be included in both
foo-group and foo-group2 groups
5. usermod -p does not update sp_lstchg

### DIFF
--- a/tests/pytest/test_error_2006_0032.py
+++ b/tests/pytest/test_error_2006_0032.py
@@ -1,0 +1,102 @@
+""" Automation of Shadow Utils tests """
+
+from __future__ import print_function
+import subprocess
+import os
+import pytest
+
+
+class TestShadowUtilsErrors():
+    """
+    Automation of Shadow Utils tests
+    """
+    def test_error_2006_0032(self):
+        """
+        :title: Tests if shadow-utils are
+         immune against bugs in 2006:0032
+        :id: TC#46675
+        :steps:
+          1. Creating tmp directory
+          2. Added two patches to shadow-utils for
+          additional usermod flag
+          3. The user foo must be included in both
+          foo-group and foo-group2 groups
+          4. usermod -p does not update sp_lstchg
+        :expectedresults:
+          1. Should succeed
+          2. Should succeed
+          3. Should succeed
+          4. Should succeed
+        """
+        # Creating tmp directory
+        mktmp = "mktemp -d"
+        process = subprocess.Popen(mktmp.split(),
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.PIPE)
+        stdout, stderr = process.communicate()
+        assert stderr.decode() == ''
+        # Added two patches to shadow-utils for
+        # additional usermod flag
+        foo_info = ["useradd foo",
+                    "groupadd foo-group",
+                    "usermod -G foo-group foo",
+                    "groupadd foo-group2",
+                    "usermod -G foo-group2 -a foo"]
+        for i in foo_info:
+            process = subprocess.Popen(i.split(),
+                                       stdout=subprocess.PIPE,
+                                       stderr=subprocess.PIPE)
+            stdout, stderr = process.communicate()
+            assert stderr.decode() == ''
+        # The user foo must be included in both
+        # foo-group and foo-group2 groups
+        cmd_id = "id foo"
+        process = subprocess.Popen(cmd_id.split(),
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.PIPE)
+        stdout, stderr = process.communicate()
+        assert stderr.decode() == ''
+        for grp in ['foo-group', 'foo-group2']:
+            assert grp in stdout.decode()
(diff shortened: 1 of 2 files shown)


tests/pytest/test_error_2006_0032.py
```diff
@@ -0,0 +1,117 @@
...
+import pytest
+
+
+class TestShadowUtilsErrors():
+    """
+    Automation of Shadow Utils tests
+    """
+    def test_error_2006_0032(self):
+        """
+        :title: Shadow-Utils: Tests if shadow-utils are
```
jakub-vavra-cz, line 15: It might be good to keep add the test prefix "IDM-SSSD-TC:" to the title.

ikerexxe:
>     2. Allowing '.' in usernames

I'm not sure whether this point affects other distributions apart from Fedora and RHEL. I'll have to check in my personal computer to see if Debian derivatives allow a period in the username.

tests/pytest/test_error_2006_0032.py
hallyn: I've googled that but still haven't found a reference for what this is addressing.

tests/pytest/test_error_2006_0032.py
```diff
@@ -0,0 +1,117 @@
...
+         immune against bugs in 2006:0032
```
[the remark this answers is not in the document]
aborah-sudo: its as per the template we are following

tests/pytest/test_error_2006_0032.py
```diff
@@ -0,0 +1,117 @@
...
+        :id: 017f615a-92a8-11eb-bca1-002b677efe14
```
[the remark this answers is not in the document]
aborah-sudo: done

tests/pytest/test_error_2006_0032.py
aborah-sudo: we are now not putting any prefix , so removing it

ikerexxe:
I have reviewed it and the period in usernames is a downstream patch for Fedora and Debian derivatives. Moreover, the requirement to ban the usage of characters outside the portable filename character set `[A-Za-z0-9._-]` seems to come from the POSIX standard. So, unless shadow-utils decides to stop following the standard, the test regarding the usage of periods in usernames will have to be dropped from the upstream project.

aborah-sudo:
> I have reviewed it and the period in usernames is a downstream patch for Fedora and Debian derivatives. Moreover, the requirement to ban the usage of characters outside the portable filename character set `[A-Za-z0-9._-]` seems to come from the POSIX standard. So, unless shadow-utils decides to stop following the standard, the test regarding the usage of periods in usernames will have to be dropped from the upstream project.

Hi i have remove that part of the script as per your suggestion .

tests/pytest/test_error_2006_0032.py
```diff
@@ -0,0 +1,104 @@
...
+    Automation of Shadow Utils tests
+    """
+    def test_error_2006_0032(self):
+        """
+        :title: Tests if shadow-utils are
+         immune against bugs in 2006:0032
+        :id: TC#46675
+        :steps:
+          1. Creating tmp directory
+          2. Allowing '.' in usernames
```
ikerexxe, line 20: Since you removed the test you should also remove the comment

tests/pytest/test_error_2006_0032.py
```diff
@@ -0,0 +1,104 @@
...
+          3. Added two patches to shadow-utils for
+          additional usermod flag
+          4. The user foo must be included in both
+          foo-group and foo-group2 groups
+          5. usermod -p does not update sp_lstchg
+        :expectedresults:
+          1. Should succeed
+          2. Should succeed
```
ikerexxe, line 28: Same comment as before

ikerexxe (commented):
Apart from two minor changes everything looks good to me

tests/pytest/test_error_2006_0032.py
aborah-sudo: done

tests/pytest/test_error_2006_0032.py
aborah-sudo: done

ikerexxe:
LGTM

kaushikub:
Any idea when this patch will be merged? Is there anything else that needs to be done for this patch?

hallyn:
1. I asked before, but I still do not see any sort of reference to what these testcase identifiers mean (2006:0032, TC#46675)
2. I don't see anything particularly esoteric here, so do not see why a new test infrastructure was introduced, rather than hooking into the existing tests infrastructure and the tests/run_some script which Travis runs.
3. The test comments say "2. Added two patches to shadow-utils for additional usermod flag".  Does that mean you're testing behavior of out of tree patches?

hallyn:
Ping?

PR #327, opened 2021-04-29
fix: create relative home path correctly

squat:
Currently, supplying a relative path via the --prefix flag to the
useradd command triggers a bug in the creation of home directories. The
code seems to unintentionally prepend a leading "/" to all paths,
quietly transforming a relative prefixed home path into an absolute
path. This can be seen in the following strace logs from running
"useradd --create-home --prefix tmp/root squat":

```
access("tmp/root//home/squat", F_OK)    = -1 ENOENT (No such file or directory)
access("/mp", F_OK)                     = 0
access("/mp/root", F_OK)                = 0
access("/mp/root/home", F_OK)           = 0
access("/mp/root/home/squat", F_OK)     = -1 ENOENT (No such file or directory)
mkdir("/mp/root/home/squat", 000)       = 0
chown("/mp/root/home/squat", 0, 0)      = 0
chmod("/mp/root/home/squat", 0755)      = 0
chown("tmp/root//home/squat", 1000, 1000) = -1 ENOENT (No such file or directory)
chmod("tmp/root//home/squat", 0700)     = -1 ENOENT (No such file or directory)
```

Note that the relative path is correctly probed in the beginning and it
is only during the recursive creation that the path is turned into an
absolute path. This invocation results in the creation of a "/mp"
hierarchy in the root of the filesystem.

Similar problems occur when using `--prefix ./tmp/root`.

This commit fixes the handling of relative paths by not assuming that
the given path is anchored with a "/".

Signed-off-by: Lucas Servén Marín <lserven@gmail.com>

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2171,7 +2171,6 @@ static void create_home (void)
 							Prog, user_home);
 			fail_exit (E_HOMEDIR);
 		}
-		++bhome;
 
 #ifdef WITH_SELINUX
 		if (set_selinux_file_context (prefix_user_home) != 0) {
@@ -2188,7 +2187,11 @@ static void create_home (void)
 		 */
 		cp = strtok (bhome, "/");
 		while (cp) {
-			strcat (path, "/");
+                        /* Avoid turning a relative path into an absolute path.
+                         */
+                        if (bhome[0] == '/' || strlen (path) != 0) {
+			        strcat (path, "/");
+                        }
 			strcat (path, cp);
 			if (access (path, F_OK) != 0) {
 				/* Check if parent directory is BTRFS, fail if requesting


squat:
Friendly ping :) is anyone available to review this bug fix?

hallyn:
I agree what it's doing now is completely wrong, but wouldn't it be better to refuse a relative path?

squat:
Hmm I don't think it's correct to refuse a relative path. The other commands, e.g. userdel, correctly operates on relative paths today; useradd seems to be the only exception to me. Refusing relative paths would seem like an artificial constraint that is inconsistent with other UNIX tools and even inconsistent with the behavior of this toolset: e.g. the `--root` flag allows users to chroot to a directory in a relative path.

squat:
Ping :))

hallyn:
Ok, thanks.  So regardless of relative paths, I want to add some tests for that.  Let me merge your patch, and then I may or may not disable relative paths everywhere i see them being supported, and I intend to pull out the code doing the path walk for mkdir and add some tests.

setharnold:
On a first thought I think allowing relative paths is dangerous and prone to abuse. What's the usecase for supporting them?

Thanks

squat:
Note: this PR did not add new support for relative paths but rather fix a dangerous bug in the current implementation of relative path support that could corrupt systems. All shadow utils commands correctly support relative paths for `--path` so this is just trying to fix the delta with useradd.

I'm not using this feature myself, but rather happened upon this bug during some experimentation.

I think the motivation is that UNIX users expect all commands to support relative paths unless the opposite is explicitly documented. Whenever tools break from convention or go against the expectations of users without proper communication, I would argue this is a bug in the UX of the software.

PR #333, opened 2021-05-06
login.defs: include HMAC_CRYPTO_ALGO key

ikerexxe:
Include the new HMAC_CRYPTO_ALGO key that is needed by pam_timestamp to
select the algorithm that is going to be used to calculate the message
authentication code.

pam_timestamp is currently using an embedded algorithm to calculate the
HMAC message, but the idea is to improve this behaviour by relying on
openssl's implementation. On top of that, the ability to change the
algorithm with a simple configuration change allows to simplify the
process of removing unsecure algorithms.

Reference to pam's pull-request: https://github.com/linux-pam/linux-pam/pull/342
Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1947294

### DIFF
--- a/etc/login.defs
+++ b/etc/login.defs
@@ -467,3 +467,13 @@ USERGROUPS_ENAB yes
 # Set to "no" to not prevent for any account (dangerous, historical default)
 
 PREVENT_NO_AUTH superuser
+
+#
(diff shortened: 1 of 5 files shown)


ikerexxe:
That was fast. Thanks for your hard work @hallyn !

PR #334, opened 2021-05-06
libmisc: retain CAP_SETFCAP when mapping uid 0

brauner:
When uid 0 maps host uid 0 into the child userns newer kernels require
CAP_SETFCAP be retained as this allows the caller to create fscaps that
are valid in the ancestor userns. This was a security issue (in very
rare circumstances). So whenever host uid 0 is mapped, retain
CAP_SETFCAP if the caller had it.
Userspace won't need to set CAP_SETFCAP on newuidmap as this is really
only a scenario that real root should be doing which always has
CAP_SETFCAP. And if they don't then they are in a locked-down userns.
(LXC sometimes maps host uid 0 during chown operations in a helper
 userns but will not rely on newuidmap for that. But we don't want to
 risk regressing callers that want to rely on this behavior.)

Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

### DIFF
--- a/libmisc/idmapping.c
+++ b/libmisc/idmapping.c
@@ -123,6 +123,25 @@ struct map_range *get_map_ranges(int ranges, int argc, char **argv)
  */
 #define ULONG_DIGITS ((((sizeof(unsigned long) * CHAR_BIT) + 9)/10)*3)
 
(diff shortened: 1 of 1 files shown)


hallyn (approved):
D'oh, good point, thanks.

PR #335, opened 2021-05-08
[WIP] libsubid: don't print error messages on stderr by default

### LINKED ISSUE #325: `libsubid` shouldn't log to `stderr`
Currently 'libsubid' can log to `stderr`:
`get_subuid_ranges()` -> `list_owner_ranges()` -> `get_subid_nss_handle()` -> `nss_init()` -> `fprintf(stderr, ...)`

For `new?idmap` this is fine to print to `stderr`, but I think `libsubid` shouldn't.

([reference](https://github.com/shadow-maint/shadow/pull/321#issuecomment-826855274) for a context)

UPD: #329 was merged here

hallyn:
Closes #325

Add a new subid_init() function which can be used to specify the
stream on which error messages should be printed.  (If you want to
get fancy you can redirect that to memory :)  If subid_init() is
not called, use stderr.  If NULL is passed, then /dev/null will
be used.

This patch also fixes up the 'Prog', which previously had to be
defined by any program linking against libsubid.  Now, by default
in libsubid it will show (subid).  Once subid_init() is called,
it will use the first variable passed to subid_init().

### DIFF
--- a/libsubid/api.c
+++ b/libsubid/api.c
@@ -32,11 +32,38 @@
 #include <stdio.h>
 #include <errno.h>
 #include <stdlib.h>
+#include <string.h>
 #include <pwd.h>
 #include <stdbool.h>
 #include "subordinateio.h"
 #include "idmapping.h"
 #include "subid.h"
+
+const char *Prog = "(libsubid)";
+extern FILE * shadow_logfd;
+
+bool libsubid_init(const char *progname, FILE * logfd)
+{
+	if (progname) {
+		progname = strdup(progname);
+		if (progname)
+			Prog = progname;
+		else
+			fprintf(stderr, "Out of memory");
+	}
+
+	if (logfd) {
+		shadow_logfd = logfd;
+		return true;
+	}
+	shadow_logfd = fopen("/dev/null", "w");
+	if (!shadow_logfd) {
+		fprintf(stderr, "ERROR opening /dev/null for error messages.  Using stderr.");
+		shadow_logfd = stderr;
+		return false;
+	}
+	return true;
+}
 
 static
 int get_subid_ranges(const char *owner, enum subid_type id_type, struct subordinate_range ***ranges)
--- a/libsubid/subid.h
+++ b/libsubid/subid.h
@@ -22,6 +22,22 @@ enum subid_status {
 };
 
 /*
+ * libsubid_init: initialize libsubid
+ *
+ * @progname: Name to display as program.  If NULL, then "(libsubid)" will be
+ *            shown in error messages.
+ * @logfd:    Open file pointer to pass error messages to.  If NULL, then
+ *            /dev/null will be opened and messages will be sent there.  The
+ *            default if libsubid_init() is not called is stderr (2).
+ *
+ * This function does not need to be called.  If not called, then the defaults
+ * will be used.
+ *
+ * Returns false if an error occurred.
+ */
+bool libsubid_init(const char *progname, FILE *logfd);
+
+/*
  * get_subuid_ranges: return a list of UID ranges for a user
  *
  * @owner: username being queried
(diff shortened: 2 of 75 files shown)


hallyn:
Drat, I meant this to be a 'draft' pull request.

Only compile-tested so far.

hallyn:
oh neato, a new 'convert to draft' button.

libsubid/subid.h
```diff
@@ -21,6 +21,20 @@ enum subid_status {
 	SUBID_STATUS_ERROR = 3,
 };
 
+/*
+ * libsubid_init: initialize libsubid
+ *
+ * @progname: Name to display as program.  If NULL, then "(libsubid)" will be
```
alexey-tikhonov, line 27: I would mention pointer is kept (not memory copy). Thus caller should not release memory.

libsubid/api.c
```diff
@@ -38,6 +38,25 @@
...
+	if (progname)
+		Prog = progname;
+
+	if (logfd) {
+		shadow_logfd = logfd;
+		return;
+	}
+	shadow_logfd = fopen("/dev/null", O_RDWR);
+	if (!shadow_logfd) {
+		fprintf(stderr, "ERROR opening /dev/null for error messages.  Using stderr.");
```
alexey-tikhonov, line 55: I'm not sure this is expected if caller explicitly requested "dev/null".
Perhaps error return code for init() function would make more sense here.

libsubid/api.c
```diff
@@ -38,6 +38,25 @@
 #include "idmapping.h"
 #include "subid.h"
 
+char *Prog = "(libsubid)";
```
alexey-tikhonov, line 41: `const char *`

libsubid/subid.h
```diff
@@ -21,6 +21,20 @@ enum subid_status {
...
+ *            shown in error messages.
+ * @logfd:    Open file pointer to pass error messages to.  If NULL, then
+ *            /dev/null will be opened and messages will be sent there.  The
+ *            default if libsubid_init() is not called is stderr (2).
```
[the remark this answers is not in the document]
hallyn: That's more difficult in this approach.  Would have to use an __init, which I'd prefer not to.  When we switch to a proper log subsystem, we can make that change.

libsubid/api.c
hallyn: I'll return false on error to help the library figure out what's going on.

libsubid/subid.h
hallyn: Hm, actually I'm doing an strdup now instead.  Prefer not to risk segfaults due to an accidental caller garbage collection.

hallyn:
@alexey-tikhonov   I agree with your feeling that a proper log subsystem would be nicer, and I intend to do that "next".  But this was doable with mostly just a 'sed' and no scary complications.  As I think you're under a time pressure to release the package, this should suffice for now, right?

libsubid/api.c
```diff
@@ -32,12 +32,39 @@
...
+extern FILE * shadow_logfd;
+
+bool libsubid_init(const char *progname, FILE * logfd)
+{
+	if (progname) {
+		progname = strdup(progname);
+		if (progname)
+			Prog = progname;
+		else
+			fprintf(stderr, "Out of memory");
```
alexey-tikhonov, line 52: `return false`?

libsubid/subid.h
alexey-tikhonov: Ok.

libsubid/subid.h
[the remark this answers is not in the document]
alexey-tikhonov: Ok.

libsubid/api.c
alexey-tikhonov: Thanks for the update with error code.

I think `return false` should be enough here.
IMO, `fprintf()` in this case is unnecessary and maybe even undesirable (taking into account "/dev/null" was requested explicitly).

alexey-tikhonov:
Yes, this would suffice. Thank you, @hallyn.

(Please take a note I didn't review changes other than in `libsubid`)

PR #340, opened 2021-05-17
Don't return owner in list_owner_ranges API call.

hallyn:
Closes: 339

struct subordinate_range is pretty closely tied to the existing
subid code and /etc/subuid format, so it includes an owner.  Dropping
that or even renaming it is more painful than I'd first thought.
So introduce a 'struct subid_range' which is only the start and
count, leaving 'struct subordinate_range' as the owner, start and
count.

### DIFF
--- a/lib/subordinateio.c
+++ b/lib/subordinateio.c
@@ -17,23 +17,6 @@
 #include <ctype.h>
 #include <fcntl.h>
 
-/* subid_free_ranges: free a subid_range
- *
- * @ranges: an array of subid_ranges to free
- * @count: number of items in the array
- *
- * The subid_range is a subordinate_range without the owner field,
- * defined in subid.h
- */
-void subid_free_ranges(struct subid_range **ranges, int count)
-{
-	int i;
-
-	for (i = 0; i < count; i++)
-		free(ranges[i]);
-	free(ranges);
-}
-
 /*
  * subordinate_dup: create a duplicate range
  *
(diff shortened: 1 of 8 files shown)


hallyn:
I was originally going to rename subordinate_range to subordinate_delegated, but the way it interacts with the commonio routines makes that painful.

hallyn:
SUB_UID_COUNT=0 is still a problem.  I'll fix that before merging this.

hallyn:
> Having `owner` in the struct was the only reason to keep it `***ranges` instead of `**ranges`.
> Since this code is being touched (and lib ABI is changed) anyway, I would propose to switch `**` (i.e. "pointer to an array of structs" instead of "pointer to an array of pointers to structs")
> 
> This would also render `subid_free_ranges()` unneeded.

Yeah that's a good point, will try that.

hallyn:
> SUB_UID_COUNT=0 is still a problem. I'll fix that before merging this.

Fixed with separate PR.

lib/subordinateio.c
```diff
@@ -308,24 +326,25 @@ static bool have_range(struct commonio_db *db,
...
 		if (!new)
 			return false;
 		*ranges = new;
 	}
 	(*ranges)[n] = NULL;
-	tmp = subordinate_dup(new);
+	tmp = malloc(sizeof(*tmp));
 	if (!tmp)
 		return false;
+	memcpy(tmp, new, sizeof(*tmp));
```
alexey-tikhonov: That doesn't matter much already, since next commit fixes this...

but `tmp` was `subid_range *` and `new` was `subordinate_range *` with `owner` as a first field, so this `memcpy` was corrupting content of `tmp`...

PR #344, opened 2021-05-23
fix newusers when nss provides subids

### LINKED ISSUE #331: Inconsistency: "nsswitch:subid has network source configured" vs "local user manipulation are being done"
What is our general approach to a case: "nsswitch:subid has network source configured" and "local user manipulation are being done"?

1) `newusers`: sub_uid_assigned() -> nsswitch::has_any_range(), so if network source is configured in `nsswitch.conf` then `newusers` consults network source for a **local** user instead of /etc/subid, and this doesn't make much sense, imo.
And this is the only use of `has_any_range()` I've found. Do we need this function in a plugin at all if only local user manipulation tools use it?

2) `useradd`: sub_uid_add() -> "if network source is configured" => -EOPNOTSUPP. Again, if network source is configured in `nsswitch.conf` then `useradd` will fail to update /etc/subuid for a **local** user. Is this expected?

3) (somewhat relevant) if we suppose that a given subid can be used by multiple users (uids), then probably usage of has_range() in user_busy.c is wrong (subrange doesn't uniquely identify owner).

hallyn:
Closes #331

1. drop 'has_any_range' nss method as it is not useful

2. do not try to create a subid range in newusers when using nss for
   subids, since that's not possible.

(cherry picked from commit 88a434adbdcf4a8640793fd58bcd2ba77598349d)

### DIFF
--- a/lib/nss.c
+++ b/lib/nss.c
@@ -116,14 +116,6 @@ void nss_init(char *nsswitch_path) {
 				subid_nss = NULL;
 				goto done;
 			}
-			subid_nss->has_any_range = dlsym(h, "shadow_subid_has_any_range");
-			if (!subid_nss->has_any_range) {
-				fprintf(shadow_logfd, "%s did not provide @has_any_range@\n", libname);
-				dlclose(h);
-				free(subid_nss);
-				subid_nss = NULL;
-				goto done;
-			}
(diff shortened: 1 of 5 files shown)


hallyn:
Oops, had forgotten to open a PR for this.

Issue #346, opened 2021-05-25
libsubid.so shouldn't require symbol `shadow_logfd`


alexey-tikhonov:
This is similar to #329:

```
libsubid/api.c:43:extern FILE * shadow_logfd;

$ nm -D /usr/lib64/libsubid.so | grep shadow_logfd
                 U shadow_logfd

lib64/libsubid.so: undefined reference to `shadow_logfd'
```

I guess `shadow_logfd` should not be `extern` in `libsubid/api.c`..?

Another very minor note not worth dedicated ticket: I think `subid.h` should ‘#include <stdio.h>’ due to 'FILE' used in API.

hallyn:
Indeed that shouldn't be extern, thanks.

PR #348, opened 2021-05-29
libsubid/api.c: make shadow_logfd not extern

### LINKED ISSUE #329: `libsubid.so` shouldn't require symbol `Prog`
```
$ grep Prog lib/prototypes.h 
extern /*@observer@*/ const char *Prog;

$ nm -D /usr/lib64/libsubid.so | grep Prog
                 U Prog
```

This is quite unexpected for applications linking with `libsubid`: `.../lib64/libsubid.so: undefined reference to 'Prog'`

hallyn:
Closes #329

Also #include stdio.h


hallyn:
> #329 is already closed, so maybe this commit should point to #346

Oh dear - fixed, thank you.

alexey-tikhonov:
Tools like [`src/list_subid_ranges.c`](https://github.com/shadow-maint/shadow/blob/a6154b8572ef163ccc743d444a9317d6181ab4ec/src/list_subid_ranges.c#L7) (that use `libsubid`) should define `shadow_logfd` anymore...

PR #352, opened 2021-06-02
usermod, newusers, prefix: enforce absolute paths for homedir

hallyn:
useradd already was enforcing this, but these were not.

### DIFF
--- a/libmisc/prefix_flag.c
+++ b/libmisc/prefix_flag.c
@@ -109,6 +109,12 @@ extern const char* process_prefix_flag (const char* short_opt, int argc, char **
 			return ""; /* if prefix is "/" then we ignore the flag option */
 		/* should we prevent symbolic link from being used as a prefix? */
 
+		if ( prefix[0] != '/') {
+			fprintf (shadow_logfd,
+				 _("%s: prefix must be an absolute path\n"),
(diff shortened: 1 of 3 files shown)


votdev:
With this change it is not possible anymore to unset the user home directory. What is the suggested workaround or is this a regression?

hallyn:
> With this change it is not possible anymore to unset the user home directory. What is the suggested workaround or is this a regression?

So if user w3m has homedir /home/w3m you want to usermod -d "" ?  I guess the suggested workaround would be to just use a nonexistent path, but yes this is a regression.  Would you mind opening an issue for it (assigned to me), or if you prefer posting a patch to fix it (by just adding a check for it being the empty string and allowing that).

votdev:
> > With this change it is not possible anymore to unset the user home directory. What is the suggested workaround or is this a regression?
> 
> So if user w3m has homedir /home/w3m you want to usermod -d "" ? I guess the suggested workaround would be to just use a nonexistent path, but yes this is a regression. Would you mind opening an issue for it (assigned to me), or if you prefer posting a patch to fix it (by just adding a check for it being the empty string and allowing that).

If opened the PR https://github.com/shadow-maint/shadow/pull/1133.

PR #353, opened 2021-06-07
man: clarify subid delegation behaviour

ikerexxe:
Following the discussion https://github.com/shadow-maint/shadow/pull/345
I have changed the documentation to clarify the behaviour of subid
delegation when any subid source except files is configured.

Please, do not merge until ack from @hallyn and @alexey-tikhonov is granted.


man/newgidmap.1.xml
```diff
@@ -90,9 +90,9 @@
       The <command>newgidmap</command> sets <filename>/proc/[pid]/gid_map</filename> based on its
       command line arguments and the gids allowed. The subid delegation can come either from files
       (<filename>/etc/subgid</filename>) or from the configured NSS subid module. Only one of them
-      can be chosen at a time. So, for example, if the subid source is configured as NSS and
-      <command>groupadd</command> is executed, then the command will fail and the entry will not be
-      created in <filename>/etc/subgid</filename>.
+      can be chosen at a time. So, for example, if any subid source except files is configured and
```
brauner, line 93: "Subgid delegation can either be managed via /etc/subgid or through the configured NSS subid module. These options are mutually exclusive.
Note, that groupadd will only create entries in /etc/subgid if subid delegation is managed via subid files."

Seems a bit clearer to me.

alexey-tikhonov:
`the configured NSS subid module`  --  do we have described _anywhere_ (in man pages) how this configuration is done?

I.e. something like:
 - a field 'subid' is read from /etc/nsswitch.conf
 - 'files' in this field means /etc/sub?id
 - any other value is treated as a plugin name with 'libsubid_$value.so' template
 - missing value or missing plugin => fall back to 'files'

man/newgidmap.1.xml
ikerexxe: Yes, thanks for the feedback. I'll update it accordingly.

ikerexxe:
> `the configured NSS subid module` -- do we have described _anywhere_ (in man pages) how this configuration is done?

Not as far as I know. Do you think the subuid.5.xml and its group sibling would be a good place to state this information? Or were you thinking about another file?

alexey-tikhonov:
> > `the configured NSS subid module` -- do we have described _anywhere_ (in man pages) how this configuration is done?
> 
> Not as far as I know. Do you think the subuid.5.xml and its group sibling would be a good place to state this information? Or were you thinking about another file?

Either this or newgidmap.1.xml . I don't know what fits better.
The former would require significant rewrite of the man page though, since currently it talks about syntax of local file specifically.

ikerexxe:
I've updated the `newgidmap.1.xml`with your proposed improvements and I have also changed `subuid.5.xml` to define the subid functionality and how to configure it. If you agree on the changes I'll port those changes to the group files.

man/subuid.5.xml
```diff
@@ -48,11 +53,37 @@
...
   <refnamediv id='name'>
     <refname>subuid</refname>
-    <refpurpose>the subordinate uid file</refpurpose>
+    <refpurpose>the configuration for subordinate user ids</refpurpose>
   </refnamediv>
 
   <refsect1 id='description'>
     <title>DESCRIPTION</title>
+    <para>
+      Subuid links a user id from a parent namespace to a list of user ids from
```
hallyn, line 62: 'links' doesn't seem like quite the right word.  Also subuid only can authorize passing uids into a child namespace, of course, which might be a bit confusing in this sentence.  How would you feel about something like

Subuid authorizes a user id to map lists of user ids from its namespace into child namespaces.

Other than that, this looks great.  Did you want to also do this for man/subgid.5.xml?

man/newgidmap.1.xml
```diff
@@ -88,11 +88,11 @@
...
       The <command>newgidmap</command> sets <filename>/proc/[pid]/gid_map</filename> based on its
-      command line arguments and the gids allowed. The subid delegation can come either from files
-      (<filename>/etc/subgid</filename>) or from the configured NSS subid module. Only one of them
-      can be chosen at a time. So, for example, if the subid source is configured as NSS and
-      <command>groupadd</command> is executed, then the command will fail and the entry will not be
-      created in <filename>/etc/subgid</filename>.
+      command line arguments and the gids allowed. Subgid delegation can either be managed via
+      <filename>/etc/subgid</filename> or through the configured NSS subid module. These options
+      are mutually exclusive.
+      Note, that <command>groupadd</command> will only create entries in
```
alexey-tikhonov, line 94: Shall this go to ` man/subgid.5.xml`?

man/subuid.5.xml
alexey-tikhonov: > Subuid authorizes a user id to map lists of user ids from its namespace into child namespaces.

This variant reads better, imo.

hallyn:
@ikerexxe if you like i can make the final changes while i commit.  Please let me know.

man/subuid.5.xml
ikerexxe: I've updated this part.

man/newgidmap.1.xml
ikerexxe: Yes, I have already updated the group files to be consistent with the changes in the user files.

ikerexxe:
I've updated the changes with all your feedback. A new round of reviews is more than welcome.

brauner (approved):

ikerexxe:
@hallyn what do you think about the `This allows the user from the parent namespace to use the user ids from the child.` sentence that I included in man/subuid.5.xml file? I've been speaking about it with @alexey-tikhonov and it can be confusing. If you don't see it fit please feel free to remove it and its counterpart in man/subgid.5.xml.

man/subgid.5.xml
```diff
@@ -48,11 +53,37 @@
...
+    </para>
+    <para>
+      The delegation of the subordinate gids can be configured via the
+      <replaceable>subid</replaceable> field in
+      <filename>/etc/nsswitch.conf</filename> file. Only one value can be set
+      as the delegation source. Setting this field to
+      <replaceable>files</replaceable> configures the delegation of gids to
+      <filename>/etc/subgid</filename>. Setting any other value treats
+      the delegation as a plugin, which name is
+      <replaceable>libsubid_$value.so</replaceable>. If the value or plugin is
```
brauner, line 74: "as a plugin following with a name of the form <replaceable>libsubid_$value.so</replaceable>"

man/subgid.5.xml
```diff
@@ -48,11 +53,37 @@
...
+      missing, then the subordinate gid delegation fall backs to
```
brauner, line 75: s/fall backs/falls back/

man/subuid.5.xml
```diff
@@ -48,11 +53,37 @@
...
-    <refpurpose>the subordinate uid file</refpurpose>
+    <refpurpose>the configuration for subordinate user ids</refpurpose>
   </refnamediv>
 
   <refsect1 id='description'>
     <title>DESCRIPTION</title>
+    <para>
+      Subuid authorizes a user id to map lists of user ids from its namespace
+      into child namespaces. This allows the user from the parent namespace to
+      use the user ids from the child.
```
brauner, line 64: I'd drop the second sentence and maybe not speak of "lists" but rather for "ranges"?:

"Subid authorizes a user id to map ranges of user ids from its namespace into child namespaces."

brauner (changes requested):

man/subgid.5.xml
ikerexxe: Fixed and also changed the sentence in `subuid.5.xml`

man/subgid.5.xml
ikerexxe: Fixed and also changed the sentence in `subuid.5.xml`

man/subuid.5.xml
ikerexxe: Fixed and also changed the sentence in `subgid.5.xml`

ikerexxe:
> @hallyn what do you think about the `This allows the user from the parent namespace to use the user ids from the child.` sentence that I included in man/subuid.5.xml file? I've been speaking about it with @alexey-tikhonov and it can be confusing. If you don't see it fit please feel free to remove it and its counterpart in man/subgid.5.xml.

As @brauner also agreed to drop the sentence I've removed it.

PR #357, opened 2021-06-12
Use crypt_gensalt(), if available in libcrypt.

besser82:
Most Linux distributions, including Fedora and RHEL 8, are shipping with libxcrypt >= 4.0.

Since that version of libxcrypt the provided family of crypt_gensalt() functions are able to use automatic entropy drawn from secure system ressources, like arc4random(), getentropy() or getrandom().

Anyways, the settings generated by crypt_gensalt() are always guaranteed to works with the crypt() function.

Using crypt_gesalt() is also needed to make proper use of newer hashing methods, like yescrypt, provided by libxcrypt.

***

See https://github.com/shadow-maint/shadow/pull/303#issuecomment-764847950 for reference.

### DIFF
--- a/libmisc/salt.c
+++ b/libmisc/salt.c
@@ -15,6 +15,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#if HAVE_SYS_RANDOM_H
+#include <sys/random.h>
+#endif
 #include "prototypes.h"
 #include "defines.h"
 #include "getdef.h"
@@ -128,19 +131,46 @@ static /*@observer@*/char *l64a (long value)
 static long read_random_bytes (void)
 {
 	long randval = 0;
+
+#ifdef HAVE_ARC4RANDOM_BUF
+	/* arc4random_buf, if it exists, can never fail.  */
+	arc4random_buf (&randval, sizeof (randval));
+	goto end;
+
+#elif defined(HAVE_GETENTROPY)
+	/* getentropy may exist but lack kernel support.  */
+	if (getentropy (&randval, sizeof (randval))) {
+		goto fail;
+	}
+
+	goto end;
+
+#elif defined(HAVE_GETRANDOM)
+	/* Likewise getrandom.  */
+	if ((size_t) getrandom (&randval, sizeof (randval), 0) != sizeof (randval)) {
+		goto fail;
+	}
+
+	goto end;
+
+#else
 	FILE *f = fopen ("/dev/urandom", "r");
 
-	if (fread (&randval, sizeof (randval), 1, f) != sizeof (randval))
-	{
-		fprintf (shadow_logfd,
-			 _("Unable to read from /dev/urandom.\n"));
-
+	if (fread (&randval, sizeof (randval), 1, f) != sizeof (randval)) {
 		fclose(f);
-		exit (1);
+		goto fail;
 	}
 
 	fclose(f);
+	goto end;
+#endif
 
+fail:
+	fprintf (shadow_logfd,
+		 _("Unable to obtain random bytes.\n"));
+	exit (1);
+
+end:
 	return randval;
 }
 
--- a/libmisc/salt.c
+++ b/libmisc/salt.c
@@ -22,6 +22,13 @@
 #include "defines.h"
 #include "getdef.h"
 
+#if (defined CRYPT_GENSALT_IMPLEMENTS_AUTO_ENTROPY && \
+     CRYPT_GENSALT_IMPLEMENTS_AUTO_ENTROPY)
+#define USE_XCRYPT_GENSALT 1
+#else
+#define USE_XCRYPT_GENSALT 0
+#endif
+
 /* Add the salt prefix. */
 #define MAGNUM(array,ch)	(array)[0]=(array)[2]='$',(array)[1]=(ch),(array)[3]='\0'
 
@@ -77,21 +84,26 @@
 
 /* local function prototypes */
 static long read_random_bytes (void);
+#if !USE_XCRYPT_GENSALT
 static /*@observer@*/const char *gensalt (size_t salt_size);
+#endif /* !USE_XCRYPT_GENSALT */
 #if defined(USE_SHA_CRYPT) || defined(USE_BCRYPT)
 static long shadow_random (long min, long max);
 #endif /* USE_SHA_CRYPT || USE_BCRYPT */
 #ifdef USE_SHA_CRYPT
-static /*@observer@*/void SHA_salt_rounds_to_buf (char *buf, /*@null@*/int *prefered_rounds);
+static /*@observer@*/const unsigned long SHA_get_salt_rounds (/*@null@*/int *prefered_rounds);
+static /*@observer@*/void SHA_salt_rounds_to_buf (char *buf, unsigned long rounds);
 #endif /* USE_SHA_CRYPT */
 #ifdef USE_BCRYPT
-static /*@observer@*/void BCRYPT_salt_rounds_to_buf (char *buf, /*@null@*/int *prefered_rounds);
+static /*@observer@*/const unsigned long BCRYPT_get_salt_rounds (/*@null@*/int *prefered_rounds);
+static /*@observer@*/void BCRYPT_salt_rounds_to_buf (char *buf, unsigned long rounds);
 #endif /* USE_BCRYPT */
 #ifdef USE_YESCRYPT
-static /*@observer@*/void YESCRYPT_salt_cost_to_buf (char *buf, /*@null@*/int *prefered_cost);
+static /*@observer@*/const unsigned long YESCRYPT_get_salt_cost (/*@null@*/int *prefered_cost);
+static /*@observer@*/void YESCRYPT_salt_cost_to_buf (char *buf, unsigned long cost);
 #endif /* USE_YESCRYPT */
 
-#ifndef HAVE_L64A
+#if !USE_XCRYPT_GENSALT && !defined(HAVE_L64A)
 static /*@observer@*/char *l64a (long value)
 {
 	static char buf[8];
@@ -125,7 +137,7 @@ static /*@observer@*/char *l64a (long value)
 
 	return buf;
 }
-#endif /* !HAVE_L64A */
+#endif /* !USE_XCRYPT_GENSALT && !defined(HAVE_L64A) */
 
 /* Read sizeof (long) random bytes from /dev/urandom. */
 static long read_random_bytes (void)
@@ -199,14 +211,10 @@ static long shadow_random (long min, long max)
 #endif /* USE_SHA_CRYPT || USE_BCRYPT */
 
 #ifdef USE_SHA_CRYPT
-/*
- * Fill a salt prefix specifying the rounds number for the SHA crypt methods
- * to a buffer.
- */
-static /*@observer@*/void SHA_salt_rounds_to_buf (char *buf, /*@null@*/int *prefered_rounds)
+/* Return the the rounds number for the SHA crypt methods. */
+static /*@observer@*/const unsigned long SHA_get_salt_rounds (/*@null@*/int *prefered_rounds)
 {
 	unsigned long rounds;
-	const size_t buf_begin = strlen (buf);
 
 	if (NULL == prefered_rounds) {
 		long min_rounds = getdef_long ("SHA_CRYPT_MIN_ROUNDS", -1);
@@ -245,6 +253,17 @@ static /*@observer@*/void SHA_salt_rounds_to_buf (char *buf, /*@null@*/int *pref
 		rounds = SHA_ROUNDS_MAX;
 	}
 
+	return rounds;
+}
+
+/*
+ * Fill a salt prefix specifying the rounds number for the SHA crypt methods
+ * to a buffer.
+ */
+static /*@observer@*/void SHA_salt_rounds_to_buf (char *buf, unsigned long rounds)
+{
+	const size_t buf_begin = strlen (buf);
+
 	/* Nothing to do here if SHA_ROUNDS_DEFAULT is used. */
 	if (rounds == SHA_ROUNDS_DEFAULT) {
 		return;
@@ -265,14 +284,10 @@ static /*@observer@*/void SHA_salt_rounds_to_buf (char *buf, /*@null@*/int *pref
 #endif /* USE_SHA_CRYPT */
 
 #ifdef USE_BCRYPT
-/*
- * Fill a salt prefix specifying the rounds number for the BCRYPT method
- * to a buffer.
- */
-static /*@observer@*/void BCRYPT_salt_rounds_to_buf (char *buf, /*@null@*/int *prefered_rounds)
+/* Return the the rounds number for the BCRYPT method. */
+static /*@observer@*/const unsigned long BCRYPT_get_salt_rounds (/*@null@*/int *prefered_rounds)
 {
 	unsigned long rounds;
-	const size_t buf_begin = strlen (buf);
 
 	if (NULL == prefered_rounds) {
 		long min_rounds = getdef_long ("BCRYPT_MIN_ROUNDS", -1);
@@ -306,6 +321,11 @@ static /*@observer@*/void BCRYPT_salt_rounds_to_buf (char *buf, /*@null@*/int *p
 		rounds = B_ROUNDS_MIN;
 	}
 
+#if USE_XCRYPT_GENSALT
+	if (rounds > B_ROUNDS_MAX) {
+		rounds = B_ROUNDS_MAX;
+	}
+#else /* USE_XCRYPT_GENSALT */
 	/*
 	 * Use 19 as an upper bound for now,
 	 * because musl doesn't allow rounds >= 20.
@@ -314,6 +334,18 @@ static /*@observer@*/void BCRYPT_salt_rounds_to_buf (char *buf, /*@null@*/int *p
 		/* rounds = B_ROUNDS_MAX; */
 		rounds = 19;
 	}
+#endif /* USE_XCRYPT_GENSALT */
+
+	return rounds;
+}
+
+/*
+ * Fill a salt prefix specifying the rounds number for the BCRYPT method
+ * to a buffer.
+ */
+static /*@observer@*/void BCRYPT_salt_rounds_to_buf (char *buf, unsigned long rounds)
+{
+	const size_t buf_begin = strlen (buf);
 
 	/*
 	 * Check if the result buffer is long enough.
@@ -330,14 +362,10 @@ static /*@observer@*/void BCRYPT_salt_rounds_to_buf (char *buf, /*@null@*/int *p
 #endif /* USE_BCRYPT */
 
 #ifdef USE_YESCRYPT
-/*
- * Fill a salt prefix specifying the cost for the YESCRYPT method
- * to a buffer.
- */
-static /*@observer@*/void YESCRYPT_salt_cost_to_buf (char *buf, /*@null@*/int *prefered_cost)
+/* Return the the cost number for the YESCRYPT method. */
+static /*@observer@*/const unsigned long YESCRYPT_get_salt_cost (/*@null@*/int *prefered_cost)
 {
 	unsigned long cost;
-	const size_t buf_begin = strlen (buf);
 
 	if (NULL == prefered_cost) {
 		cost = getdef_num ("YESCRYPT_COST_FACTOR", Y_COST_DEFAULT);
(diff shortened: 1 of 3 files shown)


besser82:
@solardiz: This is PR going to make you happy.  =)

libmisc/salt.c
```diff
@@ -12,19 +12,314 @@
...
+		}
+		setting[SETTING_SIZE - 1] = '\0';
+		fprintf (shadow_logfd,
+			 _("Invalid ENCRYPT_METHOD value: '%s'.\n"
+			   "Defaulting to DES.\n"),
+			 method);
+	}
+
+	crypt_gensalt_rn(setting, rounds, NULL, 0,
+			 result, CRYPT_GENSALT_OUTPUT_SIZE);
```
solardiz, line 310: Let's check the return value here. Return value is documented, what's in `result` upon error is not, and we're not in the same library with the implementation to make such assumptions.

solardiz (commented):
While I like where this is going (thank you!), I think it's just a first step, and I find it weird that with this PR's changes we at the same time require `crypt_gensalt_rn`, yet still have the old salt generation code somewhere else in shadow (dead code now?), the legacy per-hash-type settings, and the random rounds count in a range nonsense.

solardiz:
As you can see from the Owl and ALT patches I referenced in #303, longer-term there should be only two settings - `CRYPT_PREFIX` and `CRYPT_ROUNDS` - and nothing per-hash-type.

I understand this might not be right to "force" onto shadow users without warning (having their old settings stop working). However, in that case for consistency perhaps the dependency on libxcrypt 4.0+ should be optional for now?

Just thinking out loud.

solardiz:
Let's also hear from @ldv-alt on this.

libmisc/salt.c
besser82: Fixed in rebased commit.

besser82:
@solardiz: The changes, I applied in this PR, are meant to be as transparent as possible to the end-user and/or distribution maintainer:

  * `crypt_gensalt()` is only used, if the system provides libxcrypt >= 4.0, otherwise the old code-path will be used.
  * There are no changes needed for existing configurations and such.

libmisc/salt.c
```diff
@@ -413,7 +645,7 @@ static /*@observer@*/const char *gensalt (size_t salt_size)
 #ifdef USE_YESCRYPT
 	} else if (0 == strcmp (method, "YESCRYPT")) {
 		MAGNUM(result, 'y');
-		strcat(result, YESCRYPT_salt_cost((int *)arg));
+		strcat(result, YESCRYPT_salt_cost((long *)arg));
```
solardiz, line 648: Looks like this was actually a bug being fixed here. So at least this bug fix should get in even if this PR maybe doesn't.

libmisc/salt.c
```diff
@@ -12,19 +12,321 @@
...
+	}
+
+	retval = crypt_gensalt_rn(setting, rounds, NULL, 0,
+				  result, CRYPT_GENSALT_OUTPUT_SIZE);
+
+	if (retval == NULL) {
+		for (int i = 0; i < CRYPT_GENSALT_OUTPUT_SIZE; ++i) {
+			result[i] = '\0';
+		}
+	}
```
solardiz, line 321: Why not call `crypt_gensalt_rn` from inside the `if`, so that we wouldn't need to introduce a variable? And why not use `memset` (if zeroizing the entire buffer is somehow needed, which I doubt)?

libmisc/salt.c
```diff
@@ -12,19 +12,321 @@
...
+
+	return result;
+}
+
+#else
```
solardiz, line 326: Maybe add ` /*  USE_XCRYPT_GENSALT */` here, although I don't know if this project's style requires that we possibly negate the condition in such comment on `#else`.

libmisc/salt.c
besser82: Indeed!

libmisc/salt.c
besser82: Fixed in rebased commit.

libmisc/salt.c
besser82: Fixed in rebased commit.

libmisc/salt.c
besser82: I've opened https://github.com/shadow-maint/shadow/pull/358 to keep this in mind, if the current PR doesn't get merged.

solardiz:
BTW, why not use the non-thread-safe `crypt_gensalt` if we put the result in a static buffer anyway?

besser82:
> BTW, why not use the non-thread-safe `crypt_gensalt` if we put the result in a static buffer anyway?

Wouldn't that just reorder zeroing the static buffer, need another variable for error checking, and require to strncpy the result of `crypt_gensalt` into the static buffer on success?

solardiz:
I was thinking that we'd simply return the pointer returned by `crypt_gensalt`, with no extra buffer nor copying. As to zeroing the buffer on failure, why? If we do that because we're concerned about any sensitive information being left in there (what information? just the hash type/settings?) then it's better done inside libxcrypt, not in the callers.

besser82:
> I was thinking that we'd simply return the pointer returned by `crypt_gensalt`, with no extra buffer nor copying.

Okay, that makes sense.  Changed that in rebased commit.

ikerexxe (changes requested):
I've done a review of the c code and noted several items. Can you improve them?

By the way, I've seen that https://github.com/shadow-maint/shadow/pull/358 and this PR have several changes in common. Can you consider removing the common changes from this PR?

solardiz:
@ikerexxe I think most of your comments are about code/issues copied from elsewhere in the original file. My thinking was that improving this old code is out of scope of this PR, and increasing differences between the original vs. copied instances isn't obviously an improvement. Ideally, all of this would be obsoleted with more complete use of `crypt_gensalt` later, as I had suggested. At least the random rounds number in a range nonsense should be dropped soon, I think with a separate PR.

A possible improvement to this PR would be to avoid such code copying/duplication, if this can be done cleanly.

ldv-alt:
> Let's also hear from @ldv-alt on this.

ALT is using Owl's crypt_gensalt-based approach for almost 20 years now.
In ALT's shadow package [libmisc/salt.c](http://git.altlinux.org/gears/s/..git?p=shadow.git;a=blob;f=shadow/libmisc/salt.c) contains essentially the following (everything else is #if 0'ed):
```
const char *
crypt_make_salt(/*@null@*/const char *meth, /*@null@*/void *arg)
{
        char *retval;

        retval = crypt_gensalt(getdef_str("CRYPT_PREFIX"),
                        getdef_num("CRYPT_ROUNDS", 0), NULL, 0);
        if (!retval) {
                fprintf(stderr, "Unable to generate a salt, "
                                "check your CRYPT_PREFIX and CRYPT_ROUNDS settings.\n");
                exit(1);
        }

        return retval;
}
```

So from my point of view, the sooner all this legacy code is gone the better.

besser82:
> I've done a review of the c code and noted several items. Can you improve them?

Working on it.


> By the way, I've seen that #358 and this PR have several changes in common. Can you consider removing the common changes from this PR?

This is intentional, as I've split out a bugfix into #358, that should not be overlooked in case this PR will *not* get merged.

solardiz:
I think if you keep the bugfix in this PR as well, then it should be a separate first commit (same as what's in the other PR).

libmisc/salt.c
```diff
@@ -12,19 +12,309 @@
...
+/*
+ * Generate 8 base64 ASCII characters of random salt.  If MD5_CRYPT_ENAB
+ * in /etc/login.defs is "yes", the salt string will be prefixed by "$1$"
+ * (magic) and pw_encrypt() will execute the MD5-based FreeBSD-compatible
+ * version of crypt() instead of the standard one.
+ * Other methods can be set with ENCRYPT_METHOD
+ *
+ * The method can be forced with the meth parameter.
+ * If NULL, the method will be defined according to the MD5_CRYPT_ENAB and
+ * ENCRYPT_METHOD login.defs variables.
```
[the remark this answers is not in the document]
besser82: Clarified in rebased commit.

libmisc/salt.c
```diff
@@ -12,19 +12,309 @@
...
+ * If meth is specified, an additional parameter can be provided.
+ *  * For the SHA256 and SHA512 method, this specifies the number of rounds
+ *    (if not NULL).
+ *  * For the YESCRYPT method, this specifies the cost factor (if not NULL).
+ */
+/*@observer@*/const char *crypt_make_salt (/*@null@*//*@observer@*/const char *meth, /*@null@*/void *arg)
+{
+	const char *method;
+	char setting[SETTING_SIZE];
+        unsigned long rounds = 0;
```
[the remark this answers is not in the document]
besser82: Fixed in rebased commit.

libmisc/salt.c
```diff
@@ -12,19 +12,309 @@
...
+	if (cost > Y_COST_MAX) {
+		cost = Y_COST_MAX;
+	}
+
+	return cost;
+}
+#endif /* USE_YESCRYPT */
+
+/* Maximum size of the setting for crypt_gensalt(). */
+#define SETTING_SIZE 16
```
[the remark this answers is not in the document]
besser82: Fixed in rebased commit.

libmisc/salt.c
```diff
@@ -12,19 +12,309 @@
...
+		}
+	} else {
+		rounds = (unsigned long) *prefered_rounds;
+	}
+
+	/*
+	 * Sanity checks.
+	 * Use 19 as an upper bound for now,
+	 * because musl doesn't allow rounds >= 20.
+	 */
```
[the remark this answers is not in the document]
besser82: Removed, as obviously wrong, in rebased commit.

libmisc/salt.c
```diff
@@ -12,19 +12,309 @@
...
+			max_rounds = min_rounds;
+		}
+
+		if (min_rounds > max_rounds) {
+			max_rounds = min_rounds;
+		}
+
+		rounds = (unsigned long) shadow_random (min_rounds, max_rounds);
+	} else if (0 == *prefered_rounds) {
+		return 0;
```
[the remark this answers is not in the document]
besser82: Removed, as unneeded, in rebased commit.

libmisc/salt.c
```diff
@@ -12,19 +12,309 @@
...
+	 * Sanity checks.
+	 * Use 19 as an upper bound for now,
+	 * because musl doesn't allow rounds >= 20.
+	 */
+	if (rounds < B_ROUNDS_MIN) {
+		rounds = B_ROUNDS_MIN;
+	}
+
+	if (rounds > B_ROUNDS_MAX) {
+		/* rounds = B_ROUNDS_MAX; */
```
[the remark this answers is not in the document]
besser82: This is still needed, as `B_ROUNDS_MAX` is defined to `31`, which is really the maximum rounds of the bcrypt hash method.

besser82:
> I think if you keep the bugfix in this PR as well, then it should be a separate first commit (same as what's in the other PR).

Will fix it that way.

libmisc/salt.c
```diff
@@ -12,19 +12,302 @@
...
+
+/*
+ * Add the salt prefix.
+ */
+#define MAGNUM(array,ch)	(array)[0]=(array)[2]='$',(array)[1]=(ch),(array)[3]='\0'
+#ifdef USE_BCRYPT
+/*
+ * Using the Prefix $2b$ to enable an anti-collision safety measure in musl libc.
+ * Negatively affects a subset of passwords containing the '\xff' character,
+ * which is not valid UTF-8 (so "unlikely to cause much annoyance").
```
solardiz, line 41: When revised to say `$2b$`, this comment became factually wrong. The "anti-collision safety measure" was only enabled for `$2a$`.

solardiz:
There's a typo in the function name in "Using crypt_gesalt() is also needed" in one of the commit messages.

libmisc/salt.c
solardiz: To be clear, I think this comment should simply be dropped. In fact, it never made sense to deliberately enable the anti-collision safety measure for new password, like we do in the shadow package. That measure was only intended for authentication against old potentially-buggy hashes, not for new password hashes.

besser82: Fixed in rebased commit.

besser82:
> There's a typo in the function name in "Using crypt_gesalt() is also needed" in one of the commit messages.

Also fixed.

libmisc/salt.c
```diff
@@ -90,12 +90,11 @@ static void seedRNG (void)
  */
 #define MAGNUM(array,ch)	(array)[0]=(array)[2]='$',(array)[1]=(ch),(array)[3]='\0'
 #ifdef USE_BCRYPT
-/* 
- * Using the Prefix $2a$ to enable an anti-collision safety measure in musl libc.
- * Negatively affects a subset of passwords containing the '\xff' character,
- * which is not valid UTF-8 (so "unlikely to cause much annoyance").
+/*
+ * Using the Prefix $2b$, which is the preferred prefix to use for new bcrypt
+ * hahses for a long time now.
```
solardiz: Typo in "hahses". I'd just drop the comment, really.

solardiz:
I find it weird that this PR adds as many as ~250 lines of code, most of those being copies of legacy code. There ought to be a clean way to rework it to keep and temporarily reuse the legacy code without having it duplicated. Perhaps the `#if`/`#else`/`#endif` could be moved further down, to be after the legacy functions that we temporarily reuse?

libmisc/salt.c
besser82: Fixed in rebased commit.

besser82:
> I find it weird that this PR adds as many as ~250 lines of code, most of those being copies of legacy code. There ought to be a clean way to rework it to keep and temporarily reuse the legacy code without having it duplicated. Perhaps the `#if`/`#else`/`#endif` could be moved further down, to be after the legacy functions that we temporarily reuse?

Done so in rebased commit.

besser82:
@solardiz I've given this a complete overhaul, and split out changes for code cleanup into a seperate commit before doing the implemetation of `crypt_gensalt()`.

Please have another look, if you can spare the time.

besser82:
A breakdown of this PR can be found in the following PRs:

* Small bugfix for yescrypt impl.: https://github.com/shadow-maint/shadow/pull/358
* bcypt should use `$2b$`: https://github.com/shadow-maint/shadow/pull/360
* Cleanup of `libmisc/salt.c`: https://github.com/shadow-maint/shadow/pull/361

besser82:
@ikerexxe Would you mind to have a second look on this?

libmisc/salt.c
```diff
@@ -12,33 +12,98 @@
...
 #include <stdlib.h>
 #include <stdio.h>
 #include <assert.h>
 #include "prototypes.h"
 #include "defines.h"
 #include "getdef.h"
 
+#if (defined CRYPT_GENSALT_IMPLEMENTS_AUTO_ENTROPY && \
+     CRYPT_GENSALT_IMPLEMENTS_AUTO_ENTROPY)
+/* Maximum size of the setting for crypt_gensalt(). */
```
solardiz, line 26: This comment "Maximum size of the setting for crypt_gensalt()." looks irrelevant here.

solardiz (commented):
Great that the line count doesn't increase much anymore (except in the pot file), but now the "Sanitize code" commit makes somewhat invasive changes to legacy code, which I didn't fully review. Hopefully, they're OK.

libmisc/salt.c
besser82: Ahh, yes, that somehow slipped into there…  Will remove with next push.

besser82: Fixed.

ikerexxe:
> @ikerexxe Would you mind to have a second look on this?

Sure! I've been quite busy these days but I expect to have some time tomorrow.

ikerexxe (commented):
The code now is in a better shape and it's easier to understand. I've added several minor concerns but they should be easy to fix.

By the way, taking into account that you are already touching `salt.c` file, do you mind making the necessary changes to fix https://github.com/shadow-maint/shadow/issues/376?

libmisc/salt.c
```diff
@@ -156,196 +203,174 @@ static /*@observer@*/const char *SHA_salt_rounds (/*@null@*/int *prefered_rounds
...
+{
+	const size_t buf_begin = strlen (buf);
+
+	/* Nothing to do here if SHA_ROUNDS_DEFAULT is used. */
+	if (rounds == SHA_ROUNDS_DEFAULT) {
+		return;
+	}
+
+	/* Check if the result buffer is long enough. */
+	assert (GENSALT_SETTING_SIZE > buf_begin + 17);
```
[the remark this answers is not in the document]
besser82: Fixed. See #377 also.

libmisc/salt.c
```diff
@@ -156,196 +203,174 @@ static /*@observer@*/const char *SHA_salt_rounds (/*@null@*/int *prefered_rounds
...
+
+	(void) snprintf (buf + buf_begin, 18, "rounds=%lu$", rounds);
```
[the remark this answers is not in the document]
besser82: Fixed. See #377 also.

libmisc/salt.c
```diff
@@ -156,196 +203,174 @@ static /*@observer@*/const char *SHA_salt_rounds (/*@null@*/int *prefered_rounds
...
  */
-static /*@observer@*/const char *gensalt_bcrypt (void)
+static /*@observer@*/void BCRYPT_salt_rounds_to_buf (char *buf, unsigned long rounds)
 {
-	static char salt[32];
+	const size_t buf_begin = strlen (buf);
 
-	salt[0] = '\0';
+	/* Check if the result buffer is long enough. */
+	assert (GENSALT_SETTING_SIZE > buf_begin + 3);
```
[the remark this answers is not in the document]
besser82: Fixed. See #377 also.

libmisc/salt.c
```diff
@@ -156,196 +203,174 @@ static /*@observer@*/const char *SHA_salt_rounds (/*@null@*/int *prefered_rounds
...
-	seedRNG ();
-	strcat (salt, l64a (random()));
-	do {
-		strcat (salt, l64a (random()));
-	} while (strlen (salt) < BCRYPT_SALT_SIZE);
-
-	salt[BCRYPT_SALT_SIZE] = '\0';
-
-	return salt;
+	(void) snprintf (buf + buf_begin, 4, "%2.2lu$", rounds);
```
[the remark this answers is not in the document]
besser82: Fixed. See #377 also.

besser82:
> The code now is in a better shape and it's easier to understand. I've added several minor concerns but they should be easy to fix.

Done!


> By the way, taking into account that you are already touching `salt.c` file, do you mind making the necessary changes to fix #376?

See #377 also.

besser82:
@hallyn Last one and we're done, I promise!  :P

ikerexxe:
LGTM. Thank you for your patches!

besser82:
> LGTM. Thank you for your patches!

Please wait with merging!  I forgot to include <crypt.h>…

besser82:
> > LGTM. Thank you for your patches!
> 
> Please wait with merging! I forgot to include <crypt.h>…

@hallyn Things should be fine for merging now.

besser82:
@solardiz @ikerexxe Could you please set the review flag to approved, so @hallyn knows this PR is good to merge now?

solardiz:
@besser82 I'm not formally "approving" this for two reasons: (1) I am not a maintainer and (2) my review of changes to the legacy code wasn't sufficiently thorough and I don't feel like spending enough of my time on the legacy code that I don't use to perform its proper review and testing (and this really needs retesting of all settings, not just review).

ikerexxe (commented):
I have one last concern. Why have you included the changes on the `.po` files? Have you changed anything on those files?

besser82:
> I have one last concern. Why have you included the changes on the `.po` files? Have you changed anything on those files?

No, but I updated the pot file, since my changes have added new translatable strings.  And as the pot file was updated a simple run of `make` now updates the po files as well.  Thus I've included them here, so other developers don't get annoyed by having them listed as modified files in their clone.

besser82:
> > I have one last concern. Why have you included the changes on the `.po` files? Have you changed anything on those files?
> 
> No, but I updated the pot file, since my changes have added new translatable strings. And as the pot file was updated a simple run of `make` now updates the po files as well. Thus I've included them here, so other developers don't get annoyed by having them listed as modified files in their clone.

Anyways, I've dropped the update of the translation files from this PR, and will open a new PR for them.

ikerexxe:
> > > I have one last concern. Why have you included the changes on the `.po` files? Have you changed anything on those files?
> > 
> > 
> > No, but I updated the pot file, since my changes have added new translatable strings. And as the pot file was updated a simple run of `make` now updates the po files as well. Thus I've included them here, so other developers don't get annoyed by having them listed as modified files in their clone.
> 
> Anyways, I've dropped the update of the translation files from this PR, and will open a new PR for them.

Thanks. LGTM!

ikerexxe (approved):

besser82:
@hallyn This PR should be ready to merge, then.

PR #358, opened 2021-06-12
YESCRYPT_salt_cost() has long as parameter type.

besser82:
Simple bugfix.

I've opened this PR, so it does not get overlooked, in case https://github.com/shadow-maint/shadow/pull/357 doesn't get merged.


besser82:
This one is really an easy, brain-dead fix.  It would be nice if someone could have a look at it.

PR #360, opened 2021-06-15
bcrypt should use $2b$ as prefix for setting.

besser82:
This prefix is the recommended one for new bcrypt hashes for a long time.

### DIFF
--- a/libmisc/salt.c
+++ b/libmisc/salt.c
@@ -34,7 +34,7 @@ static /*@observer@*/const char *BCRYPT_salt_rounds (/*@null@*/int *prefered_rou
 #endif /* USE_BCRYPT */
 #ifdef USE_YESCRYPT
 static /*@observer@*/const char *gensalt_yescrypt (void);
-static /*@observer@*/const char *YESCRYPT_salt_cost (/*@null@*/long *prefered_rounds);
+static /*@observer@*/const char *YESCRYPT_salt_cost (/*@null@*/int *prefered_cost);
 #endif /* USE_YESCRYPT */
 
 #ifndef HAVE_L64A
@@ -277,7 +277,7 @@ static /*@observer@*/const char *gensalt_bcrypt (void)
 /*
  * Return a salt prefix specifying the cost for the YESCRYPT method.
  */
-static /*@observer@*/const char *YESCRYPT_salt_cost (/*@null@*/long *prefered_cost)
+static /*@observer@*/const char *YESCRYPT_salt_cost (/*@null@*/int *prefered_cost)
 {
 	static char cost_prefix[5];
 	long cost;
--- a/libmisc/salt.c
+++ b/libmisc/salt.c
@@ -90,12 +90,8 @@ static void seedRNG (void)
  */
 #define MAGNUM(array,ch)	(array)[0]=(array)[2]='$',(array)[1]=(ch),(array)[3]='\0'
 #ifdef USE_BCRYPT
-/* 
- * Using the Prefix $2a$ to enable an anti-collision safety measure in musl libc.
- * Negatively affects a subset of passwords containing the '\xff' character,
- * which is not valid UTF-8 (so "unlikely to cause much annoyance").
- */
-#define BCRYPTMAGNUM(array)	(array)[0]=(array)[3]='$',(array)[1]='2',(array)[2]='a',(array)[4]='\0'
+/* Use $2b$ as prefix for compatibility with OpenBSD's bcrypt. */
+#define BCRYPTMAGNUM(array)	(array)[0]=(array)[3]='$',(array)[1]='2',(array)[2]='b',(array)[4]='\0'
 #endif /* USE_BCRYPT */
 
 #if defined(USE_SHA_CRYPT) || defined(USE_BCRYPT)


besser82:
This one is really an easy, brain-dead fix.  It would be nice if someone could have a look at it.

hallyn:
can you provide a reference for this?  (and put it in a comment)

besser82:
> can you provide a reference for this? (and put it in a comment)

Quoting from https://www.openwall.com/crypt/:

> Version 1.3 adds support for the $2b$ prefix introduced in OpenBSD 5.5+, which behaves exactly the same as crypt_blowfish's $2y$. This way, full compatibility with OpenBSD's bcrypt is achieved at this new prefix.

See also [ChangeLog for OpenBSD 5.5](https://www.openbsd.org/plus55.html) and [bcrypt(3) from OpenBSD](https://man.openbsd.org/bcrypt.3)

@solardiz can confirm this, too.

solardiz:
Yes, I confirm. This is primarily for being friendly towards OpenBSD, where bcrypt originates, so that the different systems agree on a shared prefix and exact behavior wrt some corner cases (not reached by typical passwords) that were previously handled differently. Since the change is only for newly set/changed passwords, it shouldn't cause any breakage as long as the system has an up-to-date version of libxcrypt or musl.

musl was late to add the support for `$2b$`, though - only adding it in October 2020:

https://git.musl-libc.org/cgit/musl/commit/src/crypt/crypt_blowfish.c?id=99d5098a885feae3ae8c32b407350d8ca85dd178

So the question might be whether we can expect all musl-based systems that need recent shadow to also be on very recent musl - perhaps yes?

solardiz:
> So the question might be whether we can expect all musl-based systems that need recent shadow to also be on very recent musl - perhaps yes?

@RichFelker and @ncopa might want to chime in on this.

ericonr:
Void has unfortunately not been able to update all musl versions, but assuming that commit doesn't depend on anything else, it can easily be backported.

ncopa:
No objection from Alpine. We'll backport the `$2b$`patch for musl if it ever becomes a problem.

hallyn:
Ok - I see some ack's from distros but am not sure at what point we consider it ok with all distros :)

Is there any other distro we should wait on?  Or can I merge now?

@besser82 if you can add that comment at salt.c:93 that woudl be great, else i'll do it while i roll it in.

besser82:
@hallyn Which text shall I add as comment there?

eli-schwartz:
Gentoo has a musl option too.

Hello71:
gentoo got 1.2.2 on Jan 19 (https://github.com/gentoo/gentoo/commit/2b1af680d5afaa6c5b11302e531d784996bc15f7), stabilized April 3 (https://github.com/gentoo/gentoo/commit/f63594fb5f1e5653f831e1ff16d56f67f29bcfa6), 1.2.1 dropped April 24 (https://github.com/gentoo/gentoo/commit/c35702f082adba18351aa9aa3ccb33387cad564c), so it should be ok. aiui this should only affect new passwords, so people won't suddenly be unable to log in regardless?

hallyn:
> @hallyn Which text shall I add as comment there?

Something like:

/* Use $2b$ as prefix for compatibility with OpenBSD's bcrypt */

should be fine.

besser82:
> aiui this should only affect new passwords, so people won't suddenly be unable to log in regardless?

Correct, only new passwords (or when changing passwords) are affected by this, and only if bcrypt is used for password hashing.

besser82:
> /* Use $2b$ as prefix for compatibility with OpenBSD's bcrypt */

@hallyn Done so in rebased commit.

hallyn:
Thanks.  I will merge in an hour or two unless someone says to wait.

libmisc/salt.c
```diff
@@ -34,7 +34,7 @@ static /*@observer@*/const char *BCRYPT_salt_rounds (/*@null@*/int *prefered_rou
 #endif /* USE_BCRYPT */
 #ifdef USE_YESCRYPT
 static /*@observer@*/const char *gensalt_yescrypt (void);
-static /*@observer@*/const char *YESCRYPT_salt_cost (/*@null@*/long *prefered_rounds);
+static /*@observer@*/const char *YESCRYPT_salt_cost (/*@null@*/int *prefered_cost);
```
ericonr, line 37: I guess it's lucky this didn't appear in a release, since it would have been broken on big endian platforms... Shouldn't it be possible to fix the APIs so casting to and from `void *` to different width integer pointers isn't necessary?

(this isn't me raising an issue with this PR, just a comment)

besser82: Could you please elaborate a bit more, why this would be broken on big endian platforms?

besser82: I mean, *all* the other methods use a cast from `void *` to `int *` for the parameter…  Are they broken too in case of a big endian platform?

ericonr: I meant it was lucky the previous version didn't appear in a release. `long *` and `int *` are different widths on 64 bit platforms, and if the API users expected a pointer to `int`, interpreting as a pointer to `long *` would have read a completely different piece of memory. The big endian case would have mattered if API users passed `long *` and the function used `int *`, which is the opposite of the issue here; the previous version before your fix was always accessing memory it shouldn't.

> I mean, all the other methods use a cast from void * to int * for the parameter… Are they broken too in case of a big endian platform?

They aren't broken, but doing `void *` to `int *` instead of defining APIs to be `int *` makes it easy to let such errors pass by, since they can't be type checked.

hallyn:
Thanks all.

libmisc/salt.c
besser82: Now I get you.  =)

PR #361, opened 2021-06-15
Sanitize code in libmisc/salt.c

besser82:
* Move all pre-processor defines to the top of the file.
* Unify the gensalt() function to be useable for all supported hash methods.
* Drop the gensalt_{b,yes}crypt() functions in favor of the previous change.
* Clarify the comment about how crypt_make_salt() chooses the used hash method from the settings in the login.defs file.
* Use memset() to fill static buffers with zero before using them.
* Use a fixed amount of 16 random base64-chars for the sha{256,512}crypt hash methods, which is effectively still less than the recommendation from NIST (>= 128 bits), but the maximum those methods can effectively use (approx. 90 bits).
* Rename ROUNDS_{MIN,MAX} to SHA_ROUNDS_{MIN,MAX}.
* Bugfixes in the logic of setting rounds in BCRYPT_salt_rounds().
* Likewise for YESCRYPT_salt_cost().
* Fix formatting and white-space errors.

### DIFF
--- a/libmisc/salt.c
+++ b/libmisc/salt.c
@@ -12,12 +12,66 @@
 #ident "$Id$"
 
 #include <sys/time.h>
+#include <string.h>
 #include <stdlib.h>
 #include <stdio.h>
 #include <assert.h>
 #include "prototypes.h"
 #include "defines.h"
 #include "getdef.h"
+
+/* Add the salt prefix. */
+#define MAGNUM(array,ch)	(array)[0]=(array)[2]='$',(array)[1]=(ch),(array)[3]='\0'
+
+#ifdef USE_BCRYPT
+/* Use $2b$ as prefix for compatibility with OpenBSD's bcrypt. */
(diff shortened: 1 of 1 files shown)


jubalh:
Fitting username for such a PR ;)

besser82:
@jubalh Then reviewing this must be a pleasure to you!  ;)

@hallyn Would you mind takling this one during tomorrow?

hallyn:
Thanks.  It *looks* good, and I appreciate the improvements.  But it would be easier to review if it were broken into some more pieces.  My feeble brain doesn't do well when it has to try and decide which part of a patch does what.

I'll look a bit more and merge if I don't see any problems.

I did worry a bit about the 'prefered_rounds' casting to unsigned long, but i see it always comes from #defined values so it doesn't need any checks (and would of course have been a pre-existing problem).

besser82:
@hallyn Thank you! =)

PR #377, opened 2021-06-23
libmisc/salt.c: Obtain random bytes from /dev/urandom.

### LINKED ISSUE #376: covscan issues for 4.8.1: DC.WEAK_CRYPTO
The following issues were detected by covscan for shadow-utils version 4.8.1. I've done a first review to fix the critical ones, remove the false positives and divide the issues by type.

```
Error: DC.WEAK_CRYPTO (CWE-327): [#def69]
shadow-4.8.1/libmisc/salt.c:112: dont_call: "random" should not be used for security-related applications, because linear congruential algorithms are too easy to break.
shadow-4.8.1/libmisc/salt.c:112: remediation: Use a compliant random number generator, such as "/dev/random" or "/dev/urandom" on Unix-like systems, and CNG (Cryptography API: Next Generation) on Windows.
  110|   	long ret;
  111|   	seedRNG ();
  112|-> 	drand = (double) (max - min + 1) * random () / RANDOM_MAX;
  113|   	/* On systems were this is not random() range is lower, we favor
  114|   	 * higher numbers of salt. */
```

```
Error: DC.WEAK_CRYPTO (CWE-327): [#def70]
shadow-4.8.1/libmisc/salt.c:281: dont_call: "random" should not be used for security-related applications, because linear congruential algorithms are too easy to break.
shadow-4.8.1/libmisc/salt.c:281: remediation: Use a compliant random number generator, such as "/dev/random" or "/dev/urandom" on Unix-like systems, and CNG (Cryptography API: Next Generation) on Windows.
  279|   	        salt_size <= MAX_SALT_SIZE);
  280|   	seedRNG ();
  281|-> 	strcat (salt, l64a (random()));
  282|   	do {
  283|   		strcat (salt, l64a (random()));
```

```
Error: DC.WEAK_CRYPTO (CWE-327): [#def72]
shadow-4.8.1/libmisc/salt.c:283: dont_call: "random" should not be used for security-related applications, because linear congruential algorithms are too easy to break.
shadow-4.8.1/libmisc/salt.c:283: remediation: Use a compliant random number generator, such as "/dev/random" or "/dev/urandom" on Unix-like systems, and CNG (Cryptography API: Next Generation) on Windows.
  281|   	strcat (salt, l64a (random()));
  282|   	do {
  283|-> 		strcat (salt, l64a (random()));
  284|   	} while (strlen (salt) < salt_size);
  285|   
```

besser82:
Using the random() function to obtain pseudo-random bytes for generating salt strings is considered to be dangerous.  See [CWE-327](https://cwe.mitre.org/data/definitions/327.html).

We really should use a more reliable source for obtaining pseudo-random bytes like `/dev/urandom`.

Fixes #376.

***

Also add some clarifying comments about how the minimum required buffer length is computed inside of the *_salt_rounds_to_buf() functions.

### DIFF
--- a/libmisc/salt.c
+++ b/libmisc/salt.c
@@ -11,11 +11,10 @@
 
 #ident "$Id$"
 
-#include <sys/time.h>
-#include <string.h>
-#include <stdlib.h>
-#include <stdio.h>
 #include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 #include "prototypes.h"
 #include "defines.h"
 #include "getdef.h"
@@ -74,7 +73,7 @@
 #define GENSALT_SETTING_SIZE 100
 
 /* local function prototypes */
-static void seedRNG (void);
+static long read_random_bytes (void);
 static /*@observer@*/const char *gensalt (size_t salt_size);
 #if defined(USE_SHA_CRYPT) || defined(USE_BCRYPT)
 static long shadow_random (long min, long max);
@@ -125,23 +124,27 @@ static /*@observer@*/char *l64a (long value)
 }
 #endif /* !HAVE_L64A */
 
-static void seedRNG (void)
+/* Read sizeof (long) random bytes from /dev/urandom. */
+static long read_random_bytes (void)
 {
-	struct timeval tv;
-	static int seeded = 0;
+	long randval = 0;
+	FILE *f = fopen ("/dev/urandom", "r");
 
-	if (0 == seeded) {
-		(void) gettimeofday (&tv, NULL);
-		srandom (tv.tv_sec ^ tv.tv_usec ^ getpid ());
-		seeded = 1;
+	if (fread (&randval, sizeof (randval), 1, f) != sizeof (randval))
+	{
+		fprintf (shadow_logfd,
+			 _("Unable to read from /dev/urandom.\n"));
+
+		fclose(f);
+		exit (1);
 	}
+
+	fclose(f);
+
+	return randval;
 }
 
 #if defined(USE_SHA_CRYPT) || defined(USE_BCRYPT)
-/* It is not clear what is the maximum value of random().
- * We assume 2^31-1.*/
-#define RANDOM_MAX 0x7FFFFFFF
-
 /*
  * Return a random number between min and max (both included).
  *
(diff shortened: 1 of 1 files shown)


besser82:
@hallyn This one is easy again, I promise.

libmisc/salt.c
```diff
@@ -125,23 +124,27 @@ static /*@observer@*/char *l64a (long value)
...
 #endif /* !HAVE_L64A */
 
-static void seedRNG (void)
+/* Read sizeof (long) random bytes from /dev/urandom. */
+static long read_random_bytes (void)
 {
-	struct timeval tv;
-	static int seeded = 0;
+	long randval = 0;
+	FILE *f = fopen ("/dev/urandom", "r");
```
hallyn, line 131: I fear we need to support systems without urandom.  Ideally read_random_bytes() would read urandom if possible, and fall back if not.  I realize that may open the door to certain attacks, but an unprivileged user shouldn't be able to hide /dev/urandom, at least.

Am I wrong?  I'm actually not sure what the full list of platforms using this is.

besser82: `/dev/urandom` by the time of writing is well supported by Linux, Solaris, {Free,Net,Open}BSD, Tru64 UNIX 5.1B, AIX 5.2, and HP-UX 11i v2.  Cygwin emulates it.  I don't think relying on its availability would be a blocker here.

hallyn: Ok - I'm pretty sure you're right.  let's see if anyone objects at next release :)

Thanks

besser82:
Thank you!  =)

AdamWill:
I think this breaks `chpasswd -R`, at least when the target root doesn't have `/dev/urandom`, which seems likely to be common. See https://openqa.fedoraproject.org/tests/920397#step/_do_install_and_reboot/32 for this breaking our automated tests of Fedora Rawhide, when it landed there (we use `chpasswd -R` to set a root password for the installed system at the end of certain install tests where the installer does not offer the option during installation).

AdamWill:
`-R` is documented in the manpage thus:

       -R, --root CHROOT_DIR
           Apply changes in the CHROOT_DIR directory and use the configuration files from the CHROOT_DIR directory.

so I guess an option would be to use `/dev/urandom` from the original environment (not the CHROOT_DIR), as that text does not state or imply (as I read it) that the command would actually be expected to use `/dev/urandom` from the chroot. It's not a "change" to be "appl[ied]", and it's not a "configuration file"...

PR #378, opened 2021-06-29
Update translation files.

besser82:
With the previous PRs I've added a few new translatable strings.  Thus the translation files should be updated to reflect this.

I've also added German translations for those new strings.

***

This PR is a follow up of #357.


hallyn:
Thank you.

PR #380, opened 2021-07-04
libmisc/salt.c: Use secure system ressources to obtain random bytes.

besser82:
In a previous commit we introduced `/dev/urandom` as a source to obtain random bytes from.  This may not be available on all systems, or when operating inside of a chroot.

Almost all systems provide functions to obtain random bytes from secure system ressources.  Thus we should prefer to use these, and fall back to `/dev/urandom`, if there is no such function present, as a last resort.

***

See:  https://github.com/shadow-maint/shadow/pull/377#issuecomment-873534049

@AdamWill: This PR will likely fix your issue.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -44,18 +44,19 @@ AC_HEADER_STDBOOL
 
 AC_CHECK_HEADERS(errno.h fcntl.h limits.h unistd.h sys/time.h utmp.h \
 	utmpx.h termios.h termio.h sgtty.h sys/ioctl.h syslog.h paths.h \
-	utime.h ulimit.h sys/capability.h sys/resource.h gshadow.h lastlog.h \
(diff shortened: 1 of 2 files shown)


AdamWill:
Yeah, it looks like it would have a shot at least. can you backport it to rawhide? Thanks!

besser82:
@AdamWill Here you go: https://src.fedoraproject.org/rpms/shadow-utils/pull-request/12

besser82:
@AdamWill Imported and build in Rawhide: https://koji.fedoraproject.org/koji/buildinfo?buildID=1778000

besser82:
@AdamWill Imported and build in Rawhide:  https://koji.fedoraproject.org/koji/buildinfo?buildID=1778001

AdamWill:
Thanks, it looks like it did solve the problem.

besser82:
You're welcome!

PR #382, opened 2021-07-14
fread returns element count, not element size

StevenYGui:
fread returns element count, not element size.


hallyn:
Indeed - I suppose the author meant to swap args 2 and 3, but maybe not.

Thanks for the fix.

PR #395, opened 2021-08-02
Makefile: include libeconf dependency in new*idmap

ikerexxe:
new*idmap has a dependency with libeconf since commit
c464ec55709dc931ba2f24073b8b1a86d5209ab0. I'm just adding it to the
Makefile to be able to compile in distributions that include libeconf.


brauner (approved):

hallyn:
Thanks.

Issue #338, opened 2021-05-14
Regression: 'usermod -G' fails if user has any remote groups


mcatanzaro:
We discovered that 140510de9de4771feb3af1d859c09604043a4c9b breaks `usermod -G` if the user is a member of any remote groups. Say the user "foo" is a member of remote group "foogroup". We want to add the user to local group "bargroup".

```
$ usermod -G foogroup:bargroup -- foo
```

This will fail since 140510de9de4771feb3af1d859c09604043a4c9b because usermod claims the remote group foogroup does not exist. However, we *must* specify foogroup because if we do not, then foogroup will be removed. All we want to do is *add* bargroup while *maintaining* the existing group foogroup.

This is breaking accountsservice's org.freedesktop.Accounts.User.SetAccountType method, [see here for details](https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/issues/80#note_1105137). A workaround in this case would be to use `usermod -aG` instead, but that only works when adding groups. There's currently no way to remove a group without specifying all existing groups that we don't want to remove. Avoiding that would require a hypothetical `usermod -rG`, see #337.

CC @ikerexxe

hallyn:
Trying to wrap my head around this...

How does usermod remove the user from a remote supplementary group?

mcatanzaro:
> How does usermod remove the user from a remote supplementary group?

I don't know. Apparently it shouldn't be able to, which is what 140510d was trying to fix. Doesn't matter for the purposes of this bug. For this bug, we just need to know that to remove a group user usermod, we have to use `usermod -G` which requires manually passing every group that is to be not removed, and that is no longer possible after 140510d.

hallyn:
Well no.  I'm obviously missing something here, but if user joe is a member of remote group r1, and you want to add local group l1, and so you do 'usermod -G joe l1', it should not affect joe's membership in r1, since that membership is stored elsewhere.

Obviously something in the statement I'm making above is wrong.

hallyn:
Are you storing the remote groups in sssd?  nis?  I haven't touched this stuff before, but will have to set up a test environment.

mcatanzaro:
> Well no. I'm obviously missing something here, but if user joe is a member of remote group r1, and you want to add local group l1, and so you do 'usermod -G joe l1', it should not affect joe's membership in r1, since that membership is stored elsewhere.

[...]

PR #397, opened 2021-08-03
usermod: allow all group types with -G option

ikerexxe:
The only way of removing a group from the supplementary list is to use
-G option, and list all groups that the user is a member of except for
the one that wants to be removed. The problem lies when there's a user
that contains both local and remote groups, and the group to be removed
is a local one. As we need to include the remote group with -G option
the command will fail.

This reverts commit 140510de9de4771feb3af1d859c09604043a4c9b. This way,
it would be possible to remove the remote groups from the supplementary
list.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1967641
Resolves: https://github.com/shadow-maint/shadow/issues/338

### DIFF
--- a/src/usermod.c
+++ b/src/usermod.c
@@ -187,7 +187,6 @@ static bool sub_gid_locked = false;
 static void date_to_str (/*@unique@*//*@out@*/char *buf, size_t maxsize,
                          long int date);
 static int get_groups (char *);
-static struct group * get_local_group (char * grp_name);
 static /*@noreturn@*/void usage (int status);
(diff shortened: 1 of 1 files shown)


hallyn:
So then to replace 140510d, would you want to add an extra cli flag pair to separately specify local/remote groups?

ikerexxe:
No, I think it would be best for usermod to only check the group type if there's a real change to process. For example: to add a new group. If the group was already present nothing should be done about it.

PR #399, opened 2021-08-10
useradd: avoid generating an empty subid range

ikerexxe:
useradd generates an empty subid range when adding a new user. This is
caused because there are two variables, one local and the other one
global, that have a very similar name and they are used indistinctly in
the code. The local variable loads the SUB_*ID_COUNT configuration from
the login.defs file, while the global variable, which holds a value of
0, is used to generate the subid range. Causing the empty subid range
problem.

I've merged the two variables in the local one and removed the global
variable. I prefer to do it this way to reduce the scope of it but I'm
open to doing it the other way round.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1990653

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -142,9 +142,7 @@ static bool is_sub_gid = false;
 static bool sub_uid_locked = false;
 static bool sub_gid_locked = false;
 static uid_t sub_uid_start;	/* New subordinate uid range */
-static unsigned long sub_uid_count;
 static gid_t sub_gid_start;	/* New subordinate gid range */
-static unsigned long sub_gid_count;
 #endif				/* ENABLE_SUBIDS */
 static bool pw_locked = false;
 static bool gr_locked = false;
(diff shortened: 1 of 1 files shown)


ikerexxe:
@alexey-tikhonov FYI

hallyn:
Ah, it looks like the bug may come from 0dffc7c61200f492eeac03c29fa7e93b62d3cead.  What that did was prevent us searching calling find_new_sub_uids() if sub_uid_count == 0, but find_new_sub_uids is what sets sub_uid_count.

663824ef4ca927aa2b4319b69e0bfa68282ec719 is the better fix.

note that if I simply revert 0dffc7c61200f492eeac03c29fa7e93b62d3cead, then

a) SUB_UID_COUNT 0 in login.defs simply creates no new /etc/subuid entries
b) SUB_UID_COUNT 1000 in login.defs leads to a correct subuid entry.

Perhaps the better fix yet would have been to not do the extra local calculation of "subuid_count" at all, have find_new_sub_uids() set subuid_start to -1 if login.defs set SUB_UID_COUNT to 0, and then have src/useradd.c:usr_update() skip the call to sub_uid_add() if sub_uid_start == -1.

Let me look a bit more.  If your patch handles all the cases correctly then I'll just merge it.

Thanks!

Issue #393, opened 2021-07-26
SHA_CRYPT_{MIN/MAX}_ROUNDS vs. SHA_ROUNDS_DEFAULT


ernibert:
Issue:
chpasswd and passwd "hang" for several minutes (10-20min) at 100% cpu usage (while they finally terminate successfully).

System:
newly built Linux system with shadows installed by configure/make/make install.  SHA_CRYPT_{MIN/MAX}_ROUNDS are unset out in /etc/login.defs.

Cause:
Looks like the passwords are hashed somehow using the maximum value of encryption rounds (999999999) as the resulting password line in /etc/shadow shows:
root:$6$rounds=999999999$yBrrL1jQHN.00...iWC.:18834:0:99999:7:::

Solution:
I'm not sure if i got it all right - but this tweak made chpasswd/passwd working again:

```
diff -Naur shadow-4.9.orig/libmisc/salt.c shadow-4.9/libmisc/salt.c
--- shadow-4.9.orig/libmisc/salt.c	2021-07-22 23:55:35.000000000 +0200
+++ shadow-4.9/libmisc/salt.c	2021-07-26 19:34:49.882862891 +0200
@@ -221,7 +221,7 @@
 		long max_rounds = getdef_long ("SHA_CRYPT_MAX_ROUNDS", -1);
 
 		if ((-1 == min_rounds) && (-1 == max_rounds)) {
-			rounds = SHA_ROUNDS_DEFAULT;
+			min_rounds = SHA_ROUNDS_DEFAULT;
 		}
 
 		if (-1 == min_rounds) {

```

PR #400, opened 2021-08-14
libmisc: fix default value in SHA_get_salt_rounds()

floppym:
If SHA_CRYPT_MIN_ROUNDS and SHA_CRYPT_MAX_ROUNDS are both unspecified,
use SHA_ROUNDS_DEFAULT.

Previously, the code fell through, calling shadow_random(-1, -1). This
ultimately set rounds = (unsigned long) -1, which ends up being a very
large number! This then got capped to SHA_ROUNDS_MAX later in the
function.

The new behavior matches BCRYPT_get_salt_rounds().

Bug: https://bugs.gentoo.org/808195
Fixes: https://github.com/shadow-maint/shadow/issues/393


ikerexxe:
Good catch @floppym ! Thanks for the fix.

PR #405, opened 2021-08-18
Minor cleanups

a1346054:
Spelling, whitespace, shellcheck


hallyn:
Looking through this right now . Please don't push any more changes.

So far they all look safe, so I expect to merge this.  However, in general things like this are an easy way to sneak in a s/==/=/ sort of change (or even more subtle), so I do not want to encourage these types of things.

hallyn:
(That comment does not apply for shellcheck fixes - those are always welcome.  It's long whitespace-changing patches which, although I definitely agree they improve matters, worry me)

PR #417, opened 2021-09-20
Only free sgent if it was initialized

jubalh:
`sgent` is only initialized in `get_group()` if `is_shadowgrp` is true.
So we should also only attempt to free it if this is actually the case.

Can otherwise lead to:
```
free() double free detected in tcache 2 (gpasswd)
```

### DIFF
--- a/src/gpasswd.c
+++ b/src/gpasswd.c
@@ -1207,11 +1207,13 @@ int main (int argc, char **argv)
 	sssd_flush_cache (SSSD_DB_GROUP);
 
 #ifdef SHADOWGRP
-	if (sgent.sg_adm) {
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks!

I shouldn't have taken several of those patches.  Freeing memory right
before exit just isn't worth it.

Would the result be more future-proof if we just memset sgent to 0 at
the start?  (Or move it to BSS)

jubalh:
> Would the result be more future-proof if we just memset sgent to 0 at
the start? 

Most likely would be. I also thought about doing this for this patch, but saw that it has been done this way on some other locations and adhered to it.

hallyn:
Ok, will merge as is - thanks again.

PR #422, opened 2021-10-08
Create the home and mail directories with SELinux user label

ikerexxe:
This PR contains two patches. The first one is intended to move the creation of the folders after the SELinux user has been set, to create them with the proper SELinux user.

The second one fixes a problem with the file_contexts. As the selabel handle wasn't closed the old file_contexts.* files were still mapped and used for selabel_lookup(). In order to avoid that the selabel handle has be closed to update the file_contexts.* files so that the selabel_lookup() returns the updated SELinux user.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1993081

### DIFF
--- a/lib/prototypes.h
+++ b/lib/prototypes.h
@@ -392,6 +392,7 @@ extern /*@observer@*/const char *crypt_make_salt (/*@null@*//*@observer@*/const 
 /* selinux.c */
 #ifdef WITH_SELINUX
 extern int set_selinux_file_context (const char *dst_name, mode_t mode);
+extern void reset_selinux_handle (void);
 extern int reset_selinux_file_context (void);
 extern int check_selinux_permit (const char *perm_name);
 #endif
--- a/lib/semanage.c
+++ b/lib/semanage.c
@@ -293,6 +293,7 @@ int set_seuser (const char *login_name, const char *seuser_name)
 	}
 
 	ret = 0;
+	reset_selinux_handle();
 
 done:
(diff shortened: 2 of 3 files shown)


lib/prototypes.h
```diff
@@ -392,6 +392,7 @@ extern /*@observer@*/const char *crypt_make_salt (/*@null@*//*@observer@*/const
 /* selinux.c */
 #ifdef WITH_SELINUX
 extern int set_selinux_file_context (const char *dst_name, mode_t mode);
+extern void reset_selinux_handle ();
```
cgzones, line 395: Maybe explicit declare the number of arguments to be zero:
```c
extern void reset_selinux_handle (void);
```

lib/semanage.c
```diff
@@ -297,6 +297,7 @@ int set_seuser (const char *login_name, const char *seuser_name)
 done:
 	semanage_seuser_key_free (key);
 	semanage_handle_destroy (handle);
+	reset_selinux_handle();
```
cgzones, line 300: The call is probably only needed after a successful `semanage_commit()`, so it might also go with the `r = 0` statement.

lib/prototypes.h
ikerexxe: Done

lib/semanage.c
ikerexxe: That makes sense, so I've moved the statement

ikerexxe:
The PR is ready for another round of reviews.

PR #423, opened 2021-10-09
su: Fix never alarmed SIGKILL when session terminates

lrh2000:
The buggy code was introduced nearly 5 years ago at the
commit 08fd4b69e84364677a10e519ccb25b71710ee686. The
desired behavior is that SIGKILL will be sent to the
child if it does not exit within 2 seconds after it
receives SIGTERM. However, SIGALRM is masked while
waiting for the child so it cannot wake the program
up after 2 seconds to send SIGKILL.

An example shows the buggy behavior, which exists in
Ubuntu 18.04 LTS (with login 1:4.5-1ubuntu2).
```bash
user1@localhost:~$ su user2 -c '
_term() {
  echo SIGTERM received
}
trap _term TERM

while true; do
  sleep 1
  echo still alive
done'
Password:
still alive

Session terminated, terminating shell...Terminated
SIGTERM received
still alive
still alive
still alive
still alive
```
(SIGTERM is sent in another user1's terminal by
executing `killall su`.)

Here is the desired behavior, which shows what the
commit fixes.
```bash
user1@localhost:~$ su user2 -c '
_term() {
  echo SIGTERM received
}
trap _term TERM

while true; do
  sleep 1
  echo still alive
done'
Password:
still alive

Session terminated, terminating shell...Terminated
SIGTERM received
still alive
still alive
 ...killed.
user1@localhost:~$ echo $?
255
```

### DIFF
--- a/src/su.c
+++ b/src/su.c
@@ -397,22 +397,28 @@ static void prepare_pam_close_session (void)
 		snprintf (kill_msg, sizeof kill_msg, _(" ...killed.\n"));
 		snprintf (wait_msg, sizeof wait_msg, _(" ...waiting for child to terminate.\n"));
 
+		/* Any signals other than SIGCHLD and SIGALRM will no longer have any effect,
+		 * so it's time to block all of them. */
+		sigfillset (&ourset);
+		if (sigprocmask (SIG_BLOCK, &ourset, NULL) != 0) {
+			fprintf (stderr, _("%s: signal masking malfunction\n"), Prog);
+			kill_child (0);
+			/* Never reach (_exit called). */
+		}
+
+		/* Send SIGKILL to the child if it doesn't
+		 * exit within 2 seconds (after SIGTERM) */
 		(void) signal (SIGALRM, kill_child);
 		(void) signal (SIGCHLD, catch_signals);
 		(void) alarm (2);
 
-		sigemptyset (&ourset);
-		if ((sigaddset (&ourset, SIGALRM) != 0)
-		    || (sigprocmask (SIG_BLOCK, &ourset, NULL) != 0)) {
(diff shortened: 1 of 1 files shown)


hallyn:
> > Thanks, but did you mean to keep
> > sigprocmask (SIG_UNBLOCK, &ourset, NULL);
> > at the end of the block? I think you had it before, and I don't see it now.
> 
> Actually I didn't have it before. It is meaningless as I have mentioned that these signals will no longer make any difference in [#423 (comment)](https://github.com/shadow-maint/shadow/pull/423#issuecomment-950035155) , so it has been removed before.

Are we certain that no implementation of pam_end_session will require signals to function properly?

lrh2000:
> Are we certain that no implementation of pam_end_session will require signals to function properly?

So I want to know that what is your opinion? Even in the code before this PR, we blocked all signals other than several signals listed below:
https://github.com/shadow-maint/shadow/blob/4bec156db14f3f03e5e7fc3c220b6e13059e6a74/src/su.c#L328-L348
I am also curious whether we are certain that (and how to make sure that)  every implementation of `pam_close_session` works properly if all other signals are blocked.

If we use `sigprocmask (SIG_UNBLOCK, &ourset, NULL);`, we will unblock all signals, including signals that we blocked before. Is it OK? Or should I just save and restore the signal mask?

hallyn:
> So I want to know what is your opinion?

Yeah, I'm not sure.  I was hoping you'd looked at some implementations :)  Let me look around a bit and get a feel for it.

Actually, I'll merge this now, and if I find that signals should be unblocked, i'll do a separate PR cc:ing you.

Thanks for your patch!  And the discussion :)

PR #425, opened 2021-10-10
initial github actions attempt

### LINKED ISSUE #415: Setup github actions

hallyn:
Closes #415

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,40 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: debug
+      run: |
+        id
+        whoami
+        ps -ef
+        pwd
+        cat /proc/self/uid_map
+        cat /proc/self/status
+        systemd-detect-virt
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get -y install automake autopoint xsltproc gettext expect byacc libtool
+    - name: configure
+      run: |
+        ./autogen.sh --without-selinux --disable-man
+        grep ENABLE_ config.status
+    - name: make
+      run: make
+    - name: make install
+      run: sudo make install
+    - name: Run CI tests
+      run: (cd tests; sudo ./run_some; cat testsuite.log)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,9 @@ jobs:
     - name: debug
       run: |
         id
+        which bash
         whoami
+        env
         ps -ef
         pwd
         cat /proc/self/uid_map
@@ -32,9 +34,10 @@ jobs:
       run: |
         ./autogen.sh --without-selinux --disable-man
         grep ENABLE_ config.status
-    - name: make
-      run: make
-    - name: make install
-      run: sudo make install
-    - name: Run CI tests
-      run: (cd tests; sudo ./run_some; cat testsuite.log)
+    - run: make
+    - run: make install DESTDIR=${HOME}/rootfs
+    - run: sudo make install
+    - run: |
+        cd tests
+        sudo ./run_some
+        cat testsuite.log


.github/workflows/checkin.yml
```diff
@@ -0,0 +1,31 @@
...
+  pull_request:
+    branches: [ master ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+      - uses: actions/checkout@v2
+      - name: Set up node 
+        uses: actions/setup-node@v1
```
vapier, line 12: why does shadow need nodejs ?  this is a bit weird.

.github/workflows/checkin.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Install dependencies
+        run: |
+          id
+          whoami
+          ps -ef
+          pwd
+          sudo apt-get update -qq
+          sudo apt-get -y install -qq automake autopoint xsltproc libselinux1-dev gettext expect
+          sudo apt-get -y install -qq byacc libtool
```
vapier, line 21: why not merge into the previous line ?  it'll prob be faster to batch them all together at once.

.github/workflows/checkin.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Run tests 
+        run: |
+          cat /proc/self/uid_map
+          cat /proc/self/status
+          systemd-detect-virt
+          ./autogen.sh --without-selinux --disable-man
```
vapier, line 27: it's better to break these distinct steps up.  should have one run that does autogen, one for make, one for make install, and one for running the tests.

.github/workflows/checkin.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Set up node 
+        uses: actions/setup-node@v1
+      - name: Install dependencies
+        run: |
+          id
+          whoami
+          ps -ef
+          pwd
+          sudo apt-get update -qq
+          sudo apt-get -y install -qq automake autopoint xsltproc libselinux1-dev gettext expect
```
vapier, line 20: why install libselinux1-dev when you use `--without-selinux` below ?

.github/workflows/checkin.yml
```diff
@@ -0,0 +1,31 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up node 
+        uses: actions/setup-node@v1
+      - name: Install dependencies
+        run: |
+          id
+          whoami
+          ps -ef
+          pwd
+          sudo apt-get update -qq
```
vapier, line 19: i know `-qq` is seen in a lot of runs out there, but tbh, i don't know why.  the GH actions output will collapse steps once they're done running, so the default amount of log output isn't really a problem.

.github/workflows/checkin.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Run tests 
+        run: |
+          cat /proc/self/uid_map
+          cat /proc/self/status
+          systemd-detect-virt
+          ./autogen.sh --without-selinux --disable-man
+          grep ENABLE_ config.status
+          make
+          sudo make install
+          (cd tests; sudo ./run_some; cat testsuite.log)
```
vapier, line 31: any reason for this manual invocation instead of like `make check VERBOSE=1` ?

.github/workflows/checkin.yml
hallyn: Heh, it doesn't.  I'm copy/pasting on my first attempt.

Sorry about wasting your time - probably best to wait until tomorrow to bother reviewing :)  I'm just trying to get it to trigger anything at all.

.github/workflows/checkin.yml
hallyn: Some of these are very valid points against the pre-existing .travis.yml which I'm also copy-pasting :)  I'll fix them, thanks.

hallyn:
I'm not sure how github actions are wired up  - does the .github/workflow/X.yml only get read if it's in the target branch already, not if it's in the PR branch?

vapier:
hmm, i'm not sure about that nuance.  when i've developed these before, i fork the repo and then push directly to the master branch, and then amending commits & force pushing it to retest.  once things work, i send a PR with a single commit.

you could check https://github.com/shadow-maint/shadow/settings/actions to make sure it's all enabled now.  only admins can see settings.

.github/workflows/main.yml
```diff
@@ -0,0 +1,40 @@
...
+        sudo apt-get update
+        sudo apt-get -y install automake autopoint xsltproc gettext expect byacc libtool
+    - name: configure
+      run: |
+        ./autogen.sh --without-selinux --disable-man
+        grep ENABLE_ config.status
+    - name: make
+      run: make
+    - name: make install
+      run: sudo make install
```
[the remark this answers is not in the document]
hallyn: Looks like that may have uncovered a real bug :)  thanks.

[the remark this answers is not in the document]
hallyn: We could probably play with the path and ld_library_path, or change all the individual testscripts under tests/ to use the files and binaries from $DESTDIR, but right now that's now how it works.

PR #432, opened 2021-10-28
Add missing chmod() for shadowtcb_move()

galaxy4public:
During shadowtcb_move() the directory is temporarily changed to be
owned by root:root with permissions 0700.  After the change is done,
the ownership and permissions were supposed to be restored.  The
call for chown() was there, but the chmod() call was missing.  This
resulted in the broken TCB functionality.  The added chmod() fixes
the issue.

### DIFF
--- a/lib/tcbfuncs.c
+++ b/lib/tcbfuncs.c
@@ -523,6 +523,12 @@ shadowtcb_status shadowtcb_move (/*@NULL@*/const char *user_newname, uid_t user_
 		         Prog, tcbdir, strerror (errno));
 		goto out_free;
 	}
+	if (chmod (tcbdir, dirmode.st_mode & 07777) != 0) {
+		fprintf (shadow_logfd,
+		         _("%s: Cannot change mode of %s: %s\n"),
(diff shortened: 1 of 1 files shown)


hallyn:
Looks correct to me.  Would be great if @solardiz could confirm, but github doesn't seem to let me highlight him here.

solardiz:
@hallyn Looks good to me. Also, @galaxy4public had e-mailed me about this PR.

I've just confirmed this PR's addition matches what we had in the tcb patch in Owl, modulo style and conventions changes in the surrounding code. Here's what we had in this place in `tcb_move()`:

```c
        if (!unlink_suffs(user_newname))
                goto out_free;
        if (chown(tcbdir, user_newid, dirmode.st_gid)) {
                perror("chown");
                goto out_free;
        }
        if (chmod(tcbdir, dirmode.st_mode & 07777)) {
                perror("chmod");
                goto out_free;
        }
```

It looks like a forward-porting and upstreaming error that this `chmod()` was missing.

hallyn:
Awesome - thanks @galaxy4public  and thanks @solardiz

PR #433, opened 2021-10-29
Handle malformed lines in hushlogins file.

stoeckmann:
If a line in hushlogins file, e.g. /etc/hushlogins, starts with
'\0', then current code performs an out of boundary write.
If the line lacks a newline at the end, then another character is
overridden.

With strcspn both cases are solved.

### DIFF
--- a/libmisc/hushed.c
+++ b/libmisc/hushed.c
@@ -90,7 +90,7 @@ bool hushed (const char *username)
 		return false;
 	}
 	for (found = false; !found && (fgets (buf, (int) sizeof buf, fp) == buf);) {
-		buf[strlen (buf) - 1] = '\0';
+		buf[strcspn (buf, "\n")] = '\0';
 		found = (strcmp (buf, pw->pw_shell) == 0) ||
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks, this looks good.  As a trivial test of corner cases I tried:

```
#include <stdio.h>
#include <string.h>

int main() {
        char *x = "\0abcd\n";
        int x1 = strlen(x);
        int x2 = strcspn(x, "\n");
        char *y = "abcd\n";
        int y1 = strlen(y);
        int y2 = strcspn(y, "\n");
        char *z = "abcd";
        int z1 = strlen(z);
        int z2 = strcspn(z, "\n");
        printf("x1 is %d, x2 is %d\n", x1, x2);
        printf("y1 is %d, y2 is %d\n", y1, y2);
        printf("z1 is %d, z2 is %d\n", z1, z2);
}
```

which gives

```
x1 is 0, x2 is 0
y1 is 5, y2 is 4
z1 is 4, z2 is 4
```

PR #435, opened 2021-11-01
groupdel: fix SIGSEGV when passwd does not exist

freedge:
When using groupdel with a prefix, groupdel will attempt to read a
passwd file to look for any user in the group. When the file does not
exist it cores with segmentation fault.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1986111

### DIFF
--- a/libmisc/prefix_flag.c
+++ b/libmisc/prefix_flag.c
@@ -288,6 +288,9 @@ extern struct passwd* prefix_getpwent()
 	if (!passwd_db_file) {
 		return getpwent();
 	}
+	if (!fp_pwent) {
+		return NULL;
+	}
 	return fgetpwent(fp_pwent);
 }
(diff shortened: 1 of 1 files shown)


ikerexxe (approved):
Thanks for the patch! It looks good from my side.

I've tested it before the applying the patch:
```
$ mkdir -p $HOME/custom_root/etc/ && touch $HOME/custom_root/etc/group
$ sudo groupadd -P $HOME/custom_root test_group
$ groupdel -P $HOME/custom_root test_group
Segmentation fault (core dumped)
```

And after applying the patch:
```
$ sudo groupdel -P $HOME/custom_root test_group
```

hallyn:
Looks good, thanks.  (Had to look several times as the prefix code can be a bit backward at times)

PR #436, opened 2021-11-10
getsubids: system binary for user's sub*ids

ikerexxe:
Rename list_subid_ranges to getsubids to provide a system binary to
check the sub*ids of a user. The intention is to provide this binary
with any distribution that includes the subid feature, so that system
administrators can check the subid ranges of a given user.

Moreover, include a new configuration option "--with-subids-helpers" to
select whether the tool will be compiled as a non-installable binary or
a system binary.

Finally, add a man page to explain the behaviour of getsubids.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1980780

### DIFF
--- a/man/getsubids.1.xml
+++ b/man/getsubids.1.xml
@@ -1,0 +1,141 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Copyright (c) 2021 Iker Pedrosa
+   All rights reserved.
+
+   Redistribution and use in source and binary forms, with or without
+   modification, are permitted provided that the following conditions
+   are met:
+   1. Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+   2. Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+   3. The name of the copyright holders or contributors may not be used to
+      endorse or promote products derived from this software without
+      specific prior written permission.
+
+   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+   ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+   LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+   PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT
+   HOLDERS OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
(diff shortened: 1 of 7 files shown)


man/getsubids.1.xml
```diff
@@ -0,0 +1,124 @@
...
+      </arg>
+    </cmdsynopsis>
+  </refsynopsisdiv>
+
+  <refsect1 id='description'>
+    <title>DESCRIPTION</title>
+    <para>
+      The <command>getsubids</command> command lists the subordinate user ID
+      ranges for a given user. The subordinate group IDs can be listed using
+      the <option>-g</option> option.
```
AkihiroSuda, line 72: Could you add an example output, and explanation of the format?

configure.ac
```diff
@@ -306,6 +306,9 @@ AC_ARG_WITH(group-name-max-length,
 AC_ARG_WITH(su,
 	[AC_HELP_STRING([--with-su], [build and install su program and man page @<:@default=yes@:>@])],
 	[with_su=$withval], [with_su=yes])
+AC_ARG_WITH(subids-helpers,
+	[AC_HELP_STRING([--with-subids-helpers], [build and install subordinate ids helpers and man pages @<:@default=no@:>@])],
+	[with_subids_helpers=$withval], [with_subids_helpers=no])
```
AkihiroSuda, line 311: Do we really need a new `configure` flag?

I'd like to see the `getsubids` binary to be always installed when `newuidmap`  / `newgidmap` is installed.

ikerexxe: I added the flag just in case some distribution doesn't want to include the binary, but I wouldn't mind removing it as it would reduce the review and maintenance effort.

man/getsubids.1.xml
ikerexxe: Ive updated the PR with the example. Can you review it?

AkihiroSuda: Thanks, looks good (non-binding), but it might be better if we can make the format to be compatible with `cat /etc/subuid` so that caller programs can reuse existing subuid parsers.

```
$ getsubids user1
# 65,536 subuids (100000-165535) for user1
user1:100000:65536
```

Can be another PR with additional flag, though.

ikerexxe: I think it's best to have another option or just use `cat /etc/subuid | grep testuser`

hallyn (approved):
Ok - since I had only intended this as a test program I didn't much think through good output or sane CLI API...  But It doesn't look too bad as is and should be extensible/fixable if needed.  So let's merge this as is.

Thanks.

man/getsubids.1.xml
hallyn: Yeah if someone wants that they can add a --output=classic or something option to give that format.

configure.ac
hallyn: Up to you @ikerexxe .  It's not setuid so I don't mind it always being included.  Please shout here today if you want me to wait for an update to remove it, else I'll go ahead and merge.

ikerexxe: Then, let's include it always. Let me change that today.

ikerexxe:
I've updated the changes to remove the compilation flag, so I think it's ready for merging

hallyn:
Thanks.

PR #437, opened 2021-11-10
newgrp: fix segmentation fault

ikerexxe:
Fix segmentation fault in newgrp when xgetspnam() returns a NULL value
that is immediately freed.

The error was committed in
https://github.com/shadow-maint/shadow/commit/e65cc6aebcb4132fa413f00a905216a5b35b3d57

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2019553


hallyn:
Should spw_free check for NULL instead?

hallyn:
(Even if we do add a NULL check to spw_free, this patch still is fine of course, so let's unblock)

ikerexxe:
> Should spw_free check for NULL instead?

Yes, I've made a note to improve it.

PR #439, opened 2021-11-12
useradd: change SELinux labels for home files

ikerexxe:
Change SELinux labels for files copied from the skeleton directory to
the home directory.

This could cause gnome's graphical user adding to fail without copying
the full skeleton files.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2022658


hallyn:
Thanks.

Think I might add a #define for RESET_SELINUX v NORESET_SELINUX, because that argument list is not very helpful to a casual reader/reviewer :)

PR #445, opened 2021-11-15
pwck: fix segfault when calling fprintf()

ikerexxe:
As `shadow_logfd` variable is not set at the beginning of the program if
something fails and fprintf() is called a segmentation fault happens.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2021339


hallyn (approved):
Thanks.  Did you check other programs to see whether this is the only one with this problem?

If not I'll make a note to do so.

ikerexxe:
Yes, I checked them and they seemed fine.

PR #446, opened 2021-11-19
lib: check NULL before freeing passwd data

ikerexxe:
Add an additional NULL check condition in spw_free() and pw_free() to
avoid freeing an already empty pointer.

This is a follow-up to https://github.com/shadow-maint/shadow/pull/437#issuecomment-971622934


hallyn:
Thanks.

PR #449, opened 2021-11-27
2021 11 27/libsubid symbols

hallyn:


hallyn:
@brauner @rbalint @ikerexxe So this does change the names of the exported variables. Are we ok with that?  Do we maybe want #defines in subid.h from the old to the new names?

jubalh:
I hope it's ok to ask a quick question here: what does the `--enable-shared` actually do?

hallyn:
> I hope it's ok to ask a quick question here: what does the `--enable-shared` actually do?

It sets ENABLE_SHARED which is hooked to libsubid/Makefile.am:

  2 if ENABLE_SHARED
  3 libsubid_la_LDFLAGS = -Wl,-soname,libsubid.so.@LIBSUBID_ABI@ \
  4     -shared -version-info @LIBSUBID_ABI_MAJOR@
  5 endif

It pre-exists libsubid, but I don't believe it's doing anything else.
configure.ac always does both AM_ENABLE_STATIC and AM_ENABLE_SHARED.

rbalint:
@hallyn IMO it is perfectly fine to rename the symbols together with the major version bump. Since the library is not widely used yet IMO there is no need to provide the mapping `#define`-s. From Debian's POV the library has never been even in experimental, thus it is not known in the official archive.

ikerexxe:
I don't think that configuring the API names with `#define`s is a good idea and I prefer to keep them "hardcoded". @alexey-tikhonov, what do you think about this change? Will it affect Fedora or RHEL?

alexey-tikhonov:
Podman is already using existing API upstream and in Fedora/RHEL.
@giuseppe, what do you think?

@hallyn, what is the motivation to rename functions?

giuseppe:
> Podman is already using existing API upstream and in Fedora/RHEL.
> @giuseppe, what do you think?

yes, Podman is already using these functions so it would be a breaking change: https://github.com/containers/storage/blob/main/pkg/idtools/idtools_supported.go#L34-L38

hallyn:
> Podman is already using existing API upstream and in Fedora/RHEL. @giuseppe, what do you think?
> 
> @hallyn, what is the motivation to rename functions?

The motivation is to have a prefix for all the functions it exports, to avoid conflicts.

hallyn:
> > Podman is already using existing API upstream and in Fedora/RHEL.
> > @giuseppe, what do you think?
> 
> yes, Podman is already using these functions so it would be a breaking change: https://github.com/containers/storage/blob/main/pkg/idtools/idtools_supported.go#L34-L38

Ok, thanks.  So perhaps I could add them as aliases for now, with the intent of removing the old names "eventually"?

giuseppe:
> Ok, thanks. So perhaps I could add them as aliases for now, with the intent of removing the old names "eventually"?

sure thanks, that would help.

Thinking more of it: would it be possible to export the ABI major number as part of `subid.h`?  In this way, I could address the problem directly in the Cgo code with some macros (not sure if there is a smarter way in Go to have the equivalent of `AC_CHECK_FUNCS`) so you won't have to worry about carrying the alias.  What do you think?

alexey-tikhonov:
> > yes, Podman is already using these functions so it would be a breaking change: https://github.com/containers/storage/blob/main/pkg/idtools/idtools_supported.go#L34-L38
> 
> Ok, thanks. So perhaps I could add them as aliases for now, with the intent of removing the old names "eventually"?

When do you plan next upstream release?

rbalint:
> 



> > Ok, thanks. So perhaps I could add them as aliases for now, with the intent of removing the old names "eventually"?
> 
> sure thanks, that would help.
> 
> Thinking more of it: would it be possible to export the ABI major number as part of `subid.h`? In this way, I could address the problem directly in the Cgo code with some macros (not sure if there is a smarter way in Go to have the equivalent of `AC_CHECK_FUNCS`) so you won't have to worry about carrying the alias. What do you think?

This is redundant and I have not seen this anywhere. I'm sure there is something in Go that helps with major shared library version bumps. If you need something at run time, you can certainly use dlsym from Go.

rbalint:
> > I have not seen this anywhere
> 
> there are a bunch of `.*_ABI_VERSION` definitions under `/usr/include`.
> 
> Said so, if it looks redundant, I will find some other way to deal with the renaming.

I'm sorry, you are righ and I did not get what you meant.
I suggest setting not only the major version, but minor and micro to be able to check for it like it is done for glib: `GLIB_CHECK_VERSION(2, 68, 0)`.

hallyn:
> When do you plan next upstream release?

Possibly Dec 11.

hallyn:
@giuseppe  how do you feel about https://github.com/hallyn/shadow/commit/52302daf58b1bb7b67e037c74748f702ac3b86f3 ?

I can still add the symbols if you like, but arguably it makes things uglier by forcing us to deal with that again later.

giuseppe:
> @giuseppe how do you feel about [hallyn@0c9f641](https://github.com/hallyn/shadow/commit/0c9f64140852e8d50cdd769951eb0624e834483d) ?

@hallyn thanks for amending the patch!  That works for me and I think there is no need to keep the symbol name.

giuseppe:
> Please let me know if you run into any problems, or need me to wait (to do a release) on more testing.

no problems from my side, I've tested the current master version with https://github.com/containers/storage/pull/1086 and it works well.

PR #451, opened 2021-12-05
Update licensing info

### LINKED ISSUE #238: Licensing
Don't know if this has been answered but I'm confused by the licensing on this project that's available [here](https://github.com/shadow-maint/shadow/blob/master/COPYING) it says:

> This license has been obsoleted by the change to the BSD-style copyright.

Does this mean this is licensed under BSD and if so what variant of BSD is it ? I don't want to presume. If this is the wrong place can you redirect me to the correct place to ask this question.

hallyn:
Closes #238


jfhaugh:
Commenting to get notified on changes.

Lemme know when you're done and I'll review and approve so there's traceability that I approved nuking the Artistic License. If anyone needs a signed piece of paper, I can do that as well.

lib/commonio.c
```diff
@@ -1,33 +1,10 @@
 /*
- * Copyright (c) 1990 - 1994, Julianne Frances Haugh
- * Copyright (c) 1996 - 2001, Marek Michałkiewicz
- * Copyright (c) 2001 - 2006, Tomasz Kłoczko
- * Copyright (c) 2007 - 2011, Nicolas François
- * All rights reserved.
+ * SPDX-FileCopyrightText: 1990 - 1994, Julianne Frances Haugh
+ * SPDX-FileCopyrightText: 1996 - 2001, Marek Michałkiewicz
+ * SPDX-FileCopyrightText: 2001 - 2006, Tomasz Kłoczko
+ * SPDX-FileCopyrightText: 2007 - 2011, Nicolas François
```
andrewshadura, line 5: How about removing spaces from the year ranges?

andrewshadura: Hmm, from the rest of the files this seems like a conscious style decision, not an oversight.

andrewshadura (approved):

lib/commonio.c
hallyn: Yes, I'm maintaining the spacing from the original copyright statements.  Not sure how that fits with spdx specs, but it makes it easier to read IMO.

hallyn:
> Commenting to get notified on changes.
> 
> Lemme know when you're done and I'll review and approve so there's traceability that I approved nuking the Artistic License. If anyone needs a signed piece of paper, I can do that as well.

I'll wait on that approval and this merge before doing a next release.  That way hopefully the release can be used as straightforwardly as possible in the new debian merge.

rbalint:
I'm waiting with the Debian upload until this licensing change lands (and ideally gets in a release).

COPYING
```diff
@@ -1,118 +1,41 @@
...
-MERCHANTIBILITY AND FITNESS FOR A PARTICULAR PURPOSE.
-
-				The End
+SPDX-License-Identifier: BSD-3-Clause
+
+All files under this project either
+
+1. fall under the BSD 3 clause license (by default).
+
+2. carry an SBDX header declaring what license applies.
```
rbalint, line 7: SPDX

COPYING
```diff
@@ -1,118 +1,41 @@
...
-6.  The name of the Copyright Holder may not be used to endorse or
-promote products derived from this software without specific prior
-written permission.
-
-7.  THIS PACKAGE IS PROVIDED "AS IS" AND WITHOUT ANY EXPRESS OR IMPLIED
-WARRANTIES, INCLUDING, WITHOUT LIMITATION, THE IMPLIED WARRANTIES OF
-MERCHANTIBILITY AND FITNESS FOR A PARTICULAR PURPOSE.
-
-				The End
+SPDX-License-Identifier: BSD-3-Clause
```
rbalint, line 1: I think BSD-3-Clause is not a valid license for the whole project. I think there is none, since `contrib/atudel` is BSD-4-Clause which is not compatible with GPL https://en.wikipedia.org/wiki/BSD_licenses#4-clause_license_(original_%22BSD_License%22), yet `src/vipw.c` is licensed under GPL.
I suggest dropping `contrib/atudel` or splitting it out to a separate project.

If `vipw.c` stays in the source under GPL, then the the license of the project should be GPL (with BSD parts).

rbalint (changes requested):

COPYING
andrewshadura: I don't think so since those are all separate programs, and this situation is far from being unique; I've seen quite a few projects with non-GPL parts, e.g. an application and a client library to talk to it.

rbalint: @andrewshadura Do you have some links? I guess you question the GPL as the whole project's license and not the problem with `atudel`.

andrewshadura: Most definitely GPL is not the project’s license, but just the license of one independent program.

rbalint: @andrewshadura I agree that the project is belived to be BSD licensed, but looking at the current state the release tarballs can't be distributed under BSD. If we remove `atudel` the resulting tarball can be distributed under GPL, but not under BSD (3 clause), because it contains a GPL licensed source file. If we cut `vipw.c`, the remaining tarball can be distributed under BSD (3 clause).
From Debian's perspective I'm OK with keeping `vipw.c` in the tarball, but I have to exclude `atudel`, for sure, to keep `shadow`'s source in `main`.
It could be claimed, that `atudel` and `vipw.c` are only bundled, but this needs an explanation in the COPYING file, IMO.

andrewshadura: @rbalint, you’re misinterpreting it. `atudel` has no relationship whatsoever to `vipw` except being shipped in the same tarball. It is not a derivative work of `vipw`, it’s not based on it, linked with it or anything really except being bundled together.

You cannot claim that the license of the whole package is GPL, because it’s BSD-3-Clause for most files, GPL for `vipw` and BSD-4-Clause for `atudel`. I don’t think a concept of a "license for the whole package" makes sense at all in this case.

COPYING
andrewshadura:
```suggestion
2. carry an SPDX header declaring what license applies.
```

hallyn: Oops.  Thanks.

COPYING
hallyn: Originally I'd planned to remove the copyright statement from the files which fall under the 3 clause BSD license.  The COPYING file then just listed the default.  It explicitly says that files may list alternative licenses, which vipw does.

Perhaps I should add a paragraph to say that all new files are to be BSD-3.  But as @andrewshadura says, vipw and atudel are the exceptions.  It may make a difference from a debian packaging perspective, but for the git repo all I'm doing is documenting the existing licenses.  We will not relicense other files just because vipw is under src/.  In particular, section 2 of the GPL says:

```
In addition, mere aggregation of another work not based on the Program with the Program (or with a work based on the Program) on a volume of a storage or distribution medium does not bring the other work under the scope of this License. 
```

The git repo is (in my estimation) a storage or distribution medium.  No other programs in shadow are based on vipw.

Issue #454, opened 2021-12-14
Simplify usermod.c


alejandro-colomar:
In parsing the argument to the `--expiredate` option, there's a special case for `""`: <https://github.com/shadow-maint/shadow/blob/7687ae4dbd995f4042e375f750c75953568a2444/src/usermod.c#L1076>

```
			case 'e':
				if ('\0' != *optarg) {
					user_newexpire = strtoday (optarg);
					if (user_newexpire < -1) {
						fprintf (stderr,
						         _("%s: invalid date '%s'\n"),
						         Prog, optarg);
						exit (E_BAD_ARG);
					}
					user_newexpire *= DAY / SCALE;
				} else {
					user_newexpire = -1;
				}
				eflg = true;
				break;
```

However, (the new version of) `strtoday()` already handles that special case: <https://github.com/shadow-maint/shadow/blob/7687ae4dbd995f4042e375f750c75953568a2444/libmisc/strtoday.c#L77>

```
	if ((NULL == str) || ('\0' == *str)) {
		return -1;
	}

	...
```

So the calling code could be simplified to the following:

```
			case 'e':
				user_newexpire = strtoday (optarg);
				if (user_newexpire < -1) {
					fprintf (stderr,
					         _("%s: invalid date '%s'\n"),
					         Prog, optarg);
					exit (E_BAD_ARG);
				}
				user_newexpire *= DAY / SCALE;
				eflg = true;
				break;
```

---

There are 2 versions of `strtoday()` (actually 3), and at first I thought that the old ones might not handle correctly the special case.  However, I found what could be a bug, which I think would handle `""` correctly by accident.  Instead of returning -2 on error, the old functions return -1, and since they don't handle `""` specially, they will report an error.  Since `-1` is handled not as an error but as a 'never' by the calling code, it would work.

I'll send a pull request soon and link to this issue.

ikerexxe:
As the PR was merged I guess you could close this issue. Am I right?

alejandro-colomar:
Ahh, I forgot.  Sure

PR #455, opened 2021-12-14
usermod: Remove special case for ""

alejandro-colomar:
That special case is already handled by the called function: strtoday()
so we can simplify the calling code.

Link: <https://github.com/shadow-maint/shadow/issues/454>

--

**NOTE**: I didn't test the change, since I couldn't find instructions on how to build it in the README, and there's no `Makefile`.

### DIFF
--- a/src/usermod.c
+++ b/src/usermod.c
@@ -1073,18 +1073,14 @@ static void process_flags (int argc, char **argv)
 				}
 				break;
 			case 'e':
-				if ('\0' != *optarg) {
-					user_newexpire = strtoday (optarg);
-					if (user_newexpire < -1) {
-						fprintf (stderr,
-						         _("%s: invalid date '%s'\n"),
-						         Prog, optarg);
-						exit (E_BAD_ARG);
-					}
-					user_newexpire *= DAY / SCALE;
-				} else {
-					user_newexpire = -1;
+				user_newexpire = strtoday (optarg);
+				if (user_newexpire < -1) {
+					fprintf (stderr,
+						 _("%s: invalid date '%s'\n"),
+						 Prog, optarg);
+					exit (E_BAD_ARG);
 				}
+				user_newexpire *= DAY / SCALE;
 				eflg = true;
 				break;
 			case 'f':


hallyn:
> **NOTE**: I didn't test the change, since I couldn't find instructions on how to build it in the README, and there's no `Makefile`.

Hi,

I generally build for dev/test using:

./autogen.sh
make
DESTDIR=/tmp/shadow make install

tweaking autogen.sh as needed to tweak options.

src/usermod.c
```diff
@@ -1073,18 +1073,14 @@ static void process_flags (int argc, char **argv)
...
-				} else {
-					user_newexpire = -1;
+				user_newexpire = strtoday (optarg);
+				if (user_newexpire < -1) {
+					fprintf (stderr,
+						 _("%s: invalid date '%s'\n"),
+						 Prog, optarg);
+					exit (E_BAD_ARG);
 				}
+				user_newexpire *= DAY / SCALE;
```
hallyn, line 1083: You are correct that the '\0' case is handled in strtoday.  However, here you will take -1 and set user_newexpire = -1 * DAY / SCALE.

alejandro-colomar: Yes, I thought about that, but then this result is in the end passed to `date_to_str()`, isn't it?  And it interprets any negative number as `"never"`, right?

https://github.com/shadow-maint/shadow/blob/master/src/usermod.c#L1109

https://github.com/shadow-maint/shadow/blob/master/src/usermod.c#L631

https://github.com/shadow-maint/shadow/blob/master/src/usermod.c#L216

alejandro-colomar: However, having the special case might make it more explicit.

hallyn: My ctags file has let me to the chage.c version of date_to_str.  Which
still might have ended up ok but was less obvious.

This'll do - thank you.

alejandro-colomar: Maybe `date_to_str()` could be moved to `<libmisc/>` to have a single definition?

hallyn: > Maybe `date_to_str()` could be moved to `<libmisc/>` to have a single definition?

Looks like in each case it's to do with expiration, so I think that's a good idea.  Do you want to propose a patch?

alejandro-colomar: Sure :)

alejandro-colomar: What's the difference between `<lib/>` and `<libmisc/>`?  BTW, I also found a definition of `date_to_str()` in `<src/passwd.c>`, so I'll consider that one too.  In that one, I'll move the static buffer to a global static one to make it work with the new definition.

alejandro-colomar: Patch proposed: https://github.com/shadow-maint/shadow/pull/467

PR #460, opened 2021-12-17
script to kill subjects processes

### LINKED ISSUE #404: `userdel -f` can pose a security threat
## Ubuntu Distro
`root@dopeDroplet:~# lsb_release -a`
```
No LSB modules are available.
Distributor ID:  Ubuntu
Description:  Ubuntu 14.04.6 LTS
Release:  14.04
Codename:  trusty
```

## Security Threat
1. System Admin creates group `G`.
2. System Admin creates `userB` and `userE` under the same group `G`.
3. `userE` creates a process `P` which is a backdoor that receives commands and runs it using `nohup &`.
4. System Admin deletes `userE` using `userdel -f`.
5. Linux would delete `userE` but keep the process `P` running with user id of the group `G`.
5. System Admin later inspects group `G` and only finds `userB` in it, then decides to repurpose group `G` and gives it more permissions.
6. Deleted `userE` has a backdoor to the computer.

### Real use case
1. Schools creates users for students that have CS classes, then deletes these users automatically (potentially with `userdel -f`) when they leave the school.
2. Company has a shared server, employee leaves and still has access to company's server.

### Patch suggestion
* `userdel -f` should kill all processes for user

## Setup user
`root@dopeDroplet:~# useradd sahmoud`
`root@dopeDroplet:~# passwd sahmoud`
`root@dopeDroplet:~# su sahmoud`

## Setup script
`sahmoud@dopeDroplet:/$ vim /tmp/sahmoud.sh`
```
while [ true ]
do
  echo `whoami` `id -g` `date` >> /tmp/sahmoudoutput
  sleep 30
done
```

`sahmoud@dopeDroplet:/$ chmod 777 sahmoud.sh`
`sahmoud@dopeDroplet:/$ nohup ./sahmoud.sh &`
## Test script
wait a few minutes
`sahmoud@dopeDroplet:/$ cat /tmp/sahmoudoutput`
```
sahmoud 1004 Tue Aug 17 00:51:07 EDT 2021
sahmoud 1004 Tue Aug 17 00:51:38 EDT 2021
sahmoud 1004 Tue Aug 17 00:52:08 EDT 2021
sahmoud 1004 Tue Aug 17 00:52:38 EDT 2021
sahmoud 1004 Tue Aug 17 00:53:08 EDT 2021
```

## Check PID
`sahmoud@dopeDroplet:/$ ps -aux | grep sahmoud.sh`
```
sahmoud   5789  0.0  0.0   4448   672 pts/1    S    00:51   0:00 /bin/sh ./sahmoud.sh
sahmoud      6287  0.0  0.0  11752   924 pts/1    S+   00:53   0:00 grep sahmoud.sh
```

## Delete user
Exit as user `sahmoud`: `exit`
`root@dopeDroplet:~# userdel sahmoud`
```
userdel: user sahmoud is currently used by process 5789
```

`root@dopeDroplet:~# userdel -f sahmoud`
```
userdel: user sahmoud is currently used by process 5789
```

`root@dopeDroplet:~# userdel -f sahmoud`
```
userdel: user 'sahmoud' does not exist
```

`root@dopeDroplet:~# cat /tmp/sahmoudoutput`
```
sahmoud 1004 Tue Aug 17 00:51:07 EDT 2021
sahmoud 1004 Tue Aug 17 00:51:38 EDT 2021
sahmoud 1004 Tue Aug 17 00:52:08 EDT 2021
sahmoud 1004 Tue Aug 17 00:52:38 EDT 2021
sahmoud 1004 Tue Aug 17 00:53:08 EDT 2021
sahmoud 1004 Tue Aug 17 00:53:38 EDT 2021
sahmoud 1004 Tue Aug 17 00:54:08 EDT 2021
sahmoud 1004 Tue Aug 17 00:54:38 EDT 2021
sahmoud 1004 Tue Aug 17 00:55:08 EDT 2021
 1004 Tue Aug 17 00:55:38 EDT 2021
 1004 Tue Aug 17 00:56:08 EDT 2021
```

`root@dopeDroplet:~# ps -aux | grep sahmoud.sh`
```
1004      5789  0.0  0.0   4448   676 pts/1    S    00:51   0:00 /bin/sh ./sahmoud.sh
buba      7792  0.0  0.0  11752   924 pts/1    S+   01:01   0:00 grep sahmoud.sh
```

edneville:
Closes #404

### DIFF
--- a/etc/shadow-maint/userdel-pre.d/01-kill_user_procs.sh
+++ b/etc/shadow-maint/userdel-pre.d/01-kill_user_procs.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+
+# Check user exists, and if so, send sigkill to processes that the user owns
+
+RUNNING=`ps -eo user | grep -Fx "$SUBJECT" | wc -l`
+
+# if the user does not exist, RUNNING will be 0
+
+if [ "${RUNNING}x" = "0x" ]; then
+  exit 0
+fi
+
+ls -1 /proc | while IFS= read -r PROC; do
+  echo "$PROC" | grep -E '^[0-9]+$' >/dev/null
+  if [ $? -ne 0 ]; then
+    continue
+  fi
+  if [ -d "/proc/${PROC}" ]; then
+    USR=`stat -c "%U" /proc/${PROC}`
+    if [ "${USR}" = "${SUBJECT}" ]; then
+      echo "Killing ${SUBJECT} owned ${PROC}"
+      kill -9 "${PROC}"
+    fi
+  fi
+done
+


etc/shadow-maint/userdel-pre.d/01-kill_user_procs.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# Check user exists, and if so, send sigkill to processes that the user owns
+
+RUNNING=`/bin/ps -u "${SUBJECT}" 2>/dev/null | /usr/bin/wc -l`
```
hallyn, line 5: I'm worried about the availability of 'ps -u'.  For instance, 'busybox ps' doesn't seem to support it.

edneville:
On 2021-12-24 08:27-0800, Serge Hallyn wrote:
> I'm worried about the availability of 'ps -u'.  For instance, 'busybox 
> ps' doesn't seem to support it.

Yes that is a good point. busybox has a reduced pgrep/pkill, which would 
otherwise be ideal here. Updated for busybox compatibility.

etc/shadow-maint/userdel-pre.d/01-kill_user_procs.sh
```diff
@@ -0,0 +1,15 @@
...
+
+RUNNING=`ps -eo user | grep -Fx "$SUBJECT" | wc -l`
+
+# if the user does not exist, RUNNING will be 0
+
+if [ "${RUNNING}x" = "0x" ]; then
+  exit 0
+fi
+
+ps -eo pid,user | grep -E "^\s*[0-9]+\s+${SUBJECT}$" | awk '{ print $1 }' | xargs kill -9
```
hallyn, line 15: It's possible that my busybox is just bad, but it doesn't insert a space between the two columns, so only a few of the pids get listed (those which have my username in the cmdline).  Example output:

902725root     [kworker/5:1-mm_]

I wonder whether we're not better off just checking whether killall exists on the system, using killall -u ${SUBJECT} if so, and just leaving things be if it does not exist.

edneville:
On 2021-12-26 08:14-0800, Serge Hallyn wrote:
> +ps -eo pid,user | grep -E "^\s*[0-9]+\s+${SUBJECT}$" | awk '{ print 
> $1 }' | xargs kill -9
> 
> It's possible that my busybox is just bad, but it doesn't insert a 
> space between the two columns, so only a few of the pids get listed 
> (those which have my username in the cmdline).  Example output:
> 
> 902725root     [kworker/5:1-mm_]

I can't reproduce that formatting, how is that represented if you 
busybox ps -eo pid,user | grep kworker | xxd -

> I wonder whether we're not better off just checking whether killall 
> exists on the system, using killall -u ${SUBJECT} if so, and just 
> leaving things be if it does not exist.

killall might cause problems if it runs in Sun OS, and -u isn't present 
on busybox.

hallyn:
> I can't reproduce that formatting, how is that represented if you busybox ps -eo pid,user | grep kworker | xxd -

Well that shows nothing as kworker/ threads are owned by root here, but using postfix as an example:

00000000: 3937 3331 3234 706f 7374 6669 780a 3937  973124postfix.97
00000010: 3331 3235 706f 7374 6669 780a            3125postfix.

>> I wonder whether we're not better off just checking whether killall exists on the system, using killall -u ${SUBJECT} if so, and just leaving things be if it does not exist.
> killall might cause problems if it runs in Sun OS, and -u isn't present on busybox.

Hm, but does Sun OS use this repo?  I used to think it did, now I'm not sure.

Anyway my busybox doesn't have killall at all so I thought that would be ok,
but maybe it's an option I didn't include when building.  So, ok.  That won't
work either.

edneville:
On 2021-12-27 08:04-0800, Serge Hallyn wrote:
> Well that shows nothing as kworker/ threads are owned by root here, 
> but using postfix as an example:
> 
> 00000000: 3937 3331 3234 706f 7374 6669 780a 3937  973124postfix.97
> 00000010: 3331 3235 706f 7374 6669 780a            3125postfix.

I've changed the PR to take a /proc approach, what are your thoughts 
going in that direction?

etc/shadow-maint/userdel-pre.d/01-kill_user_procs.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# if the user does not exist, RUNNING will be 0
+
+if [ "${RUNNING}x" = "0x" ]; then
+  exit 0
+fi
+
+ls -1 /proc | while IFS= read -r PROC; do
+  if [ -d "/proc/${PROC}" ]; then
+    USR=`ls -ld "/proc/${PROC}" | awk '{print $3}'`
```
hallyn, line 17: What do you think about using 'stat -c "%U" /proc/${PROC}' instead of ls -ld | awk?  busybox (at least here) seems to do %U correctly.

edneville:
On 2021-12-27 17:02-0800, Serge Hallyn wrote:
> What do you think about using 'stat -c "%U" /proc/${PROC}' instead of 
> ls -ld | awk?  busybox (at least here) seems to do %U correctly.

That works much better I think, updated, thanks.

hallyn:
I'm ok to merge this as is, but want to point out that the [ -d "/proc/${PROC}" ] will also match for directories like /proc/acpi.  Since those will have user root it's probably fine, but do you think it's worth checking for all digits?  If you don't think it's worth it, then i'll just merge.

edneville:
On 2021-12-29 10:10-0800, Serge Hallyn wrote:
> I'm ok to merge this as is, but want to point out that the [ -d 
> "/proc/${PROC}" ] will also match for directories like /proc/acpi.  
> Since those will have user root it's probably fine, but do you think 
> it's worth checking for all digits?  If you don't think it's worth it, 
> then i'll just merge.

Yeah, that's a good point. I've updated the commit to match.

Ed

PR #461, opened 2021-12-17
passwd -l should not say that it changes expiry

### LINKED ISSUE #441: passwd --lock <user> emits misleading message
When using the option `--lock`, `passwd` will print the following status
message:
```
"passwd: password expiry information changed."
```

This is not consistent with the man page:

```
       -l, --lock
           Lock the password of the named account. This option disables a password by changing it to a value
           which matches no possible encrypted value (it adds a ´!´ at the beginning of the password).

           Note that this does not disable the account. The user may still be able to login using another
           authentication token (e.g. an SSH key). To disable the account, administrators should use usermod
           --expiredate 1 (this set the account's expire date to Jan 2, 1970).

           Users with a locked password are not allowed to change their password.


```
Originally reported at https://bugs.debian.org/951480 .

edneville:
Closes #441

### DIFF
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -1109,7 +1109,7 @@ int main (int argc, char **argv)
 			(void) printf (_("%s: password changed.\n"), Prog);
 #endif				/* USE_PAM */
 		} else {
-			(void) printf (_("%s: password expiry information changed.\n"), Prog);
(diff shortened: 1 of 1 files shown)


hallyn:
Thank you.

PR #463, opened 2021-12-20
libsubid: fix defining SONAME version

thesamesam:
We were overriding this when --enable-shared was passed. We can actually
just dump the conditional logic as libtool will do the right thing for
us here anyway.

Without this patch, libsubid is installed as .0.

Signed-off-by: Sam James <sam@gentoo.org>


floppym:
You should drop this line from configure.ac as well, since we no longer need the automake conditional.

https://github.com/shadow-maint/shadow/blob/master/configure.ac#L324

floppym (approved):
Looks good to me!

hallyn:
Thank you both.

Issue #444, opened 2021-11-14
Prog and shadow_logfd should not be extern variables


rbalint:
They should be provided via function calls to the logging API . This would solve the messy situation with linking.

hallyn:
Yeah I've been meaning to try something along those lines.  Thanks.

PR #466, opened 2021-12-21
libsubid/api.c: make Prog not static

### LINKED ISSUE #465: libsubid-4 lost some symbols
so in https://github.com/shadow-maint/shadow/commit/32f641b207f6ddff1460a22fdc654def4a9c473d ABI was changed and bumped.

but it seems it broke some consumers.

consider this code (just a build time test for other app):

```c
#include <shadow/subid.h>                                                                                                                                                                                                               
#include <stdlib.h>                                                                                                                                                                                                                     
                                                                                                                                                                                                     
int main() {                                                                                                                                                                                                                            
        struct subid_range *ranges = NULL;                                                                                                                                                                                              
#if SUBID_ABI_MAJOR >= 4                                                                                                                                                                                                                
        subid_get_uid_ranges("root", &ranges);                                                                                                                                                                                          
#else                                                                                                                                                                                                                                   
        get_subuid_ranges("root", &ranges);                                                                                                                                                                                             
#endif                                                                                                                                                                                                                                  
        free(ranges);                                                                                                                                                                                                                   
        return 0;                                                                                                                                                                                                                       
}
```

it's supposed to build with both libsubid-3 and libsubid-4

but unless those 2 are defined:
```c
const char *Prog;                                                                                                                                                                                                                       
FILE *shadow_logfd = NULL;     
```

linking fails with:
```shell-session
 gcc -lsubid t.c 
powerpc64le-unknown-linux-gnu/bin/ld: /usr/lib/gcc/powerpc64le-unknown-linux-gnu/11.2.0/../../../../lib64/libsubid.so: undefined reference to `Prog'
powerpc64le-unknown-linux-gnu/bin/ld: /usr/lib/gcc/powerpc64le-unknown-linux-gnu/11.2.0/../../../../lib64/libsubid.so: undefined reference to `shadow_logfd'
collect2: error: ld returned 1 exit status
```

my guess it used to work before because all symbols were leaking, but now consumers are required to define `Prog` and `shadow_logfd` and I'm not sure it's correct/expected.


Thoughts?


some consumer issue examples:
https://github.com/containers/podman/pull/12655
https://github.com/containers/storage/pull/1089

hallyn:
Closes #465

It'll be used by libmisc/ etc.

### DIFF
--- a/libsubid/api.c
+++ b/libsubid/api.c
@@ -39,8 +39,8 @@
 #include "idmapping.h"
 #include "subid.h"
 
-static const char *Prog = "(libsubid)";
-static FILE *shadow_logfd;
+const char *Prog = "(libsubid)";
+FILE *shadow_logfd;
 
 bool subid_init(const char *progname, FILE * logfd)
 {


hallyn:
With this:
```
ubuntu@shadow:~/shadow$ cat test.c
#include <shadow/subid.h>
#include <stdlib.h>

int main() {
        struct subid_range *ranges = NULL;
        subid_init("test", NULL);
        int i, n;
#if SUBID_ABI_MAJOR >= 4
        n = subid_get_uid_ranges("root", &ranges);
#else
        n = get_subuid_ranges("root", &ranges);
#endif
        for (i = 0; i < n; i++)
                printf("%d: %ld - %ld\n", i, ranges[i].start, ranges[i].count);

        free(ranges);
        return 0;
}
ubuntu@shadow:~/shadow$ gcc -o test test.c -lsubid
ubuntu@shadow:~/shadow$ ./test
0: 200000 - 65536
```
(I really need to get the CI tests working again)

autogen.sh
```diff
@@ -4,9 +4,9 @@ autoreconf -v -f --install || exit 1
 
 ./configure \
 	CFLAGS="-O2 -Wall" \
-	--enable-man \
+	--disable-man \
```
ikerexxe, line 7: Is this change intentional?

autogen.sh
```diff
@@ -4,9 +4,9 @@ autoreconf -v -f --install || exit 1
...
 	--enable-maintainer-mode \
 	--enable-shared \
 	--without-libpam \
-	--with-selinux \
+	--without-selinux \
```
ikerexxe, line 11: Is this change intentional?

ikerexxe (commented):
The changes in `api.c` look good to me, but I'm worried about the configuration changes.

autogen.sh
hallyn: No.  Thanks for pointing that out.

ikerexxe:
LGTM. Thanks!

ikerexxe (approved):

rbalint:
@hallyn please finish the fix #444 instead. This is still ugly. :-(

hallyn:
> @hallyn please finish the fix for #444 instead. This is still ugly and the proper fix will need an ABI bump again. :-(

I thought that was already merged!  ok let me finish that first and see where that gets us.

Then I may do another release, and hopefully set this aside for awhile so I can focus a on the rust impl.

PR #467, opened 2021-12-22
Have a single definition of date_to_str()

alejandro-colomar:
PARAMETERS:

According to the C2x charter, I reordered the parameters 'size'
and 'buf' from previously existing date_to_str() definitions.

[C2x charter](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2086.htm):
> 15. Application Programming Interfaces (APIs) should be
> self-documenting when possible.  In particular, the order of
> parameters in function declarations should be arranged such that
> the size of an array appears before the array.  The purpose is to
> allow Variable-Length Array (VLA) notation to be used.  This not
> only makes the code's purpose clearer to human readers, but also
> makes static analysis easier.  Any new APIs added to the Standard
> should take this into consideration.

I used 'long' for the date parameter, as some uses of the function
need to pass a negative value meaning "never".

FUNCTION BODY:

I didn't check `#ifdef HAVE_STRFTIME`, which old definitions did,
since [**strftime(3)**](https://www.man7.org/linux/man-pages/man3/strftime.3.html) is guaranteed by the C89 standard, and all of
the conversion specifiers that we use are also specified by that
standard, so we don't need any extensions at all.

I used [**memccpy(3)**](https://www.man7.org/linux/man-pages/man3/memccpy.3.html) instead of [**strncpy(3)**](https://www.man7.org/linux/man-pages/man3/strncpy.3.html), because strncpy(3)
unnecessarily zeroes the remaining of the buffer.  memccpy(3) is
in POSIX.1-2001, and has been added to [C2x](http://www.open-std.org/JTC1/SC22/WG14/www/docs/n2731.pdf) [too](http://www.open-std.org/JTC1/SC22/WG14/www/docs/n2349.htm).  **[strlcpy(3bsd)](https://manpages.debian.org/bullseye/libbsd-dev/strlcpy.3.en.html)**
would be better (it guarantees NUL termination, and doesn't have
a parameter for `'\0'`), but it may not be always available.

### DIFF
--- a/lib/prototypes.h
+++ b/lib/prototypes.h
@@ -134,6 +134,9 @@ extern int copy_tree (const char *src_root, const char *dst_root,
                       uid_t old_uid, uid_t new_uid,
                       gid_t old_gid, gid_t new_gid);
 
+/* date_to_str.c */
+extern void date_to_str (size_t size, char buf[size], long date);
+
 /* encrypt.c */
 extern /*@exposed@*//*@null@*/char *pw_encrypt (const char *, const char *);
 
--- a/libmisc/Makefile.am
+++ b/libmisc/Makefile.am
@@ -19,6 +19,7 @@ libmisc_la_SOURCES = \
 	cleanup_user.c \
 	console.c \
 	copydir.c \
+	date_to_str.c \
 	entry.c \
 	env.c \
 	failure.c \
--- a/libmisc/date_to_str.c
+++ b/libmisc/date_to_str.c
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2021, Alejandro Colomar <alx.manpages@gmail.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. The name of the copyright holders or contributors may not be used to
+ *    endorse or promote products derived from this software without
+ *    specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+ * PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT
+ * HOLDERS OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <string.h>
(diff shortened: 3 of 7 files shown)


alejandro-colomar:
I could test `/tmp/shadow/usr/sbin/usermod` after this change, but neither `/tmp/shadow/usr/bin/passwd` nor `/tmp/shadow/usr/bin/chage` did work (probably due to some configuration that I ignored).

alejandro-colomar:
Not yet tested (I only tested manually `usermod`).  I'm following `<tests/README>` to run the tests.

alejandro-colomar:
I have now run the tests in `<tests/>`, but there seem to be no tests for usermod or for passwd (I see a subdir `<chage/>` but no dirs for the other programs).  Also, many unrelated tests fail, so I don't know if this works.

However, I didn't notice anything specially related to this change in the test results, so I guess it works.  What one could be concerned is if it works on every system that this project supports.

alejandro-colomar:
BTW, there are preprocessor directives that I copied from other files, but I'm not sure if I should keep them or not.  Why `#define _XOPEN_SOURCE 500?  I don't think I need it; may I remove it?  Do I also need `<defines.h>` and `<config.h>`?

hallyn:
> I used memccpy(3) instead of strncpy(3), because strncpy(3) unnecessarily zeroes the remaining of the buffer

However what it appears to miss is the guarantee that \0 will be copied.

I realize that cannot happen here, but I worry about setting an example for future users.

I think I would prefer that you stick to strncpy here, and that we do a separate PR that switches from strnpcy to strlcpy (with a configure.ac-triggered custom definition for systems that need it, the way github.com/lxc/lxc does it.  What do you think?

(I also realize that asking you to rebase just to undo that seems silly, it's just that future greps for strncpy would then miss this case.  But I suppose that's no big deal)

hallyn:
> BTW, there are preprocessor directives that I copied from other files, but I'm not sure if I should keep them or not. Why #define _XOPEN_SOURCE 500? I don't think I need it; may I remove it? Do I also need <defines.h> and <config.h>?

_XOPEN_SOURCE 500 means to use the 1995 posix standard available fns :)  #define GNU_SOURCE makes that unnecessary.  AC_GNU_SOURCE in configure.ac should auto-set that when using glibc.  I would assume musl will also make then automatically available.

I'm convinced it's ok to drop the _XOPEN_SOURCE define.

Right now it doesn't look like you're using anything from lib/defines.h or config.h.  Fine to drop those too.  If we were to switch from memccpy to strlcpy optionally defined when needed, then we might need config.h again.

alejandro-colomar:
> > I used memccpy(3) instead of strncpy(3), because strncpy(3) unnecessarily zeroes the remaining of the buffer
> 
> However what it appears to miss is the guarantee that \0 will be copied.

Yes, both `strncpy(3)` and `memccpy(3)` miss that (they are not designed to use NUL-terminated strings).  That's why I prefer `strlcpy(3bsd)`.

> 
> I realize that cannot happen here, but I worry about setting an example for future users.

Agree.  Both `strncpy(3)` and memccpy(3)` set a bad example.  Let's use `strlcpy(3bsd)`.  If we can't depend on libbsd we could easily reimplement it in terms of `memccpy(3)`.

> 
> I think I would prefer that you stick to strncpy here, and that we do a separate PR that switches from strnpcy to strlcpy (with a configure.ac-triggered custom definition for systems that need it, the way github.com/lxc/lxc does it. What do you think?

Agree too.

> 
> (I also realize that asking you to rebase just to undo that seems silly, it's just that future greps for strncpy would then miss this case. But I suppose that's no big deal)

Don't worry, I was already concerned about that sudden/unexpected change in this PR.  Your concern confirms it.  I'll use `strncpy(3)` for this PR, and then open a new one, discussing how to add `strlcpy(3bsd)`.

alejandro-colomar:
> > BTW, there are preprocessor directives that I copied from other files, but I'm not sure if I should keep them or not. Why #define _XOPEN_SOURCE 500? I don't think I need it; may I remove it? Do I also need <defines.h> and <config.h>?
> 
> _XOPEN_SOURCE 500 means to use the 1995 posix standard available fns :)

Yup, I know :)
What I meant is that we don't really use any POSIX-specific behavior, do we?

We use: `gmtime(3)`, `strftime(3)`, and `memccpy(3)`.
None of them require setting any feature test macro, according to the man pages.

> #define GNU_SOURCE makes that unnecessary. AC_GNU_SOURCE in configure.ac should auto-set that when using glibc. I would assume musl will also make then automatically available.
> 
> I'm convinced it's ok to drop the _XOPEN_SOURCE define.
> 
> Right now it doesn't look like you're using anything from lib/defines.h or config.h. Fine to drop those too. If we were to switch from memccpy to strlcpy optionally defined when needed, then we might need config.h again.

Okay, I'll remove all of them.  Thanks!

alejandro-colomar:
> and that we do a separate PR that switches from strnpcy to strlcpy (with a configure.ac-triggered custom definition for systems that need it, the way github.com/lxc/lxc does it. What do you think?

However, I tend to hate ifdefs and variable dependencies.  I would prefer either always requiring a dependency that provides libbsd (libbsd, or musl, or a BSD), or just redefining it ourselves unconditionally (maybe with a custom name so that it doesn't collide (`strlcpy_`)?).

PR #472, opened 2021-12-25
lib: rename Prog to shadow_progname, with only one definition

atsampson:
After #471, the build fails with duplicate symbol errors with `-fno-common`:

```
/gar/bin/ld: ../lib/.libs/libshadow.a(libshadow_la-encrypt.o):/src/sys/shadow/work/6761cf2d7e9defeed7f915c668a87b84bac2c9c3/lib/shadowlog_internal.h:2: multiple definition of `shadow_logfd'; ../lib/.libs/libshadow.a(libshadow_la-commonio.o):/src/sys/shadow/work/6761cf2d7e9defeed7f915c668a87b84bac2c9c3/lib/shadowlog_internal.h:2: first defined here
/gar/bin/ld: ../lib/.libs/libshadow.a(libshadow_la-encrypt.o):/src/sys/shadow/work/6761cf2d7e9defeed7f915c668a87b84bac2c9c3/lib/shadowlog_internal.h:1: multiple definition of `Prog'; ../lib/.libs/libshadow.a(libshadow_la-commonio.o):/src/sys/shadow/work/6761cf2d7e9defeed7f915c668a87b84bac2c9c3/lib/shadowlog_internal.h:1: first defined here
```
(etc.)

`-fcommon` is the default in GCC 10 and later, and explicitly enabled in some distributions to catch problems like this. There were two causes:

- `Prog` and `shadow_logfd` were defined in a header file that was included in multiple other files. Fix this by defining them once in `shadowlog.c`, and having `extern` declarations in the header.

- Most of the tools (except `id`/`nologin`) also define a `Prog` variable, which is not intended to alias the one in the library. Fix this by renaming `Prog` in the library to `shadow_progname`, which also matches the new accessor functions for it.

I think this should still be OK in terms of what #465 wanted, i.e. the shared library doesn't require users to define these symbols.

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -147,7 +147,7 @@ static int do_lock_file (const char *file, const char *lock, bool log)
 		if (log) {
 			(void) fprintf (shadow_logfd,
 			                "%s: %s: %s\n",
-			                Prog, file, strerror (errno));
+			                shadow_progname, file, strerror (errno));
 		}
 		return 0;
 	}
@@ -159,7 +159,7 @@ static int do_lock_file (const char *file, const char *lock, bool log)
 		if (log) {
 			(void) fprintf (shadow_logfd,
 			                "%s: %s file write error: %s\n",
-			                Prog, file, strerror (errno));
+			                shadow_progname, file, strerror (errno));
 		}
 		(void) close (fd);
 		unlink (file);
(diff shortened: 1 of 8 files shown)


hallyn:
Thanks.

hallyn:
I think libsubid/api.c:subid_init() also needs to call log_set_progname().  Do you want to do a trivial PR for that?

PR #474, opened 2021-12-26
Default value for progname/logfd, and set progname in subid_init too

atsampson:
Set the program name in `subid_init` too, as [suggested](https://github.com/shadow-maint/shadow/pull/472#issuecomment-1001205406) in #472.

The existing code here made me think that there should be a default value in lib if the user doesn't set it, so do that as well.

### DIFF
--- a/lib/shadowlog.c
+++ b/lib/shadowlog.c
@@ -2,8 +2,8 @@
 
 #include "lib/shadowlog_internal.h"
 
-const char *shadow_progname;
-FILE *shadow_logfd;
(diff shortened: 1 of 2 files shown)


hallyn:
Thanks.  IIIUC the = NULL in lib/shadowlog.c shouldn't be needed as that's in BSS, but that's ok.

Do you mind rebasing this against master and force-pushing?  That way the new CI build should run.

atsampson:
Done!

hallyn:
Awesome, thanks.

PR #475, opened 2021-12-26
date_to_str.c: replace strncpy(3) by strlcpy(3)

alejandro-colomar:
I didn't change all occurences of strncpy(3), because that would be dangerous.  I prefer discussing each one separately.  Especially, there are a few cases where the input string in not trustworthy, and in those cases memccpy(3) would be better (or strscpy(9) even better).  I don't know which are those cases.  If you could point those out to me, I could prepare 2 separate PRs.

I didn't yet add code to require a dependency that pulls `strlcpy(3)`.  I would like to discuss how we should do it.  Follow lxc?  Require libbsd or other library that provides this function?  Define it ourselves with a different name and avoid collision?

### DIFF
--- a/lib/prototypes.h
+++ b/lib/prototypes.h
@@ -134,6 +134,9 @@ extern int copy_tree (const char *src_root, const char *dst_root,
                       uid_t old_uid, uid_t new_uid,
                       gid_t old_gid, gid_t new_gid);
 
+/* date_to_str.c */
+extern void date_to_str (size_t size, char buf[size], long date);
+
 /* encrypt.c */
 extern /*@exposed@*//*@null@*/char *pw_encrypt (const char *, const char *);
 
--- a/libmisc/Makefile.am
+++ b/libmisc/Makefile.am
@@ -19,6 +19,7 @@ libmisc_la_SOURCES = \
 	cleanup_user.c \
 	console.c \
 	copydir.c \
+	date_to_str.c \
 	entry.c \
 	env.c \
 	failure.c \
--- a/libmisc/date_to_str.c
+++ b/libmisc/date_to_str.c
@@ -38,9 +38,10 @@ void date_to_str (size_t size, char buf[size], long date)
 	time_t t;
 
 	t = date;
-	if (date < 0)
-		(void) strncpy (buf, "never", size);
-	else
+	if (date < 0) {
+		(void) strlcpy (buf, "never", size);
+	} else {
 		(void) strftime (buf, size, "%Y-%m-%d", gmtime (&t));
-	buf[size - 1] = '\0';
+		buf[size - 1] = '\0';
+	}
 }
--- a/libmisc/date_to_str.c
+++ b/libmisc/date_to_str.c
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2021, Alejandro Colomar <alx.manpages@gmail.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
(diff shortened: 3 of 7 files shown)


alejandro-colomar:
This PR is supposed to be applied after https://github.com/shadow-maint/shadow/pull/467

hallyn:
Thanks, but I guess let's wait on this.

alejandro-colomar:
Okay.  Now that you merged the other one, I'll rebase.  I'll wait for you to comment on how you prefer to define `strlcpy(3)` :)

alejandro-colomar:
https://alioth-lists.debian.net/pipermail/pkg-shadow-devel/2022-February/013142.html

hallyn:
I don't consider the 'extra work' strncpy does of filling in with 0's a bad thing necessarily.

In general I always want to see:

    ret = strncpy(dest, src, n);
    if (ret < 0 || ret >= n)
       return error;

It seems to me it would be simpler to replace strncpy with shadow_strncpy() which does:

   int shadow_strncpy(char *dest, char *src, size_t n) {
      int ret = strncpy(dest, src, n);
      if (ret >= n)
          ret = -1;
      return ret;
   }

And then callers can have a simpler rv check.

Now, having typed all that, I'm ok with your strpecpy proposal - except that it's another new strXYZ name to remember :)

But yeah I encourage you to post that patch.

Thanks.

hallyn:
https://lwn.net/Articles/905777/   lwn on getting rid of strlcpy in the kernel.

alejandro-colomar:
Hi Serge!

Sorry the long time without notice; I still plan to do this, and in fact, will probably start soon.  I just had many more things to do before.

So, the reason to stop using strlcpy(3bsd) in the kernel and replace it by strscpy(9) is not justified, IMO.

The reasons Linus states are that users misunderstand and misuse the BSD function.  The official prototypes for both are:

```c
     size_t strlcpy(char *dst, const char *src, size_t size);

     ssize_t strscpy(char *dst, const char *src, size_t size);
```

I've been recently having a discussion in the Linux man-pages mailing list about improving documentation, and referred to this precise issue: <https://lore.kernel.org/linux-man/d9865e28-fc81-9878-769b-471652a8be94@gmail.com/T/#m53ec50f1ec427aeb1d9bae90a5a52ac94dbfbacb>.

By using VLA syntax (not yet supported in this way by compilers or the standard, but it's under discussion), we could document the functions as:

```c
     size_t strlcpy(char dst[restrict .size], const char *restrict src, size_t size);

     ssize_t strscpy(char dst[restrict .size], const char src[restrict .size], size_t size);
```

which clearly shows the difference between both.  They are functions designed for completely different purposes.  The BSD function is for copying trusted (guaranteed to be NUL-terminated) strings into a buffer.  The Linux kernel function is for copying untrusted (maybe not NUL-terminated) strings into a buffer.

We should differentiate both cases within this project, and use the appropriate one in each case.  I'll need your help to determine in which cases we can trust the string and in which not.  When in doubt, I'll try to keep the current behavior, being bug-compatible.

alejandro-colomar:
I opened a different PR, since it does a slightly different thing from this one:

<https://github.com/shadow-maint/shadow/pull/569>

alejandro-colomar:
> > which clearly shows the difference between both. They are functions designed for completely different purposes. The BSD function is for copying trusted (guaranteed to be NUL-terminated) strings into a buffer. The Linux kernel function is for copying untrusted (maybe not NUL-terminated) strings into a buffer.
> 
> That's a great explanation! Ideally the function names themselves would convey this :) Failing that (we don't really wanna rename them) just documenting them as such in our definitions or include is a great start.

Thanks!

Yes, I'll copy this information into the documentation I wrote into the documentation.

BTW, @cgzones here's the reason to keep strlcpy().  But I'll try to add some more details in the form of a simple example of code, in case it also helps.  I'll do that in the currently-open PR.

PR #479, opened 2021-12-27
Remove HAVE_GETGROUPS ifdefs

alejandro-colomar:
getgroups(2) has been in POSIX since POSIX.1-2001.  It is also in
in SVr4 and in 4.3BSD (see [`getgroups(2)`](https://man7.org/linux/man-pages/man2/getgroups.2.html) and [`getgroups(3p)`](https://www.man7.org/linux/man-pages/man3/getgroups.3p.html)).
We can assume that this function is always available.


alejandro-colomar:
Please, check that I correctly removed getgroups(3) checking in the configure scripts, since I'm not used to that.

PR #480, opened 2021-12-27
Remove HAVE_MEMCPY and HAVE_MEMSET ifdefs

alejandro-colomar:
memcpy(3) has been in standard C since C89.  It is also in
POSIX.1-2001, in SVr4, and in 4.3BSD (see [`memcpy(3)`](https://man7.org/linux/man-pages/man3/memcpy.3.html) and [`memcpy(3p)`](https://man7.org/linux/man-pages/man3/memcpy.3p.html)).
We can assume that this function is always available.


alejandro-colomar:
Do the same with memset(3), for the same reasons.

PR #481, opened 2021-12-28
Assume C89 is available

alejandro-colomar:
Remove STDC_HEADERS, since C89 is nowadays available everywhere (or at least everywhere we can possibly care).

STDC_HEADERS removal allows to simplify other things.  Propagate those changes.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -38,7 +38,6 @@ dnl Checks for libraries.
 
 dnl Checks for header files.
 AC_HEADER_DIRENT
-AC_HEADER_STDC
 AC_HEADER_SYS_WAIT
 AC_HEADER_STDBOOL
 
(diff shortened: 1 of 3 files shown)


alejandro-colomar:
Oops, I started from another branch where I had the strlcpy(3) changes.  I've force-pushed to remove those.

alejandro-colomar:
I wonder if there's any system that defined a locale-dependent isdigit(3).  C89 allows for that, but at least glibc unconditionally defines it to basically the same that we do.

hallyn:
Thank you.  Sorry i forgot to merge this before tagging 4.11.

alejandro-colomar:
No problem, I have more changes pending :)

Thanks!
Alex

PR #482, opened 2021-12-29
fix segfault running useradd -D

xry111:
There are two problems:

1. `dirname` is not declared properly, so the compiler assumes it to return `int`.
2. `dirname` does not support called with a static string (see `man 3 dirname`).

Both may cause a segfault, as (1) leads to an undefined behavior and the man page explicitly says (2) can cause a segfault.

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -41,6 +41,7 @@
 #include <getopt.h>
 #include <grp.h>
 #include <lastlog.h>
+#include <libgen.h>
 #include <pwd.h>
 #ifdef ACCT_TOOLS_SETUID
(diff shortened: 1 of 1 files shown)


xry111:
@hallyn : I and @Berzerkula have really observed the segfault on our system (`useradd -D --gid 999` can trigger it).  So I think this should be reviewed ASAP.

hallyn:
Thanks, looks good.  I had actually noticed a weird build error message which may have been warning about this but hadn't had time to dig deeper.

I'd have preferred the free() not happen in two places, but moving it up means messing with errno so just not worth it.

Thanks again.

Issue #317, opened 2021-03-23
Has CVE-2013-4235 been fixed ？


cheng362432:
reference：https://bugzilla.redhat.com/show_bug.cgi?id=884658

hallyn:
I don't see any changes since 2010 to the src/remove_tree.c, so I don't think so.  However, I'm not sure the CVE is legitimate.  remove_tree() uses lstat() to check for a directory, and only descends non-link directories.  Hard links are not allowed for directories.

Knowing Florian, however, I expect I'm misunderstanding something.  Would be good to get elaboration.

Blub:
Assuming `remove_tree()` is a problem here (because it may be called on the user's home directory from `userdel` while the user is still logged in (or at least has evil programs running)), these calls happen by-name, so the time between the `lstat()` and the recursion would give the user a chance to turn the name into a symbolic link right before the `opendir()` happens. So if this is the a problematic scenario then this should be switched to `open->fstat->fdopendir` (which I think is also more efficient anyway, so not wrong either way).

hallyn:
Oh, I was thinking just the link target would be switched.  Yeah, that seems worth doing.  Thanks.

If one of you has time to float a patch, please leave a note before you start (to avoid duplicate work).

cheng362432:
Are there any plans to fix this CVE？

hallyn:
> Are there any plans to fix this CVE？

Please feel free to open a PR with a patch.

edneville:
Since the underlying issue is a user logged in whilst userdel runs, would it make sense to add a script in `/etc/shadow-maint/userdel-pre.d/` to logout the user for deletion?

sweetylz:
Does shadow install the script, or do we need to install it ourselves?

hallyn:
> Does shadow install the script, or do we need to install it ourselves?

It is not being installed by default.

jhawkins1:
Still fuzzy to me on the current state of this CVE...  Various Linux Distros still say it is open and no fix upstream...  Based on the discussion in this thread, was this fixed as of 4.11.1 ?

PR #483, opened 2022-01-01
Adding nofollow to opens

edneville:
References #317

### DIFF
--- a/libmisc/copydir.c
+++ b/libmisc/copydir.c
@@ -741,7 +741,7 @@ static int copy_file (const char *src, const char *dst,
 	char buf[1024];
 	ssize_t cnt;
 
-	ifd = open (src, O_RDONLY);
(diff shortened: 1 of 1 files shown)


libmisc/copydir.c
```diff
@@ -751,7 +751,7 @@ static int copy_file (const char *src, const char *dst,
 		return -1;
 	}
 #endif				/* WITH_SELINUX */
-	ofd = open (dst, O_WRONLY | O_CREAT | O_TRUNC, statp->st_mode & 07777);
+	ofd = open (dst, O_WRONLY | O_CREAT | O_TRUNC | O_NOFOLLOW, statp->st_mode & 07777);
```
hallyn, line 754: I was going to suggest using O_EXCL here instead, but the manpage says NFS before v3 does not support it, so maybe this is best.

hallyn:
Thanks!

PR #492, opened 2022-01-04
useradd: modify check ID range for system users

ikerexxe:
useradd warns that a system user ID less than SYS_UID_MIN is outside the
expected range, even though that ID has been specifically selected with
the "-u" option.

In my opinion all the user ID's below SYS_UID_MAX are for the system,
thus I change the condition to take that into account. Moreover, I also
modify the check for SYS_UID_MAX, so that it's always less than UID_MIN.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2004911

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2409,11 +2409,9 @@ static void check_uid_range(int rflg, uid_t user_id)
 	uid_t uid_min ;
 	uid_t uid_max ;
 	if (rflg) {
-		uid_min = (uid_t)getdef_ulong("SYS_UID_MIN",101UL);
 		uid_max = (uid_t)getdef_ulong("SYS_UID_MAX",getdef_ulong("UID_MIN",1000UL)-1);
-		if (uid_min <= uid_max) {
-			if (user_id < uid_min || user_id >uid_max)
-				fprintf(stderr, _("%s warning: %s's uid %d outside of the SYS_UID_MIN %d and SYS_UID_MAX %d range.\n"), Prog, user_name, user_id, uid_min, uid_max);
+		if (user_id > uid_max) {
+			fprintf(stderr, _("%s warning: %s's uid %d is greater than SYS_UID_MAX %d\n"), Prog, user_name, user_id, uid_max);
 		}
 	}else{
 		uid_min = (uid_t)getdef_ulong("UID_MIN", 1000UL);


ikerexxe:
> Wouldn't https://bugzilla.redhat.com/show_bug.cgi?id=2004911 also be fixed by setting SYS_UID_MIN to the default 100?

I don't think so, as there would still be a problem for ID's below the 100 value. As an example [comment 10 in the aforementioned bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2004911#c10) reports the same problem for the apache user, which has the ID 48. Another examples in my system are ftp (14) and dbus (81). I guess something similar happens in your system.

Let me back up a bit to define the ID ranges for each type of user:
* 0: reserved for root.
* UID_MIN and UID_MAX: dynamically assigned to regular users. This is the range between 1000 and 60000 for both Debian and Fedora derivatives.

I think we both agree until this point, and the next point may be where we disagree:
* 1 and 999: system user accounts. The value 999 can be easily tracked to SYS_UID_MAX in both Fedora and Debian. But what about 1? There isn't any definition in `/etc/login.defs` that tracks it. Moreover, what's the meaning of SYS_UID_MIN in that case? In my opinion, the definition of SYS_UID_MIN should be changed a little bit to indicate that this is the range of dynamically allocated system users, and the range 1 to SYS_UID_MIN is reserved for system users that need to have their IDs static.

So, in reality:
* 1 and SYS_UID_MIN (100 or 200 depending on the distribution): static system user accounts.
* SYS_UID_MIN and SYS_UID_MAX: dynamic system user accounts.

[This old email from 2009 in Fedora's mailing list](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/L7FHINNIJH4GK3DHHJOL23TQ2W32RLFQ/) and [systemd documentation](https://systemd.io/UIDS-GIDS/) seem to agree with this understanding.

What do you think? By the way, I'm not in a hurry to merge this PR as I'd like to clarify the topic in hand first.

> I must be misunderstanding your patch - you seem to be expecting SYS_UID_MAX to be > (non-sys-user-)UID_MIN. But the intent by default is that the range for system users is lower than regular users - system users being 100-999 and regular users starting at 1000...
> 
> ```
> -               uid_min = (uid_t)getdef_ulong("SYS_UID_MIN",101UL);
> +               uid_min = (uid_t)getdef_ulong("UID_MIN", 1000UL);
>                 uid_max = (uid_t)getdef_ulong("SYS_UID_MAX",getdef_ulong("UID_MIN",1000UL)-1);
> ```
> 
> So you are setting uid_min to the regular user uid_min, and uid_max to the system user uid_max, so in most normal systems uid_min would end up > uid_max.

This is just an extra check and I can remove it as the problem reported in the bugzilla isn't related with it. I just wanted to make sure that UID_MIN is greater than SYS_UID_MAX before giving any warning, just in case somebody messed with those values and they don't make any sense.

ikerexxe:
@blueskycs2c as you are the original author of the check_uid_range() function I'd also like to know your opinion.

hallyn:
Sorry, I am to look at this in detail tonight.

hallyn:
There are two separate issues to debate here, I think.

The first is that your patch does not, in my opinion, do quite what you want it to do.

The second is the question about quite the range of (1..SYS_UID_MIN) is.  You called it static system user accounts, I think i like the name 'reserved system user accounts' per https://github.com/ComplianceAsCode/content/pull/1285 .  So if apache user wants to be 48, then - following that definition - either (1) SYS_UID_MIN should be lowered to something lower than 48, or (2) apache user should be pre-defined (reserved).

So keeping in mind that SYS_UID_MIN and SYS_UID_MAX are meant as bounds to constrain automatic userid selection, I guess the question is just - do you want a warning or not.

So how about a login.defs "WARN_ON_UID_RANGE" or something, which fedora sets to false?  Like you said, if you specified a uid with -u then chances are you know what you want.  So I'd even be ok with setting WARN_ON_UID_RANGE to false by default even though that changes default behavior.  I categorically do not believe there is a script out there depending on that warning message :)

ikerexxe:
> The second is the question about quite the range of (1..SYS_UID_MIN) is. You called it static system user accounts, I think i like the name 'reserved system user accounts' per [ComplianceAsCode/content#1285](https://github.com/ComplianceAsCode/content/pull/1285) . So if apache user wants to be 48, then - following that definition - either (1) SYS_UID_MIN should be lowered to something lower than 48, or (2) apache user should be pre-defined (reserved).

'reserved system user accounts' sounds good to me. As far as I know, apache has reserved the ID 48, so option 2. Is there a way of reserving a given ID for a user apart from using the `-u` option?

> So keeping in mind that SYS_UID_MIN and SYS_UID_MAX are meant as bounds to constrain automatic userid selection, I guess the question is just - do you want a warning or not.

No, I don't think a warning is needed when at user creation the ID has been manually selected.

> So how about a login.defs "WARN_ON_UID_RANGE" or something, which fedora sets to false? Like you said, if you specified a uid with -u then chances are you know what you want. So I'd even be ok with setting WARN_ON_UID_RANGE to false by default even though that changes default behavior. I categorically do not believe there is a script out there depending on that warning message :)

But do we need to add a new option? I mean, in my opinion, if a system user is created and the ID is manually selected, then as long as it is below the SYS_UID_MAX threshold there shouldn't be any warning.

I think I'm not understanding something. Can you tell me why this warning is needed under the previous conditions?

hallyn:
> But do we need to add a new option? I mean, in my opinion, if a system user is created and the ID is manually selected, then as long as it is below the SYS_UID_MAX threshold there shouldn't be any warning.
>
> I think I'm not understanding something. Can you tell me why this warning is needed under the previous conditions?

Because there are a large number of admins out there, and I would fully expect some of them to happily type 'useradd -u 48 specialuser', then complain when later attempts to install apache fail because 48 is taken.

But then again I don't know anyone who does useradd -r -u by hand, scripts always do it, and scripts don't care about warnings.

I still do not want to take this patch: the uid_min part is misleading.  Can you update your patch to do the below?

diff --git a/src/useradd.c b/src/useradd.c
index 34376fa5..e97d867f 100644
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2409,12 +2409,9 @@ static void check_uid_range(int rflg, uid_t user_id)
    uid_t uid_min ;
    uid_t uid_max ;
    if (rflg) {
-       uid_min = (uid_t)getdef_ulong("SYS_UID_MIN",101UL);
        uid_max = (uid_t)getdef_ulong("SYS_UID_MAX",getdef_ulong("UID_MIN",1000UL)-1);
-       if (uid_min <= uid_max) {
-           if (user_id < uid_min || user_id >uid_max)
-               fprintf(stderr, _("%s warning: %s's uid %d outside of the SYS_UID_MIN %d and SYS_UID_MAX %d range.\n"), Prog, user_name, user_id, uid_min, uid_max);
-       }
+       if (user_id > uid_max)
+           fprintf(stderr, _("%s warning: %s's uid %d is greater than SYS_UID_MAX %d\n"), Prog, user_name, user_id, uid_max);
    }else{
        uid_min = (uid_t)getdef_ulong("UID_MIN", 1000UL);
        uid_max = (uid_t)getdef_ulong("UID_MAX", 6000UL);

ikerexxe:
> But then again I don't know anyone who does useradd -r -u by hand, scripts always do it, and scripts don't care about warnings.

People who install these packages will see the warning printed in their screen and they will care, as it was already notified in the [bug from the commit message](https://bugzilla.redhat.com/show_bug.cgi?id=2004911).

I wonder, now that you are considering to change the log management, if we shouldn't take another approach to their output target. My idea would be to define several log levels, two output targets (e.g. stderr and a file) and a threshold level. Those log levels below the given threshold would be printed to the file and those above the threshold to stderr and the file. This way we could avoid having warnings such as the ones provided in `check_uid_range()` as an error output. I think that something similar is being done in PAM.

> 
> I still do not want to take this patch: the uid_min part is misleading. Can you update your patch to do the below?

Done. Thanks for taking the time to review it and for accepting the change!

hallyn:
Thanks

PR #494, opened 2022-01-10
nss: get shadow_logfd with log_get_logfd()

ikerexxe:
If /etc/nsswitch.conf doesn't exist podman crashes because shadow_logfd
is NULL. In order to avoid that load the log file descriptor with the
log_get_logfd() helper function.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2038811

### DIFF
--- a/lib/nss.c
+++ b/lib/nss.c
@@ -9,6 +9,7 @@
 #include "prototypes.h"
 #include "../libsubid/subid.h"
 #include "shadowlog_internal.h"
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks.  I'm hoping to improve this at some point by replacing all those fprintf(shadow_log_fds) with a proper shadow_log() function.  The current situation is just not sustainable.

PR #498, opened 2022-01-24
fix noselinux su config bug

sulincix:
If we use **--without-selinux** su say Unknown module. pam_selinux must be optional.

### DIFF
--- a/etc/pam.d/login
+++ b/etc/pam.d/login
@@ -4,8 +4,8 @@ auth		include		system-auth
 account		required	pam_nologin.so
 account		include		system-auth
 password	include		system-auth
-session		required	pam_selinux.so close
+session [success=ok ignore=ignore module_unknown=ignore default=bad] pam_selinux.so close
 session		include		system-auth
 session		required	pam_loginuid.so
 session		optional	pam_console.so
-session		required	pam_selinux.so open
+session [success=ok ignore=ignore module_unknown=ignore default=bad] pam_selinux.so open
--- a/etc/pam.d/su
+++ b/etc/pam.d/su
@@ -7,7 +7,7 @@ auth		required	pam_wheel.so use_uid
 auth		include		system-auth
 account		include		system-auth
 password	include		system-auth
-session		required	pam_selinux.so close
+session [success=ok ignore=ignore module_unknown=ignore default=bad] pam_selinux.so close
 session		include		system-auth
-session		required	pam_selinux.so open multiple
+session [success=ok ignore=ignore module_unknown=ignore default=bad] pam_selinux.so open
 session		optional	pam_xauth.so


hallyn:
Oh, hm, I guess.

But regarding making it 'optional'.  Is it actually ok if setting the selinux context for the login process fails?  Or is that not what 'pam_selinux.so session open' does?

Based on the patch description, module_unknown seems the way to go (per http://www.linux-pam.org/Linux-PAM-html/sag-configuration-file.html)

ikerexxe:
> Based on the patch description, module_unknown seems the way to go (per http://www.linux-pam.org/Linux-PAM-html/sag-configuration-file.html)

Yes, that clearly makes sense.

sulincix:
Some distributions doesnt have selinux (like archlinux). If libpam_selinux.so file does not exists, su and login commands not working. 

Selinux must be optional. Optional modules skip if does not exists.

hallyn:
> Some distributions doesnt have selinux (like archlinux). If libpam_selinux.so file does not exists, su and login commands not working.
> 
> Selinux must be optional. Optional modules skip if does not exists.

But optional modules (iiuc) will also skip if module was there but failed.

That's bad.

> As a result of Archlinux replace shadow source pam config files: https://github.com/archlinux/svntogit-packages/blob/packages/shadow/trunk/PKGBUILD#L86

Have you *tried* something like

session [success=ok ignore=ignore module_unknown=ignore default=bad] pam_selinux.so open

?

sulincix:
> 

I will try. Thank you.

ikerexxe:
Another option would be to track the service files inside the distribution's repository, instead of relying in the upstream version. This is what Fedora has been doing for a while. Example: [password-auth](https://src.fedoraproject.org/rpms/pam/blob/f35/f/system-auth.pamd).

sulincix:
image

I tried. Top command is ignore_module config, Bottom command is required config.

bauen1:
I'm not sure if changing the order is a good idea, you should also be testing this on a system with `pam_selinux` but without selinux enabled and on a system with selinux enabled, a policy, and enforcing mode if possible.

To quote the `pam_selinux` man page:

> Adding pam_selinux into the PAM stack might disrupt behavior of other PAM modules which execute applications. To avoid that, pam_selinux.so open should be placed after such modules in the PAM stack, and pam_selinux.so close should be placed before them. When such a placement is not feasible, pam_selinux.so restore could be used to temporary restore original security contexts.

sulincix:
pam config order fixed.

hallyn:
Thanks.  Can you squash these commits into one?

sulincix:
I merged my commits :D

ikerexxe:
Could you be more specific in the commit message? As an example:
```
fix PAM service files --without-selinux

Make pam_selinux.so optional in environments where the module isn't provided.
```

Apart from that the PR looks good to me. Thanks for it.

sulincix:
Done.

ikerexxe:
LGTM

hallyn:
Thank you

PR #503, opened 2022-02-14
Update Spanish translation

feer9:
I changed the order of some words to be easier to understand


po/es.po
```diff
@@ -1058,15 +1058,14 @@ msgid ""
 "  -l, --lock-secs SEC           after failed login lock account for SEC "
 "seconds\n"
 msgstr ""
-"  -l, --lock-secs SEG           después de fallar al acceder la cuenta se\n"
-"                                bloqueará después de SEG segundos\n"
+"  -l, --lock-secs SEG           después de fallar al acceder a la cuenta\n"
+"                                se bloqueará después de SEG segundos\n"
```
ikerexxe, line 1062: IIUC from the English version, if a failed login happens then the account will be locked for SEC seconds. In my opinion, as a Spaniard, is that the second `después` doesn't make sense with the aforementioned understanding. Although I admit that it may make sense for our Latin Americans colleagues. I would suggest to change the phrase to: `después de fallar al acceder a la cuenta esta se bloqueará durante/por SEG segundos`. We should choose between `durante` or `por`. By the way, I'm open to discussion as my previous suggestion could be incorrect for other dialects.

po/es.po
```diff
@@ -654,13 +654,13 @@ msgid ""
...
-"  -d, --lastday ÚLTIMO_DÍA      establece el día del último cambio de la\n"
-"                                contraseña a ÚLTIMO_DÍA\n"
+"  -d, --lastday ÚLTIMO_DÍA      establece a ÚLTIMO_DÍA el día del último\n"
+"                                cambio de la contraseña\n"
 
 msgid ""
 "  -E, --expiredate EXPIRE_DATE  set account expiration date to EXPIRE_DATE\n"
 msgstr ""
-"  -E, --expiredate FECHA_CAD    establece la fecha de caducidad a FECHA_CAD\n"
+"  -E, --expiredate FECHA_CAD    establece a FECHA_CAD la fecha de caducidad\n"
```
ikerexxe, line 663: I'd prefer to translate `expiration` as `expiración` instead of `caducidad`. In my opinion `caducidad` should be used for edibles and not for software related stuff.

po/es.po
```diff
@@ -1971,15 +1970,14 @@ msgstr ""
 msgid ""
 "  -w, --warndays WARN_DAYS      set expiration warning days to WARN_DAYS\n"
 msgstr ""
-"  -w, --warndays DÍAS_AVISO     establece el aviso de caducidad a "
-"DÍAS_AVISO\n"
+"  -w, --warndays DÍAS_AVISO     establece a DÍAS_AVISO el aviso de caducidad\n"
```
ikerexxe, line 1973: I'd prefer to translate `expiration` as `expiración` instead of `caducidad`. In my opinion `caducidad` should be used for edibles and not for software related stuff. By the way, I think you should copy the translation from the previous `--warndays`, as it is better translated there.

ikerexxe (commented):
Some minor corrections but the overall looks good.

po/es.po
```diff
@@ -687,25 +687,25 @@ msgid ""
...
 "                                change to MAX_DAYS\n"
 msgstr ""
-"  -M, --maxdays DÍAS_MAX        establece el número máximo de días antes de\n"
-"                                cambiar la contraseña a DÍAS_MAX\n"
+"  -M, --maxdays DÍAS_MAX        establece a DÍAS_MAX el número máximo de días\n"
+"                                antes de cambiar la contraseña\n"
 
 msgid "  -R, --root CHROOT_DIR         directory to chroot into\n"
-msgstr ""
+msgstr "  -R, --root DIR_CHROOT         directorio al cual hacer chroot\n"
```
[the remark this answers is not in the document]
feer9: Seems good

po/es.po
feer9: I had not read properly the English version (just fixed the typo), but you're absolutely right. It's the same meaning for Latin America (argentinean here). I think `por` is good enough.

po/es.po
feer9: Seems fair

ikerexxe:
LGTM. Thanks for the patches.

PR #505, opened 2022-02-18
configure.ac: Default to max group name length of 32

JamiKettunen:
This used to be 16 for [historical reasons](https://github.com/shadow-maint/shadow/blob/45c6603/libmisc/chkname.c#L69) but these days basically every distro configures `--with-group-name-max-length=32` to make it match the max Linux username length, make it default.


hallyn:
Looks good to me, thanks.  Will leave it open for a bit in case someone knows of a reason why 16 has to stay the default.

PR #507, opened 2022-03-01
libmisc/xgetXXbyYY.c: Handle ERANGE error correctly

NikoDelarich:
The reentrant functions getgrgid_r, getgrnam_r, getpwnam_r, etc. all return an error code instead of setting errno. Adapt the error check accordingly.

See the man page of e.g. getpwnam_r for reference:
> On success, getpwnam_r() and getpwuid_r() return zero, and set *result to pwd.
> If no matching password record was found, these functions return 0 and store NULL in *result.
> **In case of error, an error number is returned**, and NULL is stored in *result.

### DIFF
--- a/libmisc/xgetXXbyYY.c
+++ b/libmisc/xgetXXbyYY.c
@@ -66,7 +66,6 @@
 			         "x" STRINGIZE(FUNCTION_NAME));
 			exit (13);
 		}
-		errno = 0;
 		status = REENTRANT_NAME(ARG_NAME, result, buffer,
 		                        length, &resbuf);
 		if ((0 == status) && (resbuf == result)) {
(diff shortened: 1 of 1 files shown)


hallyn:
The manpages certainly seem to agree with you.

Do you have a testcase handy (for one or all of gxgetgrgid, xgetgrnam, xgetpwnam, xgetpwuid, and xgetspnam) to show that in fact the manpage is correct?

NikoDelarich:
Not really - I stumbled upon this error on my buildroot-based system with uClibc when adding many members to a group. When a line in /etc/group becomes too long for the default buffersize (256 in uClibc), ERANGE is returned.
The same problem most likely occurs on glibc-based systems with line lengths > 1024 characters, but I haven't tested this yet.

hallyn:
After some experimentation (a few printf()s :)  I think glibc papers this over for us - status and errno are both ERANGE on my test system.  @brauner I wonder would you know offhand whether that is the case?  (e.g. for getspnam_r, that it will on ERANGE both have return value of ERANGE, and set errno to ERANGE?

PR #508, opened 2022-03-07
Fix syntax error in usermod man source

atsampson:
This was introduced in 2f30d235c2b6ba565c50d492d7be33380f73164a; it breaks the build if manpages are enabled.


hallyn:
Thanks.

PR #509, opened 2022-03-11
CI: add fedora to build

ikerexxe:


ikerexxe:
I disabled the tests because the actual test suite relies heavily on configurations coming from Ubuntu, which make it hard to use any other distribution. I guess we should think of a way of overcoming this difficulty, but for the moment I'm quite happy with the results of this PR.

hallyn:
Nice, thank you :)

PR #510, opened 2022-03-13
Fix code comment

### LINKED ISSUE #484: A comment in pwck incorrectly describes a check
The following comment in the source for `pwck` is missing a "not". It should read:
```
If uid is not system and has a home directory, then check
```
https://github.com/shadow-maint/shadow/blob/6761cf2d7e9defeed7f915c668a87b84bac2c9c3/src/pwck.c#L524

edneville:
Improving check around pw_dir which may be NULL

Closes #484

### DIFF
--- a/src/pwck.c
+++ b/src/pwck.c
@@ -498,9 +498,9 @@ static void check_pw_file (int *errors, bool *changed)
 		}
 
 		/*
-		 * If uid is system and has a home directory, then check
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks.

This is probably good enough reason to schedule another release soon, although pwck is not setuid-root and the pw field is not user-controllable.

PR #513, opened 2022-03-18
Change to strncat

stevegrubb:
ut_line is declared as a nonstring in bits/utmp.h. It might not be NUL
terminated. Limit how much it copies to the size of the array.


hallyn:
Thanks.

Issue #514, opened 2022-03-30
Segmentation fault when changing password with chpasswd


juyin:
When running the following command, chpasswd  segfaults: 
`echo "test:test" | chpasswd  -R /home/juyin/rootfs`

The backtrace:
(gdb) bt
#0  0x00007f26ab9ba570 in fread () from /lib64/libc.so.6
#1  0x00000000004040dc in read_random_bytes () at salt.c:179
#2  0x000000000040438a in gensalt (salt_size=16) at salt.c:440
#3  0x0000000000404652 in crypt_make_salt (meth=0x0, arg=0x0) at salt.c:559
#4  0x00000000004036a9 in main (argc=3, argv=0x7ffc6950ab88) at chpasswd.c:539

ikerexxe:
I guess this is happening because `chpasswd` is unable to open `/dev/urandom` in [salt.c#L177](https://github.com/shadow-maint/shadow/blob/master/libmisc/salt.c#L177), but we aren't checking the return value. Is the aforementioned file available on your system?

PR #515, opened 2022-03-30
fix segmentation fault when changing password with chpasswd

juyin:
Returns null when fopen fails. Then, using fread with a null pointer will cause a segfault.

### DIFF
--- a/src/chpasswd.c
+++ b/src/chpasswd.c
@@ -394,38 +394,36 @@ static void close_files (void)
 
 static const char *get_salt(void)
 {
-	if (   !eflg
-		&& (   (NULL == crypt_method)
-			|| (0 != strcmp (crypt_method, "NONE")))) {
-		void *arg = NULL;
+	void *arg = NULL;
 
-		if (md5flg) {
-			crypt_method = "MD5";
-		}
-#if defined(USE_SHA_CRYPT) || defined(USE_BCRYPT) || defined(USE_YESCRYPT)
-		if (sflg) {
-#if defined(USE_SHA_CRYPT)
-			if (   (0 == strcmp (crypt_method, "SHA256"))
-				|| (0 == strcmp (crypt_method, "SHA512"))) {
-				arg = &sha_rounds;
-			}
-#endif				/* USE_SHA_CRYPT */
-#if defined(USE_BCRYPT)
-			if (0 == strcmp (crypt_method, "BCRYPT")) {
-				arg = &bcrypt_rounds;
-			}
-#endif				/* USE_BCRYPT */
-#if defined(USE_YESCRYPT)
-			if (0 == strcmp (crypt_method, "YESCRYPT")) {
-				arg = &yescrypt_cost;
-			}
-#endif				/* USE_YESCRYPT */
-		}
-#endif
-		return crypt_make_salt (crypt_method, arg);
+	if (eflg || ((NULL != crypt_method) && (0 == strcmp (crypt_method, "NONE")))) {
+		return NULL;
 	}
 
-	return NULL;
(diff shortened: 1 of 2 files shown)


juyin:
The issue： #514

ikerexxe:
LGTM but since I proposed the solution I prefer that somebody else takes a look at the solution.

Thanks for the patch!

src/chpasswd.c
```diff
@@ -392,12 +392,49 @@ static void close_files (void)
 	pw_locked = false;
 }
 
+static const char *get_salt(void)
+{
+	if (   !eflg
+		&& (   (NULL == crypt_method)
+			|| (0 != strcmp (crypt_method, "NONE")))) {
```
ikerexxe, line 399: As you are changing this part of the code. Can you make it easier to read?

src/chpasswd.c
```diff
@@ -392,12 +392,49 @@ static void close_files (void)
...
+		void *arg = NULL;
+
+		if (md5flg) {
+			crypt_method = "MD5";
+		}
+#if defined(USE_SHA_CRYPT) || defined(USE_BCRYPT) || defined(USE_YESCRYPT)
+		if (sflg) {
+#if defined(USE_SHA_CRYPT)
+			if (   (0 == strcmp (crypt_method, "SHA256"))
+				|| (0 == strcmp (crypt_method, "SHA512"))) {
```
ikerexxe, line 409: Make this part also easier to read.

ikerexxe (changes requested):
I have added two minor comments.

src/chpasswd.c
juyin: I have not changed the code. I just moved this part of code in a function.

ikerexxe: This is more about the spaces/tab between `if (` and `!eflg`. I don't know why the original author did that but IMO it's difficult to read. Can you change it to something like:
```
	if (!eflg
	    && ((NULL == crypt_method) || (0 != strcmp (crypt_method, "NONE")))) {
```

This is the best I could come up with, but I am open to better solutions.

juyin: > This is more about the spaces/tab between `if (` and `!eflg`. I don't know why the original author did that but IMO it's difficult to read. Can you change it to something like:
> 
> ```
> 	if (!eflg
> 	    && ((NULL == crypt_method) || (0 != strcmp (crypt_method, "NONE")))) {
> ```
> 
> This is the best I could come up with, but I am open to better solutions.

OK, I have refactored this function

ikerexxe (approved):
LGTM! Thanks for taking care of it.

hallyn:
It seems to me there are cases of strcmp(crypt_method, X) being called the code before this patch.  Could you fix those in this PR as well?  In either case the last patch is a nice cleanup but challenging to reason about, so I want to take another look after a break.

juyin:
> It seems to me there are cases of strcmp(crypt_method, X) being called the code before this patch. Could you fix those in this PR as well? In either case the last patch is a nice cleanup but challenging to reason about, so I want to take another look after a break.

OK, we can replace it with a macro to make it cleanup.

hallyn:
That would be great - but could you have the macro first check for str not being NULL? :)

hallyn:
Oh, never mind, cflg handles it.

hallyn:
> Oh, never mind, cflg handles it.

No, it doesn't.  So I figured

```
#define IS_CRYPT_METHOD(str) ((crypt_method != NULL && strcmp(crypt_method, str) == 0) ? true : false)
```

hallyn:
Thanks!

PR #522, opened 2022-04-26
Suggest badname if name has special characters

ikerexxe:
Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2076819


ikerexxe:
Before merging the PR I have one doubt. What's the name of the option? `badname` or `badnames`?

I'm asking this question because we are using both possibilities inconsistently. Once it's clarified I may need to change other parts of the code.

hallyn:
> Before merging the PR I have one doubt. What's the name of the option? `badname` or `badnames`?
> 
> I'm asking this question because we are using both possibilities inconsistently. Once it's clarified I may need to change other parts of the code.

I suggest using badname, and editing usermod to accept both --badname and --badnames (to not break pre-existing scripts).

hallyn:
Thanks.

ikerexxe:
The PR is ready for review. I had to change both the source code, the man pages and their translations to remove all references to `badnames`.

> I suggest using badname, and editing usermod to accept both --badname and --badnames (to not break pre-existing scripts).

I left this [line](https://github.com/shadow-maint/shadow/blob/master/src/usermod.c#L981) as it was to accept both options.  Apparently, `getopt_long()` accepts that
```
The option name may be abbreviated as long as the abbreviation is unique or is
an exact match for some defined option.
```

[Source](https://patchwork.sourceware.org/project/glibc/patch/20220503045743.2887439-1-siddhesh@sourceware.org/).

The `help` option for usermod will only print `badname` though.

hallyn:
Thanks

PR #529, opened 2022-05-10
Fix issues detected by CodeQL

ikerexxe:


src/newusers.c
```diff
@@ -1130,16 +1130,23 @@ int main (int argc, char **argv)
...
+			         Prog, line, fields[0], pw_dbname ());
+			errors++;
+			continue;
+		}
+		newpw = *pw;
+
 		/* local, no need for xgetpwnam */
-		if (   (NULL == pw)
-		    && (getpwnam (fields[0]) != NULL)) {
+		if (getpwnam (fields[0]) != NULL) {
```
hallyn, line 1143: I don't think this set is right.  In the original code, these calls are only meant to be done if pw == NULL, that is, the user did not yet exist and needed to be added.  But after your patch these would always be done.

hallyn (changes requested):
I don't think the src/newusers.c patch is correct.

src/newusers.c
ikerexxe: You are completely right. I misunderstood what the code was doing.

ikerexxe:
It's ready for review again

lib/shadow.c
```diff
@@ -305,8 +305,7 @@ static struct spwd *my_sgetspent (const char *string)
 	if (fields[8][0] == '\0') {
 		spwd.sp_flag = SHADOW_SP_FLAG_UNSET;
 	} else {
-		if (getlong (fields[8], &spwd.sp_flag) == 0) {
-			/* FIXME: add a getulong function */
+		if (getulong (fields[8], &spwd.sp_flag) == 0) {
```
hallyn, line 308: Is that the right thing to do?  spwd.sp_flag is a signed int, right?  Would it be better to have a get_int_range(char *in, int *result, int min, mint max);

src/lastlog.c
```diff
@@ -88,7 +88,7 @@ static void print_one (/*@null@*/const struct passwd *pw)
 	char ptime[80];
 
 #ifdef HAVE_LL_HOST
-	int maxIPv6Addrlen;
+	int maxIPv6Addrlen = 0;
```
hallyn, line 91: Would it be better to just define a (build-time-computed) const at the top of the function?

lib/shadow.c
ikerexxe: In Fedora the structure is defined in glibc and it's an 'unsigned long int', but I guess that it could change for other distributions. Should I drop this change?

src/lastlog.c
ikerexxe: Do you mean the following?
```
#ifdef HAVE_LL_HOST
	const int maxIPv6Addrlen = 0;
#endif
```

hallyn: No, I meant

```
#ifdef HAVE_LL_HOST
	const int maxIPv6Addrlen = 25+1+IFNAMSIZ;
#endif
```

lib/shadow.c
hallyn: Oh, hm.  I guess I was looking at something like man/ja/man3/shadow.3.  I resent the fact that /usr/include/shadow.h has the { on the next line so my "grep -nrH "struct spwd {" /usr/include" didn't catch it :)

Musl also has unsigned (https://git.musl-libc.org/cgit/musl/tree/include/shadow.h).

Let's leave the patch as is, then, thanks.

libmisc/salt.c
```diff
@@ -343,7 +343,6 @@ static /*@observer@*/const unsigned long BCRYPT_get_salt_rounds (/*@null@*/int *
 	 * because musl doesn't allow rounds >= 20.
 	 */
 	if (rounds > 19) {
-		/* rounds = B_ROUNDS_MAX; */
```
hallyn, line 346: Note that I know *I* am not smart enough that, if musl ever supports > 20 rounds, rounds should be set to B_ROUNDS_MAX.  Should the comment above at line 343 include that detail?

ikerexxe: I added the information in the comment

hallyn:
Thanks!

PR #530, opened 2022-05-15
tests: replace egrep -> grep -E

thesamesam:
"egrep" is an obsolete alias for grep -E and newer greps will warn on usage
of egrep, so let's just swap it out.

Signed-off-by: Sam James <sam@gentoo.org>


hallyn:
Hm, even busybox grep supports it, so it's probably ok.

hallyn:
Thanks

thesamesam:
Thanks! If we hit issues, we can use an autoconf test for it, but I didn't want to bother with that for now given it was hardcoded to egrep anyway.

PR #531, opened 2022-05-16
Update Ukrainian translation

yurchor:
Tested for validity with "msgfmt -vc". Thanks in advance for merging.


hallyn:
Thank you.

PR #532, opened 2022-05-19
Add Ukrainian translation for man pages

yurchor:
Synced with the current XML pages, tested to compile as expected. Will be completed asap (~87% now).

Many thanks in advance for merging.


hallyn:
Thanks

PR #536, opened 2022-06-13
libmisc: use /dev/urandom as a generic fallback for read_random_bytes()

### LINKED ISSUE #512: Kernels < 3.17 not supported? No fallback to /dev/urandom
I have encountered an issue with Shadow 4.11.1 not salting the passwords at all if running a Linux kernel 3.2 (or anything under 3.17) host. Everything builds correctly, but running passwd root or any other username results in "Unable to obtain random bytes."

The code is in libmisc/salt.c. (Lines 155-177.) It tries to get random bytes from arc4random_buf (BSD systems), getentropy(), getrandom(), and finally /dev/urandom in that order, but only chooses ONE of those before exiting/failing. This is presumably for security reasons.

Getentropy and getrandom have been defined since glibc 2.25. (https://sourceware.org/bugzilla/show_bug.cgi?id=17252) However, if they lack kernel support, (i.e. < 3.17) the read_random_bytes routine exits without getting any random bytes. If running a 3.2 kernel, running even a current current glibc fails to get random bytes.

Attempting to build Shadow on with kernel 3.2 AND glibc < 2.25 doesn't have the same issue. The old glibc doesn't define getentropy and getrandom at all, and the fallback /dev/urandom is used. This is not the case with a >= glibc 2.25 inside a chroot. (Trying to build LFS.)

Any thoughts on this? Can't /dev/urandom be used as a catch-all fallback for systems with a mixed 3.2 and newer glibc?

Thank you.

xry111:
On systems with Linux kernel < 3.17, getentropy() and getrandom() may
exist but return ENOSYS.  Use /dev/urandom as a fallback to avoid a hard
requirement on Linux kernel version.

Fixes #512.

### DIFF
--- a/libmisc/salt.c
+++ b/libmisc/salt.c
@@ -156,24 +156,23 @@ static long read_random_bytes (void)
 	/* arc4random_buf, if it exists, can never fail.  */
 	arc4random_buf (&randval, sizeof (randval));
 	goto end;
+#endif
 
-#elif defined(HAVE_GETENTROPY)
+#ifdef HAVE_GETENTROPY
 	/* getentropy may exist but lack kernel support.  */
-	if (getentropy (&randval, sizeof (randval))) {
-		goto fail;
+	if (getentropy (&randval, sizeof (randval)) == 0) {
+		goto end;
 	}
+#endif
 
-	goto end;
-
-#elif defined(HAVE_GETRANDOM)
+#ifdef HAVE_GETRANDOM
 	/* Likewise getrandom.  */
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks.  Looks good as is, but should the #if #elif #elif just be a set of independent #if/#endifs?  So that it might try both getentropy and, if that fails, getrandom?

xry111:
> Thanks. Looks good as is, but should the #if #elif #elif just be a set of independent #if/#endifs? So that it might try both getentropy and, if that fails, getrandom?

I agree it's better to try them one by one.  Code updated.

hallyn:
Thanks.

PR #538, opened 2022-07-13
subordinateio: also compare the owner ID

ikerexxe:
IDs already populate /etc/subuid and /etc/subgid files so it's necessary
not only to check for the owner name but also for the owner ID of a
given range.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2093311

### DIFF
--- a/lib/subordinateio.c
+++ b/lib/subordinateio.c
@@ -17,6 +17,8 @@
 #include <ctype.h>
 #include <fcntl.h>
 
+#define ID_SIZE 31
+
 /*
  * subordinate_dup: create a duplicate range
  *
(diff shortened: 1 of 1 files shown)


ikerexxe:
@hallyn can I get a review for these changes?

lib/subordinateio.c
```diff
@@ -770,6 +799,8 @@ int list_owner_ranges(const char *owner, enum subid_type id_type, struct subid_r
 	enum subid_status status;
 	int count = 0;
 	struct subid_nss_ops *h;
+	char *id = malloc(sizeof(char) * ID_SIZE);
```
hallyn, line 802: No check for NULL?

Any reason not to just do 'char id[ID_SIZE]' ?

hallyn (changes requested):
Thanks, this will be good to have.

lib/subordinateio.c
ikerexxe: I've changed it to `char id[ID_SIZE]`

lib/subordinateio.c
```diff
@@ -770,6 +799,8 @@ int list_owner_ranges(const char *owner, enum subid_type id_type, struct subid_r
...
+	int ret;
```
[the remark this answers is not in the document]
ikerexxe: Done, and I've also changed the signature of `get_owner_id()` to return a bool instead of an int.

ikerexxe:
The PR is ready for review.

lib/subordinateio.c
```diff
@@ -745,6 +747,33 @@ gid_t sub_gid_find_free_range(gid_t min, gid_t max, unsigned long count)
...
+	struct passwd *pw;
+	struct group *gr;
+
+	switch (id_type) {
+	case ID_TYPE_UID:
+		pw = getpwnam(owner);
+		if (pw == NULL) {
+			return false;
+		}
+		snprintf(id, ID_SIZE, "%u", pw->pw_uid);
```
hallyn, line 761: Sorry could you change both snprintfs to checking the return value as is done e..g here https://github.com/lxc/lxc/blob/master/src/lxc/lsm/apparmor.c#L406 ?

There are various schools of thoughts on string printing, but when I see *that* then I know every corner case is handled and I don't have to think it through.

hallyn (approved):
Thanks!

PR #542, opened 2022-07-27
Fix E_NAME_IN_USE documentation.

jpwhiting:
Since code gives this error if username or group name is already
used the documentation should reflect that.

Found by accident when testing useradd when the username's corresponding group already existed and got 9 return code.


jpwhiting:
|Ubuntu 21.10 | Impish Indri | Release Notes | October 14, 2021 | July 14, 2022|

Looks like Impish went EOL, time to fix the CI I imagine.

hallyn:
Thanks

PR #546, opened 2022-08-06
Fix CI by testing Ubuntu 22.04 instead of 21.10.

jpwhiting:
Since Ubuntu 21.10 has gone out of support it no longer works for
testing with CI. Instead use 22.04 which is still supported.


hallyn:
Yes, thank you for that.

PR #551, opened 2022-08-16
shadow: use relaxed usernames

kanavin:
The groupadd from shadow does not allow upper case group names, the
same is true for the upstream shadow. But distributions like
Debian/Ubuntu/CentOS has their own way to cope with this problem,
this patch is picked up from CentOS release 7.0 to relax the usernames
restrictions to allow the upper case group names, and the relaxation is
POSIX compliant because POSIX indicate that usernames are composed of
characters from the portable filename character set [A-Za-z0-9._-].

Note from Alexander: Yocto project has been carrying this patch for several years, we're not aware of any reasons why it shouldn't be acceptable, but if such reasons exist, we would like to document or link to them.

### DIFF
--- a/libmisc/chkname.c
+++ b/libmisc/chkname.c
@@ -32,26 +32,44 @@ static bool is_valid_name (const char *name)
 	}
 
 	/*
-	 * User/group names must match [a-z_][a-z0-9_-]*[$]
-	 */
+         * User/group names must match gnu e-regex:
+         *    [a-zA-Z0-9_.][a-zA-Z0-9_.-]{0,30}[a-zA-Z0-9_.$-]?
+         *
+         * as a non-POSIX, extension, allow "$" as the last char for
+         * sake of Samba 3.x "add machine script"
+         *
+         * Also do not allow fully numeric names or just "." or "..".
+         */
+	int numeric;
 
-	if (('\0' == *name) ||
-	    !((('a' <= *name) && ('z' >= *name)) || ('_' == *name))) {
+	if ('\0' == *name ||
+	    ('.' == *name && (('.' == name[1] && '\0' == name[2]) ||
+			      '\0' == name[1])) ||
+	    !((*name >= 'a' && *name <= 'z') ||
+	      (*name >= 'A' && *name <= 'Z') ||
+	      (*name >= '0' && *name <= '9') ||
+	      *name == '_' ||
+	      *name == '.')) {
 		return false;
 	}
 
+	numeric = isdigit(*name);
+
 	while ('\0' != *++name) {
-		if (!(( ('a' <= *name) && ('z' >= *name) ) ||
-		      ( ('0' <= *name) && ('9' >= *name) ) ||
-		      ('_' == *name) ||
-		      ('-' == *name) ||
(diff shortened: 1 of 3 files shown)


hallyn:
I would prefer to not allow names starting with a digit.  Or especially names which are only digits, e.g. '0'.

thesamesam:
As an FYI, we've been carrying https://bugs.gentoo.org/634688 in Gentoo for years. I don't know if I agree with that decision but obviously dropping the patch would be disruptive.

hallyn:
I don't mind dots in usernames.  Again, '.' and '.. ' should not be allowed.

ikerexxe:
> The groupadd from shadow does not allow upper case group names, the same is true for the upstream shadow. But distributions like Debian/Ubuntu/CentOS has their own way to cope with this problem, this patch is picked up from CentOS release 7.0 to relax the usernames restrictions to allow the upper case group names, and the relaxation is POSIX compliant because POSIX indicate that usernames are composed of characters from the portable filename character set [A-Za-z0-9._-].

Many thanks for opening this PR!

> Note from Alexander: Yocto project has been carrying this patch for several years, we're not aware of any reasons why it shouldn't be acceptable, but if such reasons exist, we would like to document or link to them.

I agree with you. In Fedora we also have the same problem and I was thinking about tackling it.

By the way, please note that this change could be very disruptive and should be treated with great care.

> I would prefer to not allow names starting with a digit. Or especially names which are only digits, e.g. '0'.
> I don't mind dots in usernames. Again, '.' and '.. ' should not be allowed.

I completely agree. @kanavin do you mind using the [Fedora patch](https://src.fedoraproject.org/rpms/shadow-utils/blob/rawhide/f/shadow-4.8-goodname.patch) instead? It already takes care of fully numeric names or `.` or `..`. I think it's an improved version of the patch proposed for CentOS 7.

@hallyn should we get some feedback from a Debian or other distribution (Ubuntu?) representative? I think their valid names differ from the ones proposed in this PR.

kanavin:
> > I would prefer to not allow names starting with a digit. Or especially names which are only digits, e.g. '0'.
> > I don't mind dots in usernames. Again, '.' and '.. ' should not be allowed.
> 
> I completely agree. @kanavin do you mind using the [Fedora patch](https://src.fedoraproject.org/rpms/shadow-utils/blob/rawhide/f/shadow-4.8-goodname.patch) instead? It already takes care of fully numeric names or `.` or `..`. I think it's an improved version of the patch proposed for CentOS 7.

No problem. I updated the commit in this pull request.

ikerexxe:
@rbalint I'd like to hear your opinion regarding this PR. Is Debian using the same or a similar regexp when adding users and groups? Are you missing anything?

hallyn:
Thanks.  I'm good with this.  The code looks correct.

If someone has issues with it we can revert/tweak it later, but I will merge it now.

rbalint:
@ikerexxe Debian carries a patch that's even more relaxed regarding user and group names: https://salsa.debian.org/debian/shadow/-/blob/master/debian/patches/506_relaxed_usernames

I plan carrying the patch forward to avoid regressions in Debian and derivatives, but thanks for asking!

Issue #552, opened 2022-08-16
useradd - [libsemanage]: MLS is disabled, but MLS range s0 was found - Cannot commit SELinux transaction


genbtc:
Shadow v. 4.11.1 - (useradd)
OS: Gentoo amd64.
SELinux 3.4 (also tested 3.2, the same), Policy: Strict (not MLS)

Description: 
All I wanted to do was make a user, with the -Z option, but it fails with this:

`genr8too ~ # useradd -m -Z staff_u superuser7`
`[libsemanage]: MLS is disabled, but MLS range s0 was found for Unix user superuser7`
`[libsemanage]: seuser mapping [superuser7 -> (staff_u, s0)] is invalid`
`[libsemanage]: could not iterate over records`
`Cannot commit SELinux transaction`
`useradd: warning: the user name superuser7 to staff_u SELinux user mapping failed.`
`useradd: /home/superuser7 was created, but could not be removed`

FIX: shadow-useradd-selinux-MLS-option.patch
really just two lines, adding a `if (semanage_mls_enabled(handle)) {` check guarding around the part where it calls semanage_seuser_set_mlsrange() requesting a DEFAULT_SERANGE of s0, so it just skips this part unless MLS already active

```

From a0870a771b2ae371a65ac0b2d105646b744dd46c Mon Sep 17 00:00:00 2001
From: genbtc <genbtc+NOBOTS@gmx.com>
Date: Tue, 16 Aug 2022 17:54:41 -0400
Subject: [PATCH] MLS is optional, check active policy first

---
 lib/semanage.c | 30 +++++++++++++++++-------------
 1 file changed, 17 insertions(+), 13 deletions(-)

diff --git a/lib/semanage.c b/lib/semanage.c
index 1240160..b66d00b 100644
--- a/lib/semanage.c
+++ b/lib/semanage.c
@@ -121,12 +121,14 @@ static int semanage_user_mod (semanage_handle_t *handle,
 		goto done;
 	}
 
-	ret = semanage_seuser_set_mlsrange (handle, seuser, DEFAULT_SERANGE);
-	if (ret != 0) {
-		fprintf (shadow_logfd,
-		         _("Could not set serange for %s\n"), login_name);
-		ret = 1;
-		goto done;
+	if (semanage_mls_enabled(handle)) {
+		ret = semanage_seuser_set_mlsrange (handle, seuser, DEFAULT_SERANGE);
+		if (ret != 0) {
+			fprintf (shadow_logfd,
[...]

PR #556, opened 2022-08-22
useradd: check MLS enablement before setting serange

genbtc:
useradd: check MLS enablement before setting serange

Resolves: https://github.com/shadow-maint/shadow/issues/552

This fix wraps the calls to set the s0 MLS range in an `if (semanage_mls_enabled(handle)) { }` block to ask the real libsemanage to verify MLS is actually in use.
- now MLS detection is dynamic and optional, not every SELinux user uses MLS

### DIFF
--- a/lib/semanage.c
+++ b/lib/semanage.c
@@ -122,12 +122,14 @@ static int semanage_user_mod (semanage_handle_t *handle,
 		goto done;
 	}
 
-	ret = semanage_seuser_set_mlsrange (handle, seuser, DEFAULT_SERANGE);
-	if (ret != 0) {
-		fprintf (shadow_logfd,
-		         _("Could not set serange for %s\n"), login_name);
-		ret = 1;
-		goto done;
+	if (semanage_mls_enabled(handle)) {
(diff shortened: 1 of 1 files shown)


ikerexxe:
Can you improve the wording for the commit message? I'm finding it difficult to understand. Use the following as an example:
```
useradd: check MLS enablement before setting serange

Resolves: https://github.com/shadow-maint/shadow/issues/552
```

hallyn:
Where do I find documentation for this?  I don't see a manpage for it under https://github.com/SELinuxProject/selinux/tree/master/libsemanage/man/man3 .  And policycoreutils has a Changelog entry saying they use it, but their code only calls is_selinux_mls_enabled().    What's the official place to look?

genbtc:
> is_selinux_mls_enabled()
This is it. https://github.com/SELinuxProject/selinux/blob/master/libselinux/man/man3/is_selinux_enabled.3#L6
https://github.com/SELinuxProject/selinux/blob/master/libselinux/src/enabled.c#L29

hallyn:
Thanks for the links and the patch.

PR #560, opened 2022-08-27
fix `usermod -rG x y` while user `y` is not in group `x` will cause user `y` add into group `x`

flyxyz123:
fix `usermod -rG x y` while user `y` is not in group `x` will cause user `y` add into group `x`

I'll explain more in details later in an issue thread. I'm not familiar with the code. I'm not sure if my changes consider all edge cases. But it do fix the issue.

### DIFF
--- a/src/usermod.c
+++ b/src/usermod.c
@@ -716,7 +716,7 @@ static void update_group (void)
 		* If rflg+Gflg  is passed in AKA -rG invert is_member flag, which removes
 		* mentioned groups while leaving the others.
 		*/
-		if (Gflg && rflg && was_member) {
+		if (Gflg && rflg) {
 			is_member = !is_member;
 		}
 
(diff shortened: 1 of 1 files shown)


flyxyz123:
Fix #561

flyxyz123:
# I'm not familiar with the code, so use the patch with care!

I only did some limited testing with `usermod -rG` and `usermod -rG`. The issue seems fixed and everything looks find. But I can't guarantee if there are still edge cases not considered. I'm not familiar with the code. I'm not familiar with C. I don't understand the code fully. My fix is based on guessing and trial and error. Use it with care. There should be better way to fix the issue. I'm not familiar with the standard testing procedure of this project. I'm too tired to do more testing and optimization right now. I may or may not work on it further in the future.

hallyn:
Wow.  I really hate this code.

Thanks for the patch.

I believe you're exactly correct.

hallyn:
This could might be easier to understand if we renamed is_member to something like want_member.  First we set want_member to true if the group is on the grouplist.  Then if we find -r, we set want_member to false as we know we want it specifically to not be a member.  'is_member' after that point is too confusing for my little brain.

PR #562, opened 2022-08-29
lib/commonio: make lock failures more detailed

lucab:
This tweaks the database locking logic so that failures in the
link-checking paths are more detailed.
The rationale for this is that I've experienced a non-deterministic
bug which seems to be coming from this logic, and I'd like to get
more details about the actual failing condition.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1250

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -32,7 +32,7 @@
 
 /* local function prototypes */
 static int lrename (const char *, const char *);
-static int check_link_count (const char *file);
+static int check_link_count (const char *file, bool log);
 static int do_lock_file (const char *file, const char *lock, bool log);
 static /*@null@*/ /*@dependent@*/FILE *fopen_set_perms (
 	const char *name,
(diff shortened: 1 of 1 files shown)


lib/commonio.c
```diff
@@ -93,15 +93,25 @@ int lrename (const char *old, const char *new)
...
 	if (stat (file, &sb) != 0) {
+		if (log) {
+			(void) fprintf (shadow_logfd,
+			                "%s: %s file stat error: %s\n",
+			                shadow_progname, file, strerror (errno));
+		}
 		return 0;
 	}
 
 	if (sb.st_nlink != 2) {
```
lucab, line 109: I haven't yet managed to reproduce the failure in a controlled environment, but I'm suspicious of this `st_nlink` check.
In particular, I think it is theoretically possible on a multi-processor system to race the inode update and observe a stale link-count of `1` here after the `link()` call succeeded.
I've seen other comments hinting at similar caching effects, e.g. https://github.com/miquels/liblockfile/blob/v1.17/lockfile.c#L319-L320.

What is this check actually meant to protect against? Would it be possible to safely drop it? Or maybe perform a `fsync()` on the file before this?

hallyn: If two processes are calling commonio_lock() at the same time, the first will link %s to %s.$pid , causing %s to have link count2;  then the second will link %s to %s.pid2, causing %s to have link count 3.  (either $pid or $pid2 could end up finding the link count 3, or both).

I would be very surprised if the kernel's vs code was keeping the link count high, causing a race, but your results are interesting.  One suspect would be overlay, if you're using that.

lucab: Hmm, but here the file-linking goes in the other direction, i.e. `link(dbname.pid, dbname.lock)` will create `dbname.lock` as a new name for the existing `dbname.pid`.
At that point a second call like `link(dbname.pid2, dbname.lock)` will directly fail with `EEXIST` if the `.lock` file already exists, without even reaching this point.

I don't think the link count here could ever reach anything greater than `2`. On the other hand I think a (wrong/cached) value of `1` could be observed here.
Yes it is happening in a bit of complex environment with bindmounts and overlays involved, but I suspect the real culprit is some kind of weakly-synchronized caching on SMP machines.

lucab:
@hallyn while we aren't fully aligned on where to pinpoint the real issue and how to fix it, I think this logging-only PR should be safe to land without introducing further havoc and may help shedding some light on the low-level behavior at play. Could you maybe review/merge this?

PR #565, opened 2022-09-04
copy_tree: Fix regressions

ferivoz:
This pull request fixes regressions in copy_tree:
- `./configure --without-acl` leads to a usermod which does not work due to invalid chmod call.
- FIFOs in user home directories prevent `usermod -md /new_home user` from working.

Please see commit messages for more details.

PR #566, opened 2022-09-05
copy_tree: carefully treat permissions

ferivoz:
The setuid, setgid, and sticky bits are not copied during copy_tree.

Also start with very restrictive permissions before setting ownerships.

This prevents situations in which users in a group with less permissions
than others could win a race in opening the file before permissions are
removed again.

Proof of concept:

```
$ echo $HOME
/home/uwu
$ install -o uwu -g fandom -m 604 /dev/null /home/uwu/owo
$ ls -l /home/uwu/owo
-rw----r-- 1 uwu fandom 0 Sep  4 00:00 /home/uwu/owo
```

If /tmp is on another filesystem, then "usermod -md /tmp/uwu uwu" leads
to this temporary situation:

```
$ ls -l /tmp/uwu/owo
-rw----r-- 1 root root  0 Sep  4 00:00 /tmp/uwu/owo
```

This means that between openat and chownat_if_needed a user of group
fandom could open /tmp/uwu/owo and read the content when it is finally
written into the file.

### DIFF
--- a/libmisc/copydir.c
+++ b/libmisc/copydir.c
@@ -522,15 +522,14 @@ static int copy_dir (const struct path_info *src, const struct path_info *dst,
 		return -1;
 	}
 #endif				/* WITH_SELINUX */
-	if (   (mkdirat (dst->dirfd, dst->name, statp->st_mode) != 0)
+	if (   (mkdirat (dst->dirfd, dst->name, 0700) != 0)
 	    || (chownat_if_needed (dst, statp,
 	                         old_uid, new_uid, old_gid, new_gid) != 0)
+	    || (fchmodat (dst->dirfd, dst->name, statp->st_mode & 07777, AT_SYMLINK_NOFOLLOW) != 0)
 #ifdef WITH_ACL
 	    || (   (perm_copy_path (src, dst, &ctx) != 0)
 	        && (errno != 0))
-#else				/* !WITH_ACL */
-	    || (fchmodat (dst->dirfd, dst->name, statp->st_mode & 07777, AT_SYMLINK_NOFOLLOW) != 0)
-#endif				/* !WITH_ACL */
+#endif				/* WITH_ACL */
 #ifdef WITH_ATTR
 	/*
 	 * If the third parameter is NULL, all extended attributes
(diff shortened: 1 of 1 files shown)


libmisc/copydir.c
```diff
@@ -719,11 +718,10 @@ static int copy_special (const struct path_info *src, const struct path_info *ds
 	if (   (mknodat (dst->dirfd, dst->name, statp->st_mode & ~07777U, statp->st_rdev) != 0)
```
cgzones, line 718: Why not a mode of 0600 here?

libmisc/copydir.c
```diff
@@ -522,14 +522,13 @@ static int copy_dir (const struct path_info *src, const struct path_info *dst,
 		return -1;
 	}
 #endif				/* WITH_SELINUX */
-	if (   (mkdirat (dst->dirfd, dst->name, statp->st_mode) != 0)
+	if (   (mkdirat (dst->dirfd, dst->name, 0700) != 0)
 	    || (chownat_if_needed (dst, statp,
 	                         old_uid, new_uid, old_gid, new_gid) != 0)
+	    || (chmod (dst, statp->st_mode) != 0)
```
cgzones, line 528: This chmod call needs to be fixed, similar to https://github.com/shadow-maint/shadow/pull/565

libmisc/copydir.c
ikerexxe: I'm also missing a more restrictive mode here.

ikerexxe (commented):
You should rebase on top of the current master branch as I merged [PR565](https://github.com/shadow-maint/shadow/pull/565)

libmisc/copydir.c
ferivoz: The mode is already very restrictive. Please not that the mask is inverted 07777. Only the highest bits are preserved which are required for mknodat to figure out the actual type. See its manual page for more details.

libmisc/copydir.c
ferivoz: I have adjusted this pull request now. The other one was strictly about correcting regressions. This one is strictly about fixing the permissions. To keep these pull requests separated by topic, I had to accept that they contain this merge conflict. But thank you for pointing out, so it wouldn't be missed!

ferivoz:
The pull request has been updated now.

ikerexxe (approved):
LGTM. Thanks for the patches.

@cgzones do you have any other concern? If not, I will merge this PR.

libmisc/copydir.c
```diff
@@ -522,14 +522,13 @@ static int copy_dir (const struct path_info *src, const struct path_info *dst,
...
+	if (   (mkdirat (dst->dirfd, dst->name, 0700) != 0)
 	    || (chownat_if_needed (dst, statp,
 	                         old_uid, new_uid, old_gid, new_gid) != 0)
+	    || (fchmodat (dst->dirfd, dst->name, statp->st_mode & 07777, AT_SYMLINK_NOFOLLOW) != 0)
 #ifdef WITH_ACL
 	    || (   (perm_copy_path (src, dst, &ctx) != 0)
 	        && (errno != 0))
-#else				/* !WITH_ACL */
-	    || (fchmodat (dst->dirfd, dst->name, statp->st_mode & 07777, AT_SYMLINK_NOFOLLOW) != 0)
 #endif				/* !WITH_ACL */
```
cgzones, line 532: These `#endif` comments are now out-of-sync; otherwise LGTM.

hallyn: With that fixed LGTM thanks.

ferivoz: Thank you @cgzones for the careful review! I have adjusted the pull request accordingly.

ikerexxe:
LGTM so merging. Thanks for the patch!

PR #569, opened 2022-09-05
Use safer and simpler string copy functions.

alejandro-colomar:
Please review that str_l_cpy() and STR_L_CPY() are only used with input strings that are guaranteed to be NUL-terminated, and also review that str_s_cpy() and STR_S_CPY() are only used with input strings that can be not terminated under some circumstances.  I'm almost sure that there are cases where I used the _s_ function where we could use the _l_ one.  Another pair of eyes will help for sure.

Also: not tested.  I didn't remember how to build the project, so I couldn't run the tests.  I hope CI helps me in this.

### DIFF
--- a/contrib/pwdauth.c
+++ b/contrib/pwdauth.c
@@ -223,8 +223,7 @@ main(argc, argv)
 		fprintf(stderr, "Who are you?\n");
 		exit(2);
 	}
-	strncpy(myname, pw->pw_name, sizeof myname - 1);
-	myname[sizeof myname - 1] = '\0';
+	STRLCPY (myname, pw->pw_name);
 	name = myname;
 
 	if (argc > 1) {
--- a/libmisc/utmp.c
+++ b/libmisc/utmp.c
@@ -209,13 +209,11 @@ static void updwtmpx (const char *filename, const struct utmpx *utx)
 
 	if (   (NULL != host)
 	    && ('\0' != host[0])) {
-		hostname = (char *) xmalloc (strlen (host) + 1);
-		strcpy (hostname, host);
+		hostname = xstrdup (host);
 #ifdef HAVE_STRUCT_UTMP_UT_HOST
 	} else if (   (NULL != ut)
 	           && ('\0' != ut->ut_host[0])) {
-		hostname = (char *) xmalloc (sizeof (ut->ut_host) + 1);
-		USTR2STR (hostname, ut->ut_host);
+		hostname = xstrdup (ut->ut_host);
 #endif				/* HAVE_STRUCT_UTMP_UT_HOST */
 	}
 
@@ -353,14 +351,12 @@ int setutmp (struct utmp *ut)
 
 	if (   (NULL != host)
 	    && ('\0' != host[0])) {
-		hostname = (char *) xmalloc (strlen (host) + 1);
-		strcpy (hostname, host);
+		hostname = xstrdup (host);
 #ifdef HAVE_STRUCT_UTMP_UT_HOST
 	} else if (   (NULL != ut)
 	           && (NULL != ut->ut_host)
 	           && ('\0' != ut->ut_host[0])) {
-		hostname = (char *) xmalloc (sizeof (ut->ut_host) + 1);
-		USTR2STR (hostname, ut->ut_host);
+		hostname = xstrdup (ut->ut_host);
 #endif				/* HAVE_STRUCT_UTMP_UT_TYPE */
 	}
 
--- a/libmisc/utmp.c
+++ b/libmisc/utmp.c
@@ -46,8 +46,7 @@ static bool is_my_tty (const char *tty)
 	if ('\0' == tmptty[0]) {
 		const char *tname = ttyname (STDIN_FILENO);
 		if (NULL != tname) {
-			(void) strncpy (tmptty, tname, sizeof tmptty);
-			tmptty[sizeof (tmptty) - 1] = '\0';
+			STRLCPY (tmptty, tname);
 		}
 	}
 
@@ -216,8 +215,7 @@ static void updwtmpx (const char *filename, const struct utmpx *utx)
 	} else if (   (NULL != ut)
 	           && ('\0' != ut->ut_host[0])) {
 		hostname = (char *) xmalloc (sizeof (ut->ut_host) + 1);
-		strncpy (hostname, ut->ut_host, sizeof (ut->ut_host));
-		hostname[sizeof (ut->ut_host)] = '\0';
+		USTR2STR (hostname, ut->ut_host);
 #endif				/* HAVE_STRUCT_UTMP_UT_HOST */
 	}
 
@@ -235,25 +233,25 @@ static void updwtmpx (const char *filename, const struct utmpx *utx)
 	utent->ut_type = USER_PROCESS;
 #endif				/* HAVE_STRUCT_UTMP_UT_TYPE */
 	utent->ut_pid = getpid ();
-	strncpy (utent->ut_line, line,      sizeof (utent->ut_line) - 1);
+	STRLCPY (utent->ut_line, line);
 #ifdef HAVE_STRUCT_UTMP_UT_ID
 	if (NULL != ut) {
 		strncpy (utent->ut_id, ut->ut_id, sizeof (utent->ut_id));
 	} else {
 		/* XXX - assumes /dev/tty?? */
-		strncpy (utent->ut_id, line + 3, sizeof (utent->ut_id) - 1);
+		STRLCPY (utent->ut_id, line + 3);
 	}
 #endif				/* HAVE_STRUCT_UTMP_UT_ID */
 #ifdef HAVE_STRUCT_UTMP_UT_NAME
-	strncpy (utent->ut_name, name,      sizeof (utent->ut_name));
+	STRLCPY (utent->ut_name, name);
 #endif				/* HAVE_STRUCT_UTMP_UT_NAME */
 #ifdef HAVE_STRUCT_UTMP_UT_USER
-	strncpy (utent->ut_user, name,      sizeof (utent->ut_user) - 1);
+	STRLCPY (utent->ut_user, name);
 #endif				/* HAVE_STRUCT_UTMP_UT_USER */
 	if (NULL != hostname) {
 		struct addrinfo *info = NULL;
 #ifdef HAVE_STRUCT_UTMP_UT_HOST
-		strncpy (utent->ut_host, hostname, sizeof (utent->ut_host) - 1);
+		STRLCPY (utent->ut_host, hostname);
 #endif				/* HAVE_STRUCT_UTMP_UT_HOST */
 #ifdef HAVE_STRUCT_UTMP_UT_SYSLEN
 		utent->ut_syslen = MIN (strlen (hostname),
@@ -362,8 +360,7 @@ int setutmp (struct utmp *ut)
 	           && (NULL != ut->ut_host)
 	           && ('\0' != ut->ut_host[0])) {
 		hostname = (char *) xmalloc (sizeof (ut->ut_host) + 1);
-		strncpy (hostname, ut->ut_host, sizeof (ut->ut_host));
-		hostname[sizeof (ut->ut_host)] = '\0';
+		USTR2STR (hostname, ut->ut_host);
 #endif				/* HAVE_STRUCT_UTMP_UT_TYPE */
 	}
 
@@ -378,22 +375,22 @@ int setutmp (struct utmp *ut)
 
 	utxent->ut_type = USER_PROCESS;
 	utxent->ut_pid = getpid ();
-	strncpy (utxent->ut_line, line,      sizeof (utxent->ut_line));
+	STRLCPY (utxent->ut_line, line);
 	/* existence of ut->ut_id is enforced by configure */
 	if (NULL != ut) {
 		strncpy (utxent->ut_id, ut->ut_id, sizeof (utxent->ut_id));
 	} else {
 		/* XXX - assumes /dev/tty?? */
-		strncpy (utxent->ut_id, line + 3, sizeof (utxent->ut_id));
+		STRLCPY (utxent->ut_id, line + 3);
 	}
 #ifdef HAVE_STRUCT_UTMPX_UT_NAME
-	strncpy (utxent->ut_name, name,      sizeof (utxent->ut_name));
+	STRLCPY (utxent->ut_name, name);
 #endif				/* HAVE_STRUCT_UTMPX_UT_NAME */
-	strncpy (utxent->ut_user, name,      sizeof (utxent->ut_user));
+	STRLCPY (utxent->ut_user, name);
 	if (NULL != hostname) {
 		struct addrinfo *info = NULL;
 #ifdef HAVE_STRUCT_UTMPX_UT_HOST
-		strncpy (utxent->ut_host, hostname, sizeof (utxent->ut_host));
+		STRLCPY (utxent->ut_host, hostname);
 #endif				/* HAVE_STRUCT_UTMPX_UT_HOST */
 #ifdef HAVE_STRUCT_UTMPX_UT_SYSLEN
 		utxent->ut_syslen = MIN (strlen (hostname),
--- a/src/logoutd.c
+++ b/src/logoutd.c
@@ -54,8 +54,7 @@ static int check_login (const struct utmp *ut)
 	/*
 	 * ut_user may not have the terminating NUL.
 	 */
-	strncpy (user, ut->ut_user, sizeof (ut->ut_user));
-	user[sizeof (ut->ut_user)] = '\0';
+	USTR2STR (user, ut->ut_user);
 
 	(void) time (&now);
 
@@ -246,8 +245,7 @@ int main (int argc, char **argv)
 				kill (-ut->ut_pid, SIGKILL);
 			}
 
-			strncpy (user, ut->ut_user, sizeof (user) - 1);
-			user[sizeof (user) - 1] = '\0';
+			USTR2STR (user, ut->ut_user);
 
 			SYSLOG ((LOG_NOTICE,
 				 "logged off user '%s' on '%s'", user,
--- a/contrib/shadow-anonftp.patch
+++ b/contrib/shadow-anonftp.patch
@@ -81,7 +81,7 @@ diff -uNr shadow-980403.orig/src/newusers.c shadow-980403/src/newusers.c
 +	while ((flag = getopt (argc, argv, "p:h")) != EOF) {
 +		switch (flag) {
 +		case 'p':
-+			STRFCPY(anonproto, optarg);
++			STRLCPY(anonproto, optarg);
 +			break;
 +		case 'h':
 +		default:
--- a/lib/defines.h
+++ b/lib/defines.h
@@ -241,6 +241,8 @@ static inline void memzero(void *ptr, size_t size)
 #define SCALE DAY
 #endif
 
+#define NITEMS(arr)  (sizeof((arr)) / sizeof((arr)[0]))
+
 /* Copy string pointed by B to array A with size checking.  It was originally
    in lmain.c but is _very_ useful elsewhere.  Some setuid root programs with
    very sloppy coding used to assume that BUFSIZ will always be enough...  */
--- a/lib/defines.h
+++ b/lib/defines.h
@@ -251,6 +251,11 @@ static inline void memzero(void *ptr, size_t size)
 #define STRFCPY(A,B) \
 	(strncpy((A), (B), sizeof(A) - 1), (A)[sizeof(A) - 1] = '\0')
 #define STRLCPY(dst, src)  (strlcpy((dst), (src), NITEMS(dst)) >= NITEMS(src))
+#define USTR2STR(dst, src)  do \
+{ \
+	static_assert(NITEMS(dst) > NITEMS(src), ""); \
+	ustr2str((dst), (src), NITEMS(src)); \
+} while (0)
 
 #ifndef PASSWD_FILE
 #define PASSWD_FILE "/etc/passwd"
--- a/lib/defines.h
+++ b/lib/defines.h
@@ -250,6 +250,7 @@ static inline void memzero(void *ptr, size_t size)
 					/* danger - side effects */
 #define STRFCPY(A,B) \
 	(strncpy((A), (B), sizeof(A) - 1), (A)[sizeof(A) - 1] = '\0')
+#define STRLCPY(dst, src)  (strlcpy((dst), (src), NITEMS(dst)) >= NITEMS(src))
 
 #ifndef PASSWD_FILE
 #define PASSWD_FILE "/etc/passwd"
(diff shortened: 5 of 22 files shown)


alejandro-colomar:
In the end I didn't use my own `stpecpy()` function, which is very similar to `strlcpy(3bsd)`, because for the use cases in this project, `strlcpy(3bsd)` has a simpler interface (my function is especially useful for concatenation, but shadow-utils doesn't do that).

alejandro-colomar:
BTW, I plan to use `ARRAY_SIZE()` and derivatives instead of `sizeof()` for `STR[LS]CPY()`. See <https://stackoverflow.com/questions/37538/how-do-i-determine-the-size-of-my-array-in-c/57537491#57537491>.  That will be for a separate PR, in which I'll also check other places where that might be useful.

alejandro-colomar:
Do you know why are those linker errors?  Do I need to modify the build system?

alejandro-colomar:
Please, tell me what do I need to do to fix the linker errors.  I tried the same changes as when I added `date_to_str.c`.

lslebodn:
> Please, tell me what do I need to do to fix the linker errors. I tried the same changes as when I added date_to_str.c.
https://builds.sr.ht/~hallyn/job/839071 says:
```
/usr/bin/ld: ../lib/.libs/libshadow.a(libshadow_la-gshadow.o): in function `sgetsgent':
gshadow.c:(.text+0x187): undefined reference to `shdw_strlcpy'
```

You try to use shdw_strlcpy on two places in libshadow.
* https://github.com/shadow-maint/shadow/pull/569/files#diff-31ab4957a0dc40664087c923ea54ad7a772f55d80d7178fe9061bcbde519b5dfR103
* https://github.com/shadow-maint/shadow/pull/569/files#diff-3afff04b0dd0b1757f2732586602eb6fe60c0445ee3d00b596a8025b3ca907e4R127

But libshadow is not linked with libmisc
* https://github.com/shadow-maint/shadow/blob/master/lib/Makefile.am

libsuid is linked with both
* https://github.com/shadow-maint/shadow/blob/master/libsubid/Makefile.am#L22

alejandro-colomar:
> > Please, tell me what do I need to do to fix the linker errors. I tried the same changes as when I added date_to_str.c.
> > https://builds.sr.ht/~hallyn/job/839071 says:
> 
> ```
> /usr/bin/ld: ../lib/.libs/libshadow.a(libshadow_la-gshadow.o): in function `sgetsgent':
> gshadow.c:(.text+0x187): undefined reference to `shdw_strlcpy'
> ```
> 
> You try to use shdw_strlcpy on two places in libshadow.
> 
>     * https://github.com/shadow-maint/shadow/pull/569/files#diff-31ab4957a0dc40664087c923ea54ad7a772f55d80d7178fe9061bcbde519b5dfR103
> 
>     * https://github.com/shadow-maint/shadow/pull/569/files#diff-3afff04b0dd0b1757f2732586602eb6fe60c0445ee3d00b596a8025b3ca907e4R127
> 
> 
> But libshadow is not linked with libmisc
> 
>     * https://github.com/shadow-maint/shadow/blob/master/lib/Makefile.am
> 
> 
> libsuid is linked with both
> 
>     * https://github.com/shadow-maint/shadow/blob/master/libsubid/Makefile.am#L22

Hmmm, should I put `shdw_strlcpy()` in a different library?  I hope I don't have to duplicate it.  How is the structure of the libs?  Is any of them pulled by every other?

alejandro-colomar:
Actually, I realized of an issue of this change.  Since now we're not zeroing the rest of the buffer when copying strings, we may be allowing to leak password lengths, if one is able to read a buffer that once contained a password.  We zero the string after it is used, but that zeroing is only until the NUL terminator with strzero().  This change needs to be accompanied by a full bzero(3)ing of the buffer.

cgzones:
> Actually, I realized of an issue of this change. Since now we're not zeroing the rest of the buffer when copying strings, we may be allowing to leak password lengths, if one is able to read a buffer that once contained a password. We zero the string after it is used, but that zeroing is only until the NUL terminator with strzero(). This change needs to be accompanied by a full bzero(3)ing of the buffers.

As part of that `getpass(3)`, which is deprecated and removed in in POSIX.1-2001, should then probably replaced by a custom implementation, putting the password in a buffer of known size.

libmisc/strscpy.c
```diff
@@ -0,0 +1,82 @@
...
+ *
+ *	If the input string didn't contain a NUL character, -E2BIG is
+ *	returned.
+ *
+ * CAVEATS
+ *	This function can hide bugs in the input string, by not reading
+ *	until the end of the string.  strlcpy(3bsd) and stpecpy()
+ *	read the input string until the end of it, which normally
+ *	results in SIGSEGV if the input string was not correctly
+ *	NUL-terminated.
```
hallyn, line 65: Should also add the caveat that results will be undefined if memory areas overlap (per memccpy manpage)

alejandro-colomar: That should be obvious by the use of `restrict` in the prototype.  I also updated the Linux man-pages recently to document restrict in the SYNOPSIS sections consistently.  I think the only function that explicitly states in text that overlapping memory is UB is memcpy(3), due to the historical bugs about it.

I mean, we could add the text, but my first impression is that I prefer to keep it short.  `restrict` is clear (to those who know the keyword, that is).

contrib/pwdauth.c
```diff
@@ -223,8 +223,7 @@ main(argc, argv)
 		fprintf(stderr, "Who are you?\n");
 		exit(2);
 	}
-	strncpy(myname, pw->pw_name, sizeof myname - 1);
-	myname[sizeof myname - 1] = '\0';
+	strscpy(myname, pw->pw_name, sizeof myname);
```
hallyn: I see no reason not to check for E2BIG here.

alejandro-colomar: I tried to not modify behavior in the same commit.

If we do that, I'd do it in a separate one.

I'm fine with checking truncation, but I don't know in which cases it would be important.  So, if you mark all of them, I can fix them in a continuation commit.

Thanks!

libmisc/strlcpy.c
```diff
@@ -0,0 +1,78 @@
...
+ *	slow, since it reads the whole 'dst' string, to find the NUL
+ *	character.  stpecpy() is both easy to use and fast.
+ *
+ *	This function requires that the input string is NUL-terminated.
+ *	If this cannot be guaranteed, other functions should be used,
+ *	such as strscpy().
+ */
+
+size_t
+shdw_strlcpy(char *dst, const char *restrict src, size_t n)
```
cgzones, line 66: Missing `restrict` for first argument

libmisc/utmp.c
```diff
@@ -215,8 +214,7 @@ static void updwtmpx (const char *filename, const struct utmpx *utx)
 	} else if (   (NULL != ut)
 	           && ('\0' != ut->ut_host[0])) {
 		hostname = (char *) xmalloc (sizeof (ut->ut_host) + 1);
-		strncpy (hostname, ut->ut_host, sizeof (ut->ut_host));
-		hostname[sizeof (ut->ut_host)] = '\0';
+		shdw_strscpy (hostname, ut->ut_host, sizeof (ut->ut_host) + 1);
```
cgzones, line 217: xstrdup()?

libmisc/utmp.c
```diff
@@ -361,8 +359,7 @@ int setutmp (struct utmp *ut)
 	           && (NULL != ut->ut_host)
 	           && ('\0' != ut->ut_host[0])) {
 		hostname = (char *) xmalloc (sizeof (ut->ut_host) + 1);
-		strncpy (hostname, ut->ut_host, sizeof (ut->ut_host));
-		hostname[sizeof (ut->ut_host)] = '\0';
+		shdw_strscpy (hostname, ut->ut_host, sizeof (ut->ut_host) + 1);
```
cgzones, line 362: xstrdup()?

src/logoutd.c
```diff
@@ -54,8 +54,7 @@ static int check_login (const struct utmp *ut)
 	/*
 	 * ut_user may not have the terminating NUL.
 	 */
-	strncpy (user, ut->ut_user, sizeof (ut->ut_user));
-	user[sizeof (ut->ut_user)] = '\0';
+	shdw_strscpy (user, ut->ut_user, sizeof (ut->ut_user) + 1);
```
cgzones, line 57: `STRSCPY (user, ut->ut_user)`?

libmisc/strlcpy.c
alejandro-colomar: Yup! forgot.

alejandro-colomar:
Changes:

- Added a missing restrict [@cgzones]
- wsfix

alejandro-colomar:
Rebased to master.

libmisc/utmp.c
alejandro-colomar: BTW, just to confirm that there are no other places than the 2 you suggested, I run the following:

```c
$ grep -rnI -A5 xmalloc | grep -i -B5 str.cpy
libmisc/utmp.c-212-		strcpy (hostname, host);
libmisc/utmp.c-213-#ifdef HAVE_STRUCT_UTMP_UT_HOST
libmisc/utmp.c-214-	} else if (   (NULL != ut)
libmisc/utmp.c-215-	           && ('\0' != ut->ut_host[0])) {
libmisc/utmp.c:216:		hostname = (char *) xmalloc (sizeof (ut->ut_host) + 1);
libmisc/utmp.c-217-		shdw_strscpy (hostname, ut->ut_host, sizeof (ut->ut_host) + 1);
--
libmisc/utmp.c-357-#ifdef HAVE_STRUCT_UTMP_UT_HOST
libmisc/utmp.c-358-	} else if (   (NULL != ut)
libmisc/utmp.c-359-	           && (NULL != ut->ut_host)
libmisc/utmp.c-360-	           && ('\0' != ut->ut_host[0])) {
libmisc/utmp.c:361:		hostname = (char *) xmalloc (sizeof (ut->ut_host) + 1);
libmisc/utmp.c-362-		shdw_strscpy (hostname, ut->ut_host, sizeof (ut->ut_host) + 1);
```

alejandro-colomar: Oh, wait, there are a few more, when we consider `strcpy(3)`:

```c
$ grep -rnI -A5 x.alloc | grep -i -B5 -e str.cpy -e strcpy
src/su.c-245-		targs[n_args+2] = NULL;
src/su.c-246-		while (1 != n_args) {
--
src/su.c:1180:		arg0 = xmalloc (strlen (cp) + 2);
src/su.c-1181-		arg0[0] = '-';
src/su.c-1182-		strcpy (arg0 + 1, cp);
--
src/login.c-934-			if ('\0' == username[0]) {
src/login.c-935-				/* Prompt for a new login */
--
src/passwd.c:520:		char *newpw = xmalloc (strlen (cp) + 2);
src/passwd.c-521-
src/passwd.c-522-		strcpy (newpw, "!");
--
libmisc/utmp.c-133-
libmisc/utmp.c-134-	endutxent ();
libmisc/utmp.c-135-
--
libmisc/utmp.c:211:		hostname = (char *) xmalloc (strlen (host) + 1);
libmisc/utmp.c-212-		strcpy (hostname, host);
libmisc/utmp.c-213-#ifdef HAVE_STRUCT_UTMP_UT_HOST
libmisc/utmp.c-214-	} else if (   (NULL != ut)
libmisc/utmp.c-215-	           && ('\0' != ut->ut_host[0])) {
libmisc/utmp.c:216:		hostname = (char *) xmalloc (sizeof (ut->ut_host) + 1);
libmisc/utmp.c-217-		shdw_strscpy (hostname, ut->ut_host, sizeof (ut->ut_host) + 1);
--
libmisc/utmp.c-229-
libmisc/utmp.c-230-
libmisc/utmp.c-231-#ifdef HAVE_STRUCT_UTMP_UT_TYPE
--
libmisc/utmp.c:355:		hostname = (char *) xmalloc (strlen (host) + 1);
libmisc/utmp.c-356-		strcpy (hostname, host);
libmisc/utmp.c-357-#ifdef HAVE_STRUCT_UTMP_UT_HOST
libmisc/utmp.c-358-	} else if (   (NULL != ut)
libmisc/utmp.c-359-	           && (NULL != ut->ut_host)
libmisc/utmp.c-360-	           && ('\0' != ut->ut_host[0])) {
libmisc/utmp.c:361:		hostname = (char *) xmalloc (sizeof (ut->ut_host) + 1);
libmisc/utmp.c-362-		shdw_strscpy (hostname, ut->ut_host, sizeof (ut->ut_host) + 1);
--
libmisc/xmalloc.c-31-	void *ptr;
libmisc/xmalloc.c-32-
libmisc/xmalloc.c-33-	ptr = malloc (size);
libmisc/xmalloc.c-34-	if (NULL == ptr) {
--
libmisc/xmalloc.c:45:	return strcpy (xmalloc (strlen (str) + 1), str);
--
libmisc/xmalloc.c-48-void xfree(void *ap)
libmisc/xmalloc.c-49-{
libmisc/xmalloc.c-50-	if (ap) {
--
libmisc/obscure.c:161:	wrapped = xmalloc (strlen (oldmono) * 2 + 1);
libmisc/obscure.c-162-	strcpy (wrapped, oldmono);
```

contrib/pwdauth.c
hallyn: Sounds good, thanks, will do.

alejandro-colomar:
Changes: Add commit merging `xmalloc()` + `str*cpy()` into `xstrdup()`.

alejandro-colomar:
Changes: Deprecated my gmail email account in favor of the kernel address.

cgzones:
Since the kernel is planing to get rid of `strlcpy()`, https://lwn.net/Articles/905777/, would it make sense to just introduce `strscpy()`?

libmisc/strlcpy.c
```diff
@@ -0,0 +1,78 @@
...
+{
+	char    *d = dst;
+	size_t  i;
+
+	for (i = 0; src[i] != '\0'; i++) {
+		if (i < n)
+			*d++ = src[i];
+	}
+	if (d - dst < n)
+		*d = '\0';
```
hallyn, line 76: I don't think this is right.  If n = 2, and strlen(src) > 2, then *d here will (if i'm thinking right) be pointing at *(d[2]).

alejandro-colomar: Yeah, you're right; I was off by one.  I guess it should be
```c
if (n)
        dst[n - 1] = '\0';
```

alejandro-colomar: BTW, I would wait to merge the agetpass() thing, and then we can just pick strlcpy(3bsd) from libbsd.

alejandro-colomar:
> Since the kernel is planing to get rid of `strlcpy()`, https://lwn.net/Articles/905777/, would it make sense to just introduce `strscpy()`?

I think so.

Technically, in code with zero bugs, `strscpy()` is a valid implementation of `strlcpy()` (except for the return value), since they handle valid input strings exactly in the same way.  It's in how they handle non strings where they differ:  `strscpy()` also handles non-strings, while `strlcpy()` makes it Undefined Behavior.  When working with strings which are (in theory) guaranteed by the programmer to be strings, it might be better to invoke some undefined behavior on purpose to detect incorrect assumptions.

```c
char longbuf[BUFSIZ];
char buf[4];
char string[7] = "string";
char non_string[10] "non-string";  // note no space for '\0'

// Notice that for correct usage of strscpy() one needs to use MIN(),
// or risk copying random bytes.

// All three copy the same substring, and report truncation.
strlcpy(buf, string, sizeof(buf));
strscpy(buf, string, MIN(sizeof(buf), sizeof(string));
strscpy(buf, string, sizeof(buf));

// All three copy the full string.
strlcpy(longbuf, string, sizeof(longbuf));
strscpy(longbuf, string, MIN(sizeof(longbuf), sizeof(string)));
strscpy(longbuf, string, sizeof(longbuf));

// Here they differ:

// Might crash, or might succeed with truncation.
strlcpy(buf, non_string, sizeof(buf));
// Both succeed with truncation.
strscpy(buf, non_string, MIN(sizeof(buf), sizeof(non_string)));
strscpy(buf, non_string, sizeof(buf));

// And in the case of longbuf:

// Might crash, or might (incorrectly) succeed with or without truncation,
// after copying some random bytes.
strlcpy(longbuf, non_string, sizeof(longbuf));
// Succeeds.
strscpy(longbuf, non_string, MIN(sizeof(longbuf), sizeof(non_string)));
// (Incorrectly) succeeds with or without truncation,
// after copying some random bytes.
strscpy(longbuf, non_string, sizeof(longbuf));
```

That crash is something desirable, to avoid continuing with an incorrect assumption and invoke another (more dangerous) class of UB.  Also, for correct use of `strscpy()`, that is, for non strings, one needs to use `MIN()`; forgetting to use it is only correct if the destination buffer is guaranteed to be at least as large as the source buffer.

alejandro-colomar:
Changes:

- Add `LENGTHOF()` and use it instead of `sizeof()` within `STRSCPY()` and `STRLCPY()`.
- Use `MIN()` within `STRSCPY()`.

alejandro-colomar:
Rebased to master.

src/logoutd.c
alejandro-colomar: So, the array is of size `UT_NAMESIZE`, which in my sistem happens to be 32.  `utmp(7)` says that the string is terminated iff the name is shorter than 32 chars.  If the name is 32 chars, we're out of luck.  I think it was a bad decission to not include the null byte inconditionally, and thus limiting the name to 31 bytes.

This made me realize that `strscpy()` seems to not be useful for this case: it's introducing a bug.

When the name is 32 chars, we were passing 33 to `strscpy()`, but that might crash, or would copy 33 chars from the source string and then zero the last one.

For a correct function here, we would need a new function which accepts 2 different sizes, one for the dest buffer, and one for the source buffer.

alejandro-colomar:
@cgzones @hallyn I'll add yet another strXYZ function :p

`strncpy_s()` seems to be necessary.  I'll also write a wrapper `STRNCPY_S()` to use it more comfortably.

alejandro-colomar:
Changes: Add buf2str()

alejandro-colomar:
Changes:

- Replace strscpy() by strlcpy() or strncpy() or buf2str().
  strscpy() is only good when a program can't crash on invalid input, which may be valid for the kernel, but not for us.
- Fix some bugs introduced by changing string copy functions in previous iterations.  The fix is done by changing to a different string copy function.
- Improve documentation for the new functions.

alejandro-colomar:
Changes:

- Replace strscpy() by strlcpy() or strncpy() or buf2str().
  strscpy() is only good when a program can't crash on invalid input, which may be valid for the kernel, but not for us.
- Fix some bugs introduced by changing string copy functions in previous iterations.  The fix is done by changing to a different string copy function.
- Improve documentation for the new functions.

alejandro-colomar:
Changes: remove strscpy()

alejandro-colomar:
This is a slightly modified version of the table above, taking `buf2str()` into account.  Use of `MIN()` with `strscpy()` was incorrect, and it showed me the need for `buf2str()`.

```c
char longbuf[BUFSIZ];
char buf[4];
char string[7] = "string";
char non_string[10] "non-string";  // note no space for '\0'

// 1)

// Both copy the same substring, and report truncation:
strlcpy(buf, string, sizeof(buf));
strscpy(buf, string, sizeof(buf));
// Undefined behavior:
buf2str(buf, string, sizeof(string));  // nonsense

// 2)

// All three copy the full string:
strlcpy(longbuf, string, sizeof(longbuf));
strscpy(longbuf, string, sizeof(longbuf));
buf2str(longbuf, string, sizeof(string));  // nonsense

// 3)

// Might crash, or might succeed with truncation:
strlcpy(buf, non_string, sizeof(buf));
// Succeeds with truncation:
strscpy(buf, non_string, sizeof(buf));
// Undefined behavior:
buf2str(buf, non_string, sizeof(non_string));  // nonsense

// 4)

// Might crash, or might (incorrectly) succeed with or without truncation,
// after copying some random bytes:
strlcpy(longbuf, non_string, sizeof(longbuf));
// (Incorrectly) succeeds with or without truncation,
// after copying some random bytes:
strscpy(longbuf, non_string, sizeof(longbuf));  // nonsense
buf2str(longbuf, non_string, sizeof(non_string));
```

alejandro-colomar:
Changes: Actually remove strscpy().

alejandro-colomar:
I'm going to close this until agetpass(3) gets merged.

alejandro-colomar:
Since we're close to having libbsd support in the master branch, I'll reopen this PR.  It's been a long time since I last had a look at it, so I'll transform it as a draft (if github allows me), and then check that everything makes sense.  There are a few things that I will do:

- Rename buf2str().  I've had some discussions about it with a collegue at nginx, and we came up with a better name that more precisely tells users what it does.

- Remove my bogus definition of strlcpy(3), and use libbsd's.

- While the agetpass() PR isn't merged, I'll cherry-pick the commit that adds libbsd support in the build system.

And probably a few more changes.

alejandro-colomar:
Rebased to master.

alejandro-colomar:
Changes: Rename LENGTHOF() to NITEMS().

Rationale:

Length is typically associated with strlen(3).  It's better to have a distinct term for number of elements of an array.  See discussion here: <https://lore.kernel.org/linux-man/7931044a-b707-5a70-86c2-be298c35aa57@gmail.com/T/#m794ad2a3173a19099625ee1dec7ea11ab754513d>.

alejandro-colomar:
Changes:

- Rename buf2str() to ustr2str().
- Change the type of its second param (the unterminated string) to `u_char *`.
- Update documentation of the function with the new changes.

alejandro-colomar:
Changes:

- Use libbsd's strlcpy(3).

My definition was bogus.  Prefer not reinventing delicate stuff.

alejandro-colomar:
Changes: Use better names for macro's parameters.

alejandro-colomar:
Changes:

- Add a dummy message to static_assert(3).

Only since C23 the second argument is optional.  C11 required a string literal as argument.  Pass `""`, since it's useless (the context already provides more information than a message could do).

See: <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/man3/static_assert.3#n41>

alejandro-colomar:
Changes:

- Use memcpy(3) instead of memccpy(3).

memccpy(3) is unnecessary.  The fixed-width buffers that we handle are very small, so copying more bytes than necessary is not a performance problem, and neither a security problem, since we guarantee that (1) we don't overrun buffers (by using safe copy functions, such as strlcpy(3)), and we also guarantee that (2) we don't keep old sensitive data in freed buffers (by using agetpass() and erase_pass()).

Instead, using memcpy(3) may have some performance gains, since we don't need to be checking for the (possibly missing) NUL byte.  It's also less code to read, so it only has benefits.

alejandro-colomar:
Changes:

- Remove the libbsd hook, since it wasn't working either.  We'll need to wait to the agetpass() PR, and still need some more changes to the build system (or maybe library organization?).

alejandro-colomar:
I'll close again, now that the patches are cleaner, until agetpass() is merged.

alejandro-colomar:
Heh, I force pushed before reopening, and now github doesn't allow me to reopen.  Too bad.  I'll have to open a new one, which for auditability, I'll initially open at the same commit that this one was: <https://github.com/shadow-maint/shadow/commit/49fcdcf95f7c3dd2f94ea4df3946ec0d020c2edd>.  I'll use this opportunity to use a branch, instead of master :)

PR #573, opened 2022-09-27
agetpass() to replace getpass(3)

alejandro-colomar:
```
There are several issues with getpass(3).

Many implementations of it share the same issues that the infamous
gets(3).  In glibc it's not so terrible, since it's a wrapper
around getline(3).  But it still has an important bug:

If the password is long enough, getline(3) will realloc(3) memory,
and prefixes of the password will be laying around in some
deallocated memory.

See the getpass(3) manual page for more details, and especially
the commit that marked it as deprecated, which links to a long
discussion in the linux-man@ mailing list.

So, readpassphrase(3bsd) is preferrable, which is provided by
libbsd on GNU systems.  However, using readpassphrase(3) directly
is a bit verbose, so we can write our own wrapper with a simpler
interface similar to that of getpass(3).

One of the benefits of writing our own interface around
readpassphrase(3) is that we can hide there any checks that should
be done always and which would be error-prone to repeat every
time.  For example, check that there was no truncation in the
password.

Also, use malloc(3) to get the buffer, instead of using a global
buffer.  We're not using a multithreaded program (and it wouldn't
make sense to do so), but it's nice to know that the visibility of
our passwords is as limited as possible.

erase_pass() is a clean-up function that handles all clean-up
correctly, including zeroing the entire buffer, and then
free(3)ing the memory.  By using [[gnu::malloc(erase_pass)]], we
make sure that we don't leak the buffers in any case, since the
compiler will be able to enforce clean up.

Link: <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/commit?id=7ca189099d73bde954eed2d7fc21732bcc8ddc6b>
Reported-by: Christian Göttsche <cgzones@googlemail.com>
```

I still haven't added autoconf code to add the libbsd dependency, since I don't have much experience in using autoconf.  If someone wants to help with that, please do :)  Otherwise, gimme some time to do that.

I also reopened an old thread in the libc-alpha@ mailing list to discuss the possible addition of readpassphrase(3) to glibc (and also discuss the agetpass(3) interface, in case they want to add it too).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -405,6 +405,21 @@ AC_SUBST(LIYESCRYPT)
 AC_CHECK_LIB(crypt, crypt, [LIYESCRYPT=-lcrypt],
 	[AC_MSG_ERROR([crypt() not found])])
 
+AC_SEARCH_LIBS([readpassphrase], [bsd], [], [
+	AC_MSG_ERROR([readpassphrase() is missing, either from libc or libbsd])
+])
+AS_IF([test "$ac_cv_search_readpassphrase" = "-lbsd"], [
+	PKG_CHECK_MODULES([LIBBSD], [libbsd-overlay])
+])
+dnl Make sure either the libc or libbsd provide the header.
+save_CFLAGS="$CFLAGS"
+CFLAGS="$CFLAGS $LIBBSD_CFLAGS"
+AC_CHECK_HEADERS([readpassphrase.h])
+AS_IF([test "$ac_cv_header_readpassphrase_h" != "yes"], [
+	AC_MSG_ERROR([readpassphrase.h is missing])
+])
+CFLAGS="$save_CFLAGS"
+
 AC_SUBST(LIBACL)
 if test "$with_acl" != "no"; then
 	AC_CHECK_HEADERS(acl/libacl.h attr/error_context.h, [acl_header="yes"], [acl_header="no"])
--- a/lib/defines.h
+++ b/lib/defines.h
@@ -330,6 +330,12 @@ extern char *strerror ();
 /* Maximum length of passwd entry */
 #define PASSWD_ENTRY_MAX_LENGTH 32768
 
+#if (__GNUC__ >= 11) && !defined(__clang__)
+# define ATTR_MALLOC(deallocator)  [[gnu::malloc(deallocator)]]
+#else
+# define ATTR_MALLOC(deallocator)
+#endif
+
 #ifdef HAVE_SECURE_GETENV
 #  define shadow_getenv(name) secure_getenv(name)
 # else
--- a/lib/prototypes.h
+++ b/lib/prototypes.h
@@ -47,7 +47,7 @@ extern int expire (const struct passwd *, /*@null@*/const struct spwd *);
 
 /* agetpass.c */
 extern void erase_pass(char *pass);
-[[gnu::malloc(erase_pass)]]
+ATTR_MALLOC(erase_pass)
 extern char *agetpass(const char *prompt);
 
 /* isexpired.c */
--- a/lib/prototypes.h
+++ b/lib/prototypes.h
@@ -44,6 +44,12 @@ extern int add_groups (const char *);
 /* age.c */
 extern void agecheck (/*@null@*/const struct spwd *);
 extern int expire (const struct passwd *, /*@null@*/const struct spwd *);
+
+/* agetpass.c */
+extern void erase_pass(char *pass);
+[[gnu::malloc(erase_pass)]]
+extern char *agetpass(const char *prompt);
+
 /* isexpired.c */
 extern int isexpired (const struct passwd *, /*@null@*/const struct spwd *);
 
--- a/libmisc/agetpass.c
+++ b/libmisc/agetpass.c
@@ -26,6 +26,8 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <config.h>
+
 #include <limits.h>
 #include <readpassphrase.h>
 #include <stdio.h>
--- a/libmisc/agetpass.c
+++ b/libmisc/agetpass.c
@@ -1,0 +1,150 @@
+/*
+ * Copyright (c) 2022, Alejandro Colomar <alx@kernel.org>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. The name of the copyright holders or contributors may not be used to
+ *    endorse or promote products derived from this software without
+ *    specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+ * PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT
+ * HOLDERS OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <limits.h>
+#include <readpassphrase.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ident "$Id$"
+
+#include "prototypes.h"
+
+
+#if !defined(PASS_MAX)
+#define PASS_MAX  BUFSIZ
+#endif
+
+
+/*
+ * SYNOPSIS
+ *	[[gnu::malloc(erase_pass)]]
+ *	char *agetpass(const char *prompt);
+ *
+ *	void erase_pass(char *pass);
+ *
+ * ARGUMENTS
+ *   agetpass()
+ *	prompt	String to be printed before reading a password.
+ *
+ *   erase_pass()
+ *	pass	password previously returned by agetpass().
+ *
+ * DESCRIPTION
+ *   agetpass()
+ *	This function is very similar to getpass(3).  It has several
+ *	advantages compared to getpass(3):
+ *
+ *	- Instead of using a static buffer, agetpass() allocates memory
+ *	  through malloc(3).  This makes the function thread-safe, and
+ *	  also reduces the visibility of the buffer.
+ *
+ *	- agetpass() doesn't call realloc(3) internally.  Some
+ *	  implementations of getpass(3), such as glibc, do that, as a
+ *	  consequence of calling getline(3).  That's a bug in glibc,
+ *	  which allows leaking prefixes of passwords in freed memory.
+ *
+ *	- agetpass() doesn't overrun the output buffer.  If the input
+ *	  password is too long, it simply fails.  Some implementations
+ *	  of getpass(3), share the same bug that gets(3) has.
+ *
+ *	As soon as possible, the password obtained from agetpass() be
+ *	erased by calling erase_pass(), to avoid possibly leaking the
+ *	password.
+ *
+ *   erase_pass()
+ *	This function first clears the password, by calling
+ *	explicit_bzero(3) (or an equivalent call), and then frees the
+ *	allocated memory by calling free(3).
+ *
+ *	NULL is a valid input pointer, and in such a case, this call is
+ *	a no-op.
+ *
+ * RETURN VALUE
+ *	agetpass() returns a newly allocated buffer containing the
+ *	password on success.  On error, errno is set to indicate the
+ *	error, and NULL is returned.
+ *
+ * ERRORS
+ *   agetpass()
+ *	This function may fail for any errors that malloc(3) or
+ *	readpassphrase(3) may fail, and in addition it may fail for the
+ *	following errors:
+ *
+ *	ENOBUFS
+ *		The input password was longer than PASS_MAX.
+ *
+ * CAVEATS
+ *	If a password is passed twice to erase_pass(), the behavior is
+ *	undefined.
+ */
+
+
+char *
+agetpass(const char *prompt)
+{
+	char    *pass;
+	size_t  len;
+
+	pass = malloc(PASS_MAX);
+	if (pass == NULL)
+		return NULL;
+
+	if (readpassphrase(prompt, pass, PASS_MAX, RPP_REQUIRE_TTY) == NULL)
+		goto fail;
+
+	len = strlen(pass);
+
+	if (len == 0)
+		return pass;
+
+	if (pass[len - 1] != '\n') {
+		errno = ENOBUFS;
+		goto fail;
+	}
+
+	pass[len - 1] = '\0';
+
+	return pass;
+
+fail:
+	memzero(pass, PASS_MAX);
+	free(pass);
+	return NULL;
+}
+
+
+void
+erase_pass(char *pass)
+{
+	if (pass == NULL)
+		return;
+	memzero(pass, PASS_MAX);
+	free(pass);
+}
--- a/src/sulogin.c
+++ b/src/sulogin.c
@@ -202,10 +202,9 @@ static void catch_signals (unused int sig)
 			execl (PATH_TELINIT, "telinit", RUNLEVEL, (char *) 0);
 #endif
 			exit (0);
-		} else {
-			STRFCPY (pass, cp);
-			strzero (cp);
 		}
+		STRFCPY (pass, cp);
+		strzero (cp);
 		if (valid (pass, &pwent)) {	/* check encrypted passwords ... */
 			break;	/* ... encrypted passwords matched */
 		}
--- a/src/sulogin.c
+++ b/src/sulogin.c
@@ -182,7 +182,7 @@ static void catch_signals (unused int sig)
 		 */
 
 		/* get a password for root */
-		cp = getpass (_(
+		cp = agetpass (_(
 "\n"
 "Type control-d to proceed with normal startup,\n"
 "(or give root password for system maintenance):"));
@@ -193,6 +193,7 @@ static void catch_signals (unused int sig)
 		 * --marekm
 		 */
 		if ((NULL == cp) || ('\0' == *cp)) {
+			erase_pass (cp);
 #ifdef	USE_SYSLOG
 			SYSLOG (LOG_INFO, "Normal startup\n");
 			closelog ();
@@ -204,7 +205,8 @@ static void catch_signals (unused int sig)
 			exit (0);
 		}
 		STRFCPY (pass, cp);
-		strzero (cp);
+		erase_pass (cp);
+
 		if (valid (pass, &pwent)) {	/* check encrypted passwords ... */
 			break;	/* ... encrypted passwords matched */
 		}
--- a/libmisc/Makefile.am
+++ b/libmisc/Makefile.am
@@ -5,6 +5,7 @@ AM_CPPFLAGS = -I$(top_srcdir)/lib -I$(top_srcdir) $(ECONF_CPPFLAGS)
 
 noinst_LTLIBRARIES = libmisc.la
 
+libmisc_la_CFLAGS = $(LIBBSD_CFLAGS)
 libmisc_la_SOURCES = \
 	addgrps.c \
 	age.c \
--- a/libmisc/Makefile.am
+++ b/libmisc/Makefile.am
@@ -8,6 +8,7 @@ noinst_LTLIBRARIES = libmisc.la
 libmisc_la_SOURCES = \
 	addgrps.c \
 	age.c \
+	agetpass.c \
 	audit_help.c \
 	basename.c \
 	chkname.c \
--- a/src/gpasswd.c
+++ b/src/gpasswd.c
@@ -887,24 +887,24 @@ static void change_passwd (struct group *gr)
 	printf (_("Changing the password for group %s\n"), group);
 
 	for (retries = 0; retries < RETRIES; retries++) {
-		cp = getpass (_("New Password: "));
+		cp = agetpass (_("New Password: "));
 		if (NULL == cp) {
 			exit (1);
 		}
 
 		STRFCPY (pass, cp);
-		strzero (cp);
-		cp = getpass (_("Re-enter new password: "));
+		erase_pass (cp);
+		cp = agetpass (_("Re-enter new password: "));
 		if (NULL == cp) {
 			exit (1);
 		}
 
 		if (strcmp (pass, cp) == 0) {
-			strzero (cp);
+			erase_pass (cp);
 			break;
 		}
 
-		strzero (cp);
+		erase_pass (cp);
 		memzero (pass, sizeof pass);
 
 		if (retries + 1 < RETRIES) {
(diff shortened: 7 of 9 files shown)


hallyn:
You'll need to add libbsd-dev to build deps right?

alejandro-colomar:
> You'll need to add libbsd-dev to build deps right?

Yep.  And in RPM-based distros I guess they call it libbsd-devel.

libmisc/agetpass.c
```diff
@@ -0,0 +1,84 @@
...
+agetpass(const char *prompt)
+{
+	char    *pass;
+	size_t  len;
+
+	pass = malloc(PASS_MAX);
+	if (pass == NULL)
+		return NULL;
+
+	if (readpassphrase(prompt, pass, PASS_MAX, 0) == NULL)
```
hallyn, line 55: Seems like you should pass RPP_REQUIRE_TTY in as flags, esp since you are replacing getpass which requires a controlling terminal.  Any reason not to do that?

alejandro-colomar: Good catch!  I'll fix that.  Thanks.

alejandro-colomar:
Changes in last push:

readpassphrase(3): Add RPP_REQUIRE_TTY to flags field.

alejandro-colomar:
Rebased to master.

src/sulogin.c
```diff
@@ -202,10 +203,10 @@ static void catch_signals (unused int sig)
 			execl (PATH_TELINIT, "telinit", RUNLEVEL, (char *) 0);
 #endif
 			exit (0);
-		} else {
-			STRFCPY (pass, cp);
-			strzero (cp);
 		}
+		STRFCPY (pass, cp);
+		erase_pass (cp);
+
```
alejandro-colomar, line 209: BTW, I maybe should have splitted this into a separate commit:

Don't `else` after a `noreturn` call (`exit(3)`).

alejandro-colomar: I will split it.  Definitely.  I don't want someone auditing the commits in the future and think "why the hell did this change?  no commit message at all about it...".

alejandro-colomar:
Changes:

- Split a commit that was conceptually completely different from the others.
- wsfix

alejandro-colomar:
> You'll need to add libbsd-dev to build deps right?

In which file(s) do you define the build deps?  And also the runtime deps if at all?

libmisc/agetpass.c
```diff
@@ -0,0 +1,84 @@
...
+
+	pass[len - 1] = '\0';
+
+	return pass;
+
+truncated:
+	memzero(pass, PASS_MAX);
+fail:
+	free(pass);
+	return NULL;
```
ikerexxe, line 74: Can you reduce the number of goto definitions to one? It helps reading the code and the future maintenance tasks.

Moreover, if you are using them it would be nice to have just one return sentence (or maybe two if we can take into account the malloc error).

src/sulogin.c
```diff
@@ -202,10 +203,10 @@ static void catch_signals (unused int sig)
 			execl (PATH_TELINIT, "telinit", RUNLEVEL, (char *) 0);
 #endif
 			exit (0);
-		} else {
-			STRFCPY (pass, cp);
-			strzero (cp);
 		}
+		STRFCPY (pass, cp);
```
ikerexxe, line 207: What would happen if `cp` is NULL?

ikerexxe (commented):
I added some comments.

> Yep. And in RPM-based distros I guess they call it libbsd-devel.

You are right.

@hallyn do we have any policy regarding indentation (tabs vs spaces)? And regarding `if` and mandatory block sequences?

libmisc/agetpass.c
alejandro-colomar: Yes, we can reduce the `goto`s to one by calling `erase_pass()`.  Since I wrote the code for `agetpass()` before deciding to accept `NULL` in the cleanup, it wasn't possible, but now it is.  It'll also reduce code size a little bit.  I had this consideration before pushing the code, but in the end decided not to reduce the number of `goto`s, as I think maintainability will be better with two `goto`s, as the code won't depend on if `erase_pass(3)` accepts `NULL` or not (arguably this is unlikely to change), and also because the `truncated` label makes it clear the error condition without having to add a comment and without forcing the maintainer to guess it by the code.

Consider the change to accomodate to the suggestion:

```c
diff --git a/libmisc/agetpass.c b/libmisc/agetpass.c
index 619c9f36..e5be64a8 100644
--- a/libmisc/agetpass.c
+++ b/libmisc/agetpass.c
@@ -61,16 +61,14 @@ agetpass(const char *prompt)
                return pass;
 
        if (pass[len - 1] != '\n')
-               goto truncated;
+               goto fail;
 
        pass[len - 1] = '\0';
 
        return pass;
 
-truncated:
-       memzero(pass, PASS_MAX);
 fail:
-       free(pass);
+       erase_pass(pass);
        return NULL;
 }
 
```

I'd like to ask you to reconsider the suggestion, as that code is less readable to me (but of course this is subjective, and if you confirm that you like it more, I'll change it).  I strongly oppose to adding comments to the code, considering the `goto` label did a great job of distinguishing the error cases.

Regarding the one return sentence (not accounting the `malloc(3)` error), I'm not sure how to achieve it in a not-too-deeply-indented way.  Of course it can be done with lots of `if`-`else`, but I'd prefer to avoid that.  See <https://www.kernel.org/doc/html/latest/process/coding-style.html#centralized-exiting-of-functions>.

src/sulogin.c
alejandro-colomar: That you'd have entered in the above conditional, and if `execl(3)` is successful, you'd have changed the current process into `telinit`, so this code would not be executed.  If that call failed, it would return -1 (correctly ignored), and then we'd have run `exit(EXIT_SUCCESS)`, which would have immediately quitted the program, and also the code would not be executed.

I question why `exit(3)` is being passed `0` instead of `1`, but I didn't care too much to investigate.

But it's clear that this code is not being executed if `cp == NULL`.

libmisc/agetpass.c
alejandro-colomar: BTW, I just asked to OpenBSD maintainers if `explicit_bzero(3)` is required after `readpassphrase(3)` failed.  It's not so clear as I first thought.

<https://marc.info/?l=openbsd-misc&m=166437328100790&w=2>

ikerexxe: I'd suggest the following:
```
char *
agetpass(const char *prompt)
{
	char    *pass;
	size_t  len;
	bool err = false;

	pass = malloc(PASS_MAX);
	if (pass == NULL)
		return NULL;

	if (readpassphrase(prompt, pass, PASS_MAX, RPP_REQUIRE_TTY) == NULL) {
		err = true;
		goto done;
	}

	len = strlen(pass);

	if (len == 0)
		goto done;

	if (pass[len - 1] != '\n') {
		memzero(pass, PASS_MAX);
		err = true;
		goto done;
	}

	pass[len - 1] = '\0';

done:
	if (err == true) {
	    free(pass);
	}

	return pass;
}
```

What do you think?

src/sulogin.c
ikerexxe: You are right. Thanks for clarifying.

libmisc/agetpass.c
alejandro-colomar: That has a bug: we should return `NULL` on error, not a `free(3)`d pointer.

alejandro-colomar:
Changes: Rework indentation in `erase_pass()`, by returning early.  That way, it's more explicit the intention that passing NULL is a no-op.

src/sulogin.c
alejandro-colomar: You're welcome to ask :)

lib/prototypes.h
```diff
@@ -44,6 +44,12 @@ extern int add_groups (const char *);
 /* age.c */
 extern void agecheck (/*@null@*/const struct spwd *);
 extern int expire (const struct passwd *, /*@null@*/const struct spwd *);
+
+/* agetpass.c */
+extern void erase_pass(char *p);
+__attribute__((malloc(erase_pass)))
```
cgzones, line 50: This function annotation does not work with Clang 15:
```
In file included from basename.c:21:
../lib/prototypes.h:50:16: error: 'malloc' attribute takes no arguments
__attribute__((malloc(erase_pass)))
               ^
1 error generated.
```

libmisc/agetpass.c
```diff
@@ -0,0 +1,143 @@
...
+	if (pass[len - 1] != '\n')
+		goto truncated;
+
+	pass[len - 1] = '\0';
+
+	return pass;
+
+truncated:
+	errno = ENOBUFS;
+	memzero(pass, PASS_MAX);
```
cgzones, line 129: Why not memzero on failure to avoid potential partial leaks?
The buffer was allocated by us, so it is safe to access.

cgzones (changes requested):
The build system needs to be updated.

Draft:
```diff
diff --git a/configure.ac b/configure.ac
index 61f2e588..399bb0cb 100644
--- a/configure.ac
+++ b/configure.ac
@@ -393,6 +393,11 @@ if test "$enable_subids" != "no"; then
 fi
 AM_CONDITIONAL(ENABLE_SUBIDS, test "x$enable_subids" != "xno")
 
+AC_SEARCH_LIBS([readpassphrase], [bsd],,[AC_MSG_ERROR([readpassphrase() function not found])])
+AC_CHECK_HEADERS([readpassphrase.h], [], [
+       AC_CHECK_HEADERS([bsd/readpassphrase.h], [], [AC_MSG_ERROR([readpassphrase.h header file not found])])
+])
+
 AC_SUBST(LIBCRYPT)
 AC_CHECK_LIB(crypt, crypt, [LIBCRYPT=-lcrypt],
        [AC_MSG_ERROR([crypt() not found])])
diff --git a/libmisc/Makefile.am b/libmisc/Makefile.am
index 3d319cd8..636cc439 100644
--- a/libmisc/Makefile.am
+++ b/libmisc/Makefile.am
@@ -8,6 +8,7 @@ noinst_LTLIBRARIES = libmisc.la
 libmisc_la_SOURCES = \
        addgrps.c \
        age.c \
+       agetpass.c \
        audit_help.c \
        basename.c \
        chkname.c \
diff --git a/libmisc/agetpass.c b/libmisc/agetpass.c
index f6eca4be..edee4e19 100644
--- a/libmisc/agetpass.c
+++ b/libmisc/agetpass.c
@@ -26,8 +26,14 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <config.h>
+
 #include <limits.h>
+#ifdef HAVE_READPASSPHRASE_H
 #include <readpassphrase.h>
+#elif defined HAVE_BSD_READPASSPHRASE_H
+#include <bsd/readpassphrase.h>
+#endif
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
```

lib/prototypes.h
alejandro-colomar: Confirmed:

```c
alx@asus5775:~/tmp$ cat malloc.c 
#include <stdlib.h>


void my_free(void *p);
[[gnu::malloc(my_free)]] void *my_malloc(size_t s);


void my_free(void *p)
{
	free(p);
}

void *my_malloc(size_t s)
{
	return malloc(s);
}
```
```c
alx@asus5775:~/tmp$ gcc -Wall -Wextra -S malloc.c 
```
```c
alx@asus5775:~/tmp$ clang -Wall -Wextra -S malloc.c 
malloc.c:5:2: error: expected expression
[[gnu::malloc(my_free)]] void *my_malloc(size_t s);
 ^
malloc.c:5:26: error: expected identifier or '('
[[gnu::malloc(my_free)]] void *my_malloc(size_t s);
                         ^
2 errors generated.
```
We'll need to hide it in an `ATTR_MALLOC` macro.

libmisc/agetpass.c
alejandro-colomar: Yup, now that I added the `ENOBUFS` thing, I'll rework it a little bit to also remove one label.

lib/prototypes.h
alejandro-colomar: BTW, the fact that this results in an error is due to a GCC bogus design decision.  They called to completely different attributes with the same name.

If they called it differently, one should be able to use `[[]]` syntax and force the compiler to ignore unknown attributes.

libmisc/agetpass.c
alejandro-colomar: Done.

alejandro-colomar:
Changes:
- Reduce `goto` labels to 1 instead of two.  Now that we set `errno` to `ENOBUFS`, that's self-documenting enough to remove the labels.  This also has the side effect that we don't need to care if `readpassphrase(3)` can fail with a partially-filled buffer anymore.  Now we always clear the buffer before free(3)ing it. [@cgzones, @ikerexxe]

alejandro-colomar:
Changes:

    Reduce goto labels to one instead of two. Now that we set errno to ENOBUFS, that's self-documenting enough to remove the labels. This also has the side effect that we don't need to care if readpassphrase(3) can fail with a partially-filled buffer anymore. Now we always clear the buffer before free(3)ing it. [@cgzones, @ikerexxe]

alejandro-colomar:
> The build system needs to be updated.
> 
> Draft:
> 

[...]

> diff --git a/libmisc/agetpass.c b/libmisc/agetpass.c
> index f6eca4be..edee4e19 100644
> --- a/libmisc/agetpass.c
> +++ b/libmisc/agetpass.c
> @@ -26,8 +26,14 @@
>   * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
>   */
>  
> +#include <config.h>
> +
>  #include <limits.h>
> +#ifdef HAVE_READPASSPHRASE_H
>  #include <readpassphrase.h>
> +#elif defined HAVE_BSD_READPASSPHRASE_H
> +#include <bsd/readpassphrase.h>
> +#endif
>  #include <stdio.h>
>  #include <stdlib.h>
>  #include <string.h>
> ```

You should always use the `<readpassphrase.h>` form.  See[ `libbsd(7)`](https://manpages.debian.org/testing/libbsd-dev/libbsd.7.en.html):

```
     The library can be used in an overlay mode, which is the
     preferred way, so that the code is portable and requires no
     modification to the original BSD code.  This can be done
     easily with the pkg‐config(1) library named libbsd‐overlay.
     Or by adding the system‐specific include directory with the
     bsd/ suffix to the list of system include paths.  With gcc
     this could be ‐isystem ${includedir}/bsd.  In addition the
     LIBBSD_OVERLAY pre‐processor variable needs to be defined.
     The includes in this case should be the usual system ones,
     such as <unistd.h>.

     The other way to use the library is to use the namespaced
     headers, this is less portable as it makes using libbsd
     mandatory and it will not work on BSD‐based systems, and
     requires modifying original BSD code.  This can be done
     with the pkg‐config(1) library named libbsd.  The includes
     in this case should be namespaced with bsd/, such as
     <bsd/unistd.h>.
```

libmisc/agetpass.c
alejandro-colomar: I updated the code to have only one label.

alejandro-colomar:
Changes: Add `agetpass.c` to the Makefile.  [@cgzones]

alejandro-colomar:
Changes: Hide `[[gnu::malloc(x)]]` in a macro to workaround the issues in clang.

libmisc/agetpass.c
```diff
@@ -0,0 +1,142 @@
...
+ *   agetpass()
+ *	This function may fail for any errors that malloc(3) or
+ *	readpassphrase(3) may fail, and in addition it may fail for the
+ *	following errors:
+ *
+ *	ENOBUFS	The input password was longer than PASS_MAX.
+ *
+ * CAVEATS
+ *	If a password is passed twice to erase_pass(), the behavior is
+ *	undefined, as it will be passed twice to free(3).
```
ikerexxe, line 98: You already fixed this, didn't you?

libmisc/agetpass.c
ikerexxe: Thanks!

ikerexxe (commented):
One minor comment left.

Apart from that you should add `libbsd-dev` or `libbsd-devel` to the package dependencies in the `.build` folder. An example of such a file is [.builds/ubuntu-focal.yml](https://github.com/shadow-maint/shadow/blob/master/.builds/ubuntu-focal.yml).

libmisc/agetpass.c
alejandro-colomar: Nope, that caveat still applies.

alejandro-colomar:
Changes: Add `libbsd-dev[el]` to `<.build/>` files.

alejandro-colomar:
> One minor comment left.
> 
> Apart from that you should add `libbsd-dev` or `libbsd-devel` to the package dependencies in the `.build` folder. An example of such a file is [.builds/ubuntu-focal.yml](https://github.com/shadow-maint/shadow/blob/master/.builds/ubuntu-focal.yml).

And also, we still need to use the following stuff in the build system, but I don't know how:

```sh
$ pkg-config --cflags-only-I libbsd-overlay
-isystem /usr/include/bsd
$ pkg-config --cflags-only-other libbsd-overlay
-DLIBBSD_OVERLAY
$ pkg-config --libs libbsd-overlay
-lbsd
```

lib/defines.h
```diff
@@ -335,6 +335,13 @@ extern char *strerror ();
 # endif
 #endif
 
+/* [[gnu::malloc(deallocator)]] */
+#if defined(__GNUC__) && !defined(__clang__)
```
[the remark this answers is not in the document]
alejandro-colomar: Hmm, didn't know this is so new.

If it is so new, I'll go directly for the ISO C attribute syntax `[[]]`, I think, since I think they are already implemented in GCC 11.  I don't have a gcc-11 instance, so if you can confirm, it would be great.

alejandro-colomar:
Changes:

- Check the GCC version for ATTR_MALLOC()
- Use ISO C attribute syntax

alejandro-colomar:
Rebase to master.

ikerexxe:
I think that you need to change `#include <readpassphrase.h>` to `#include <bsd/readpassphrase.h>` to compile the changes. At least this is working in Fedora.

On top of that, you should add libbsd to the linking stage in the `configure.ac` file. You could use libcrypt as an [example](https://github.com/shadow-maint/shadow/blob/master/configure.ac#L397).

alejandro-colomar:
> I think that you need to change `#include <readpassphrase.h>` to `#include <bsd/readpassphrase.h>` to compile the changes. At least this is working in Fedora.

You could do that, but libbsd was designed so that you could use the same includes as you would use in a BSD system (it's a compatibility library), which is why they provide the following pkg-config configuration:

```
$ pkg-config --cflags libbsd-overlay 
-DLIBBSD_OVERLAY -isystem /usr/include/bsd
```

That `-isystem /usr/include/bsd` is what allows not specifying the directory.

> 
> On top of that, you should add libbsd to the linking stage in the `configure.ac` file. You could use libcrypt as an [example](https://github.com/shadow-maint/shadow/blob/master/configure.ac#L397)

How would you use the pkg-config configuration from the configure scripts?  Does it change anything?

alejandro-colomar:
Rebased to master.

alejandro-colomar:
> > I think that you need to change `#include <readpassphrase.h>` to `#include <bsd/readpassphrase.h>` to compile the changes. At least this is working in Fedora.
> 
> You could do that, but libbsd was designed so that you could use the same includes as you would use in a BSD system (it's a compatibility library), which is why they provide the following pkg-config configuration:
> 
> ```
> $ pkg-config --cflags libbsd-overlay 
> -DLIBBSD_OVERLAY -isystem /usr/include/bsd
> ```
> 
> That `-isystem /usr/include/bsd` is what allows not specifying the directory.
> 
> > On top of that, you should add libbsd to the linking stage in the `configure.ac` file. You could use libcrypt as an [example](https://github.com/shadow-maint/shadow/blob/master/configure.ac#L397)
> 
> How would you use the pkg-config configuration from the configure scripts? Does it change anything?

@hallyn Could you please with this?  I tried, but I have no clue.

ikerexxe:
Sorry, I've been very busy lately and this message must have slipped into my inbox.

> How would you use the pkg-config configuration from the configure scripts? Does it change anything?

This is the [solution](https://github.com/ikerexxe/shadow/commit/0c0970fe342807d56219dc84d861054fa5b940f8) that comes to my mind.

By the way, I only tested it on my machine and I think it should work for Linux distributions, but I wasn't able to test it in a BSD system.

alejandro-colomar:
> Sorry, I've been very busy lately and this message must have slipped into my inbox.
> 
> > How would you use the pkg-config configuration from the configure scripts? Does it change anything?
> 
> This is the [solution](https://github.com/ikerexxe/shadow/commit/0c0970fe342807d56219dc84d861054fa5b940f8) that comes to my mind.
> 
> By the way, I only tested it on my machine and I think it should work for Linux distributions, but I wasn't able to test it in a BSD system.

Hi @ikerexxe ,  I wanted to avoid having conditional includes, but since I don't have any idea of how to do it better, I'll take it.

Thanks!

configure.ac
```diff
@@ -401,6 +401,9 @@ AC_SUBST(LIYESCRYPT)
 AC_CHECK_LIB(crypt, crypt, [LIYESCRYPT=-lcrypt],
 	[AC_MSG_ERROR([crypt() not found])])
 
+AC_CHECK_HEADERS(/usr/include/bsd/readpassphrase.h,
```
guillemj, line 404: Hmm, please do not use absolute pathnames, this hardcodes filesystem policy. I'll try to prepare a proposal later today for autodetection using the libbsd-overlay.

alejandro-colomar: Hi @guillemj ,

Just a friendly ping, in case you have some time to have a look at this :)

Thanks,
Alex

guillemj: Sorry, right this slipped my mind. Probably something like https://github.com/guillemj/shadow/commit/b1476818de79ff74c8250a6fd04e65b2a52cedff

alejandro-colomar: Thanks!  Will try :)

alejandro-colomar:
Rebased to master.

alejandro-colomar:
Changes: silence a `-Wmismatched-dealloc` warning.

alejandro-colomar:
@hallyn @ikerexxe 

Locally it seems to build fine, thanks to @guillemj 's patch.  There seems to be some missing piece in the CI.  Could you please have a look at it?

alejandro-colomar:
Changes: Added SYNOPSIS to the documentation.

ikerexxe:
> Locally it seems to build fine, thanks to @guillemj 's patch. There seems to be some missing piece in the CI. Could you please have a look at it?

Fixed in https://github.com/shadow-maint/shadow/pull/596

alejandro-colomar:
Changes: cosmetic fix in function documentation

alejandro-colomar:
Changes:
- Better name for `erase_pass()`'s parameter in the prototype.
- Use `libmisc:` prefix in a commit message.

alejandro-colomar:
Changes: Remove CI bits (already handled in #596 )

alejandro-colomar:
Rebase to master

alejandro-colomar:
Changes: reword caveats. 


calling erase_pass() twice is actually much worse than calling free(3) twice.  It's calling memset(3) on freed memory.  That's critical undefined behavior.

alejandro-colomar:
I'm happy with the patch set now.  If it passes the autotests, please review again.

ikerexxe:
LGTM. Thanks for the patches!

@hallyn any opinion before I merge it?

libmisc/agetpass.c
```diff
@@ -0,0 +1,152 @@
...
+ * HOLDERS OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <config.h>
```
alejandro-colomar, line 29: @ikerexxe @cgzones @hallyn Is this include necessary?

guillemj: I'm not any of the highlighted, but on autotools-based projects that's required yes, so that any of the checks performed by configure apply here, including extensions such as LFS support and similar.

alejandro-colomar: > I'm not any of the highlighted,

I forgot to CC you.  Good that you were notified anyway :)

> but on autotools-based projects that's required yes, so that any of the checks performed by configure apply here, including extensions such as LFS support and similar.

Thanks!

It was a bit unintuitive to me that I'd need to include a header if I wan't using any of its definitions (since the build system usually finds the location of libbsd headers in projectsI use, and then it's transparent to me; you can guess I don't use autotools :p).

ikerexxe:
Moving this forward to stop blocking you from other work you are doing.

Thanks for your hard work!

alejandro-colomar:
> Moving this forward to stop blocking you from other work you are doing.
> 
> Thanks for your hard work!

Thanks!  I'll rebase the currently-open PRs.

Cheers,

Alex

PR #574, opened 2022-09-28
Don't test for NULL before calling free(3)

alejandro-colomar:
```
free(3) accepts NULL, since the oldest ISO C.  I guess the paranoid code was taking
care of prehistoric implementations of free(3).  I've never known of an implementation
that doesn't conform to this, so let's simplify this.

Remove xfree(3), which was effectively an equivalent of free(3).

```

I found this while merging `xmalloc()`+`str*cpy()` into `xstrdup()`.

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -85,9 +85,7 @@ int lrename (const char *old, const char *new)
 	res = rename (old, new);
 
 #ifdef __GLIBC__
-	if (NULL != r) {
-		free (r);
-	}
+	free (r);
 #endif				/* __GLIBC__ */
 
 	return res;
@@ -337,9 +335,7 @@ static void free_linked_list (struct commonio_db *db)
 		p = db->head;
 		db->head = p->next;
 
-		if (NULL != p->line) {
-			free (p->line);
-		}
+		free (p->line);
 
 		if (NULL != p->eptr) {
 			db->ops->free (p->eptr);
@@ -395,10 +391,8 @@ int commonio_lock_nowait (struct commonio_db *db, bool log)
 		err = 1;
 	}
 cleanup_ENOMEM:
-	if (file)
-		free(file);
-	if (lock)
-		free(lock);
+	free(file);
+	free(lock);
 	return err;
 }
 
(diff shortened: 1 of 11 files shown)


ikerexxe:
The code looks good to me but I'm a little bit preoccupied about the error found by CodeQL.

alejandro-colomar:
> The code looks good to me but I'm a little bit preoccupied about the error found by CodeQL.

Yeah, I have no idea how that's failing now, but I'll have a look in case there's a bug that has been uncovered by this trivial change.

However, there is clearly a bug in the static analyzer, since a no-op change like this shouldn't trigger any changes in the analysis.

alejandro-colomar:
I didn't check all paths that are being reported by CodeQL, but:

One of the reports is:
`This 'call to strcpy' with input from tzbuf may overflow the destination.`

Let's investigate it:

```c
$ grep -rnI -e strdup -e strcpy -e str.cpy | grep tzbuf
libmisc/tz.c:43:		strcpy (tzbuf, def_tz);
```

A single match; it should be easy to find out if the report is correct.
Let's print the whole function:

```c
/*@observer@*/const char *tz (const char *fname)
{
	FILE *fp = NULL;
	static char tzbuf[BUFSIZ];
	const char *def_tz = "TZ=CST6CDT";

	fp = fopen (fname, "r");
	if (   (NULL == fp)
	    || (fgets (tzbuf, (int) sizeof (tzbuf), fp) == NULL)) {
		def_tz = getdef_str ("ENV_TZ");
		if ((NULL == def_tz) || ('/' == def_tz[0])) {
			def_tz = "TZ=CST6CDT";
		}

		strcpy (tzbuf, def_tz);
	} else {
		tzbuf[strlen (tzbuf) - 1] = '\0';
	}

	if (NULL != fp) {
		(void) fclose (fp);
	}

	return tzbuf;
}
```

No entry in the table of definitions (`def_table`) is bigger than `BUFSIZ`, so at least this one is not a real bug.

ikerexxe:
> I wonder why this has been triggered now, if there's no call to anything `free(3)` in this function :/

Who knows. I'll merge the changes. Thanks for your patch.

hallyn:
Thanks - indeed the manpage also "guarantees" null safety

PR #576, opened 2022-10-05
chage: Fix regression in print_date

### LINKED ISSUE #564: Incorrect date format in chage.c
There has been a change in the chage.c/print_date function between 4.11.1 and 4.12.3, and the date format is « wrong » because of double % :

```
diff -c shadow-4.11.1/work/shadow-4.11.1/src/chage.c shadow-4.12.3/work/shadow-4.12.3/src/chage.c 
*** shadow-4.11.1/work/shadow-4.11.1/src/chage.c        2022-01-03 01:46:53.000000000 +0100
--- shadow-4.12.3/work/shadow-4.12.3/src/chage.c        2022-08-09 00:30:40.000000000 +0200
***************
*** 223,242 ****
  {
        struct tm *tp;
        char buf[80];
-       char format[80];
- 
-       if (iflg) {
-               (void) snprintf (format, 80, "%%Y-%%m-%%d");
-       }
-       else {
-               (void) snprintf (format, 80, "%%b %%d, %%Y");
-       }
  
        tp = gmtime (&date);
        if (NULL == tp) {
                (void) printf ("time_t: %lu\n", (unsigned long)date);
        } else {
!               (void) strftime (buf, sizeof buf, format, tp);
                (void) puts (buf);
        }
  }
--- 223,234 ----
  {
        struct tm *tp;
        char buf[80];
  
        tp = gmtime (&date);
        if (NULL == tp) {
                (void) printf ("time_t: %lu\n", (unsigned long)date);
        } else {
!               (void) strftime (buf, sizeof buf, iflg ? "%%Y-%%m-%%d" : "%%b %%d, %%Y", tp);
                (void) puts (buf);
        }
  }
```

Thus, when running chage -l <user>, we do not get the dates displayed but the format :

```
% chage -l <someuser>
Last password change                                    : %b %d, %Y
Password expires                                        : %b %d, %Y
```

Xiami2012:
Introduced by c6c8130db4319613a91dd07bbb845f6c33c5f79f .
After removing snprintf, the format string should get unescaped once.

Fixes #564

Reporter and patch author: DerMouse (github.com/DerMouse)

### DIFF
--- a/src/chage.c
+++ b/src/chage.c
@@ -228,7 +228,7 @@ static void print_date (time_t date)
 	if (NULL == tp) {
 		(void) printf ("time_t: %lu\n", (unsigned long)date);
 	} else {
-		(void) strftime (buf, sizeof buf, iflg ? "%%Y-%%m-%%d" : "%%b %%d, %%Y", tp);
+		(void) strftime (buf, sizeof buf, iflg ? "%Y-%m-%d" : "%b %d, %Y", tp);
 		(void) puts (buf);
 	}
 }


ikerexxe (approved):
LGTM. Thanks for the patch.

PR #577, opened 2022-10-05
usermod: report error if homedir does not exist

ikerexxe:
Report error if usermod asked for moving homedir and it does not exist.


hallyn:
I'm not necessarily opposed, but would you actually find this useful?

PR #578, opened 2022-10-06
lastlog: check for localtime() return value

ikerexxe:


src/lastlog.c
```diff
@@ -150,9 +150,12 @@ static void print_one (/*@null@*/const struct passwd *pw)
 
 	ll_time = ll.ll_time;
 	tm = localtime (&ll_time);
-	strftime (ptime, sizeof (ptime), "%a %b %e %H:%M:%S %z %Y", tm);
-	cp = ptime;
-
+	if (tm == NULL) {
+		cp = "(unknown)";
```
hallyn, line 154: Do you think it's worth doing _("(unknown)") ?

ikerexxe: I don't mind changing the message but we definitely need to set something, as `cp` will be printed several lines below.

hallyn: All right well I'll merge it as is for now, thanks.

PR #582, opened 2022-10-07
Raise limit for passwd and shadow entry length

ikerexxe:
Moreover, include checks to prevent writing entries longer than the length limit.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1422497


hallyn (approved):

ikerexxe:
As nobody has complained and it has been approved I'm merging this PR.

PR #586, opened 2022-10-18
Allow supplementary groups to be added via config file

AZaugg:
Allow supplementary groups to be set via the /etc/default/useradd config file. Allowing an administrator to set additonal groups via the GROUPS configurable and control the default behaviour of useradd.


hallyn:
Hm, convince me...

What's an example where I'd really want this, where every time in the future that I add a user I'll get those groups?  It seems more likely I'd want to have a list of groups which I wanted added to this next batch of student users I'm creating now, but then in two weeks when the other admin is adding a user to run some script, this GROUP entry will have been forgotten and now it has unintended access.

hallyn:
Well I guess it's optional and default-empty, so it's ok.  Thanks.

PR #588, opened 2022-10-23
useradd: Fix buffer overflow when using a prefix

dm0-:
Running `useradd --prefix ...` aborts on openSUSE due to their hardening flags with a `buffer overflow detected` message.  It looks like this buffer is one byte short.


ikerexxe:
LGTM but would prefer @hallyn to review it as well.

hallyn:
Sure is (a byte short) - thanks.

PR #590, opened 2022-11-07
Don't warn when failed to open /etc/nsswitch.conf

### LINKED ISSUE #557: Avoid nss warning/error with musl
The libnss configuration file is only installed when glibc is used. The inexistence of it on a musl-based rootfs, will make shadow complain
about it:

> Failed opening /etc/nsswitch.conf

This is because shadow will try to use `nsswich.conf` when dealing with subordinate IDs and the message is just a warning as the tool will still generate them correctly in subuid/subgid files. To avoid confusion and needless errors on a musl system, could we do the nss support glibc specific?

hallyn:
Maybe we should have a debug mode where it's still printed, but we don't, so let's be quieter.

Closes #557

### DIFF
--- a/lib/nss.c
+++ b/lib/nss.c
@@ -59,7 +59,6 @@ void nss_init(const char *nsswitch_path) {
 	//   subid:	files
 	nssfp = fopen(nsswitch_path, "r");
 	if (!nssfp) {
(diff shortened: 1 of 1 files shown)


ikerexxe:
LGTM.

PR #591, opened 2022-11-09
Add support for skeleton files from /usr/etc/skel

jubalh:
This patch is used by openSUSE to make useradd look for skeleton files in /usr/etc/skel additionally to /etc/skel in accordance with
https://uapi-group.org/specifications/specs/base_directory_specification/

### DIFF
--- a/libmisc/copydir.c
+++ b/libmisc/copydir.c
@@ -449,6 +449,14 @@ static int copy_entry (const struct path_info *src, const struct path_info *dst,
 		}
 
 		/*
+		 * If the destination already exists do nothing.
+		 * This is after the copy_dir above to still iterate into subdirectories.
+		 */
+		if (fstatat(dst->dirfd, dst->name, &sb, AT_SYMLINK_NOFOLLOW) != -1) {
+			return 0;
+		}
+
+		/*
 		 * Copy any symbolic links
 		 */
 
@@ -507,6 +515,7 @@ static int copy_dir (const struct path_info *src, const struct path_info *dst,
                      gid_t old_gid, gid_t new_gid)
 {
 	int err = 0;
+	struct stat dst_sb;
 
 	/*
 	 * Create a new target directory, make it owned by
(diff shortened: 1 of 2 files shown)


brauner (approved):

hallyn:
I don't understand - I looked at https://uapi-group.org/specifications/specs/base_directory_specification/ mentioned in the commit message, but don't see /usr/etc/skell mentioned there.  What I do see there is talk of masking ...

Can you quote the exact text?

Also, if the goal is for /usr/etc/skel to be overridable by local changes in /etc/skel, wouldn't you rather have DSKEL handled after DUSRSKEL?  Or does that not practically ever matter (or am I misunderstanding the goal altogether)?

jubalh:
>  I looked at https://uapi-group.org/specifications/specs/base_directory_specification/ 
Sorry that was a copy/paste fail. Actually I wanted to post `https://uapi-group.org` only. I'll update the commit (after more review).

The goal is to make it possible to separate config files provided by the distro vs provided by the admin. And help also with read only /etc.

I think https://en.opensuse.org/openSUSE:Packaging_UsrEtc might explain what is done.

hallyn:
Thanks.  I support separating out distributed from admin specified configuration, but again wouldn't you want to rsync from /usr/etc (distro provided) first and then from /etc (admin provided, local updates)?

hallyn:
> Thanks. I support separating out distributed from admin specified configuration, but again wouldn't you want to rsync from /usr/etc (distro provided) first and then from /etc (admin provided, local updates)?

Hi,

when you get a chance, I'm still waiting for a reply to this.  If I'm wrong on my guess, let me know and I'll merge as is.

Thanks.

jubalh:
Hi @hallyn !
Sorry for not responding, I'm on vacation right now and I promised myself not to check Emails etc :)
Yet here I am...
But only wanted to let you know that I'll take a look and answer in more detail when I'm back home.

hallyn:
> Hi @hallyn ! Sorry for not responding, I'm on vacation right now and I promised myself not to check Emails etc :) Yet here I am... But only wanted to let you know that I'll take a look and answer in more detail when I'm back home.

Thanks - sorry about interrupting your vacation!

jubalh:
Back now.
AFAIK currently we do that if  /etc/skel entry exists do nothing with /usr/etc/skel otherwise use later one. Unfortunately I wasn't around when the original maintainer wrote this patch and I don't have much other info about it.

hallyn (approved):

hallyn:
Note - would be great if someone had time to add testcases for this :)

pothos:
> Can you quote the exact text?

It's a bit hidden: _If `$XDG_CONFIG_DIRS` is either not set or empty, a value equal to `/etc:/usr/etc:/etc/xdg` should be used._

Vogtinator:
> AFAIK currently we do that if /etc/skel entry exists do nothing with /usr/etc/skel otherwise use later one. Unfortunately I wasn't around when the original maintainer wrote this patch and I don't have much other info about it.

Not entirely true - first it copies `/etc/skel` into the target home and then the content of `/usr/etc/skel`, skipping all files/directories that already exists. Effectively `/etc/skel` and `/usr/etc/skel` get merged, with `/etc/skel` having preference.

PR #592, opened 2022-11-17
useradd: check if subid range exists for user

ikerexxe:
According to useradd man page: If /etc/subuid exists, the commands useradd and newusers (**unless the user already have subordinate user IDs**) allocate SUB_UID_COUNT unused user IDs from the range SUB_UID_MIN to SUB_UID_MAX for each new user.

Thus, I've added a condition to check if a user already has a subid/subgid range before assigning one.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2012929


hallyn (approved):
Thanks, looks good.

Issue #561, opened 2022-08-27
`usermod -rG x y` while user `y` is not in group `x` will cause user `y` add into group `x`


flyxyz123:
`usermod -rG x y` while user `y` is not in group `x` will cause user `y` add into group `x`

For better testing, I wrote a script `t` to get the group informations: (`xyz` is my username)
```sh
#!/bin/sh
# script t
id xyz
printf '\n'
getent group games
printf '\n'
grep games /etc/group
printf '\n'
sudo grep games /etc/gshadow
printf '\n'
grep xyz /etc/group
printf '\n'
sudo grep xyz /etc/gshadow
```

Here's the initial state:
```
$ t
uid=1000(xyz) gid=998(wheel) groups=998(wheel)

games:x:50:

games:x:50:

games:!*::


```

Then I run `sudo usermod -rG games xyz`
```
$ sudo usermod -rG games xyz
$ t
uid=1000(xyz) gid=998(wheel) groups=50(games),998(wheel)

games:x:50:xyz

games:x:50:xyz

games:!*::xyz

games:x:50:xyz

games:!*::xyz
```

Expected output of script `t` should be the same as the initial state. Actual output of script `t` shows that user `xyz` is added to group `games`.

User `xyz` should not be added to group `games` because `-r` flag of `usermod` is to "Remove the user from named supplementary group(s)." (quote from man page).

# My fix for this issue see #560

flyxyz123:
# I'm not familiar with the code, here are my understanding of the code but maybe WRONG!

I find that these if condition code causes the problem:
<https://github.com/shadow-maint/shadow/blob/7136e24f8a4a541dbe0ddd24f5a6b2de583ebb65/src/usermod.c#L715-L721>
<https://github.com/shadow-maint/shadow/blob/7136e24f8a4a541dbe0ddd24f5a6b2de583ebb65/src/usermod.c#L838-L844>

There are four states I can think of for now
```
was_number  is_number initial   is_number after actual  is_number after expected
0           0                   0                       0
0           1                   1                       0 and/or error
1           0                   1                       1
1           1                   0                       0
```

`is_number initial` means the initialization of variable `is_number` here:
<https://github.com/shadow-maint/shadow/blob/7136e24f8a4a541dbe0ddd24f5a6b2de583ebb65/src/usermod.c#L708-L709>
[...]

PR #593, opened 2022-11-17
Add regression test coverage for 'usermod -rG' (supplemental group removal)

jayaddison:
This attempts to add regression test coverage relating to #561.

## Test info

The configuration here creates two users, `foo` and `bar`:

- Both belong to `users`
- `foo` initially belongs to `floppy`
- `bar` initially belongs to `fax`

The test case uses `usermod -rG` twice, attempting to remove `foo` from _both_ `floppy` (valid) and from `fax` (invalid).

Afterwards the assertions check the group/user state, including that:

- Both users belong to `users`
- `foo` does not belong to any other groups
- `bar` continues to belong to the `fax` group

## Status

:warning: I have managed to get this to succeed locally in a `chroot` using `usermod` from `shadow` v4.13 (with the fix from #560 applied), but I have **not** yet run confirmed that the regression test fails when running with a version where the bug existed.

If anyone can help out with doing that, it'd be much appreciated!


hallyn:
Please add it also to tests/run_some , as that is the set that CI runs.

Then once that's done, we can just open another draft pr where the fixing commit gets reverted, to verify that it correctly fails.

jayaddison:
Thanks @hallyn - that's done in #594 (draft pr), and I think that that demonstrates the expected test failure (please verify, though).

hallyn:
> 

Sorry I didn't notice 594.  Github tells me that you deleted issue-561/verify-test-coverage, and I don't see the test results, only shellcheck results.  Could you re-push that branch?

jayaddison:
Yep - done; (another) empty commit required to get the continuous integration to run again - pushing that in a few moments.

PR #594, opened 2022-11-18
Draft: verify regression test coverage added in #593

jayaddison:
This branches from 3e3c1801ca7d467714b25475d4cedbcf56e8f11d, with the following changes:

- Adds the additional test case to the `run_some` test suite script so that the test is run within continuous integration
- Reverts the fix for #561 to attempt to prove that the (regression) test coverage does catch the original issue


hallyn:
```
###############################################################################
#
# Status of test ./usertools/62_usermod_remove_supplementary_groups/usermod.test: FAILURE
#
###############################################################################
+ false
```

Thank you!

jayaddison:
Thanks @hallyn!

PR #595, opened 2022-11-21
Fix HAVE_SHADOWGRP configure check

fweimer-rh:
The missing #include <gshadow.h> causes the configure check to fail spuriously, resulting in HAVE_SHADOWGRP not being defined even on systems that actually have sgetsgent (such as current glibc).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -116,6 +116,10 @@ if test "$ac_cv_header_shadow_h" = "yes"; then
 		ac_cv_libc_shadowgrp,
 		AC_RUN_IFELSE([AC_LANG_SOURCE([
 				#include <shadow.h>
+				#ifdef HAVE_GSHADOW_H
+				#include <gshadow.h>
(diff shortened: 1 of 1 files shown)


ikerexxe:
Florian is this affecting Fedora somehow?

@hallyn any objection to merge this?

ikerexxe (approved):

hallyn (approved):
Oh, interesting.  Thanks.

alejandro-colomar:
@fweimer-rh I don't find sgetsgent documented (neither man-pages nor glibc manual).  Is there any documentation for it?

alejandro-colomar:
Hmmm; thanks!

BTW, do you know if (and since when) musl supports it?  Can we rely on it and remove our fallback implementation?

BTW, if you would be able to write some manual page for it, it'd be great.  Even just for the HISTORY section it would be quite interesting.

PR #596, opened 2022-11-22
CI: add libbsd and pkg-config dependencies

ikerexxe:


.builds/ubuntu-focal.yml
```diff
@@ -3,11 +3,14 @@ packages:
   - automake
   - autopoint
   - xsltproc
+  - libbsd0
```
guillemj, line 6: Please do not add libbsd0, as that is an implementation detail of the ABI which is not relevant during builds. The libbsd-dev build-dependency is enough and will make sure to pull in whatever specific shared library is required.

.builds/ubuntu-jammy.yml
```diff
@@ -3,11 +3,14 @@ packages:
   - automake
   - autopoint
   - xsltproc
+  - libbsd0
```
guillemj, line 6: Ditto here.

.github/actions/install-dependencies/action.yml
```diff
@@ -6,7 +6,7 @@ runs:
   - shell: bash
     run: |
       sudo apt-get update -y
-      sudo apt-get install -y ubuntu-dev-tools
+      sudo apt-get install -y ubuntu-dev-tools libbsd0 libbsd-dev
```
guillemj, line 9: And here.

guillemj (changes requested):

.builds/ubuntu-focal.yml
ikerexxe: Removed `libbsd0` as suggested

hallyn (approved):

hallyn:
Looks fine, but were there PRs breaking due to this?  (I'm not seeing them in the Actions view)

ikerexxe:
Sorry, I forgot to mention that this patch fixes the issues found in https://github.com/shadow-maint/shadow/pull/573

alejandro-colomar:
I think there are a few files that should also get these changes, right?:

.builds/alpine-edge.yml
.builds/fedora.yml

ikerexxe:
No need for that as they are already being installed by default or pulled from another dependency.

alejandro-colomar:
> No need for that as they are already being installed by default or pulled from another dependency.

Makes sense.  I had some changes about it in my PR, so I've removed them just to make sure that it still works without them.  I'll now pick your changes to my PR and see if it works fine.

alejandro-colomar:
It failed in fedora and alpine, and passed in all others, so I guess you need to add that.

alejandro-colomar:
The following seems to work in the other PR:

```diff
$ git show
commit 85aa8209f265cb32d17acfc90462325a1f49d91f (HEAD -> agetpass, gh/agetpass)
Author: Alejandro Colomar <alx@kernel.org>
Date:   Wed Nov 23 13:22:29 2022 +0100

    test

diff --git a/.builds/alpine-edge.yml b/.builds/alpine-edge.yml
index ec5355bb..6ab4d7c6 100644
--- a/.builds/alpine-edge.yml
+++ b/.builds/alpine-edge.yml
@@ -9,9 +9,11 @@ packages:
   - gettext
   - gettext-dev
   - gettext-lang
+  - libbsd-dev
   - libcap-dev
   - libtool
   - linux-pam-dev
+  - pkgconfig
   - sed
 sources:
   - https://github.com/shadow-maint/shadow
diff --git a/.builds/fedora.yml b/.builds/fedora.yml
index 9c426631..522c4611 100644
--- a/.builds/fedora.yml
+++ b/.builds/fedora.yml
@@ -8,10 +8,12 @@ packages:
   - gettext
   - gettext-devel
   - git
+  - libbsd-devel
   - libselinux-devel
   - libsemanage-devel
   - libtool
   - libxslt
+  - pkg-config
 sources:
   - https://github.com/shadow-maint/shadow
 tasks:
```

alejandro-colomar:
If you add that to the patch, please append:

```
Co-developed-by: Alejandro Colomar <alx@kernel.org>
```

ikerexxe:
Thanks for the patch Alejandro

alejandro-colomar:
LGTM.  As soon as you merge, I'll do some final tweaks to my patches, and we'll have the new `agetpass()` :)

alejandro-colomar:
Actually, I've been told that the future of pkg-config is pkgconf.  pkf-config is only provided for backwards compatibility.

So, we can use `pkgconf` in all systems (unless any of them starts failing).

ikerexxe:
> I've been told that the future of pkg-config is pkgconf. pkg-config is only provided for backwards compatibility.
> 
> So, we can use `pkgconf` in all systems.
> 
> I believe it's available in the 4 CI systems we have.

Yes, that's working properly. Thanks!

alejandro-colomar:
Great!

Issue #600, opened 2022-12-02
Raise minimum ISO C to C99, or even further to POSIX.1-2001 (or 2008) and C11


alejandro-colomar:
Hi!

Supporting C89 is burdensome.  I did a lot of cleanup in the last year dropping support for pre-ISO C, which we can assume to be dead.  However, there are still lots of code (and configure code too) for supporting C89, like for example conditionals on having <stdbool.h>.  Since these days all compilers support C99, would you agree that we can raise the bar to ISO C99, and fail to compile on anything older than that?

Even the oldest Debian and CentOS that are on extended LTS have partial support for ISO C11, and full support for ISO C99 (the oldest compiler alive on those systems is some gcc 4.x, as I checked some time ago).

Moreover, ISO C23 will make `true` and `false` be predefined macros (no need to include `<stdbool.h>` for having them), so most pre-C99 code that defines them will turn Undefined Behavior.  Why is it UB?  Redefining macros is valid in C, except that the redefinitions must be identical (excluding comments and whitespace); since we wrap them in parentheses and glibc doesn't, our redefinition is invalid.  We could address this by wrapping our redefinitions in yet more ifdef preprocessor, but I'd rather burn it all, and require C99.

Would you like a round of cleanup based on this?

Cheers,
Alex

ikerexxe:
I like this idea, and I think it would be good to move forward to more actual standards.

I'd like to ask something though, what's your exact proposal? Move to ISO C99 and POSIX.1-2001? I don't see any issue for the latter. I'm not implying that you should open one, as we could perfectly use this issue for both efforts. Is there any additional work that you have in mind? I'd like to have a clear idea before starting to review anything in this area.

alejandro-colomar:
Hi Iker!

> I like this idea, and I think it would be good to move forward to more actual standards.
> 
> I'd like to ask something though, what's your exact proposal? Move to ISO C99 and POSIX.1-2001? I don't see any issue for the latter. I'm not implying that you should open one, as we could perfectly use this issue for both efforts. Is there any additional work that you have in mind? I'd like to have a clear idea before starting to review anything in this area.

Heh, I didn't even had POSIX.1-2001 in mind when I opened this issue.  It was while doing the C99 cleanup when I realised that POSIX.1-2001 would bring a much greater cleanup, so I did it.  The next standards are:

[...]

PR #601, opened 2022-12-02
Raise the bar to ISO C99 (drop support for C89)

alejandro-colomar:
<https://github.com/shadow-maint/shadow/issues/600>

### DIFF
--- a/lib/defines.h
+++ b/lib/defines.h
@@ -18,15 +18,7 @@ extern char * textdomain (const char * domainname);
 # define ngettext(Msgid1, Msgid2, N) \
     ((N) == 1 ? (const char *) (Msgid1) : (const char *) (Msgid2))
 #else
-#ifdef HAVE_LOCALE_H
-# include <locale.h>
-#else
-# undef setlocale
-# define setlocale(category, locale)	(NULL)
-# ifndef LC_ALL
-#  define LC_ALL	6
-# endif
-#endif
+#include <locale.h>
 
 #define gettext_noop(String) (String)
 /* #define gettext_def(String) "#define String" */
--- a/lib/defines.h
+++ b/lib/defines.h
@@ -48,9 +48,7 @@ extern char * textdomain (const char * domainname);
 #include <stdlib.h>
 #include <string.h>
 
-#if HAVE_ERRNO_H
-# include <errno.h>
-#endif
+#include <errno.h>
 
 #include <sys/stat.h>
 #include <sys/types.h>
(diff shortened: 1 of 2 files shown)


alejandro-colomar:
I think this is all for C99.

After this is merged, I'll suggest raising the bar even further, to POSIX.1-2001.

alejandro-colomar:
Rebased to master.

lib/defines.h
```diff
@@ -250,7 +250,6 @@ static inline void memzero(void *ptr, size_t size)
 #ifdef sun			/* hacks for compiling on SunOS */
 # ifndef SOLARIS
 extern char *strdup ();
-extern char *strerror ();
```
ikerexxe, line 278: `configure.ac` also contains a reference to `strerror()` which I believe can be removed.

ikerexxe (commented):
Apart from a little comment inline, I'd also like to ask you to add the reference to the issue fixed in the commit messages. Something like:
```
Resolves: https://github.com/shadow-maint/shadow/issues/600
```

lib/defines.h
alejandro-colomar: Thanks!

alejandro-colomar:
> Apart from a little comment inline, I'd also like to ask you to add the reference to the issue fixed in the commit messages. Something like:
> 
> ```
> Resolves: https://github.com/shadow-maint/shadow/issues/600
> ```

Yep, I think that would be `Closes:`.  Although since the issue really covers more than just C99, I think I'll just use `Link:`.  Are you happy with it?

alejandro-colomar:
Changes:

- Remove reference to strerror(3) in configure.ac.  [Iker]

alejandro-colomar:
Okay, now it should be done.

Changes:

- Also remove checks for 3 other functions required by C99 (and in some cases even C89).

ikerexxe:
> Yep, I think that would be conventionally `Closes:`. Although since the issue really covers more than just C99, I think I'll just use `Link:`. Are you happy with it?

As long as you reference the issue I'm fine.

LGTM. Thanks for the patches!

alejandro-colomar:
:-)

alejandro-colomar:
I'm happy to help.  And BTW, I learnt a lot with these patches.  Just FYI, I started rewriting some of the Linux man-pages about strings, and plan a new `string(7)` manual page that details all known string handling functions (and some more that I've only seen customly written in high-performance projects such as nginx), and add details when you should use each of them.

As a starter, I already killed `strncpy(3)`, in the hope that nobody ever uses it again ;)
<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/man3/strncpy.3#n26>

alejandro-colomar:
Great! :-)

PR #605, opened 2022-12-06
changing lock mechanism

edneville:
Systems can suffer power interruptions whilst .lock files are in /etc, preventing scripts and other automation tools from updating shadow's files which persist across boots.

This commit replaces that mechanism with file locking to avoid problems of power interruption/crashing.

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -32,8 +32,6 @@
 
 /* local function prototypes */
 static int lrename (const char *, const char *);
-static int check_link_count (const char *file, bool log);
-static int do_lock_file (const char *file, const char *lock, bool log);
 static /*@null@*/ /*@dependent@*/FILE *fopen_set_perms (
 	const char *name,
 	const char *mode,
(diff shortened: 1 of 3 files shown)


hallyn:
I like the idea, but I'm worried about filesystem support.  For instance http://0pointer.net/blog/projects/locking.html raises concerns about some NFS servers not supporting it.  I don't know if there are any other filesystems which are problematic.

lib/commonio.c
```diff
@@ -361,6 +215,31 @@ bool commonio_present (const struct commonio_db *db)
...
+	fd = open (file, O_CREAT | O_WRONLY, 0600);
+	if (-1 == fd) {
+		if (log) {
+				(void) fprintf (shadow_logfd, "%s: %s: %s\n",
+				                shadow_progname, file, strerror (errno));
+		}
+		return 0;
+	}
+
+	lck.l_type = F_WRLCK | F_RDLCK;
```
cgzones, line 237: why first set `l_type` to the passed parameter and then override before usage?

lib/commonio.c
```diff
@@ -361,6 +215,31 @@ bool commonio_present (const struct commonio_db *db)
...
+{
+	int fd;
+	struct flock lck = {
+		.l_type = type,
+		.l_whence = SEEK_SET,
+		.l_start = 0,
+		.l_len = 0,
+	};
+
+	fd = open (file, O_CREAT | O_WRONLY, 0600);
```
cgzones, line 228: Shouldn't the file to lock exist already?

lib/commonio.c
```diff
@@ -385,7 +264,8 @@ int commonio_lock_nowait (struct commonio_db *db, bool log)
 	snprintf (file, file_len, "%s.%lu",
 	          db->filename, (unsigned long) getpid ());
 	snprintf (lock, lock_file_len, "%s.lock", db->filename);
```
cgzones, line 266: The `lock` variable now seems unused.

edneville:
@brauner, thanks for taking a look. I appreciate the review. What are your thoughts overall, would switching from current lock flag file to file locking present more problems, or do you see it helping to avoid race and stale issues?

PR #606, opened 2022-12-15
Fix useradd audit event logging of ID field

jubalh:
When useradd sends its ADD_USER event, it is filling in the id field. This is not yet written to disk. When auditd sees the event and the log format is enriched, auditd tries to lookup the user name but it does not exist. This causes the event to never be resolvable since ausearch relies on the lookup information attached by auditd.

The fix is to not send the id information for any event until after close_files() is called. Just the acct field is all that is

Patch by Steve Grubb (afaik).

Reported at https://bugzilla.redhat.com/show_bug.cgi?id=1713432

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2225,9 +2225,14 @@ static void usr_update (unsigned long subuid_count, unsigned long subgid_count)
 #endif				/* ENABLE_SUBIDS */
 
 #ifdef WITH_AUDIT
+	/*
+	 * Even though we have the ID of the user, we won't send it now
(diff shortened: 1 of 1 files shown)


hallyn:
lgtm, thanks.  It looks like you're posting the patch as verbatim as possible, so you may not want to change this, but "When useradd sends it's ADD_USER"  s/it's/its/ did catch my eye...  But like i say if you don't want to change that, I understand.

jubalh:
Corrected

hallyn:
Thanks

PR #607, opened 2022-12-16
Groupmod fixes

alejandro-colomar:
This is a preparation patch before some more fixes.  I found some problems in this file, which are not yet addressed by these fixes.  This is only cleaning up so that the more important fixes are more readable.

@hallyn  @ikerexxe  I need to add functions that are usable in the entire project; we've seen that some parts of the project don't use libmisc, right?  In which library should I put them?  Or can we make the entire project depend on libmisc?

### DIFF
--- a/src/groupmod.c
+++ b/src/groupmod.c
@@ -578,11 +578,11 @@ static void prepare_failure_reports (void)
 	                 "group %s/%ju", group_name, (uintmax_t) group_id);
 #ifdef	SHADOWGRP
 	(void) snprintf (info_gshadow.action,
-	                 512 - strlen (info_group.audit_msg),
+	                 512 - strlen (info_gshadow.audit_msg),
 	                 "group %s", group_name);
 #endif
 	(void) snprintf (info_passwd.action,
-	                 512 - strlen (info_group.audit_msg),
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
Changes: added link to PR in commit message

ikerexxe (approved):
LGTM

> @hallyn @ikerexxe I need to add functions that are usable in the entire project; we've seen that some parts of the project don't use libmisc, right? In which library should I put them? Or can we make the entire project depend on libmisc?

I think it's fine to add those general functions in libmisc and make the entire project depend on this library.

alejandro-colomar:
> LGTM

Thanks!

> 
> > @hallyn @ikerexxe I need to add functions that are usable in the entire project; we've seen that some parts of the project don't use libmisc, right? In which library should I put them? Or can we make the entire project depend on libmisc?
> 
> I think it's fine to add those general functions in libmisc and make the entire project depend on this library.

Could you please do that step of making libmisc globally available?  I have no idea of the organization of the project.  I'll start writing patches that don't need that, while I wait for that to be done.

ikerexxe:
Unfortunately, I don't think I'll have time this year, but I'll try to accommodate some time.

PR #609, opened 2022-12-16
Use strlcpy(3) instead of its pattern

alejandro-colomar:
-  Since strncpy(3) is not designed to write strings, but rather
   (null-padded) character sequences (a.k.a. unterminated strings), we
   had to manually append a '\0'.  strlcpy(3) creates strings, so they
   are always terminated.  This removes dependencies between lines, and
   also removes chances of accidents.

-  Repurposing strncpy(3) to create strings requires calculating the
   location of the terminating null byte, which involves a '-1'
   calculation.  This is a source of off-by-one bugs.  The new code has
   no '-1' calculations, so there's almost-zero chance of these bugs.

-  strlcpy(3) doesn't padd with null bytes.  Padding is relevant when
   writing fixed-width buffers to binary files, or when sending them
   to other processes or through the network.  This is not the case,
   so padding is effectively ignored.

-  strlcpy(3) requires that the input string is really a string;
   otherwise it crashes (SIGSEGV).  Let's check if the input strings are
   really strings:

   -  contrib/pwdauth.c:
      -  struct passwd::pw_name is a string (see getpwent(3)).

   -  lib/fields.c:
      -  'cp' was assigned from 'newft', and 'newft' comes from fgets(3).

   -  lib/gshadow.c:
      -  strlen(string) is calculated a few lines above.

   -  libmisc/console.c:
      -  'cons' comes from getdef_str, which is a bit cryptic, but seems
         to generate strings, I guess.1

   -  libmisc/date_to_str.c:
      -  It receives a string literal.  :)

   -  libmisc/utmp.c:
      -  'tname' comes from ttyname(3), which returns a string.

   -  src/su.c:
      -  'tmp_name' has been passed to strcmp(3) a few lines above.

### DIFF
--- a/src/su.c
+++ b/src/su.c
@@ -653,8 +653,7 @@ static /*@only@*/struct passwd * check_perms (void)
 		SYSLOG ((LOG_INFO,
 		         "Change user from '%s' to '%s' as requested by PAM",
 		         name, tmp_name));
-		strncpy (name, tmp_name, sizeof(name) - 1);
-		name[sizeof(name) - 1] = '\0';
+		strlcpy (name, tmp_name, sizeof(name));
 		pw = xgetpwnam (name);
 		if (NULL == pw) {
 			(void) fprintf (stderr,
--- a/lib/fields.c
+++ b/lib/fields.c
@@ -100,8 +100,7 @@ void change_field (char *buf, size_t maxsize, const char *prompt)
 			cp++;
 		}
 
-		strncpy (buf, cp, maxsize - 1);
-		buf[maxsize - 1] = '\0';
+		strlcpy (buf, cp, maxsize);
 	}
 }
 
--- a/lib/gshadow.c
+++ b/lib/gshadow.c
@@ -15,6 +15,7 @@
 #ident "$Id$"
 
 #include <stdio.h>
+#include <string.h>
 #include "prototypes.h"
 #include "defines.h"
 static /*@null@*/FILE *shadow;
@@ -124,8 +125,7 @@ void endsgent (void)
 		sgrbuflen = len;
 	}
 
-	strncpy (sgrbuf, string, len);
-	sgrbuf[len-1] = '\0';
+	strlcpy (sgrbuf, string, len);
 
 	cp = strrchr (sgrbuf, '\n');
 	if (NULL != cp) {
--- a/libmisc/console.c
+++ b/libmisc/console.c
@@ -44,8 +44,7 @@ static bool is_listed (const char *cfgin, const char *tty, bool def)
 
 	if (*cons != '/') {
 		char *pbuf;
-		strncpy (buf, cons, sizeof (buf));
-		buf[sizeof (buf) - 1] = '\0';
+		strlcpy (buf, cons, sizeof (buf));
 		pbuf = &buf[0];
 		while ((s = strtok (pbuf, ":")) != NULL) {
 			if (strcmp (s, tty) == 0) {
--- a/libmisc/date_to_str.c
+++ b/libmisc/date_to_str.c
@@ -38,9 +38,10 @@ void date_to_str (size_t size, char buf[size], long date)
 	time_t t;
 
 	t = date;
-	if (date < 0)
-		(void) strncpy (buf, "never", size);
-	else
+	if (date < 0) {
+		(void) strlcpy (buf, "never", size);
+	} else {
 		(void) strftime (buf, size, "%Y-%m-%d", gmtime (&t));
-	buf[size - 1] = '\0';
+		buf[size - 1] = '\0';
+	}
 }
--- a/libmisc/utmp.c
+++ b/libmisc/utmp.c
@@ -40,10 +40,8 @@ static bool is_my_tty (const char *tty)
 
 	if ('\0' == tmptty[0]) {
 		const char *tname = ttyname (STDIN_FILENO);
-		if (NULL != tname) {
-			(void) strncpy (tmptty, tname, sizeof tmptty);
-			tmptty[sizeof (tmptty) - 1] = '\0';
-		}
+		if (NULL != tname)
+			(void) strlcpy (tmptty, tname, sizeof tmptty);
 	}
 
 	if ('\0' == tmptty[0]) {


alejandro-colomar:
I decided to split the string changes in smaller PRs, so it's easier to review each one.

hallyn:
I'm sorry, I know I'm being slow, but I'd really like to review all string parsing patches closely myself.  I'll aim to get to this by EOD Monday.

ikerexxe (commented):
Some general remarks. `strlcpy` returns the total length of the new string, so I think this should be checked. I know it wasn't done before, but as we are improving the code I think it would be good.

On top of that, you changed `pwdauth.c` file, do you know what's its purpose? I mean, I don't see this binary in my distribution and it isn't in use by any test (that's why it's in `contrib` folder, right?). I know you didn't add this file but as I was reviewing it I got curious.

alejandro-colomar:
> Some general remarks. `strlcpy` returns the total length of the new string, so I think this should be checked. I know it wasn't done before, but as we are improving the code I think it would be good.

Yes, `strlcpy(3)` return value should almost always be checked (it's only possibly irrelevant when logging).  However, I do not know what's the appropriate action for such cases (which is why I intended to do it in a different PR).  Should we just abort?  If so, I'd add a function `xstrlcpy()` similar to `xmalloc()`, and let it die on truncation.  Another option is to report the truncation,. but that's only useful for user input.  In case of a programming error, I think we should just abort.

> 
> On top of that, you changed `pwdauth.c` file, do you know what's its purpose? I mean, I don't see this binary in my distribution and it isn't in use by any test (that's why it's in `contrib` folder, right?). I know you didn't add this file but as I was reviewing it I got curious.

I know the same as you :)  I guess it's some old file that was used in the past.

```sh
alx@debian:~$ apt-file find pwdauth
alx@debian:~$
```

It seems to be extinct.

How about removing extinct files?  They are in the git history anyway, so if someone wants to recover them, they just need to dig in the git history.  It would remove a few thousands of LoC I guess :)

hallyn:
> BTW, I still haven't received the phone I "bought" 2+ years ago, so I guess I can wait a few > days for a review ;)

Oh, you too are waiting for a pro1x?

alejandro-colomar:
> > BTW, I still haven't received the phone I "bought" 2+ years ago, so I guess I can wait a few > days for a review ;)
> 
> Oh, you too are waiting for a pro1x?

Yep :)
<https://s3hh.wordpress.com/2020/09/07/sxmo-on-pinephone/#comment-7201>
I already forgot when I ordered it...

ikerexxe:
> Yes, `strlcpy(3)` return value should almost always be checked (it's only possibly irrelevant when logging). However, I do not know what's the appropriate action for such cases (which is why I intended to do it in a different PR). Should we just abort? If so, I'd add a function `xstrlcpy()` similar to `xmalloc()`, and let it die on truncation. Another option is to report the truncation,. but that's only useful for user input. In case of a programming error, I think we should just abort. (However, I have concerns about crashing in the middle of a shadow program, since a password might end up in a core dump...)

Following the example of `xmalloc()` seems to me to be the most sensible idea.

> How about removing extinct files? They are in the git history anyway, so if someone wants to recover them, they just need to dig in the git history. It would remove a few thousands of LoC I guess :)

I like this idea, especially if we don't use these lines for anything useful.

alejandro-colomar:
@hallyn 

There's one thing which I'm not 100% certain of it.  I guess we don't need padding null bytes after the terminating null byte.  However, I think that some systems require them in some functions (probably those that write things like utmp structures to files.  You'll probably know better than me when we require that and when we don't.  I'd like you to be especially careful about those when reviewing.  Thanks! :)

alejandro-colomar:
> > Yes, `strlcpy(3)` return value should almost always be checked (it's only possibly irrelevant when logging). However, I do not know what's the appropriate action for such cases (which is why I intended to do it in a different PR). Should we just abort? If so, I'd add a function `xstrlcpy()` similar to `xmalloc()`, and let it die on truncation. Another option is to report the truncation,. but that's only useful for user input. In case of a programming error, I think we should just abort. (However, I have concerns about crashing in the middle of a shadow program, since a password might end up in a core dump...)
> 
> Following the example of `xmalloc()` seems to me to be the most sensible idea.

Okay.  But I'll do that in a separate PR.  I prefer to first make sure this small patch is safe, and then send another one.  I'm a bit worried about these changes.

> 
> > How about removing extinct files? They are in the git history anyway, so if someone wants to recover them, they just need to dig in the git history. It would remove a few thousands of LoC I guess :)
> 
> I like this idea, especially if we don't use these lines for anything useful.

alejandro-colomar:
Rebased to master

hallyn:
I see, strlcpy comes in with libbsd-dev.  I was trying to figure out why I didn't have the manpage here.

hallyn:
> @hallyn
> 
> There's one thing which I'm not 100% certain of it. I guess we don't need padding null bytes after the terminating null byte. However, I think that some systems require them in some functions that accept fixed-width buffers (probably those that write things like utmp structures to files). You'll probably know better than me when we require that and when we don't. I'd like you to be especially careful about those when reviewing. Thanks! :)

Well honestly there is nothing performance sensitive about any of these, so starting with a memset(dst, 0, sizeof(dst)) would be fine.

src/su.c
```diff
@@ -655,8 +655,7 @@ static /*@only@*/struct passwd * check_perms (void)
 		SYSLOG ((LOG_INFO,
 		         "Change user from '%s' to '%s' as requested by PAM",
 		         name, tmp_name));
-		strncpy (name, tmp_name, sizeof(name) - 1);
```
hallyn, line 658: So you're right in the patch description that

> tmp_name' has been passed to strcmp(3) a few lines above.

however that could have been counting on $name being null terminated, not $tmp_name.

Now I find it hard to believe that pam_get_item would return a non-null-terminated string with PAM_SUCCESS, however I :sadface: notice that the manpage declines to guarantee that for us :(

alejandro-colomar: Taking into account that there's no size parameter, it necessarily is a string.  Otherwise, it would need to document some kind of size that the user needs to use to read from the pointer, to avoid overrunning the buffer.

I guess they didn't document it believing it would be obvious...  I'd prefer it documented though.  Just saying that it's a string would be enough :/

alejandro-colomar: BTW, `strcmp(3)` requires that both arguments are strings.  If one of them is not a string, then `strncmp(3)` should be called.

alejandro-colomar:
> > @hallyn
> > There's one thing which I'm not 100% certain of it. I guess we don't need padding null bytes after the terminating null byte. However, I think that some systems require them in some functions that accept fixed-width buffers (probably those that write things like utmp structures to files). You'll probably know better than me when we require that and when we don't. I'd like you to be especially careful about those when reviewing. Thanks! :)
> 
> Well honestly there is nothing performance sensitive about any of these, so starting with a memset(dst, 0, sizeof(dst)) would be fine.

I would very much prefer not to call bzero(3) (or memset(3)) unless necessary.  The main issue is readability.  If I read some call to bzero(3), I default to believe it is necessary, and try to understand the reason.  Zeroing just because I'm not sure how libc works, makes me believe that the solution would be to make sure that libc documents its behavior.

Since I mentioned Jakub in another thread today, I guess I can abuse a little bit of him (and a few more people) and ask...

@jakubjelinek @fweimer @zatrazz Is there any libc function (or syscall wrapper) that expects a null-padded fixed-width character sequence (I'm talking about things like `struct utmp.ut_user`) where padding null bytes (I'm not talking about the terminating null byte when the character sequence is shorter than the buffer, but null bytes _after_ the terminating one) are significant?

I'll quote the utmp(5) manual page here:

```
STANDARDS
       POSIX.1 does not specify a utmp structure, but rather one  named  utmpx
       (as  part  of  the  XSI  extension), with specifications for the fields
       ut_type, ut_pid, ut_line, ut_id, ut_user, and ut_tv.  POSIX.1 does  not
       specify the lengths of the ut_line and ut_user fields.

       Linux defines the utmpx structure to be the same as the utmp structure.

   Comparison with historical systems
       ...

       In Linux (as in System V), the ut_id  field  of  a  record  will  never
       change once it has been set, which reserves that slot without needing a
       configuration file.  Clearing ut_id may result in race conditions lead‐
       ing  to  corrupted utmp entries and potential security holes.  Clearing
       the abovementioned fields by filling them with null bytes  is  not  re‐
       quired  by  System  V semantics, but makes it possible to run many pro‐
       grams which assume BSD semantics and which do not modify  utmp.   Linux
       uses the BSD conventions for line contents, as documented above.

       ...
```

I'd like to know, not only about utmp(5), but about any API in glibc, if you know.  I'm guessing padding nulls won't be significant, but we better make sure.  Also, it would be nice if you know all of the APIs that have meaningful padding in other systems; I'd like to document them more clearly; probably in CAVEATS.

Thanks!

hallyn (approved):

alejandro-colomar:
Rebased to master.

hallyn:
Have looked a bit more at all the callers, am convinced this is fine.

Thank you

alejandro-colomar:
Thanks :)

alejandro-colomar:
> > @jakubjelinek @fweimer @zatrazz Is there any libc function (or syscall wrapper) that expects a null-padded fixed-width character sequence (I'm talking about things like `struct utmp.ut_user`) where padding null bytes (I'm not talking about the terminating null byte when the character sequence is shorter than the buffer, but null bytes _after_ the terminating one) are significant?
> 
> @alejandro-colomar Not sure if I understand the question. Why would bytes after the object, outside its boundaries, be significant?

Not outside the object, but outside the string.

> 
> There are some struct fields where it is not entirely clear (particularly across operating systems) whether a fixed-length string usually used with `strncpy` must include a terminating null byte or not (that is, if the maximum string length is one less than the size of the byte array, or exactly the array size).

Yep, that's the kind of information I'd like to know.  I'd like to document in the manual pages what kind of promises/restrictions are imposed on GNU/Linux systems, and what kind of promises one would get in other Unix systems (well, I care only about common free systems mostly, which restricts the search to the BSDs basically; nonfree systems are out of luck).

So, for the previous paragraph where you had a doubt, let's imagine the following case:

```c
char array[11] = "foo\0garbage";
char array[11] = "foo";  // filled with '\0'
```

For the fixed-width buffers you mentioned (e.g., utmp(5)), would both contents be equally valid?  In which systems would the garbage be something to avoid?

Thanks!

alejandro-colomar:
@fweimer Who might know?  I'd like to know if the glibc APIs read the entire array, or if they just read until the first NUL and then fill with null bytes even if the buffer contained garbagge.

alejandro-colomar:
> I don't think you can expect implementations to clear the padding before writing the data to disk.

Thanks.  I guess we can do the following to be safe:

-  Never malloc(3), but rather calloc(3) all utmp structures.
-  Always zero those structures between uses if they are reused.

PR #613, opened 2022-12-21
Remove dead code

alejandro-colomar:


ikerexxe:
LGTM!

Is that all for this PR? If so I'll proceed to merge it.

alejandro-colomar:
I think this is all of it.  Since we were using syslog (AFAICT), there's not much we can remove, apart from the conditionals themselves.

alejandro-colomar:
I still want to remove the funny comment in source code that I introduced, so I'll investigate the `RLOGIN` and `RUSEROK` macros, but in a new PR...

ikerexxe:
Thanks and good work!

PR #615, opened 2022-12-21
Don't redefine errno(3)

alejandro-colomar:
It is Undefined Behavior to declare errno (see NOTES in its manual page). Instead of using the errno dummy declaration, use one that doesn't need a comment.


ikerexxe:
Is this part of some additional work? Or can I merge it as it is?

alejandro-colomar:
No, this is all; you can merge it.
I'm trying to be consistent in the use of draft pull requests until they're ready.

PR #616, opened 2022-12-22
Raise the bar to POSIX.1-2008

alejandro-colomar:
However, don't stop using functions that were once required by POSIX.1-2001 that were good functions, and which have been removed by POSIX.1-2008 without good reasons.  For example, continue using `bzero(3)`.  However, there may be functions that were removed with a good reason.  We should stop using those.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -54,7 +54,6 @@ AC_CHECK_FUNCS(arc4random_buf futimes \
 AC_SYS_LARGEFILE
 
 dnl Checks for typedefs, structures, and compiler characteristics.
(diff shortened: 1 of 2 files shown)


alejandro-colomar:
I have one more change that could be included in this patch set, since it is part of upgrading to POSIX.1-2008, but since it's a quite complex change, and and important side effects, I'll keep it in a separate PR.  So, this is ready!  :)

ikerexxe (approved):
LGTM

hallyn (approved):

PR #617, opened 2022-12-22
Use getaddrinfo(3) instead of gethostbyname(3)

alejandro-colomar:
gethostbyname(3) was removed in POSIX.1-2008.  It has been obsoleted,
and replaced by getaddrinfo(3), which is superior in several ways:

-  gethostbyname(3) is not reentrant.  There's a GNU extension,
   gethostbyname_r(3) which is reentrant, but it's not likely to be
   standardized for the following reason.  And we don't care too much
   about this point either.

-  gethostbyname(3) only supports IPv4, but getaddrinfo(3) supports both
   IPv4 and IPv6 (and may support other address families in the future).

We don't care about reentrancy, so for keeping the code simple (i.e.,
not touch call site to add code to free(3) an allocated buffer), I added
a static buffer for inet_ntop(3).  We could address that in the future,
but I don't think it's worth it.

BTW, we also replace inet_ntoa(3) by inet_ntop(3), as a consequence of
using getaddrinfo(3).  inet_ntoa(3) is also marked as deprecated, but
that deprecation seems to have been documented only in the manual page,
and POSIX doesn't mark it as deprecated.  The deprecation notice goes
back to when the inet_ntop(3) manual page was added by Sam Varshavchik
to the Linux man-pages in version 1.30 (year 2000).

So, this, apart from updating the code to POSIX.1-2008, is also adding
support for IPv6 :)  Although, probably many other parts of the code are
written for IPv4 only, so I wouldn't yet claim support for it.

A few notes:

-  I didn't check the return value of inet_ntop(3), since it can't fail
   for the given input:

   -  EAFNOSUPPORT:  We only call it with AF_INET and AF_INET6.
   -  ENOSPC:  We calculate the size of the buffer to be wide enough:
               MAX(INET_ADDRSTRLEN, INET6_ADDRSTRLEN) so it always fits.

Link: <https://github.com/shadow-maint/shadow/issues/600>
Cc: Dave Hagewood <admin@arrowweb.com>
Cc: Sam Varshavchik
Cc: Jakub Jelinek <jakub@redhat.com> @jakubjelinek

### DIFF
--- a/libmisc/inet_sockaddr2str.c
+++ b/libmisc/inet_sockaddr2str.c
@@ -1,0 +1,78 @@
+/*
+ * SPDX-FileCopyrightText: 2022, Alejandro Colomar <alx@kernel.org>
+ *
+ * SPDX-License-Identifier:  BSD-3-Clause
+ */
+
+#include <config.h>
+
+#include <arpa/inet.h>
+#include <errno.h>
+#include <netinet/in.h>
+#include <sys/param.h>
+#include <sys/socket.h>
+
+#ident "$Id$"
+
+#include "defines.h"
+#include "prototypes.h"
+
+
+#if !defined(INET_ADDRSTRLENMAX)
+#define INET_ADDRSTRLENMAX  MAX(INET_ADDRSTRLEN, INET6_ADDRSTRLEN)
+#endif
+
+
+/*
+ * SYNOPSIS
+ *	const char *inet_sockaddr2str(const struct sockaddr *sa);
+ *
+ * DESCRIPTION
+ *	This function is similar to inet_ntop(3).  It transforms an address
+ *	in 'struct in_addr' or 'struct in6_addr' form into a human-readable
+ *	string.
+ *
+ *	It receives a sockaddr structure, which is simpler to pass after
+ *	receiving it from getaddrinfo(3).  However, this function is not
+ *	reentrant, and like inet_ntoa(3), it uses an internal buffer, for
+ *	simplicity; anyway we're not in a multithreaded program, and it
+ *	doesn't contain any sensitive data, so it's fine to use static
+ *	storage here.
+ *
+ * RETURN VALUE
+ *	This function returns a pointer to a statically allocated buffer,
+ *	which subsequent calls will overwrite.
+ *
+ *	On error, it returns NULL.
+ *
+ * ERRORS
+ *	EAFNOSUPPORT
+ *		The address family in sa->sa_family is not AF_INET or AF_INET6.
+ *
+ * CAVEATS
+ *	This function is not reentrant.
+ */
+
+
+const char *
+inet_sockaddr2str(const struct sockaddr *sa)
+{
+	struct sockaddr_in   *sin;
+	struct sockaddr_in6  *sin6;
+
+	static char          buf[INET_ADDRSTRLENMAX];
+
+	switch (sa->sa_family) {
+	case AF_INET:
+		sin = (struct sockaddr_in *) sa;
+		inet_ntop(AF_INET, &sin->sin_addr, buf, NITEMS(buf));
+		return buf;
+	case AF_INET6:
+		sin6 = (struct sockaddr_in6 *) sa;
+		inet_ntop(AF_INET6, &sin6->sin6_addr, buf, NITEMS(buf));
+		return buf;
+	default:
+		errno = EAFNOSUPPORT;
+		return NULL;
+	}
+}
--- a/src/login_nopam.c
+++ b/src/login_nopam.c
@@ -52,6 +52,7 @@
 #include <unistd.h>
 #include <stdlib.h>
 #include <sys/socket.h>
+#include <sys/param.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>		/* for inet_ntoa() */
 
@@ -68,6 +69,7 @@
 /* Delimiters for fields and for lists of users, ttys or hosts. */
 static char fs[] = ":";		/* field separator */
 static char sep[] = ", \t";	/* list-element separator */
+static char inet_ntop_buffer[MAX(INET_ADDRSTRLEN, INET6_ADDRSTRLEN)];
 
 static bool list_match (char *list, const char *item, bool (*match_fn) (const char *, const char *));
 static bool user_match (const char *tok, const char *string);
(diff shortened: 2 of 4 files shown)


alejandro-colomar:
Rebased to master.

alejandro-colomar:
Abort on IPv7 :)

alejandro-colomar:
Changes:

- Rebased to master.
- Remove NITEMS() from this PR. (So, tests will fail, but the PR will be simpler, once the other is merged).

alejandro-colomar:
Rebased to master.

src/login_nopam.c
```diff
@@ -265,33 +263,18 @@ static bool user_match (const char *tok, const char *string)
...
-		inet_ntop(AF_INET6, &sin6->sin6_addr,
-		          inet_ntop_buffer, NITEMS(inet_ntop_buffer));
-		p = inet_ntop_buffer;
-		break;
-	default:
-		p = "Hypothetical future inet version?";
+	p = inet_sockaddr2str(addrs[0].ai_addr);
+	if (p == NULL) {
+		SYSLOG ((LOG_ERR, "Hypothetical future IPv7?"));
+		abort();
```
hallyn, line 277: I assume this is a WIP, but abort()ing here is a change in behavior.

src/login_nopam.c
```diff
@@ -263,18 +265,40 @@ static bool user_match (const char *tok, const char *string)
 
 static const char *resolve_hostname (const char *string)
 {
-	/*
-	 * Resolve hostname to numeric IP address, as suggested
-	 * by Dave Hagewood <admin@arrowweb.com>.  --marekm
-	 */
-	struct hostent *hp;
+	int                  gai;
```
ikerexxe: I think it would make more sense to rename this variable. Something like `gai_error` would make more sense.

src/login_nopam.c
```diff
@@ -263,18 +265,40 @@ static bool user_match (const char *tok, const char *string)
...
+	char                 *p;
```
ikerexxe: Can you rename this variable?

libmisc/inet_sockaddr2str.c
```diff
@@ -0,0 +1,100 @@
...
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+ * PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT
+ * HOLDERS OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
ikerexxe, line 27: We are using SPDX to indicate the software license.

libmisc/inet_sockaddr2str.c
```diff
@@ -0,0 +1,100 @@
...
+ * SYNOPSIS
+ *	const char *inet_sockaddr2str(const struct sockaddr *sa);
+ *
+ * DESCRIPTION
+ *	This function is similar to inet_ntop(3).  It transforms an address
+ *	in 'struct in_addr' or 'struct in6_addr' form into a human-readable
+ *	string.
+ *
+ *	It receives a sockaddr structure, which is simpler to pass after
+ *	eceiving it from getaddrinfo(3).  However, this function is not
```
ikerexxe, line 58: You missed an `r` in `receiving`

ikerexxe (changes requested):
There are some minor issues to address before getting my approval.

src/login_nopam.c
alejandro-colomar: No, it's not a WIP.

Yes, this is an important change in behavior, which is why I kept it separate from the cleanups for POSIX.1-2008:
<https://github.com/shadow-maint/shadow/pull/616#issuecomment-1362720236>.

The change in behavior (documented in the commit message) is:

The old gethostbyname(3) only supports IPv4.  The replacement, getaddrinfo(3) supports both IPv4 and IPv6 (and is not inherently limited to those, so a hypothetical IPv7 could also be supported in the future).  By changing the function, we're adding support for IPv6, at least in that part of the code.

Because the function is not inherently limited to IPv4/IPv6, I wanted to be cautious, and abort(3) on anything that's not 4 or 6.  So, if IPv7 is ever a thing, this code will gracefully abort (there's nothing better we can do), instead of doing harmful stuff.

alejandro-colomar: Although, I should have aborted in the first commit already.  I forgot to do that.  Will fix in a moment.

alejandro-colomar: Fixed.  Now I'm aborting since the first commit.

alejandro-colomar:
Changes:

-  abort(3) in the first commit.

libmisc/inet_sockaddr2str.c
alejandro-colomar: Great.  I think I copied it from another file, so I'll check that.

libmisc/inet_sockaddr2str.c
alejandro-colomar: Thanks.

alejandro-colomar:
Changes:

-  Use SPDX
-  Fix typo (missing r)

src/login_nopam.c
alejandro-colomar: Yep.  I used gai_err, to not make a line too long.

alejandro-colomar:
Changes: Rename gai -> gai_err

alejandro-colomar:
Changes: Rename p -> addr_str

src/login_nopam.c
alejandro-colomar: Yep

src/login_nopam.c
```diff
@@ -265,41 +263,25 @@ static bool user_match (const char *tok, const char *string)
...
-	case AF_INET6:
-		sin6 = (struct sockaddr_in6 *) sa;
-		inet_ntop(AF_INET6, &sin6->sin6_addr,
-		          inet_ntop_buffer, NITEMS(inet_ntop_buffer));
-		p = inet_ntop_buffer;
-		break;
-	default:
+	addr_str = inet_sockaddr2str(addrs[0].ai_addr);
+	if (addr_str == NULL) {
 		SYSLOG ((LOG_ERR, "Hypothetical future IPv7?"));
```
ikerexxe, line 276: You don't need to print this here as it was already printed in `inet_sockaddr2str()`.

alejandro-colomar: No, I didn't log in inet_sockaddr2str();  I only set errno.

src/login_nopam.c
hallyn: Hm, well, "gracefully abort" is an interesting phrase, but I see, if a site has user and no from table entries, the login should proceed, so it's ok.

Thanks.

src/login_nopam.c
alejandro-colomar: Ahh, I see where you got confused.  The log in `inet_sockaddr2str()` is for when no addresses are found.  This one is for when an IPv7? address is found.

alejandro-colomar:
Changes: print strerror(); since I documented the meaning of errno for inet_sockaddr2str(), it's simpler to just print strerror().

src/login_nopam.c
```diff
@@ -263,18 +265,41 @@ static bool user_match (const char *tok, const char *string)
...
 
-	hp = gethostbyname (string);
-	if (NULL != hp) {
-		return inet_ntoa (*((struct in_addr *) *(hp->h_addr_list)));
+	gai = getaddrinfo(string, NULL, NULL, &addrs);
+	if (gai != 0)
+		goto notfound;
+
+	sa = addrs[0].ai_addr;
+	switch (addrs[0].ai_addr->sa_family) {
```
hallyn: Just for readability, imo it would be nice if this used switch (sa->sa_family)

hallyn: Never mind - i see it doesn't matter :)

hallyn (approved):
Thanks.

hallyn:
@ikerexxe  I'll wait for you to merge, as I don't know if you had some unfinished thoughts on this.  But I'm happy with it.

src/login_nopam.c
alejandro-colomar: Well, it can matter slightly for someone reading the diff, so I fixed it.  Anyway, it's a smaller diff now. :)

alejandro-colomar:
Changes:
Changed an intermediate diff.  Nothing change in the global diff.

ikerexxe (approved):
Now that my minor comments were addressed I'm happy with the patches.

PR #619, opened 2022-12-23
chfn: fix the change_field results printed

hallyn:
When chfn calls change_field on some fields, if it is not allowed to print some fields, it instead prints them out - except it always prints the fullname.  Fix the field being printed in that case.

### DIFF
--- a/src/chfn.c
+++ b/src/chfn.c
@@ -175,19 +175,19 @@ static void new_fields (void)
 	if (may_change_field ('r')) {
 		change_field (roomno, sizeof roomno, _("Room Number"));
 	} else {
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
It took me some time to understand the commit message.  I had to look at the code.  Maybe if you state that it's a typo fix, it would be easier to understand.

hallyn:
> It took me some time to understand the commit message. I had to look at the code. Maybe if you state that it's a typo fix, it would be easier to understand.

Yeah I struggled a bit with how to phrase it.

alejandro-colomar:
It reads better now.

ikerexxe (approved):
LGTM. Thanks for the patch! I'm merging it.

PR #621, opened 2022-12-27
shadow: Fix can not print full login timeout message

SoumyaWind:
Login timed out message prints only first few bytes when write is immediately followed by exit. Calling exit from new handler provides enough time to display full message.

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -120,6 +120,7 @@ static void get_pam_user (char **ptr_pam_user);
 
 static void init_env (void);
 static void alarm_handler (int);
+static void exit_handler (int);
 
 /*
  * usage - print login command usage and exit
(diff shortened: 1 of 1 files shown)


hallyn:
What about simply doing an fflush on stdout and stderr?

SoumyaWind:
> What about simply doing an fflush on stdout and stderr?

fflush isn't helping in this case. i.e., even with fflush, issue still persists.

hallyn (approved):

hallyn:
> > What about simply doing an fflush on stdout and stderr?
> 
> fflush isn't helping in this case. i.e., even with fflush, issue still persists.

In fact fflush is not listed as being safe for use in a signal handler, so that was a bad idea anyway.  Thanks.

PR #624, opened 2022-12-30
Remove bias from functions that get randomness, simplify their implementation, and move to separate file

alejandro-colomar:


libmisc/random.c
```diff
@@ -0,0 +1,103 @@
...
+
+
+/*
+ * Return a uniformly-distributed random number smaller than upper_bound:
+ * [0, upper_bound)
+ *
+ * Similar to arc4random_uniform(3), but returns a u_long.
+ */
+unsigned long
+shadow_random_uniform(unsigned long upper_bound)
```
zx2c4, line 79: The algo I implemented in Linux might be a bit more efficient if the cost of multiplication is low compared to the cost of drawing a new random sample, which is usually the case: 

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=e9a688bcb19348862afe30d7c85bc37c4c293471
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=7f576b2593a978451416424e75f69ad1e3ae4efe

alejandro-colomar: Hi Jason,

Hmmm, those Linux commits do too much magic to my taste.  Even with the comment and the commit message, I can't still understand why the multiplication works.

So, for an initial implementation, I'd keep the simple truncation with `&=`, even if it's slower.

After that, well, for shadow, I don't think the performance is worth making the code cryptic.  Moreover, since we use 64-bit numbers in this code, we would need 128-bit multiplication, which is not something supported everywhere.

However, it might be an interesting thing to do in glibc.  We could discuss improving the glibc implementation of arc4random_uniform(3) if you think it is an improvement over the current glibc code (please check).  If we do that, I suggest that you prepare a very detailed description of why the multiplication works, because at least in my case (and I suspect for many other programmers too), I already forgot the details of how integer multiplication works bitwise from when I studied that at school, and it's something I never use, so it's hard to understand why it serves the purpose here.

But, don't hurry; after Theo's email, I'm starting to strongly believe that the arc4random_uniform(3) interface is really bad, and you were right when you suggested a few months ago that we should reconsider if we want to add it, or maybe design a better API.  I will propose deprecating arc4random*(3) functions in glibc, and come up with a better set of interfaces, maybe implemented entirely in the kernel.

I'll also send some draft patches to the kernel, to start discussion about new functions that improve over the current get_random*() that you showed, and from that discussion we can maybe come up with new syscalls.

Cheers,

Alex

zx2c4: Lemire has proofs: https://arxiv.org/abs/1805.10941

Your previous comment had a mention of expanding the kernel API. Not sure what you're proposing, but I've got a patchset under review that will make arc4random() and company both safe and fast.

alejandro-colomar: > Lemire has proofs: https://arxiv.org/abs/1805.10941

Thanks!

> Your previous comment had a mention of expanding the kernel API. Not sure what you're proposing,

I'll restore the message here, since it seems that arc4random*(3) can't be fixed because the bugs have percolated to users that now depend on the design issues.  That's fair, though; nothing surprising in C's history.

The removed part of my message was:

"
But, don't hurry; after Theo's email, I'm starting to strongly believe that the arc4random_uniform(3) interface is really bad, and you were right when you suggested a few months ago that we should reconsider if we want to add it, or maybe design a better API. I will propose deprecating arc4random*(3) functions in glibc, and come up with a better set of interfaces, maybe implemented entirely in the kernel.

I'll also send some draft patches to the kernel, to start discussion about new functions that improve over the current get_random*() that you showed, and from that discussion we can maybe come up with new syscalls.
"

> but I've got a patchset under review that will make arc4random() and company both safe and fast.

Please add me to the loop; I'm curious.  What's the link?

alejandro-colomar:
Changes:

- Rename shadow_random_uniform() to shadow_random_below().

Since our function behaves slightly different than arc4random_uniform(3), it's better not to have a name that would cause confusion.  Instead, use a name similar to the Linux kernel internal functions that behave like ours.

alejandro-colomar:
Changes:

-  Fix undefined behavior in include guard (leading underscore + uppercase is reserved).

-  Add bit_ceil_wrapul().

-  Rename functions to distance from arc4random(3), and at the same time, make more clear that the functions produce cryptographically-secure values.  However, keep some similarity with arc4random() in what is useful, for some consistency.

-  Improve comments documenting these functions.

-  Fix Undefined Behavior in implementation of shadow_random_uniform()/csrand_uniform(), which was due to a right shift by 32.  Use instead bit_ceil_wrapul(), which has no UB.

alejandro-colomar:
I think this is ready for review.  This time, I didn't add extensive documentation for these functions, because I think the one-liner is clear enough.  In fact, the manual page I wrote for the arc4random(3) functions is similarly short (see below), and most of the differences are noting the bugs in the libc functions.  If you consider some more documentation is merited, please comment.

```
arc4random(3)              Library Functions Manual              arc4random(3)

NAME
       arc4random,  arc4random_uniform, arc4random_buf - cryptographically‐se‐
       cure pseudorandom number generator

LIBRARY
       Standard C library (libc, ‐lc)

SYNOPSIS
       #include <stdlib.h>

       uint32_t arc4random(void);
       uint32_t arc4random_uniform(uint32_t upper_bound);
       void arc4random_buf(void buf[.n], size_t n);

DESCRIPTION
       These functions give cryptographically‐secure pseudorandom numbers.

       arc4random() returns a uniformly‐distributed value.

       arc4random_uniform() returns a uniformly‐distributed  value  less  than
       upper_bound (see BUGS).

       arc4random_buf()  fills  the  memory pointed to by buf, with n bytes of
       pseudorandom data.

       The rand(3) and drand48(3) families of functions should  only  be  used
       where  the  quality  of  the  pseudorandom numbers is not a concern and
       there’s a need for repeatability of the results.  Unless you meet  both
       of those conditions, use the arc4random() functions.

RETURN VALUE
       arc4random() returns a pseudorandom number.

       arc4random_uniform()  returns  a  pseudorandom  number  less  than  up‐
       per_bound for valid input, or 0 when upper_bound is invalid.

ATTRIBUTES
       For an explanation of the terms  used  in  this  section,  see  attrib‐
       utes(7).
       ┌────────────────────────────────────────────┬───────────────┬─────────┐
       │Interface                                   │ Attribute     │ Value   │
       ├────────────────────────────────────────────┼───────────────┼─────────┤
       │arc4random(), arc4random_uniform(),         │ Thread safety │ MT‐Safe │
       │arc4random_buf()                            │               │         │
       └────────────────────────────────────────────┴───────────────┴─────────┘

STANDARDS
       These nonstandard functions are present on several Unix systems.

BUGS
       An  upper_bound  of  0  doesn’t make sense in a call to arc4random_uni‐
       form().  Such a call will fail, and return 0.  Be careful,  since  that
       value  is  not less than upper_bound.  In some cases, such as accessing
       an array, using that value could result in Undefined Behavior.

SEE ALSO
       getrandom(3), rand(3), drand48(3), random(7)

Linux man‐pages (unreleased)        (date)                       arc4random(3)
```

alejandro-colomar:
> I have several questions before being able to further review this PR.
> 
>     * What does UB mean?

Ahh, sorry.  UB is an abbreviation for Undefined Behavior (aka Nasal Demons; now I wonder if someone every wrote a program `nasald` :P ).  It's a common abbreviation in some forums, and I sometimes assume it's common everywhere :)

> 
>     * What will bit_ceil_wrapul() and leading_zerosul() do? A search on the internet hasn't provided any meaningful answer.

They are similar to planned additions to ISO C23.

<https://thephd.dev/c23-is-coming-here-is-what-is-on-the-menu#n3022---modern-bit-utilities>
<https://www.open-std.org/JTC1/SC22/WG14/www/docs/n3054.pdf#subsection.7.18.3>
<https://www.open-std.org/JTC1/SC22/WG14/www/docs/n3054.pdf#subsection.7.18.16>

> 
> 
> In the `Add csrand_uniform() ` commit you wrote `reutrns` instead of `returns`.
> 
> Finally, a request. Can you move the bit manipulation functions commit? It took me a context switch to understand why you were adding those functions, and then you didn't use the functions provided in those commits in the next commit, so I had another context switch.

Ahh, yes, I thought I had put it just before its use.  I'll move it.

Happy new year :)

 Alex

alejandro-colomar:
Changes:

-  Reorder commits.
-  Change comment to expand UB into "Undefined Behavior".

libmisc/csrand.c
```diff
@@ -59,3 +65,23 @@ csrand(void)
...
 }
+
+
+/*
+ * Return a uniformly-distributed CS random value in the interval [0, n-1].
+ */
+unsigned long
+csrand_uniform(unsigned long n)
+{
+	unsigned long  r, max, mask;
```
ikerexxe, line 76: Can you rename `r` to something more meaningful? `random` perhaps?

libmisc/csrand.c
```diff
@@ -59,3 +65,23 @@ csrand(void)
...
+
+	max = n - 1;
+	mask = bit_ceil_wrapul(n) - 1;
+
+	do {
+		r = csrand();
+		r &= mask;  // optimization
+	} while (r > max);
```
ikerexxe, line 84: From lack of knowledge, how long would it take on average to execute this loop? In other words, is it possible that it would take a long time to execute and that the user experience would suffer?

ikerexxe (commented):
Some inline questions to better understand the code.

> In the `Add csrand_uniform() ` commit you wrote `reutrns` instead of `returns`.

This comment is still valid.

ikerexxe:
> ```
> arc4random(3)              Library Functions Manual              arc4random(3)
> 
> NAME
>        arc4random,  arc4random_uniform, arc4random_buf - cryptographically‐se‐
>        cure pseudorandom number generator
> 
> ```

I guess this manual page was updated recently. Am I right? I cannot see it like that in my system.

alejandro-colomar:
> > ```
> > arc4random(3)              Library Functions Manual              arc4random(3)
> > 
> > NAME
> >        arc4random,  arc4random_uniform, arc4random_buf - cryptographically‐se‐
> >        cure pseudorandom number generator
> > ```
> 
> I guess this manual page was updated recently. Am I right? I cannot see it like that in my system.

Yes, I added it in 2023 :)

```
alx@debian:~/src/linux/man-pages/man-pages/main$ git log -- man3/arc4random.3
commit 3638ab3f148ad661e731e4d5d744c2fd32c463f0
Date:   Thu Jan 5 20:14:19 2023 +0100

    arc4random.3: Raise the severity of the CAVEATS to BUGS
    
    This is a misdesign that the original OpenBSD developers fail to
    acknowledge.
    
    Link: <https://inbox.sourceware.org/libc-alpha/068b01c4-d0c4-0849-eabb-09c020a1480b@gmail.com/T/>
    Cc: "Jason A. Donenfeld" <Jason@zx2c4.com>
    Cc: Arsen Arsenović <arsen@aarsen.me>
    Cc: Sam James <sam@gentoo.org>
    Cc: "Serge E. Hallyn" <serge@hallyn.com>

commit 477c35bc1b0a785bd0b1eb5d57943b0483390519
Date:   Mon Jan 2 12:08:31 2023 +0100

    arc4random.3: Be consistent in uses of pseudorandom
    
commit 848404feb5bfbdcb9b8f693ec29159231ee2a7c8
Date:   Sun Jan 1 17:18:22 2023 +0100

    arc4random.3: New page documenting the arc4random(3) family of functions
    
    arc4random(3)
    arc4random_uniform(3)
    arc4random_buf(3)
    
```

alejandro-colomar:
> Some inline questions to better understand the code.
> 
> > In the `Add csrand_uniform() ` commit you wrote `reutrns` instead of `returns`.
> 
> This comment is still valid.

Oops, fixed.

lib/bit.h
```diff
@@ -0,0 +1,41 @@
...
+		return 0;
+
+	return 1 + (ULONG_MAX >> leading_zerosul(x));
+}
+
+/* stdc_leading_zerosul(3) */
+inline int
+leading_zerosul(unsigned long x)
+{
+	return (x == 0) ? ULONG_WIDTH : __builtin_clz(x);
```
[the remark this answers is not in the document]
alejandro-colomar: I know that at least Clang also supports it.  Probably any GNU-C-compatible compilers also support it, since it's there since GCC 3.x.

I don't know about compilers other than GCC or Clang, but if anyone uses a compiler that doesn't have it, probably there will be some alternative name for the same functionallity, so it would be easy to patch it to use the alternative.

libmisc/csrand.c
alejandro-colomar: the thing is `rand` and `random` are both reserved names (from `<stdlib.h>`).  Maybe I could better clarify it in a different way if it's not very clear?

libmisc/csrand.c
alejandro-colomar: Hmm, I'll add some documentation in the code clarifying that.

It's a [geometric distribution](https://en.wikipedia.org/wiki/Geometric_distribution), where `p` is somewhere between `0.5` and `1`, depending on the value of `max` (if `powerof2(max + 1) == 1`, `p == 1`; if `powerof2(max) == 1`, then `p` is close to `0.5`).

A geometric distribution, theoretically, can loop forever; however, that's very unlikely (the probability of that is an infinitesimal).

Being realistic, and assuming a worst case (the worst case is for `csrand_uniform(LONG_MAX + 2ul)`, so `max == LONG_MAX + 1ul`, so `p` ~= `0.5`), let's get some numbers:

Pr(X = k) = (1 - p)^(k - 1) * p ~= 0.5^(k - 1) * 0.5 = 0.5^k

Now, let's assume that every human in the world (around 2^33) would need a million random numbers per day (around 2^20) for 10000 years (around 2^22 days), we see that k (the number of iterations in the loop) might reach 75 some day in the next 10k years.

However, that's just for seeing what is likely to happen some day on earth.  On average (even for a worst case input), it's likely to loop once, or just a few times.  It can never be worse than tossing a coin and waiting for the first heads.

alejandro-colomar: I'm still trying to understand why the multiplication that Jason suggested is an optimization.  If you want me to do that on this PR, we can wait.

libmisc/random.c
alejandro-colomar: I finally understood how the optimization works.  Here's some reduction to 3 bits, which (at least to me) clarifies how the rejection works.

image

Since it is on average much faster than my approach, I'll apply it as an optimization, in a following patch.

alejandro-colomar: This is a graph of the probability of looping, for the algorithm with multiplication:

image

From those saw lines, only the last one goes up to 0.5.  As you go closer to 0, they are smaller.  This is compared to my implementation with masks, where _all_ the saw lines go up to 0.5.

alejandro-colomar: @ikerexxe , @hallyn 

This is a considerable performance improvement for low values of N.  The only downside is that it requires using a type wider than long, which would need to be `__int128`.  It's a GNU extension, so I don't know how much it's viable.  The current approach of masking is not so bad, it will normally just call a handful of times getrandom(2), but it could be much better.

alejandro-colomar:
Changes:

-  Optimize csrand_uniform().

This comes with a caveat: it uses `unsigned __int128`.  I pushed the patch so that we can review it and see if we want it or not.  However, maybe we are fine with just 32 bits of random numbers instead of 64, in which case we wouldn't need more than 64 bits for the internal multiplication.  It depends on what is the maximum number of rounds for `*_get_salt_rounds()`.  If anybody knows, please speak.

alejandro-colomar:
Changes:

- Added comment with probability of looping for the masking implementation.

alejandro-colomar:
Changes:

-  Add some details to the commit message, link to Lemire's paper (I forgot to do that previously), and while I'm at it, CC Lemire too, in case he wants to have a look).

alejandro-colomar:
Changes:

-  Remove repeated CCs in commit message
-  Use `_` for the unused variable name

libmisc/random.c
alejandro-colomar: Here goes an (incomplete) 4-bit visual explanation of how it works, to compare it to the 3-bit one from above.  We can see that the rejection is smaller for low values of N compared to the same values of N when the original random number generator has less bits.

image

libmisc/csrand.c
```diff
@@ -0,0 +1,104 @@
...
+
+fail:
+	fprintf(log_get_logfd(), _("Unable to obtain random bytes.\n"));
+	exit(1);
+}
+
+
+/*
+ * Return a uniformly-distributed CS random value in the interval [0, n-1].
+ */
```
lemire, line 73:
```suggestion
 * Fast Random Integer Generation in an Interval
 * ACM Transactions on Modeling and Computer Simulation 29 (1), 2019
 * https://arxiv.org/abs/1805.10941
 */
```

lemire:
It could be useful to link to the research paper. Some people find that useful.

alejandro-colomar:
> It could be useful to link to the research paper. Some people find that useful.

I did in the commit message (please check if that link is accurate):  <https://github.com/shadow-maint/shadow/pull/624/commits/7949a8e194e4624363b1717035074e6a203f0b1b>

Jason linked to it in the PR conversation: <https://github.com/shadow-maint/shadow/pull/624#discussion_r1059692301>

Thanks for reviewing! :)

alejandro-colomar:
Changes:

-  Apply Daniel Lemire's suggestion of linking to the paper in source code.

lib/bit.h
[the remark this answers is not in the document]
ikerexxe: That should be enough. Thanks!

alejandro-colomar:
Changes:

-  Remove unused include.

libmisc/csrand.c
ikerexxe: ok, no problem

ikerexxe:
I like this PR a lot! It explains all the changes clearly, and it dives in the more difficult aspects providing the necessary information to understand them. Thanks for your hard work.

Is this the final version? I think @hallyn should take a look at it.

alejandro-colomar:
> I like this PR a lot! It explains all the changes clearly, and it dives in the more difficult aspects providing the necessary information to understand them. Thanks for your hard work.
> 
> Is this the final version? I think @hallyn should take a look at it.

I just had an idea to not require 128-bit integers.  And it will probably optimize a bit for high values of N.  Stay tuned :)
It will be easy to implement anyway, so in a few minutes or hours I'll push it.

hallyn:
> I like this PR a lot! It explains all the changes clearly, and it dives in the more difficult aspects providing the necessary information to understand them. Thanks for your hard work.
> 
> Is this the final version? I think @hallyn should take a look at it.

Thanks, it's on my list.

alejandro-colomar:
Changes: rebase to master

alejandro-colomar:
Changes:

- Clarify why `-n%n` works.

hallyn:
I'm still reading the paper, but let's merge now :)

thanks.  this has been educational for me.

hallyn:
Actually, I'll wait for a final ok from @alejandro-colomar  - you're ready for this to merge?

alejandro-colomar:
Hi Serge!

On Sat, Jan 28, 2023, 03:04 Serge Hallyn ***@***.***> wrote:

> Actually, I'll wait for a final ok from @alejandro-colomar
> <https://github.com/alejandro-colomar> - you're ready for this to merge?
>

Sure :)

Unless you'd like me to clarify anything a bit more in the commit messages
or comments, I think it's fine.

Cheers,
Alex

PR #629, opened 2023-01-19
Use getnameinfo(3) to avoid Undefined Behavior in our code

alejandro-colomar:
I didn't know getnameinfo(3) existed, so I implemented it, or something similar to it as inet_sockaddr2str().  Guess what?  I missed some small detail.  I implemented it using Undefined Behavior (UB). :)

What might be more worrying is that this function can't be implemented without relying on UB, and that's fault of the original design of the struct sockaddr_* types.  In fact, glibc's implementation also relies on UB, which might be worrying.

Anyway, it's safer to just rely on the standard API, so let's do it. And doing that simplifies our source, so win-win.

Link: <https://inbox.sourceware.org/libc-alpha/0f25d60f-f183-b518-b6c1-6d46aa63ee57@gmail.com/T/>
Link: <https://stackoverflow.com/a/42190913/6872717>
Cc: Zack Weinberg <zack@owlfolio.org>


alejandro-colomar:
Hmmm, it seems I was a bit too much paranoic.  This wasn't Undefined Behavior.  However, the patch is still good, even if only for simplifying code. :)

<https://software.codidact.com/posts/287748>

hallyn (approved):

hallyn:
thanks, it lgtm.  Waiting to merge since you still may be actively making changes.

alejandro-colomar:
> thanks, it lgtm. Waiting to merge since you still may be actively making changes.

Thanks!  I don't expect to make any more changes, so you can merge already.

Cheers,

Alex

PR #632, opened 2023-01-27
Read whole line in yes_or_no

ferivoz:
Do not stop after 79 characters. Read the complete line to avoid arbitrary limitations.

Proof of Concept:

```
cat > passwd-poc << EOF
root:x:0:0:root:/root:/bin/bash
root:x:0:0:root:/root:/bin/bash
root:x:0:0:root:/root:/bin/bash
EOF
python -c "print(80*'y')" | pwck passwd-poc
```

Two lines should still be within the file because we agreed only once to remove a duplicated line.

Signed-off-by: Samanta Navarro <ferivoz@riseup.net>

### DIFF
--- a/libmisc/yesno.c
+++ b/libmisc/yesno.c
@@ -28,7 +28,8 @@
  */
 bool yes_or_no (bool read_only)
 {
-	char buf[80];
+	int c;
+	bool result;
 
 	/*
 	 * In read-only mode all questions are answered "no".
@@ -46,11 +47,13 @@ bool yes_or_no (bool read_only)
 	/*
 	 * Get a line and see what the first character is.
 	 */
+	c = fgetc(stdin);
 	/* TODO: use gettext */
-	if (fgets (buf, sizeof buf, stdin) == buf) {
-		return buf[0] == 'y' || buf[0] == 'Y';
-	}
+	result = (c == 'y' || c == 'Y');
 
-	return false;
+	while (c != '\n' && c != EOF)
+		c = fgetc(stdin);
+
+	return result;
 }
 


alejandro-colomar:
I think we should use rpmatch(3) here.

<https://man.archlinux.org/man/rpmatch.3.en>

alejandro-colomar:
How about something like the following?

```c
bool yes_or_no (bool read_only)
{
	bool  ret;
	char  *buf;

	/*
	 * In read-only mode all questions are answered "no".
	 */
	if (read_only) {
		(void) puts (_("No"));
		return false;
	}

	/*
	 * Typically, there's a prompt on stdout, sometimes unflushed.
	 */
	(void) fflush (stdout);

	ret = false;
	if (getline (&buf, NULL, stdin) == NULL) {
		goto out;
	}

	ret = rpmatch (buf) == 1;

out:
	free(buf);
	return ret;
}
```

hallyn:
Does musl provide rpmatch?

alejandro-colomar:
> Does musl provide rpmatch?

It seems not :(

guillemj:
@alejandro-colomar hmm I'm a bit conflicted on this one, while I see it is available on FreeBSD, DragonflyBSD, AIX and GNU, this would seem like an SVID extension (at least going by the first commit introducing it into glibc and the feature macro protecting it). My concern is that libbsd should IMO not become gnulib, and I'm uncertain whether this one would go in that direction or what. :/

alejandro-colomar:
Hi Guillem,

On 4/9/23 19:12, Guillem Jover wrote:
> @alejandro-colomar hmm I'm a bit conflicted on this one, while I see it is available on FreeBSD, DragonflyBSD, AIX and GNU, this would seem like an SVID extension (at least going by the first commit introducing it into glibc and the feature macro protecting it). My concern is that libbsd should IMO not become gnulib, and I'm uncertain whether this one would go in that direction or what. :/
Fair enough; especially if we want to push for it being an essential
package.

The weird situation we have here is that shadow is internationalized,
but understands y/n answers only; it might create confusion in some
languages...

However, since that didn't happen yet, let's postpone that bug to the
future.

Thanks for having a look,
Alex

-- 
GPG key fingerprint: A9348594CE31283A826FBDD8D57633D441E25BB5

alejandro-colomar:
Hi Serge!

I think it would still be good to call getline(3), even if we later only
use the first character.  I think it's more readable than manually looping.

Samanta, what do you think?  Would you send an updated patch using
getline(3)?

Cheers,
Alex

--

alejandro-colomar:
Hi Serge,

Considering that Samanta didn't respond to the suggestions, and since
her patch is already an improvement (it fixes a bug), I suggest we
apply it.  You can add

	Reviewed-by: Alejandro Colomar ***@***.***>

I'll prepare myself a patch to use getline(3) after this one lands on
master.

Cheers,
Alex

-- 
GPG key fingerprint: A9348594CE31283A826FBDD8D57633D441E25BB5

hallyn:
> Hi Serge, Considering that Samanta didn't respond to the suggestions, and since her patch is already an improvement (it fixes a bug), I suggest we apply it. You can add Reviewed-by: Alejandro Colomar ***@***.***> I'll prepare myself a patch to use getline(3) after this one lands on master. Cheers, Alex
> […](#)
> -- <http://www.alejandro-colomar.es/> GPG key fingerprint: A9348594CE31283A826FBDD8D57633D441E25BB5

Sounds good, thanks.

PR #633, opened 2023-01-27
Correctly handle illegal system file in tz

ferivoz:
If the file referenced by ENV_TZ has a zero length string, then an out of boundary write occurs. Also the result can be wrong because it is assumed that the file will always end with a newline.

Only override a newline character with '\0' to avoid these cases.

This cannot be considered to be security relevant because login.defs and its contained references to system files should be trusted to begin with.

Proof of Concept:

1. Compile shadow's su with address sanitizer and --without-libpam

2. Setup your /etc/login.defs to contain ENV_TZ=/etc/tzname

3. Prepare /etc/tzname to contain a '\0' byte at the beginning

`python -c "print('\x00')" > /etc/tzname`

4. Use su

`su -l`

You can see the following output:

`tz.c:45:8: runtime error: index 18446744073709551615 out of bounds for type 'char [8192]'`

Signed-off-by: Samanta Navarro <ferivoz@riseup.net>

### DIFF
--- a/lib/port.c
+++ b/lib/port.c
@@ -130,8 +130,8 @@ static struct port *getportent (void)
       again:
 
 	/*
-	 * Get the next line and remove the last character, which
-	 * is a '\n'.  Lines which begin with '#' are all ignored.
+	 * Get the next line and remove optional trailing '\n'.
+	 * Lines which begin with '#' are all ignored.
 	 */
 
 	if (fgets (buf, (int) sizeof buf, ports) == 0) {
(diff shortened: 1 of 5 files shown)


hallyn:
Hi,

thanks for pointing this out.  FWIW even if someone did "echo -n America/Chicago > /etc/timezone" it would do the wrong thing!

Can you add a pithy comment above it to the effect of "remove \n if it's there"?

ferivoz:
> Can you add a pithy comment above it to the effect of "remove \n if it's there"?

Hi @hallyn,

I have added the comment and two commits which address the same issue in other files as well. I have discovered them after looking for more cases in which strlen is used with arrays. I hope that my analysis based on grep was sufficient to detect all of them.

Sincerely,
Samanta

hallyn:
Thank you

PR #641, opened 2023-01-31
leading_zerosul(): Fix bug

alejandro-colomar:


alejandro-colomar:
Changes:

-  Fix commit message.

I found this bug while looking at `bit_ceil_wrapul()`, which is probably what induced my mind to write that bogus commit message :P

ikerexxe:
Good catch

PR #643, opened 2023-02-01
Use freezero(3) where suitable

alejandro-colomar:
It originated in OpenBSD, and is available in libbsd.


ikerexxe:
LGTM. Thanks for the patch!

PR #645, opened 2023-02-02
Fix memory leaks by replacing realloc(3) with reallocf(3)

alejandro-colomar:

### DIFF
--- a/lib/groupmem.c
+++ b/lib/groupmem.c
@@ -86,33 +86,3 @@ void gr_free (/*@out@*/ /*@only@*/struct group *grent)
 	gr_free_members(grent);
 	free (grent);
 }
-
-bool gr_append_member(struct group *grp, char *member)
-{
-	int i;
-
-	if (NULL == grp->gr_mem || grp->gr_mem[0] == NULL) {
-		grp->gr_mem = (char **)malloc(2 * sizeof(char *));
-		if (!grp->gr_mem) {
-			return false;
-		}
-		grp->gr_mem[0] = strdup(member);
-		if (!grp->gr_mem[0]) {
-			return false;
-		}
-		grp->gr_mem[1] = NULL;
-		return true;
-	}
-
-	for (i = 0; grp->gr_mem[i]; i++) ;
-	grp->gr_mem = realloc(grp->gr_mem, (i + 2) * sizeof(char *));
-	if (NULL == grp->gr_mem) {
-		return false;
-	}
-	grp->gr_mem[i] = strdup(member);
-	if (NULL == grp->gr_mem[i]) {
-		return false;
-	}
-	grp->gr_mem[i + 1] = NULL;
-	return true;
-}
--- a/src/newusers.c
+++ b/src/newusers.c
@@ -1200,9 +1200,9 @@ int main (int argc, char **argv)
 #ifdef USE_PAM
 		/* keep the list of user/password for later update by PAM */
 		nusers++;
-		lines     = realloc (lines,     sizeof (lines[0])     * nusers);
-		usernames = realloc (usernames, sizeof (usernames[0]) * nusers);
-		passwords = realloc (passwords, sizeof (passwords[0]) * nusers);
+		lines     = reallocf (lines,     sizeof (lines[0])     * nusers);
+		usernames = reallocf (usernames, sizeof (usernames[0]) * nusers);
+		passwords = reallocf (passwords, sizeof (passwords[0]) * nusers);
 		lines[nusers-1]     = line;
 		usernames[nusers-1] = strdup (fields[0]);
 		passwords[nusers-1] = strdup (fields[1]);
--- a/src/newusers.c
+++ b/src/newusers.c
@@ -1203,6 +1203,13 @@ int main (int argc, char **argv)
 		lines     = reallocf (lines,     sizeof (lines[0])     * nusers);
 		usernames = reallocf (usernames, sizeof (usernames[0]) * nusers);
 		passwords = reallocf (passwords, sizeof (passwords[0]) * nusers);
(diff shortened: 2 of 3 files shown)


ikerexxe:
I know you are just replacing `realloc()` by `reallocf()`, but as you are already touching these lines of code. Can you add a new commit to check for the output of `reallocf()` in `newusers.c` and fail (`errors++ and continue`) if it returns NULL?

alejandro-colomar:
> I know you are just replacing `realloc()` by `reallocf()`, but as you are already touching these lines of code. Can you add a new commit to check for the output of `reallocf()` in `newusers.c` and fail (`errors++ and continue`) if it returns NULL?

Sure!  They make sense in the same PR.

alejandro-colomar:
I guess we'll need some po file for the error message.  Or maybe we can reuse some other one?

ikerexxe:
LGTM. Thanks for that patches!

src/newusers.c
```diff
@@ -1203,6 +1203,13 @@ int main (int argc, char **argv)
 		lines     = reallocf (lines,     sizeof (lines[0])     * nusers);
 		usernames = reallocf (usernames, sizeof (usernames[0]) * nusers);
 		passwords = reallocf (passwords, sizeof (passwords[0]) * nusers);
+		if (lines == NULL || usernames == NULL || passwords == NULL) {
+			fprintf (stderr,
+			         _("%s: line %d: %s\n"),
+			         Prog, line, strerror(errno));
+			errors++;
```
hallyn, line 1210: Any reason not to just fail_exit here()?

lib/groupmem.c
```diff
@@ -105,7 +105,7 @@ bool gr_append_member(struct group *grp, char *member)
 	}
 
 	for (i = 0; grp->gr_mem[i]; i++) ;
```
hallyn, line 107: Huh.  Does anything use gr_append_member?  I wonder if we can drop it.

src/newusers.c
ikerexxe: Just to be consistent with the rest of the code in the same function.

lib/groupmem.c
ikerexxe: Not that I'm aware of

src/newusers.c
alejandro-colomar: If you think it's simpler to ffail_exit(), we might want to do it everywhere in that file.  That would probably simplify a lot of code.  And we can do that in many other places.  For now, I have a bunch of patches in my local repo that will imnprove use of allocation functions, making it safer and more consistent.

alejandro-colomar:
Changes:

- Remove unused function [@hallyn ]

alejandro-colomar:
Changes:

- Rebase to master

hallyn:
BTW as a more general comment / guidance - in the gr_append_member() case, since that was a library function - you were of course right that the existing code was bad and needed to be fixed, but switching to reallocf() was IMO not the best way ,because the caller might know that the field was non-null, and now as a side effect this fn might have made it null.

Code that doesn't use a temporary buffer for realloc() is really bad though, so thanks for finding that.

(I'll take a close look at the updates in a few hours)

alejandro-colomar:
> BTW as a more general comment / guidance - in the gr_append_member() case, since that was a library function - you were of course right that the existing code was bad and needed to be fixed, but switching to reallocf() was IMO not the best way ,because the caller might know that the field was non-null, and now as a side effect this fn might have made it null.

I'm not sure I understand "the caller might know that the field was non-null".

The input may or may not be NULL; both realloc(3) and reallocf(3) interpret that in the same manner: act as if malloc(3).

The only thing reallocf(3) does different to realloc(3) is that in case of failure, it also calls free(3) on the input pointer so that you don't need the tmp pointer and an extra conditional.

> 
> Code that doesn't use a temporary buffer for realloc() is really bad though, so thanks for finding that.

Thanks :)

> 
> (I'll take a close look at the updates in a few hours)

Okay.

hallyn:
> The input may or may not be NULL, it's not a problem; both realloc(3) and reallocf(3) interpret that in the same manner: act as if malloc(3).

Yes :)  But I'm saying that the code SHOULD have done:
```
   void *tmp = realloc(grp->gr_mem, (i + 2) * sizeof(char *));
    if (NULL == tmp)
       return false;
    grp->gr_mem = tmp;
```
That way grp->gr_mem is not corrupted in case of failure.  No side effects.  Not important for a main program, but is for a library.

src/newusers.c
hallyn: Since it's going to exit at the bottom of that fn anyway, I do think it's better to exit immediately.  There isn't any partially written state to clean up, right?

alejandro-colomar: > Since it's going to exit at the bottom of that fn anyway, I do think it's better to exit immediately.

Agree.  If one line fails, I think it's safer to exit immediately.  Otherwise, some error might carry on, and result in some unexpected behavior.

> There isn't any partially written state to clean up, right?

It seems not.  `continue;` goes directly to the next input line, so the current line seems completely resolved.  And right after the loop we exit.  So, it seems a behavior improvement to break early and break hard.  I'll open a separate PR for that.

alejandro-colomar:
> > The input may or may not be NULL, it's not a problem; both realloc(3) and reallocf(3) interpret that in the same manner: act as if malloc(3).
> 
> Yes :) But I'm saying that the code SHOULD have done:
> 
> ```
>    void *tmp = realloc(grp->gr_mem, (i + 2) * sizeof(char *));
>     if (NULL == tmp)
>        return false;
>     grp->gr_mem = tmp;
> ```
> 
> That way grp->gr_mem is not corrupted in case of failure. No side effects. Not important for a main program, but is for a library.

Ahh, yep, you're right :)  Missed that.

PR #649, opened 2023-02-04
Improve allocations

alejandro-colomar:


lib/malloc.h
```diff
@@ -0,0 +1,134 @@
...
+	if (q == NULL && nmemb != 0 && size != 0)
+		free(p);
+	return q;
+}
+
+
+inline char *
+xstrdup(const char *str)
+{
+	return strcpy(XMALLOCARRAY(strlen(str) + 1, char), str);
```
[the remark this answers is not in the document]
alejandro-colomar: Wow :)

This'll take some time to make sure there's nothing.

alejandro-colomar:
Regarding the "high" report from CodeQL (`Missing return-value check for a 'scanf'-like function`):

Like what?  That's exactly what we do.  This analyzer be kidding us :)

alejandro-colomar:
Rebased to master.

```diff
$ git range-diff gh/master..gh/alloc  master..alloc 
 1:  85b5ccaa <  -:  -------- Remove superfluous casts
 2:  b277bf98 <  -:  -------- ttytype(): Fix race
 3:  41cc3c4c =  1:  03aa642e Rely on realloc(NULL, ...) being equivalent to malloc(...)
 4:  6b5cb73a =  2:  b9355fd6 malloc(3) already sets errno to ENOMEM
 5:  2bc8c628 =  3:  fd163a61 Use reallocf(3) instead of its pattern
 6:  502221f2 =  4:  8fa33624 Use reallocarray(3) instead of its pattern
 7:  8011b006 =  5:  60e1bee6 Use calloc(3) instead of its pattern
 8:  b0fade29 =  6:  ebc30525 Add safer allocation functions
 9:  3e936455 =  7:  31471ecb Use xcalloc(3) instead of its pattern
10:  9630f967 =  8:  918a3957 Use the new header for xstrdup()
11:  ed5af2c5 =  9:  1676ebad Use *array() allocation functions where appropriate
12:  aa91c6a3 = 10:  f87ad6ef Use reallocarrayf() instead of its pattern
13:  759afd46 = 11:  6066d8fd Use xreallocarray() instead of its pattern
14:  c4820476 = 12:  e7416270 Add safer allocation macros
15:  06ea8fb8 = 13:  6eaf62e8 Use safer allocation macros
16:  f37f1470 = 14:  ea048a41 Fix use-after-free of pointer after realloc(3)
```

ikerexxe:
Is this PR ready for review or are you planning to add more commits? I'd prefer to review everything in a row.

alejandro-colomar:
> Is this PR ready for review or are you planning to add more commits? I'd prefer to review everything in a row.

Yes, I think it's ready :)

libmisc/malloc.c
```diff
@@ -20,27 +21,53 @@
...
 
-/*@maynotreturn@*/ /*@only@*//*@notnull@*/char *xstrdup (const char *str)
+
+void *
+xreallocarray(void *p, size_t nmemb, size_t size)
 {
-	return strcpy (xmalloc (strlen (str) + 1), str);
+	p = reallocarrayf(p, nmemb, size);
+	if (p == NULL)
+		goto x;
```
ikerexxe, line 65: Rename `x` to something more meaningful, i.e. done, end, fail...

libmisc/malloc.c
```diff
@@ -20,27 +21,53 @@
...
-		                _("%s: failed to allocate memory: %s\n"),
-		                log_get_progname(), strerror (errno));
-		exit (13);
-	}
-	return ptr;
+	void  *p;
+
+	p = calloc(nmemb, size);
+	if (p == NULL)
+		goto x;
```
ikerexxe, line 49: Rename `x` to something more meaningful, i.e. done, end, fail...

lib/subordinateio.c
```diff
@@ -319,7 +319,7 @@ static bool append_range(struct subid_range **ranges, const struct subordinate_r
 			return false;
 	} else {
 		struct subid_range *alloced;
-		alloced = realloc(*ranges, (n + 1) * (sizeof(struct subid_range)));
+		alloced = reallocarray(*ranges, n+1, sizeof(struct subid_range));
```
ikerexxe: s/n+1/n + 1

lib/malloc.h
```diff
@@ -20,6 +20,58 @@
...
+
+#define ALLOCA(type)           ALLOCARRAY(1, type)
+#define MALLOC(type)           MALLOCARRAY(1, type)
+#define XMALLOC(type)          XMALLOCARRAY(1, type)
+#define REALLOC(ptr, type)     REALLOCARRAY(ptr, 1, type)
+#define REALLOCF(ptr, type)    REALLOCARRAYF(ptr, 1, type)
+
+#define REALLOCARRAY(ptr, n, type)                                            \
+({                                                                            \
+	__auto_type  p_ = (ptr);                                              \
```
ikerexxe, line 36: Is `__auto_type` standard C?

libmisc/env.c
```diff
@@ -128,12 +128,14 @@ void addenv (const char *string, /*@null@*/const char *value)
...
 	if ((newenvc & (NEWENVP_STEP - 1)) == 0) {
-		char **__newenvp;
+		bool  update_environ;
+		char  **__newenvp;
 
 		/*
 		 * If the resize operation succeeds we can
 		 * happily go on, else print a message.
 		 */
+		update_environ = (environ == newenvp);
```
ikerexxe, line 138: Why did you add this line here and not before the condition?

ikerexxe (changes requested):
Two major comments apart from the ones inline. The first one is related to the `malloc.*` files naming. As they contain more functionality apart from `malloc` I wonder if we should rename it to something more generic.

The second one is related to the new macros provided in https://github.com/shadow-maint/shadow/pull/649/commits/e7416270b22e6110b4dbb8ae6d05c32b50400f95. Wouldn't it be better to add this functionality to the functions themselves? We are already adding an abstraction and I'd prefer to avoid having a second one.

By the way, I didn't review properly the last three commits because I wanted to tackle my previous concern first.

alejandro-colomar:
Changes:

-  Rebased to master

```diff
$ git range-diff gh/alloc...alloc
 -:  -------- >  1:  ca9e309d Fix VPATH build
 -:  -------- >  2:  a187ad8e agetpass.c: Use SPDX tags
 -:  -------- >  3:  7e213cfb Add stpeprintf()
 -:  -------- >  4:  46610792 Use stpeprintf() where appropriate
 -:  -------- >  5:  8a9285aa Remove unnecessary NUL terminators
 -:  -------- >  6:  e0e9e57a Add mempcpy(3)
 -:  -------- >  7:  709e6b44 Add stpecpy()
 -:  -------- >  8:  5956cea1 Use stpecpy() where appropriate
 1:  03aa642e =  9:  231aa1cc Rely on realloc(NULL, ...) being equivalent to malloc(...)
 2:  b9355fd6 = 10:  820e65a4 malloc(3) already sets errno to ENOMEM
 3:  fd163a61 = 11:  27681f5e Use reallocf(3) instead of its pattern
 4:  8fa33624 = 12:  1ca3a789 Use reallocarray(3) instead of its pattern
 5:  60e1bee6 = 13:  56e751b2 Use calloc(3) instead of its pattern
 6:  ebc30525 ! 14:  7c4f8c73 Add safer allocation functions
    @@ libmisc/Makefile.am: libmisc_la_SOURCES = \
        loginprompt.c \
        mail.c \
     +  malloc.c \
    +   mempcpy.c \
        motd.c \
        myname.c \
    -   obscure.c \
     @@ libmisc/Makefile.am: libmisc_la_SOURCES = \
        xgetgrnam.c \
        xgetgrgid.c \
 7:  31471ecb = 15:  82dd2db7 Use xcalloc(3) instead of its pattern
 8:  918a3957 = 16:  52c661cd Use the new header for xstrdup()
 9:  1676ebad = 17:  d9da1bd0 Use *array() allocation functions where appropriate
10:  f87ad6ef = 18:  6fec8ba7 Use reallocarrayf() instead of its pattern
11:  6066d8fd = 19:  15e1695c Use xreallocarray() instead of its pattern
12:  e7416270 = 20:  09518b7b Add safer allocation macros
13:  6eaf62e8 ! 21:  87aeb91d Use safer allocation macros
    @@ libmisc/idmapping.c
     +
     +#include "malloc.h"
      #include "prototypes.h"
    + #include "stpeprintf.h"
      #include "idmapping.h"
    - #if HAVE_SYS_CAPABILITY_H
     @@ libmisc/idmapping.c: struct map_range *get_map_ranges(int ranges, int argc, char **argv)
                return NULL;
        }
    @@ libmisc/idmapping.c: void write_mapping(int proc_dir_fd, int ranges, const struc
        bufsize = ranges * ((ULONG_DIGITS + 1) * 3);
     -  pos = buf = xmalloc(bufsize);
     +  pos = buf = XMALLOCARRAY(bufsize, char);
    +   end = buf + bufsize;
      
        /* Build the mapping command */
    -   mapping = mappings;
     
      ## libmisc/list.c ##
     @@
    @@ src/groupmod.c: static void prepare_failure_reports (void)
      #endif
        info_passwd.name  = group_name;
      
    --  info_group.audit_msg   = xmalloc (512);
    -+  info_group.audit_msg   = XMALLOCARRAY (512, char);
    +-  gr                     = xmalloc (512);
    ++  gr                     = XMALLOCARRAY(512, char);
    +   info_group.audit_msg   = gr;
    +   gr_end                 = gr + 512;
      #ifdef    SHADOWGRP
    --  info_gshadow.audit_msg = xmalloc (512);
    -+  info_gshadow.audit_msg = XMALLOCARRAY (512, char);
    +-  sgr                    = xmalloc (512);
    ++  sgr                    = XMALLOCARRAY(512, char);
    +   info_gshadow.audit_msg = sgr;
    +   sgr_end                = sgr + 512;
      #endif
    --  info_passwd.audit_msg  = xmalloc (512);
    -+  info_passwd.audit_msg  = XMALLOCARRAY (512, char);
    +-  pw                     = xmalloc (512);
    ++  pw                     = XMALLOCARRAY(512, char);
    +   info_passwd.audit_msg  = pw;
    +   pw_end                 = pw + 512;
      
    -   (void) snprintf (info_group.audit_msg, 512,
    -                    "changing %s; ", gr_dbname ());
     
      ## src/groups.c ##
     @@
14:  ea048a41 = 22:  c86a6b73 Fix use-after-free of pointer after realloc(3)
```

libmisc/env.c
alejandro-colomar: Because if realloc moves the memory, old pointers are made invalid (which is the use-after-free bug I mention in the commit message).  Technically, the pointer contains an indeterminate value, and it shouldn't be read.
Reading the pointer is Undefined Behavior, AFAIR.

See also:
<https://inbox.sourceware.org/gcc/8ed6d28c-69dc-fed8-5ab5-99f685f06fac@gmail.com/T/#u>
<https://stackoverflow.com/questions/26072752/realloc-dangling-pointers-and-undefined-behavior>

After reading the last link, I realized that it's worse than that: even if the new object has the same address, it's still a new object, so comparing is still Undefined Behavior.  So, this fix is not really complete, but I'm not sure how to fix it completely :(

alejandro-colomar: If realloc(3) succeeds, it creates a new object, even if it may have the same address as the old one.  The lifetime of old objects has expired.  So, all pointers need to be updated to be derived from the return value of realloc(3).

lib/malloc.h
alejandro-colomar: No.  It's GNU C, and available in GCC and Clang.  ISO C23 will likely add `auto` with the same meaning, AFAIK (and hopefully deprecate the old meaning of `auto`).

alejandro-colomar: BTW, another GNU extension used here (also available at least in GCC and Clang) is `({ ... })`: <https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html>

lib/sgetgrent.c
```diff
@@ -37,22 +37,16 @@ static char **list (char *s)
...
 
 	i = 0;
 	for (;;) {
 		/* check if there is room for another pointer (to a group
 		   member name, or terminating NULL).  */
 		if (i >= size) {
 			size = i + 100;	/* at least: i + 1 */
-			rbuf = realloc (members, size * sizeof (char *));
-			if (!rbuf) {
-				free (members);
```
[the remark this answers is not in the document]
alejandro-colomar: That's why I used realloc_f_(3), which is like realloc(3) but it free(3)s the memory on failure.

See: <https://manpages.debian.org/testing/libbsd-dev/reallocf.3.en.html>

lib/subordinateio.c
alejandro-colomar: Fixed

alejandro-colomar:
Changes:

-  whitespace fix [@ikerexxe ]

```diff
$ git range-diff c86a6b73...alloc 
1:  d9da1bd0 ! 1:  648fc0ad Use *array() allocation functions where appropriate
    @@ lib/subordinateio.c: static bool append_range(struct subid_range **ranges, const
        } else {
                struct subid_range *alloced;
     -          alloced = realloc(*ranges, (n + 1) * (sizeof(struct subid_range)));
    -+          alloced = reallocarray(*ranges, n+1, sizeof(struct subid_range));
    ++          alloced = reallocarray(*ranges, n + 1, sizeof(struct subid_range));
                if (!alloced)
                        return false;
                *ranges = alloced;
2:  6fec8ba7 = 2:  8710f813 Use reallocarrayf() instead of its pattern
3:  15e1695c = 3:  9fffea5d Use xreallocarray() instead of its pattern
4:  09518b7b = 4:  a6d04d28 Add safer allocation macros
5:  87aeb91d ! 5:  02ef2cc4 Use safer allocation macros
    @@ lib/subordinateio.c: static bool have_range(struct commonio_db *db,
                        return false;
        } else {
                struct subid_range *alloced;
    --          alloced = reallocarray(*ranges, n+1, sizeof(struct subid_range));
    +-          alloced = reallocarray(*ranges, n + 1, sizeof(struct subid_range));
     +          alloced = REALLOCARRAY(*ranges, n + 1, struct subid_range);
                if (!alloced)
                        return false;
6:  c86a6b73 = 6:  5d215a3a Fix use-after-free of pointer after realloc(3)
```

alejandro-colomar:
> Two major comments apart from the ones inline. The first one is related to the `malloc.*` files naming. As they contain more functionality apart from `malloc` I wonder if we should rename it to something more generic.

I thought about calling it `alloc.c`.  Does it sound better to you?

> 
> The second one is related to the new macros provided in [e741627](https://github.com/shadow-maint/shadow/commit/e7416270b22e6110b4dbb8ae6d05c32b50400f95). Wouldn't it be better to add this functionality to the functions themselves? We are already adding an abstraction and I'd prefer to avoid having a second one.

I put most of the functionality in the functions, as they are safer, and then only put in macros what can't be put there, which is the cast and `sizeof(type)`.  It's not possible to implement that as functions in C.  Type-generic code can only go in macros (C++ would have templates for that, although they don't convince me very much).

> 
> By the way, I didn't review properly the last three commits because I wanted to tackle my previous concern first.

Sure.

alejandro-colomar:
Changes:

-  Rebase to master

alejandro-colomar:
Changes:

-  Move malloc.c to alloc.c (also, do it in a separate commit, to simplify patches)
-  Fix typo in intermediate commit (it was not a problem, since it had been fixed in another commit)

```
$ git range-diff 5956cea1..5d215a3a master..alloc 
 1:  231aa1cc =  1:  8b33eced Rely on realloc(NULL, ...) being equivalent to malloc(...)
 2:  820e65a4 =  2:  303882fb malloc(3) already sets errno to ENOMEM
 3:  27681f5e =  3:  b055d69c Use reallocf(3) instead of its pattern
 4:  1ca3a789 !  4:  1cc8ce7d Use reallocarray(3) instead of its pattern
    @@ libmisc/env.c: void addenv (const char *string, /*@null@*/const char *value)
     -          newsize = (newenvc + NEWENVP_STEP) * sizeof (char *);
     -          __newenvp = (char **) realloc (newenvp, newsize);
     +          __newenvp = (char **) reallocarray (newenvp, newenvc + NEWENVP_STEP,
    -+                                              sizeof (char *);
    ++                                              sizeof (char *));
      
                if (NULL != __newenvp) {
                        /*
 5:  56e751b2 =  5:  1e37217a Use calloc(3) instead of its pattern
 -:  -------- >  6:  3fedd07c libmisc: Move xmalloc.c to alloc.c
 6:  7c4f8c73 !  7:  fdfeb00b Add safer allocation functions
    @@ Metadata
     
      ## Commit message ##
    -    Add safer allocation functions
    +    libmisc: Add safer allocation functions
     
    - ## lib/malloc.h (new) ##
    + ## lib/alloc.h ##
     @@
     +/*
     + * SPDX-FileCopyrightText:  2023, Alejandro Colomar <alx@kernel.org>
    @@ lib/malloc.h (new)
     +
     +#include <assert.h>
     +#include <errno.h>
    -+#include <stddef.h>
    + #include <stddef.h>
     +#include <stdint.h>
     +#include <stdlib.h>
     +
    @@ lib/malloc.h (new)
     +  return strcpy(XMALLOCARRAY(strlen(str) + 1, char), str);
     +}
     +
    -+
    -+#endif  // include guard
    -
    - ## lib/prototypes.h ##
    -@@ lib/prototypes.h: extern int setutmp (struct utmp *ut);
    - /* valid.c */
    - extern bool valid (const char *, const struct passwd *);
      
     -/* xmalloc.c */
     -extern /*@maynotreturn@*/ /*@only@*//*@out@*//*@notnull@*/void *xmalloc (size_t size)
     -  /*@ensures MaxSet(result) == (size - 1); @*/;
     -extern /*@maynotreturn@*/ /*@only@*//*@notnull@*/char *xstrdup (const char *);
    --
    - /* xgetpwnam.c */
    - extern /*@null@*/ /*@only@*/struct passwd *xgetpwnam (const char *);
    - /* xgetpwuid.c */
    -
    - ## libmisc/Makefile.am ##
    -@@ libmisc/Makefile.am: libmisc_la_SOURCES = \
    -   list.c log.c \
    -   loginprompt.c \
    -   mail.c \
    -+  malloc.c \
    -   mempcpy.c \
    -   motd.c \
    -   myname.c \
    -@@ libmisc/Makefile.am: libmisc_la_SOURCES = \
    -   xgetgrnam.c \
    -   xgetgrgid.c \
    -   xgetspnam.c \
    --  xmalloc.c \
    -   yesno.c
    - 
    - if WITH_BTRFS
    ++#endif  // include guard
     
    - ## libmisc/xmalloc.c => libmisc/malloc.c ##
    + ## libmisc/alloc.c ##
     @@
       * SPDX-FileCopyrightText: 1996 - 1998, Marek Michałkiewicz
       * SPDX-FileCopyrightText: 2003 - 2006, Tomasz Kłoczko
    @@ libmisc/xmalloc.c => libmisc/malloc.c
      #ident "$Id$"
      
     -#include <stdio.h>
    -+#include "malloc.h"
    ++#include "alloc.h"
     +
      #include <errno.h>
     +#include <stddef.h>
 7:  82dd2db7 =  8:  9ecfbc12 Use xcalloc(3) instead of its pattern
 8:  52c661cd <  -:  -------- Use the new header for xstrdup()
 9:  648fc0ad =  9:  9e64c102 Use *array() allocation functions where appropriate
10:  8710f813 = 10:  f7114806 Use reallocarrayf() instead of its pattern
11:  9fffea5d = 11:  a5cbc5f5 Use xreallocarray() instead of its pattern
12:  a6d04d28 ! 12:  775f0978 Add safer allocation macros
    @@ Metadata
     
      ## Commit message ##
    -    Add safer allocation macros
    +    libmisc: Add safer allocation macros
     
         This macros have several benefits over the standard functions:
     
    @@ Commit message
         Cc: Valentin V. Bartenev <vbartenev@gmail.com>
     
    - ## lib/malloc.h ##
    + ## lib/alloc.h ##
     @@
      #include "defines.h"
      
13:  02ef2cc4 ! 13:  d15ee245 Use safer allocation macros
    @@ lib/commonio.c
      #include <stdio.h>
      #include <signal.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "nscd.h"
      #include "sssd.h"
      #ifdef WITH_TCB
    @@ lib/getdef.c
      #include <libeconf.h>
      #endif
     +
    ++#include "alloc.h"
      #include "getdef.h"
    -+#include "malloc.h"
      #include "shadowlog_internal.h"
     +
      /*
    @@ lib/groupio.c
      #include <assert.h>
      #include <stdio.h>
      
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "defines.h"
      #include "commonio.h"
    @@ lib/groupmem.c
      
      #ident "$Id$"
      
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "defines.h"
      #include "groupio.h"
    @@ lib/gshadow.c
      #include <stdio.h>
      #include <string.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "defines.h"
     +
    @@ lib/nss.c
      #include <ctype.h>
      #include <stdatomic.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "../libsubid/subid.h"
      #include "shadowlog_internal.h"
    @@ lib/pwmem.c
      
      #include <stdio.h>
     +
    ++#include "alloc.h"
      #include "defines.h"
    -+#include "malloc.h"
      #include "prototypes.h"
      #include "pwio.h"
    - 
     @@
      {
        struct passwd *pw;
    @@ lib/run_part.c
      #include <unistd.h>
      #include <lib/prototypes.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "run_part.h"
      #include "shadowlog_internal.h"
      
    @@ lib/sgetgrent.c
      #include <sys/types.h>
      #include <grp.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "defines.h"
      #include "prototypes.h"
      
    @@ lib/sgroupio.c
      
      #ident "$Id$"
      
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "defines.h"
      #include "commonio.h"
    @@ lib/shadowmem.c
      #include <shadow.h>
      #include <stdio.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "shadowio.h"
      
      /*@null@*/ /*@only@*/struct spwd *__spw_dup (const struct spwd *spent)
    @@ lib/sssd.c
      #include <sys/wait.h>
      #include <sys/types.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "exitcodes.h"
      #include "defines.h"
      #include "prototypes.h"
    @@ lib/subordinateio.c
      #include <ctype.h>
      #include <fcntl.h>
      
    -+#include "malloc.h"
    ++#include "alloc.h"
     +
      #define ID_SIZE 31
      
    @@ libmisc/addgrps.c
      #include <grp.h>
      #include <errno.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "shadowlog.h"
      
      #ident "$Id$"
    @@ libmisc/agetpass.c
      
      #ident "$Id$"
      
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      
      
    @@ libmisc/agetpass.c: agetpass(const char *prompt)
      
     
      ## libmisc/copydir.c ##
    -@@
    - #include <sys/time.h>
    - #include <fcntl.h>
    - #include <stdio.h>
    -+
    -+#include "malloc.h"
    - #include "prototypes.h"
    - #include "defines.h"
    - #ifdef WITH_SELINUX
     @@ libmisc/copydir.c: static /*@exposed@*/ /*@null@*/struct link_name *check_link (const char *name, c
                return NULL;
        }
    @@ libmisc/copydir.c: static int copy_symlink (const struct path_info *src, const s
                                 oldlink + strlen (src_orig));
     
      ## libmisc/env.c ##
    -@@
    - #include <stdio.h>
    - #include <stdlib.h>
    - #include <string.h>
    -+
    -+#include "malloc.h"
    - #include "prototypes.h"
    - #include "defines.h"
    - #include "shadowlog.h"
     @@ libmisc/env.c: static const char *const noslash[] = {
       */
      void initenv (void)
    @@ libmisc/env.c: void addenv (const char *string, /*@null@*/const char *value)
                 */
      
     -          __newenvp = (char **) reallocarray (newenvp, newenvc + NEWENVP_STEP,
    --                                              sizeof (char *);
    +-                                              sizeof (char *));
     +          __newenvp = REALLOCARRAY(newenvp, newenvc + NEWENVP_STEP, char *);
      
                if (NULL != __newenvp) {
    @@ libmisc/find_new_gid.c
      #include <stdio.h>
      #include <errno.h>
      
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "groupio.h"
      #include "getdef.h"
    @@ libmisc/find_new_uid.c
      #include <stdio.h>
      #include <errno.h>
      
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "pwio.h"
      #include "getdef.h"
    @@ libmisc/idmapping.c
      #include <stdlib.h>
      #include <stdio.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "stpeprintf.h"
      #include "idmapping.h"
    @@ libmisc/list.c
      
      #include <assert.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "defines.h"
      /*
    @@ libmisc/loginprompt.c
      #include <signal.h>
      #include <ctype.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "defines.h"
      #include "getdef.h"
    @@ libmisc/mail.c
      #include <stdio.h>
      #include <string.h>
      
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "getdef.h"
      
      #ident "$Id$"
    @@ libmisc/obscure.c
      #include <ctype.h>
      #include <stdio.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "defines.h"
      #include "getdef.h"
    @@ libmisc/pam_pass_non_interactive.c
      #include <stdlib.h>
      #include <security/pam_appl.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "shadowlog.h"
      
    @@ libmisc/prefix_flag.c
      #include <assert.h>
     +
      #include "defines.h"
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      /*@-exitarg@*/
      #include "exitcodes.h"
    @@ libmisc/setupenv.c
      #include <stdio.h>
      #include <ctype.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "defines.h"
      #include <pwd.h>
    @@ libmisc/utmp.c
      #include <netdb.h>
      #include <stdio.h>
      
    -+#include "malloc.h"
    ++#include "alloc.h"
     +
      #ident "$Id$"
      
    @@ libmisc/xgetXXbyYY.c
      #include <stdio.h>
      #include <errno.h>
     +
    -+#include "malloc.h"
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "shadowlog.h"
      
    @@ src/gpasswd.c
      #include <stdio.h>
      #include <sys/types.h>
     +
    ++#include "alloc.h"
      #include "defines.h"
      #include "groupio.h"
    -+#include "malloc.h"
      #include "nscd.h"
    - #include "sssd.h"
    - #include "prototypes.h"
     @@ src/gpasswd.c: static void get_group (struct group *gr)
      
                        sg->sg_mem = dup_list (gr->gr_mem);
    @@ src/gpasswd.c: static void get_group (struct group *gr)
     
      ## src/groupmems.c ##
     @@
    + #include "pam_defs.h"
      #endif                            /* USE_PAM */
      #include <pwd.h>
    ++
    ++#include "alloc.h"
      #include "defines.h"
    -+#include "malloc.h"
      #include "prototypes.h"
      #include "groupio.h"
    - #ifdef SHADOWGRP
     @@ src/groupmems.c: static void add_user (const char *user,
                        static struct sgrp sgrent;
                        sgrent.sg_name = xstrdup (newgrp->gr_name);
    @@ src/groupmems.c: static void purge_members (const struct group *grp)
     
      ## src/groupmod.c ##
     @@
    + #include <pwd.h>
    + #endif                            /* USE_PAM */
    + #endif                            /* ACCT_TOOLS_SETUID */
    ++
    ++#include "alloc.h"
      #include "chkname.h"
      #include "defines.h"
      #include "groupio.h"
    -+#include "malloc.h"
    - #include "pwio.h"
    - #include "nscd.h"
    - #include "sssd.h"
     @@ src/groupmod.c: static void grp_update (void)
                        // requested to replace the existing groups
                        if (NULL != grp.gr_mem[0])
    @@ src/groups.c
      #include <pwd.h>
      #include <stdio.h>
     +
    ++#include "alloc.h"
      #include "defines.h"
    -+#include "malloc.h"
      #include "prototypes.h"
      #include "shadowlog.h"
     +
    @@ src/id.c
      #include <stdio.h>
      #include <sys/types.h>
     +
    ++#include "alloc.h"
      #include "defines.h"
    -+#include "malloc.h"
     +
      /* local function prototypes */
      static void usage (void);
    @@ src/login.c
      #include <sys/ioctl.h>
      #include <assert.h>
     +
    ++#include "alloc.h"
      #include "defines.h"
      #include "faillog.h"
      #include "failure.h"
    - #include "getdef.h"
    -+#include "malloc.h"
    - #include "prototypes.h"
    - #include "pwauth.h"
    - /*@-exitarg@*/
     @@ src/login.c: int main (int argc, char **argv)
      #ifdef RLOGIN
        if (rflg) {
    @@ src/newgrp.c
      #include <stdio.h>
      #include <assert.h>
     +
    ++#include "alloc.h"
      #include "defines.h"
      #include "getdef.h"
    -+#include "malloc.h"
      #include "prototypes.h"
    - /*@-exitarg@*/
    - #include "exitcodes.h"
     @@ src/newgrp.c: int main (int argc, char **argv)
        /* don't use getgroups(0, 0) - it doesn't work on some systems */
        i = 16;
    @@ src/newgrp.c: int main (int argc, char **argv)
     
      ## src/newusers.c ##
     @@
    - #include "defines.h"
    - #include "getdef.h"
    - #include "groupio.h"
    -+#include "malloc.h"
    - #include "nscd.h"
    - #include "sssd.h"
    - #include "pwio.h"
    + #include <ctype.h>
    + #include <errno.h>
    + #include <string.h>
    ++
    ++#include "alloc.h"
    + #ifdef ACCT_TOOLS_SETUID
    + #ifdef USE_PAM
    + #include "pam_defs.h"
     @@ src/newusers.c: int main (int argc, char **argv)
      #ifdef USE_PAM
                /* keep the list of user/password for later update by PAM */
    @@ src/passwd.c
      #include <sys/types.h>
      #include <time.h>
     +
    ++#include "alloc.h"
      #include "defines.h"
      #include "getdef.h"
    -+#include "malloc.h"
      #include "nscd.h"
    - #include "sssd.h"
    - #include "prototypes.h"
     @@ src/passwd.c: static char *update_crypt_pw (char *cp)
        }
      
    @@ src/su.c
      #include <fcntl.h>
      #endif                            /* !USE_PAM */
     +
    ++#include "alloc.h"
      #include "prototypes.h"
      #include "defines.h"
    -+#include "malloc.h"
      #include "pwauth.h"
    - #include "getdef.h"
    - #ifdef USE_PAM
     @@ src/su.c: static void execve_shell (const char *shellname,
                while (NULL != args[n_args]) {
                        n_args++;
    @@ src/useradd.c
      #include <time.h>
      #include <unistd.h>
     +
    ++#include "alloc.h"
      #include "chkname.h"
      #include "defines.h"
      #include "faillog.h"
    - #include "getdef.h"
    - #include "groupio.h"
    -+#include "malloc.h"
    - #include "nscd.h"
    - #include "sssd.h"
    - #include "prototypes.h"
     @@ src/useradd.c: static void get_defaults (void)
                int wlen;
      
    @@ src/useradd.c: int main (int argc, char **argv)
     
      ## src/userdel.c ##
     @@
    - #include "defines.h"
    - #include "getdef.h"
    - #include "groupio.h"
    -+#include "malloc.h"
    - #include "nscd.h"
    - #include "sssd.h"
    - #include "prototypes.h"
    + #include <sys/stat.h>
    + #include <sys/types.h>
    + #include <unistd.h>
    ++
    ++#include "alloc.h"
    + #ifdef ACCT_TOOLS_SETUID
    + #ifdef USE_PAM
    + #include "pam_defs.h"
     @@ src/userdel.c: static int remove_mailbox (void)
        }
      
    @@ src/usermod.c
      #include <sys/types.h>
      #include <time.h>
     +
    ++#include "alloc.h"
      #include "chkname.h"
      #include "defines.h"
      #include "faillog.h"
    - #include "getdef.h"
    - #include "groupio.h"
    -+#include "malloc.h"
    - #include "nscd.h"
    - #include "sssd.h"
    - #include "prototypes.h"
     @@ src/usermod.c: static int prepend_range(const char *str, struct ulong_range_list_entry **head)
        if (range.first > range.last)
                return 0;
    @@ src/vipw.c
      #include <unistd.h>
      #include <utime.h>
     +
    ++#include "alloc.h"
      #include "defines.h"
      #include "groupio.h"
    -+#include "malloc.h"
      #include "nscd.h"
    - #include "sssd.h"
    - #include "prototypes.h"
     @@ src/vipw.c: vipwedit (const char *file, int (*file_lock) (void), int (*file_unlock) (void))
                                continue;
                }
14:  5d215a3a = 14:  18655c0d Fix use-after-free of pointer after realloc(3)
```

libmisc/malloc.c
ikerexxe: so that's the reason? In that case I withdraw my comment

libmisc/env.c
ikerexxe: Understood. Thanks for the explanation

lib/alloc.h
```diff
@@ -20,6 +20,58 @@
...
+({                                                                            \
+	type    *p_;                                                          \
+	size_t  n_ = (n);                                                     \
+                                                                              \
+	if (n_ != 0 && n_ > SSIZE_MAX / n_)                                   \
+		p_ = NULL;                                                    \
+	else                                                                  \
+		p_ = alloca(sizeof(type) * n_);                               \
+                                                                              \
+	p_;                                                                   \
```
ikerexxe, line 71: If you wanted to align the `\` characters at the end of the line you missed your goal.

ikerexxe:
> I thought about calling it `alloc.c`. Does it sound better to you?

Yes :smile: 

> I put most of the functionality in the functions, as they are safer, and then only put in macros what can't be put in functions, which is the cast and `sizeof()` and `typeof()`. It's not possible to implement that as functions in C. Type-generic code can only go in macros (C++ would have templates for that, although they don't convince me very much).

C++ templates are so handy...

> > By the way, I didn't review properly the last three commits because I wanted to tackle my previous concern first.

Reviewed.

libmisc/malloc.c
alejandro-colomar: To be more precise, `xmalloc()` has been called that way historically (probably) because it was MALLOC or eXit (on failure).  The x comes from exit(3), and corresponds to the error path.  So yes, naming the label `x` makes sense in my head.

lib/alloc.h
alejandro-colomar: I align them to the 79 column on purpose :)

I learnt that from NGINX, where they do that.  The reason is that in a 80 column terminal you see that there's nothing to the right of the escape character.  If you put `\` in column 80, then you could accidentally have some character in column 81, and you might not see it, because it's after the right margin.  So if you put it in column 79, you still have 1 column empty, which is the proof that the escape indeed escapes the newline, and not some hidden character.

alejandro-colomar:
Rebased to master:

```
$ git range-diff 8b33eced^..gh/alloc master..alloc 
 1:  8b33eced =  1:  cbf47ca9 Rely on realloc(NULL, ...) being equivalent to malloc(...)
 2:  303882fb =  2:  a921c9db malloc(3) already sets errno to ENOMEM
 3:  b055d69c =  3:  4cc9c43c Use reallocf(3) instead of its pattern
 4:  1cc8ce7d =  4:  27768bd7 Use reallocarray(3) instead of its pattern
 5:  1e37217a =  5:  2d558082 Use calloc(3) instead of its pattern
 6:  3fedd07c =  6:  7de755fc libmisc: Move xmalloc.c to alloc.c
 7:  fdfeb00b =  7:  9431db74 libmisc: Add safer allocation functions
 8:  9ecfbc12 =  8:  f95b54f4 Use xcalloc(3) instead of its pattern
 9:  9e64c102 =  9:  9b6aca1f Use *array() allocation functions where appropriate
10:  f7114806 = 10:  4f7d7a2e Use reallocarrayf() instead of its pattern
11:  a5cbc5f5 = 11:  6b2e8392 Use xreallocarray() instead of its pattern
12:  775f0978 = 12:  5ab68fe9 libmisc: Add safer allocation macros
13:  d15ee245 ! 13:  1285b23b Use safer allocation macros
    @@ libmisc/agetpass.c
       *          consequence of calling getline(3).  That's a bug in glibc,
       *          which allows leaking prefixes of passwords in freed memory.
     @@ libmisc/agetpass.c: agetpass(const char *prompt)
    -   char    *pass;
    -   size_t  len;
    - 
    --  pass = malloc(PASS_MAX);
    -+  pass = MALLOCARRAY(PASS_MAX, char);
    +    * Let's add one more byte, and if the password uses it, it
    +    * means the introduced password was longer than PASS_MAX.
    +    */
    +-  pass = malloc(PASS_MAX + 2);
    ++  pass = MALLOCARRAY(PASS_MAX + 2, char);
        if (pass == NULL)
                return NULL;
      
14:  18655c0d = 14:  9002f1f7 Fix use-after-free of pointer after realloc(3)
```

ikerexxe:
I have my doubts about adding a new GNU extension. Everything else seems fine to me, but there are a lot of changes and I may have missed something. Because of those two things I think someone else should review the PR.

alejandro-colomar:
Let's see what @hallyn thinks about it :)

alejandro-colomar:
Changes:

-  Add include guard in intermediate commit.

```
$ git range-diff master..gh/alloc master..alloc 
 1:  cbf47ca9 =  1:  cbf47ca9 Rely on realloc(NULL, ...) being equivalent to malloc(...)
 2:  a921c9db =  2:  a921c9db malloc(3) already sets errno to ENOMEM
 3:  4cc9c43c =  3:  4cc9c43c Use reallocf(3) instead of its pattern
 4:  27768bd7 =  4:  27768bd7 Use reallocarray(3) instead of its pattern
 5:  2d558082 =  5:  2d558082 Use calloc(3) instead of its pattern
 6:  7de755fc !  6:  bac52609 libmisc: Move xmalloc.c to alloc.c
    @@ Commit message
     
      ## lib/alloc.h (new) ##
     @@
    ++#ifndef SHADOW_INCLUDE_LIB_MALLOC_H_
    ++#define SHADOW_INCLUDE_LIB_MALLOC_H_
    ++
    ++
     +#include <stddef.h>
     +
     +/* xmalloc.c */
     +extern /*@maynotreturn@*/ /*@only@*//*@out@*//*@notnull@*/void *xmalloc (size_t size)
     +  /*@ensures MaxSet(result) == (size - 1); @*/;
     +extern /*@maynotreturn@*/ /*@only@*//*@notnull@*/char *xstrdup (const char *);
    ++
    ++
    ++#endif  // include guard
     
      ## lib/prototypes.h ##
     @@ lib/prototypes.h: extern int setutmp (struct utmp *ut);
 7:  9431db74 !  7:  fb2bbb75 libmisc: Add safer allocation functions
    @@ lib/alloc.h
     + */
     +
     +
    -+#ifndef SHADOW_INCLUDE_LIB_MALLOC_H_
    -+#define SHADOW_INCLUDE_LIB_MALLOC_H_
    -+
    -+
    + #ifndef SHADOW_INCLUDE_LIB_MALLOC_H_
    + #define SHADOW_INCLUDE_LIB_MALLOC_H_
    + 
    + 
     +#include <config.h>
     +
     +#include <assert.h>
    @@ lib/alloc.h
     +  return q;
     +}
     +
    -+
    + 
    +-/* xmalloc.c */
    +-extern /*@maynotreturn@*/ /*@only@*//*@out@*//*@notnull@*/void *xmalloc (size_t size)
    +-  /*@ensures MaxSet(result) == (size - 1); @*/;
    +-extern /*@maynotreturn@*/ /*@only@*//*@notnull@*/char *xstrdup (const char *);
     +inline char *
     +xstrdup(const char *str)
     +{
     +  return strcpy(XMALLOCARRAY(strlen(str) + 1, char), str);
     +}
    -+
      
    --/* xmalloc.c */
    --extern /*@maynotreturn@*/ /*@only@*//*@out@*//*@notnull@*/void *xmalloc (size_t size)
    --  /*@ensures MaxSet(result) == (size - 1); @*/;
    --extern /*@maynotreturn@*/ /*@only@*//*@notnull@*/char *xstrdup (const char *);
    -+#endif  // include guard
    + 
    + #endif  // include guard
     
      ## libmisc/alloc.c ##
     @@
 8:  f95b54f4 =  8:  16cbd0f1 Use xcalloc(3) instead of its pattern
 9:  9b6aca1f =  9:  bdd5c8d1 Use *array() allocation functions where appropriate
10:  4f7d7a2e = 10:  9e00e7b5 Use reallocarrayf() instead of its pattern
11:  6b2e8392 = 11:  b6f9bb55 Use xreallocarray() instead of its pattern
12:  5ab68fe9 = 12:  9d6f4149 libmisc: Add safer allocation macros
13:  1285b23b = 13:  a383e44e Use safer allocation macros
14:  9002f1f7 = 14:  7339875c Fix use-after-free of pointer after realloc(3)
```

lib/sgetgrent.c
```diff
@@ -45,14 +45,7 @@ static char **list (char *s)
...
 			size = i + 100;	/* at least: i + 1 */
-			if (members) {
-				rbuf =
-				    realloc (members, size * sizeof (char *));
-			} else {
-				/* for old (before ANSI C) implementations of
-				   realloc() that don't handle NULL properly */
-				rbuf = malloc (size * sizeof (char *));
-			}
+			rbuf = realloc (members, size * sizeof (char *));
```
hallyn: Ah, yes, been doing this in my code for ages.  Thanks.

lib/groupio.c
```diff
@@ -313,7 +313,6 @@ static /*@null@*/struct commonio_entry *merge_group_entries (
 	new_line_len = strlen (gr1->line) + strlen (gr2->line) +1;
 	new_line = (char *)malloc (new_line_len + 1);
 	if (NULL == new_line) {
-		errno = ENOMEM;
```
hallyn, line 316: Hm, but the manpage doesn't guarantee that for malloc()

hallyn: (I'm probably ok with it, just curious what you have to say about it :-)

lib/sgetgrent.c
alejandro-colomar: :-)

lib/groupio.c
alejandro-colomar: I guess you refer to this?:

`malloc(3)`:
```
   Nonportable behavior
       The behavior of these functions when the  requested  size  is  zero  is
       glibc  specific;  other implementations may return NULL without setting
       errno, and portable POSIX programs should tolerate such behavior.   See
       realloc(3p).

       POSIX  requires  memory allocators to set errno upon failure.  However,
       the C standard does not require this, and applications portable to non-
       POSIX platforms should not assume this.
```

If so, POSIX requires setting errno on error.

Another thing is what happens when we try to allocate 0 bytes, which I hope we don't do, but even if we do, I'm not convinced it's an error, and POSIX isn't either:

`malloc(3posix)`:
```
RETURN VALUE
       Upon successful completion with size not equal to 0, malloc() shall re‐
       turn a pointer to the allocated space. If size is 0, either:

        *  A  null pointer shall be returned and errno may be set to an imple‐
           mentation-defined value, or

        *  A pointer to the allocated space shall be returned. The application
           shall ensure that the pointer is not used to access an object.

       Otherwise, it shall return a null pointer and set errno to indicate the
       error.
```

lib/sgetgrent.c
```diff
@@ -37,22 +37,16 @@ static char **list (char *s)
...
 		/* check if there is room for another pointer (to a group
 		   member name, or terminating NULL).  */
 		if (i >= size) {
 			size = i + 100;	/* at least: i + 1 */
-			rbuf = realloc (members, size * sizeof (char *));
-			if (!rbuf) {
-				free (members);
-				members = NULL;
-				size = 0;
+			members = reallocf (members, size * sizeof (char *));
```
hallyn: Ah, yes, this is a good use of reallocf, as there's no resulting side effect to the caller.

lib/groupio.c
hallyn: I'm just referring to the fact that the manpage does explicitly say that reallocarray sets errno to ENOMEM but does not say that about malloc.

But clearly a local testcase proves that it does, and so as long as you're comfortable with it please feel free to resolve this :)

lib/sgetgrent.c
alejandro-colomar: BTW, note for myself tomorrow, I need to say _before_ rather than _after_ in the commit message.  I was confused just now reading it myself (didn't remember what I meant).  :D

lib/groupio.c
alejandro-colomar: Maybe I have a too new manual page; can you please quote what you're seeing?  Mine says:

```
RETURN VALUE
       The malloc(), calloc(), realloc(), and reallocarray() functions  return
       a  pointer  to  the allocated memory, which is suitably aligned for any
       type that fits into the requested size or less.  On error, these  func‐
       tions  return  NULL  and  set  errno.
```

alejandro-colomar: > But clearly a local testcase proves that it does, and so as long as you're comfortable with it please feel free to resolve this :)

Yeah, I am.  I am rather confused about the manual page itself, and maybe I need to fix some inconsistency there :)

hallyn:
Huh.  I don't know where I was looking before, but now I see:

> SUSv2  requires  malloc(),  calloc(), and realloc() to set errno to ENOMEM upon failure.

( https://manpages.ubuntu.com/manpages/jammy/en/man3/malloc.3.html )

So all's good, thanks.

alejandro-colomar:
> Huh. I don't know where I was looking before, but now I see:
> SUSv2 requires malloc(), calloc(), and realloc() to set errno to ENOMEM upon failure.
> ( https://manpages.ubuntu.com/manpages/jammy/en/man3/malloc.3.html ) So all's good, thanks.

Thank you!

BTW, that text is a bit old for me :D  It was largely rewritten in

```
commit cfc381be29fb7b4f642b8b333882f22628112abe
Author: Paul Eggert <eggert@cs.ucla.edu>
Date:   Tue Aug 10 12:37:07 2021 -0700

    malloc.3: Modernize for glibc 2.34
    
    glibc has tightened up its rules for replacing the memory
    allocator.  I went through the malloc man page and looked for how
    it documented malloc() and related functions, and fixed
    discrepancies with glibc malloc() documentation and/or
    implementation.  I also reorganized the portability discussion so
    that portability issues can be seen more clearly.
    
    Signed-off-by: Michael Kerrisk <mtk.manpages@gmail.com>
```

libmisc/addgrps.c
```diff
@@ -88,14 +88,12 @@ int add_groups (const char *list)
 			fputs (_("Warning: too many groups\n"), shadow_logfd);
 			break;
 		}
-		tmp = (gid_t *) reallocarray (grouplist, (size_t)ngroups + 1, sizeof (GETGROUPS_T));
-		if (NULL == tmp) {
-			free (grouplist);
+		grouplist = (gid_t *) reallocarrayf (grouplist, (size_t)ngroups + 1, sizeof (GETGROUPS_T));
```
hallyn: I'm not terribly worried about this, but just thought I'd point out:  reallocarrayf on failure will call free(), and I see that free not setting errno on failure is promised in the future, it is not right now https://sourceware.org/bugzilla/show_bug.cgi?id=17924 .  So maybe reallocarrayf() should save and re-set errno.

alejandro-colomar: Hmm, that bug seems outdated.  The manual page was updated (I didn't check but I guess it was also @eggert in the same patch mentioned earlier) to say:

```
       The free() function returns no value, and preserves errno.

...

       free() preserved errno starting in glibc 2.33.
```

Paul, could you close the bug ticket?

lib/groupio.c
hallyn: Hm, I replied in email, but I don't see the reply here.  In any case - I've found text that does say errno will be set, so I'll resolve this, thanks.

hallyn:
> Let's see what @hallyn thinks about it :)

Mostly looks good to me, thanks - I still need to look through 12-14, but have to run now.

lib/groupio.c
alejandro-colomar: Email replies don't go threaded (github sucks in many ways).  I answered below :)

lib/alloc.h
```diff
@@ -20,6 +20,58 @@
...
+#define MALLOC(type)           MALLOCARRAY(1, type)
+#define XMALLOC(type)          XMALLOCARRAY(1, type)
+#define REALLOC(ptr, type)     REALLOCARRAY(ptr, 1, type)
+#define REALLOCF(ptr, type)    REALLOCARRAYF(ptr, 1, type)
+
+#define REALLOCARRAY(ptr, n, type)                                            \
+({                                                                            \
+	__auto_type  p_ = (ptr);                                              \
+                                                                              \
+	static_assert(__builtin_types_compatible_p(typeof(p_), type *), "");  \
```
hallyn, line 38: static_assert() appears to be a c++ thing?

lib/alloc.h
```diff
@@ -20,6 +20,58 @@
...
+#define MALLOCARRAY(n, type)   ((type *) mallocarray(n, sizeof(type)))
+#define XMALLOCARRAY(n, type)  ((type *) xmallocarray(n, sizeof(type)))
+
+#define ALLOCA(type)           ALLOCARRAY(1, type)
+#define MALLOC(type)           MALLOCARRAY(1, type)
+#define XMALLOC(type)          XMALLOCARRAY(1, type)
+#define REALLOC(ptr, type)     REALLOCARRAY(ptr, 1, type)
+#define REALLOCF(ptr, type)    REALLOCARRAYF(ptr, 1, type)
+
+#define REALLOCARRAY(ptr, n, type)                                            \
```
hallyn, line 34: Once these are merged, it might be worth having a CI step that looks for non-modified users so we can catch them before they go in.

lib/alloc.h
```diff
@@ -20,6 +20,58 @@
...
+                                                                              \
+	(type *) xreallocarray(p_, n, sizeof(type));                          \
+})
+
+#define ALLOCARRAY(n, type)                                                   \
+({                                                                            \
+	type    *p_;                                                          \
+	size_t  n_ = (n);                                                     \
+                                                                              \
+	if (n_ != 0 && n_ > SSIZE_MAX / n_)                                   \
```
hallyn, line 66: What exactly is the intent here?

libmisc/env.c
hallyn: No.  This conclusion cannot be correct.

The second link is asking about pointers *into* the realloc'd memory, so does not apply.

The first is I believe wrongly interpreted.  Note if I compile the exact same code now in the same way (on ubuntu 22.04), there are no warnings.  I'm convinced that was a compiler bugL most likely it thought %p was %s, or in any case somehow the piece that warned assumed you would be printing contents, not the pointer itself.

When you do:

```
char *a, *b = NULL;

a = realloc(b, 16);
```
After the realloc you cannot safely dereference b, but you can print the value of the pointer b all you want.

The code after the patch is correct as well, so whatever - but the original code is ok.  So I'm making a strong case here because, well, if I'm wrong - which I'm convinced I'm not - then I'd better understand why or I'll be dangerous.

alejandro-colomar: > No. This conclusion cannot be correct.

It can ;)

> 
> The second link is asking about pointers _into_ the realloc'd memory, so does not apply.

I didn't understand very well the question in the second link, but the answer is quite clear.  Pointers passed to `free(3)` or equivalently to a successful `realloc(3)` (`realloc(3)` frees the input pointer on success, and you can think of it as equivalent to `malloc(3)` && `memmove(3)` && `free(3)`) become invalid, since the object they pointed to has died.  A read of an invalid pointer is UB.

> 
> The first is I believe wrongly interpreted. Note if I compile the exact same code now in the same way (on ubuntu 22.04), there are no warnings.

Which code?  The one I posted on gcc@, or shadow?

Did you try `-Wuse-after-free=3`, as suggested by GCC developers?

I've sent a patch to put that warning in `-Wall` in GCC 14.
<https://sourceware.org/pipermail/gcc-patches/2023-February/612268.html>

> I'm convinced that was a compiler bugL most likely it thought %p was %s, or in any case somehow the piece that warned assumed you would be printing contents, not the pointer itself.

Yes, the warning that I saw in my post to GCC was due to `-Wuse-after-free=2`, which only warns about uses, and it considers passing a pointer to a function a plausible use (since it loses visibility of what happens inside the call).  However, that's not really a bug, since using the pointer value itself (now I'm not referring to the pointee) is also UB.  The actual bug was that GCC didn't warn about such bugs in `-Wall` nor `-Wextra`.

> 
> When you do:
> 
> ```
> char *a, *b = NULL;
> 
> a = realloc(b, 16);
> ```
> 
> After the realloc you cannot safely dereference b, but you can print the value of the pointer b all you want.

No, reading the value of the pointer is UB.  Please read the entire discussion in the GCC list (which prompted me to move the warning in `-Wall` to `=3`); that explains better.  Here's a tl;dr:

A pointer value is only valid as long as the pointee's lifetime exists.  Once the object it points to dies, the pointer value can't be read (you can't even print the pointer value with `%p`, to be clear), or thou shalt invoke the power of nasal demons :)

> 
> The code after the patch is correct as well, so whatever - but the original code is ok. So I'm making a strong case here because, well, if I'm wrong - which I'm convinced I'm not - 

Let's try to unconvince you :D

> then I'd better understand why or I'll be dangerous.

Yup, if you need some further clarification, please ask.  the GCC discussion is very interesting.  A codidact Q&A may also help (although I didn't make it about invalid pointers, but rather uninitialized variables, to simplify):
<https://software.codidact.com/posts/287905>

lib/alloc.h
alejandro-colomar: Avoid overflow due to multiplication.  Although, now that I think, `alloca(3)` can only allocate very little memory, so this test should be far from true.  I'll just remove it (now that I remember, I think I already considered this and removed it, but for some reason, it seems I didn't).

lib/alloc.h
alejandro-colomar: It's in C11:

`static_assert(3)`:
```
static_assert(3)           Library Functions Manual           static_assert(3)

NAME
       static_assert, _Static_assert - fail compilation if assertion is false

LIBRARY
       Standard C library (libc)

SYNOPSIS
       #include <assert.h>

       void static_assert(scalar constant-expression, const char *msg);

       /* Since C23: */
       void static_assert(scalar constant-expression);

DESCRIPTION
       This  macro is similar to assert(3), but it works at compile time, gen‐
       erating a compilation error (with an optional message) when  the  input
       is false (i.e., compares equal to zero).

       If the input is nonzero, no code is emitted.

       msg must be a string literal.  Since C23, this argument is optional.

       There's  a keyword, _Static_assert(), that behaves identically, and can
       be used without including <assert.h>.

RETURN VALUE
       No value is returned.

VERSIONS
       In C11, the second argument (msg) was mandatory; since C23, it  can  be
       omitted.

STANDARDS
       C11 and later.

EXAMPLES
       static_assert()  can't  be  used  in  some  places, like for example at
       global scope.  For that, a macro must_be() can be written in  terms  of
       static_assert().   The following program uses the macro to get the size
       of an array safely.

           #include <assert.h>
           #include <stddef.h>
           #include <stdint.h>
           #include <stdio.h>
           #include <stdlib.h>
           #include <string.h>

           /*
            * This macro behaves like static_assert(), failing to
            * compile if its argument is not true.  However, it always
            * returns 0, which allows using it everywhere an expression
            * can be used.
            */
           #define must_be(e)                                        \
           (                                                         \
               0 * (int) sizeof(                                     \
                   struct {                                          \
                       static_assert(e);                             \
                       int  ISO_C_forbids_a_struct_with_no_members;  \
                   }                                                 \
               )                                                     \
           )

           #define is_same_type(a, b)  \
               __builtin_types_compatible_p(typeof(a), typeof(b))

           #define is_array(arr)       (!is_same_type((arr), &*(arr)))
           #define must_be_array(arr)  must_be(is_array(arr))

           #define sizeof_array(arr)   (sizeof(arr) + must_be_array(arr))
           #define nitems(arr)         (sizeof((arr)) / sizeof((arr)[0]) \
                                        + must_be_array(arr))

           int     foo[10];
           int8_t  bar[sizeof_array(foo)];

           int
           main(void)
           {
               for (size_t i = 0; i < nitems(foo); i++) {
                   foo[i] = i;
               }

               memcpy(bar, foo, sizeof_array(bar));

               for (size_t i = 0; i < nitems(bar); i++) {
                   printf("%d,", bar[i]);
               }

               exit(EXIT_SUCCESS);
           }

SEE ALSO
       assert(3)

Linux man-pages (unreleased)        (date)                    static_assert(3)
```
<https://manpages.debian.org/testing/manpages-dev/static_assert.3.en.html>

hallyn: Hm, my 22.04 system does not have that manpage, making me very unhappy.

However, it seems to be compiling fine here so we'll roll with it.

(I probably shouldn't admit that whenever I see C11 I think "but that's so recent"... )

alejandro-colomar: > Hm, my 22.04 system does not have that manpage,

I wrote it on:

4a96ff90d - Fri, 2 Dec 2022 16:11:46 +0100 (3 months ago)
            _Static_assert.3, static_assert.3: Add documentation for static_assert() - Alejandro Colomar

so it's to be expected :)

> making me very unhappy.

You can always get them from upstream ;)

(or use Debian Sid, which picks them pretty fast.  I guess you maybe need some drivers for a laptop that only Ubuntu has though?)

> 
> However, it seems to be compiling fine here so we'll roll with it.
> 
> (I probably shouldn't admit that whenever I see C11 I think "but that's so recent"... )

I think the opposite when I see C89 (it fells like "nobody really took the time to update this in the last 24 years?" :p).
I learnt C with C11 already out there (it was around that year that I started with C, +/-1), so I feels natural to me.

libmisc/env.c
hallyn: This is my problem with standards.  They not infrequently come in after the fact to change semantics unnecessarily (as with certain freedesktop ones).

The leap from "accessing an object outside of its lifetime is undefined behavior" to "reading a pointer to an object outside of the lifetime of that object is undefined behavior" should not have been allowed.  It may be a good feature of other languages, but it was wrong here.

But I guess that wraps up the last of my concerns, so please feel free to merge the set any time.  Not sure if you were going to make a change for ALLOCA

hallyn (approved):

alejandro-colomar:
Changes:

-  Simplify ALLOCARRAY
```
$ git range-diff master..gh/alloc master..alloc 
 1:  cbf47ca9 =  1:  cbf47ca9 Rely on realloc(NULL, ...) being equivalent to malloc(...)
 2:  a921c9db =  2:  a921c9db malloc(3) already sets errno to ENOMEM
 3:  4cc9c43c =  3:  4cc9c43c Use reallocf(3) instead of its pattern
 4:  27768bd7 =  4:  27768bd7 Use reallocarray(3) instead of its pattern
 5:  2d558082 =  5:  2d558082 Use calloc(3) instead of its pattern
 6:  bac52609 =  6:  bac52609 libmisc: Move xmalloc.c to alloc.c
 7:  fb2bbb75 =  7:  fb2bbb75 libmisc: Add safer allocation functions
 8:  16cbd0f1 =  8:  16cbd0f1 Use xcalloc(3) instead of its pattern
 9:  bdd5c8d1 =  9:  bdd5c8d1 Use *array() allocation functions where appropriate
10:  9e00e7b5 = 10:  9e00e7b5 Use reallocarrayf() instead of its pattern
11:  b6f9bb55 = 11:  b6f9bb55 Use xreallocarray() instead of its pattern
12:  9d6f4149 ! 12:  a7ac2a00 libmisc: Add safer allocation macros
    @@ lib/alloc.h
      #include "defines.h"
      
      
    ++#define ALLOCARRAY(n, type)    ((type *) alloca(sizeof(type) * (n)))
     +#define CALLOC(n, type)        ((type *) calloc(n, sizeof(type)))
     +#define XCALLOC(n, type)       ((type *) xcalloc(n, sizeof(type)))
     +#define MALLOCARRAY(n, type)   ((type *) mallocarray(n, sizeof(type)))
    @@ lib/alloc.h
     +  (type *) xreallocarray(p_, n, sizeof(type));                          \
     +})
     +
    -+#define ALLOCARRAY(n, type)                                                   \
    -+({                                                                            \
    -+  type    *p_;                                                          \
    -+  size_t  n_ = (n);                                                     \
    -+                                                                              \
    -+  if (n_ != 0 && n_ > SSIZE_MAX / n_)                                   \
    -+          p_ = NULL;                                                    \
    -+  else                                                                  \
    -+          p_ = alloca(sizeof(type) * n_);                               \
    -+                                                                              \
    -+  p_;                                                                   \
    -+})
    -+
     +
      ATTR_MALLOC(free)
      inline void *xmalloc(size_t size);
13:  a383e44e = 13:  5282a360 Use safer allocation macros
14:  7339875c = 14:  edf3a6ab Fix use-after-free of pointer after realloc(3)
```

libmisc/addgrps.c
alejandro-colomar: My bad; it seems already closed: 

```
Status: RESOLVED FIXED
```

lib/sgetgrent.c
alejandro-colomar: Fixed.

alejandro-colomar:
Changes:

-  Fix commit message

```
$ git range-diff master..gh/alloc master..alloc 
 1:  cbf47ca9 =  1:  cbf47ca9 Rely on realloc(NULL, ...) being equivalent to malloc(...)
 2:  a921c9db =  2:  a921c9db malloc(3) already sets errno to ENOMEM
 3:  4cc9c43c !  3:  eac5a5ff Use reallocf(3) instead of its pattern
    @@ Metadata
      ## Commit message ##
         Use reallocf(3) instead of its pattern
     
    -    In addition, don't set local variables just after return.
    +    In addition, don't set local variables just before return.
     
 4:  27768bd7 =  4:  42f13e8d Use reallocarray(3) instead of its pattern
 5:  2d558082 =  5:  18b3d83a Use calloc(3) instead of its pattern
 6:  bac52609 =  6:  c431a72c libmisc: Move xmalloc.c to alloc.c
 7:  fb2bbb75 =  7:  63f959f7 libmisc: Add safer allocation functions
 8:  16cbd0f1 =  8:  a85b0bda Use xcalloc(3) instead of its pattern
 9:  bdd5c8d1 =  9:  0bb64e37 Use *array() allocation functions where appropriate
10:  9e00e7b5 = 10:  8f706fa9 Use reallocarrayf() instead of its pattern
11:  b6f9bb55 = 11:  061a04d5 Use xreallocarray() instead of its pattern
12:  a7ac2a00 = 12:  186ca04e libmisc: Add safer allocation macros
13:  5282a360 = 13:  9646f692 Use safer allocation macros
14:  edf3a6ab = 14:  11d92302 Fix use-after-free of pointer after realloc(3)
```

libmisc/env.c
alejandro-colomar: Thanks for reviewing!

Re: pointers: yeah, at least for logging, it would have been easier to just allow it.  Maybe comparing them to other pointers would still be UB since you can only compare pointers derived from the same object, but for printing a value, it would make sense to allow it. You can still memcpy the representation into a `uintptr_t` (according to what Martin said in the list), but it's still weird.  Also, I'm not sure about it 100%.

ALLOCA stuff fixed.  Feel free to merge.

alejandro-colomar: @ikerexxe Serge seems happy with the patch set, would you like to merge?

tests/libsubid/04_nss/libsubid_zzz.c
```diff
@@ -76,7 +76,7 @@ static uid_t getnamuid(const char *name) {
 }
 
 static int alloc_uid(uid_t **uids, uid_t id) {
-	*uids = malloc(sizeof(uid_t));
+	*uids = MALLOC(uid_t);
```
hallyn, line 79: Oh, this is actually breaking build of libsubid_zzz during tests.

Not sure yet whether just adding #include "alloc.h" at the top suffices.

alejandro-colomar: Yes, that should work.

alejandro-colomar: I'm not sure if that library links to libmisc, though, which might be a problem if not.

PR #650, opened 2023-02-06
Fix VPATH build

nertpinx:
When trying to build shadow in a different directory I stumbled upon few issues, this commit aims to fix all of them:

- The `subid.h` file is generated and hence in the build directory and not in the source directory, so use `$(builddir)` instead of `$(srcdir)`.

- Using `$<` instead of filenames utilises autotools to locate the files in either the source or build directory automatically.

- `xsltproc` needs to access the files in login.defs.d in either the source directory or the symlink in a language subdirectory, but it does not interpret the `--path` as prefix of the entity path, but rather a path under which to locate the basename of the entity from the XML file.  So specify the whole path to login.defs.d.

- The above point could be used to make the symlinks of login.defs.d and entity path specifications in the XMLs obsolete, but I trying not to propose possibly disrupting patches, so for the sake of simplicity just specify `$(srcdir)` when creating the symlink.


nertpinx:
Sorry, forgot S-o-b in the commit message, hence the update with no difference.

ikerexxe:
LGTM. Thanks!

PR #651, opened 2023-02-06
find_new_[ug]id: Skip over IDs that could overflow to -1

nertpinx:
Due to compatibility reasons neither UID nor GID should overflow to -1, be it on systems with 16-bit or 32-bit uid_t/gid_t since -1 is special.  It has different meaning for example as a chown(2) parameter, could become 65535 in programs compiled when uid_t was 16bit and possibly more issues.

### DIFF
--- a/libmisc/find_new_gid.c
+++ b/libmisc/find_new_gid.c
@@ -99,6 +99,7 @@ static int get_ranges (bool sys_group, gid_t *min_id, gid_t *max_id,
  *
  * On success, return 0
  * If the ID is in use, return EEXIST
+ * If the ID might clash with -1, return EINVAL
  * If the ID is outside the range, return ERANGE
  * In other cases, return errno from getgrgid()
  */
@@ -110,6 +111,11 @@ static int check_gid (const gid_t gid,
 	/* First test that the preferred ID is in the range */
 	if (gid < gid_min || gid > gid_max) {
 		return ERANGE;
+	}
+
+	/* Check for compatibility with 16b and 32b gid_t error codes */
+	if (gid == UINT16_MAX || gid == UINT32_MAX) {
+		return EINVAL;
 	}
 
 	/*
@@ -183,10 +189,10 @@ int find_new_gid (bool sys_group,
 			 * gr_locate_gid() found the GID in an as-yet uncommitted
 			 * entry. We'll proceed below and auto-set a GID.
 			 */
-		} else if (result == EEXIST || result == ERANGE) {
+		} else if (result == EEXIST || result == ERANGE || result == EINVAL) {
 			/*
 			 * Continue on below. At this time, we won't
-			 * treat these two cases differently.
+			 * treat these three cases differently.
 			 */
 		} else {
 			/*
@@ -297,8 +303,11 @@ int find_new_gid (bool sys_group,
 				*gid = id;
 				free (used_gids);
 				return 0;
-			} else if (result == EEXIST) {
-				/* This GID is in use, we'll continue to the next */
+			} else if (result == EEXIST || result == EINVAL) {
+				/*
+				 * This GID is in use or unusable, we'll
+				 * continue to the next.
+				 */
 			} else {
 				/*
 				 * An unexpected error occurred.
@@ -340,8 +349,11 @@ int find_new_gid (bool sys_group,
 					*gid = id;
 					free (used_gids);
 					return 0;
-				} else if (result == EEXIST) {
-					/* This GID is in use, we'll continue to the next */
+				} else if (result == EEXIST || result == EINVAL) {
+					/*
+					 * This GID is in use or unusable, we'll
+					 * continue to the next.
+					 */
 				} else {
 					/*
 					 * An unexpected error occurred.
@@ -400,8 +412,11 @@ int find_new_gid (bool sys_group,
 				*gid = id;
 				free (used_gids);
 				return 0;
-			} else if (result == EEXIST) {
-				/* This GID is in use, we'll continue to the next */
+			} else if (result == EEXIST || result == EINVAL) {
+				/*
+				 * This GID is in use or unusable, we'll
+				 * continue to the next.
+				 */
 			} else {
 				/*
 				 * An unexpected error occurred.
@@ -443,8 +458,11 @@ int find_new_gid (bool sys_group,
 					*gid = id;
 					free (used_gids);
 					return 0;
-				} else if (result == EEXIST) {
-					/* This GID is in use, we'll continue to the next */
+				} else if (result == EEXIST || result == EINVAL) {
+					/*
+					 * This GID is in use or unusable, we'll
+					 * continue to the next.
+					 */
 				} else {
 					/*
 					 * An unexpected error occurred.
--- a/libmisc/find_new_uid.c
+++ b/libmisc/find_new_uid.c
@@ -99,6 +99,7 @@ static int get_ranges (bool sys_user, uid_t *min_id, uid_t *max_id,
  *
  * On success, return 0
  * If the ID is in use, return EEXIST
+ * If the ID might clash with -1, return EINVAL
  * If the ID is outside the range, return ERANGE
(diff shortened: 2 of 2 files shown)


libmisc/find_new_gid.c
```diff
@@ -112,6 +113,11 @@ static int check_gid (const gid_t gid,
 		return ERANGE;
 	}
 
+  /* Check for compatibility with 16b and 32b gid_t error codes */
```
ikerexxe, line 116: Indentation issue

libmisc/find_new_gid.c
```diff
@@ -183,7 +189,7 @@ int find_new_gid (bool sys_group,
 			 * gr_locate_gid() found the GID in an as-yet uncommitted
 			 * entry. We'll proceed below and auto-set a GID.
 			 */
-		} else if (result == EEXIST || result == ERANGE) {
+		} else if (result == EEXIST || result == ERANGE || result == EINVAL) {
 			/*
 			 * Continue on below. At this time, we won't
 			 * treat these two cases differently.
```
ikerexxe, line 195: You should update this comment as you added a new case

libmisc/find_new_gid.c
```diff
@@ -112,6 +113,11 @@ static int check_gid (const gid_t gid,
 		return ERANGE;
 	}
 
+  /* Check for compatibility with 16b and 32b gid_t error codes */
+	if (gid == (u_int16_t) -1 || gid == (u_int32_t) -1) {
```
ikerexxe, line 117: Where is `u_int16_t` defined? Maybe it's a typo for `uint16_t`?

ikerexxe (changes requested):
Added some comments that need to be addressed.

libmisc/find_new_gid.c
alejandro-colomar: BTW, you could get those values as `UINT16_MAX` and `UINT32_MAX` and avoid any casts.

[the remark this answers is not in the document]
ikerexxe: That value (65535) is also reserved in 32-bit systems, and several applications have problems handling it. As an example there's [systemd](https://systemd.io/UIDS-GIDS/), check the `Special Linux UIDs` section.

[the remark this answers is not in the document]
alejandro-colomar: Hmmm, makes sense.  Then I suggest using `UINT16_MAX` and `UINT32_MAX`.

libmisc/find_new_gid.c
nertpinx: Sorry, I did not set up everything properly looks like. Is the preferred default the same as with linux (indentation with tabs of width 8)?

libmisc/find_new_gid.c
nertpinx: I rather went ahead and adjusted all of them. It looked like we might not need to differentiate between `EEXIST` and `EINVAL`, but I went for the safer route.

libmisc/find_new_gid.c
nertpinx: To be honest I have no idea, probably just a completion typo or something. Looks like it is both in glibc headers and in linux, but It was meant to be `uintX_t`.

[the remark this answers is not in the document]
nertpinx: Sure, I can switch to that, it's just that the -1 cast is more intuitive that I went for that instead.

libmisc/find_new_gid.c
ikerexxe: I don't think we have any preference, we just try to keep the same indentation throughout the file.

ikerexxe (approved):
LGTM, but as I provided the general idea for the solution I'd like somebody else to review it before merging. @alejandro-colomar wdyt?

alejandro-colomar:
LGTM, but I'd add something more to the commit message.  Having compatibility to 16-bit programs seems crazy these days, IMO.  I don't think distros ship those anymore.  The real issue seems to be that some other (64-bit) programs have had compatibility to 16-bit programs (and keep that compatibility even today), and so we need to be compatible with those programs that are compatible to 16-bit programs.  And a link to systemd's docs would also help.

alejandro-colomar:
> > LGTM, but I'd add something more to the commit message. Having compatibility to 16-bit programs seems crazy these days, IMO. I don't think distros ship those anymore. The real issue seems to be that some other (64-bit) programs have had compatibility to 16-bit programs (and keep that compatibility even today), and so we need to be compatible with those programs that are compatible to 16-bit programs. And a link to systemd's docs would also help.
> 
> Some IOT devices however might?

But the problem is not 16-bit id_t, because that should already be caught by the normal range check (I guess that would result in `ERANGE`).  The problem is 16-bit applications in 32-bit systems, or 32-bit applications in 64-bit systems, if I understood this right.

I don't expect IOT devices to have these problems (although they might).  I expect those problems to only exist in very old archs such as x86, which has many traces of legacy stuff.  Whatever it is, I'd like it to be documented in the commit message, so that 10 years from now whoever picks the code knows why those weird tests are there. :)

hallyn:
> > > LGTM, but I'd add something more to the commit message. Having compatibility to 16-bit programs seems crazy these days, IMO. I don't think distros ship those anymore. The real issue seems to be that some other (64-bit) programs have had compatibility to 16-bit programs (and keep that compatibility even today), and so we need to be compatible with those programs that are compatible to 16-bit programs. And a link to systemd's docs would also help.
> > 
> > 
> > Some IOT devices however might?
> 
> But the problem is not 16-bit id_t, because that should already be caught by the normal range check (I guess that would result in `ERANGE`). The problem is 16-bit id_t applications in 32-bit id_t systems, or 32-bit id_t applications in 64-bit id_t systems, if I understood this right.

I think you're right.

nertpinx:
> > > > LGTM, but I'd add something more to the commit message. Having compatibility to 16-bit programs seems crazy these days, IMO. I don't think distros ship those anymore. The real issue seems to be that some other (64-bit) programs have had compatibility to 16-bit programs (and keep that compatibility even today), and so we need to be compatible with those programs that are compatible to 16-bit programs. And a link to systemd's docs would also help.
> > > 
> > > 
> > > Some IOT devices however might?
> > 
> > 
> > But the problem is not 16-bit id_t, because that should already be caught by the normal range check (I guess that would result in `ERANGE`). The problem is 16-bit id_t applications in 32-bit id_t systems, or 32-bit id_t applications in 64-bit id_t systems, if I understood this right.
> 
> I think you're right.

Yes, that's the issue, at least how I understood it. I can add a link to https://systemd.io/UIDS-GIDS/ in the commit message to make it clear I did not come up with it out of nowhere ;)

alejandro-colomar:
I guess that's enough, but let me suggest some alternative text:

```
Some programs don't support `(uint16_t) -1` or `(uint32_t) -1` as user
or group IDs.  This is because `-1` is used as an error code, and in the
past, `gid_t` and `uid_t` have changed width.  For legacy reasons, those
values have been kept reserved in programs today (for example systemd
does this; see the documentation in the link below).

This should not be confused with catching overflow in the ID values,
since that is already caught by our ERANGE checks.  This is about not
using reserved values that have been reserved for legacy reasons.

Link: <https://systemd.io/UIDS-GIDS/>
Cc: ...
Reviewed-by: ...
Signed-off-by: ...
```

nertpinx:
Thanks for the suggestion, I went with your version, only I took the liberty of adding one more piece of information just for completeness sake in the following sentence:

This is because `-1` is used as an error code or as an unspecified ID, e.g. in chown(2) parameters, and in the past, `gid_t` and `uid_t` have changed width.

ikerexxe:
@alejandro-colomar any additional feedback before I merge?

alejandro-colomar:
LGTM :-)

--
Alejandro Colomar
Linux man-pages comaintainer; https://www.kernel.org/doc/man-pages/
http://www.alejandro-colomar.es/

ikerexxe:
Thanks for the patch!

PR #652, opened 2023-02-07
newuidmap and newgidmap: support passing pid as fd

### LINKED ISSUE #635: pidfd for `new?idmap`
Would it be possible to offer an interface where we pass a pidfd instead of a PID number to `newuidmap` and `newgidmap`? For the program I designed (I can provide more info if requested), I can't really reserve PIDs (reaping must be automatic through `SIGCHLD`) until `newuidmap` is done, so there's a chance for a race that is unlikely, but I'd like to avoid anyway. I do have pidfds (and pidns fds) so if it's possible for me to feed them to `newuidmap` that would solve my problem.

hallyn:
newuidmap and newgidmap currently take an integner pid as the first argument, determining the process id on which to act.  Accept also "fd:N", where N must be an open file descriptor to the /proc/pid directory for the process to act upon.  This way, if you

exec 10</proc/99
newuidmap fd:10 100000 0 65536

and pid 99 dies and a new process happens to take pid 99 before newuidmap happens to do its work, then since newuidmap will use openat() using fd 10, it won't change the mapping for the new process.

Closes #635

### DIFF
--- a/lib/get_pid.c
+++ b/lib/get_pid.c
@@ -10,6 +10,9 @@
 
 #include "prototypes.h"
 #include "defines.h"
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
 
 int get_pid (const char *pidstr, pid_t *pid)
 {
@@ -29,3 +32,51 @@ int get_pid (const char *pidstr, pid_t *pid)
 	return 1;
 }
 
+/*
+ * If use passed in fd:4 as an argument, then return the
+ * value '4', the fd to use.
+ */
+int get_pidfd_from_fd(const char *pidfdstr)
+{
+	long long int val;
+	char *endptr;
+
+	errno = 0;
+	val = strtoll (pidfdstr, &endptr, 10);
+	if (   ('\0' == *pidfdstr)
+	    || ('\0' != *endptr)
+	    || (ERANGE == errno)
+	    || (/*@+longintegral@*/val != (pid_t)val)/*@=longintegral@*/) {
+		return 0;
+	}
+
+	return (int)val;
+}
+
+int open_pidfd(const char *pidstr)
+{
+	int proc_dir_fd;
+	int written;
+	char proc_dir_name[32];
+	pid_t target;
+
+	if (get_pid(pidstr, &target) == 0)
+		return -ENOENT;
+
+	/* max string length is 6 + 10 + 1 + 1 = 18, allocate 32 bytes */
+	written = snprintf(proc_dir_name, sizeof(proc_dir_name), "/proc/%u/",
+		target);
+	if ((written <= 0) || ((size_t)written >= sizeof(proc_dir_name))) {
+		fprintf(stderr, "snprintf of proc path failed for %u: %s\n",
+			target, strerror(errno));
+		return -EINVAL;
+	}
+
+	proc_dir_fd = open(proc_dir_name, O_DIRECTORY);
+	if (proc_dir_fd < 0) {
+		fprintf(stderr, _("Could not open proc directory for target %u: %s\n"),
+			target, strerror(errno));
+		return -EINVAL;
+	}
+	return proc_dir_fd;
+}
--- a/src/newgidmap.c
+++ b/src/newgidmap.c
@@ -69,7 +69,7 @@ static void verify_ranges(struct passwd *pw, int ranges,
 
 static void usage(void)
 {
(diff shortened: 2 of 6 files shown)


hallyn:
Making this a draft pr because I haven't even proof-read the patch yet, let alone tested.  Will look at this again in the next few days.

lib/get_pid.c
```diff
@@ -29,3 +32,51 @@ int get_pid (const char *pidstr, pid_t *pid)
...
+	/* max string length is 6 + 10 + 1 + 1 = 18, allocate 32 bytes */
+	written = snprintf(proc_dir_name, sizeof(proc_dir_name), "/proc/%u/",
+		target);
+	if ((written <= 0) || ((size_t)written >= sizeof(proc_dir_name))) {
+		fprintf(stderr, "snprintf of proc path failed for %u: %s\n",
+			target, strerror(errno));
+		return -EINVAL;
+	}
+
+	proc_dir_fd = open(proc_dir_name, O_DIRECTORY);
```
brauner, line 75: Please prefer `pidfd_open()` over going through `/proc`. :)

lib/get_pid.c
```diff
@@ -29,3 +32,51 @@ int get_pid (const char *pidstr, pid_t *pid)
...
+	    || ('\0' != *endptr)
+	    || (ERANGE == errno)
+	    || (/*@+longintegral@*/val != (pid_t)val)/*@=longintegral@*/) {
+		return 0;
+	}
+
+	return (int)val;
+}
+
+int open_pidfd(const char *pidstr)
```
alejandro-colomar, line 56: Could you maybe factor out into a separate function in a prepatch and then do a smaller patch that implements the change?

ikerexxe:
I know this is still in draft state, but please update the man page with the new information before merging.

lib/get_pid.c
hallyn: I will do that, but as a separate PR as Alejandro suggested.  I'll create an issue for it right now.

hallyn:
> I know this is still in draft state, but please update the man page with the new information before merging.

Yup, done.

hallyn:
Hm, interesting (and disappointing) - the test results are still better on the self-hosted runner.

src/newgidmap.c
```diff
@@ -168,25 +165,19 @@ int main(int argc, char **argv)
...
-	if (proc_dir_fd < 0) {
-		fprintf(stderr, _("%s: Could not open proc directory for target %u\n"),
-			Prog, target);
-		return EXIT_FAILURE;
+	target_str = argv[1];
+	if (strlen(target_str) > 3 && strncmp(target_str, "fd:", 3) == 0) {
+		/* the user passed in a /proc/pid fd for the process */
+		target_str = &target_str[3];
+		proc_dir_fd = get_pidfd_from_fd(target_str);
+		if (proc_dir_fd < 0)
```
vinipsmaker, line 174: Why is it comparing against smaller than 0? `get_pidfd_from_fd` returns 0 on error, not `-1`.

hallyn: Huh, the bug is in get_pidfd_from_fd() IMO - that was supposed to return -1. Thanks for noticing that.

vinipsmaker:
So as far as I see, the only checking to validate `proc_dir_fd` occurs in these lines:

```c
if ((getuid() != pw->pw_uid) ||
    (!getdef_bool("GRANT_AUX_GROUP_SUBIDS") && (getgid() != pw->pw_gid)) ||
    (pw->pw_uid != st.st_uid) ||
    (getgid() != st.st_gid)) {
```

I could create a directory (owned by me, so it passes on the above checks) which has a symlink named `uid_map` pointing to any file owned by root. Later the program will call:

```c
write_mapping(proc_dir_fd, ranges, mappings, "uid_map", pw->pw_uid);
```

And then (inside `write_mapping()`):

```c
openat(proc_dir_fd, map_file, O_WRONLY);
```

If I'm not missing anything, this would allow me to overwrite any file owned by root. What we should be doing instead is to also ensure `st.st_dev == 22 && st.st_rdev == 0` before we proceed so we only operate on files within `/proc`.

hallyn:
You're right, this is only intended for pidfds and so we should make sure about the filesystem.  Did you want to post a patch for that?

vinipsmaker:
>  Did you want to post a patch for that?

Yes. Give me a few minutes.

PR #653, opened 2023-02-07
Assume getutent(3) exists (remove dead code)

alejandro-colomar:
Assume getutent(3) exists (remove dead code)
    
    Recently, we removed support for 'struct utmpx'.  We did it because utmp
    and utmpx are identical, and while POSIX specifies utmpx (and not utmp),
    GNU/Linux documentation seems to favor utmp.  Also, this project
    defaulted to utmp, so changing to utmpx would be more dangerous than
    keeping old defaults, even if it's supposed to be the same.
    
    Now, I just found more code that didn't make much sense: lib/utent.c
    provides definitions for getutent(3) and friends in case the system
    doesn't provide them, but we don't provide prototypes for those
    definitions, so code using the functions would have never compiled.
    
    Let's just remove these definitions as dead code.

Fixes: 3be7b9d75a6b ("Remove traces of utmpx")
Fixes: 170b76cdd1a9 ("Disable utmpx permanently")


hallyn (approved):

ikerexxe (approved):
LGTM. Merging right now.

PR #654, opened 2023-02-07
noreturn

alejandro-colomar:


hallyn (approved):
I would have assumed that /*@noreturn@*/ was in fact doing something, but I'm not finding any documentation to that effect.

alejandro-colomar:
> I would have assumed that /_@NoReturn@_/ was in fact doing something, but I'm not finding any documentation to that effect.

Same here =)

In fact, I searched what is /*@observer@*/ and some others which don't seem obvious to me, but found nothing with two search engines.  Maybe some old maintainer is still around and can comment?

ikerexxe:
Not an expert but I assumed that it was used as an annotation for a C linter that a function doesn't return.

PR #659, opened 2023-02-19
libmisc: agetpass(): Fix bug detecting truncation

alejandro-colomar:
On 2/19/23 18:09, David Mudrich wrote:
> I am working on a RAM based Linux OS from source, and try to use
> latest versions of all software.  I found shadow needs libbsd's
> readpassphrase(3) as superior alternative to getpass(3).  While
> considering if I a) include libbsd, or include libbsd's code of
> readpassphrase(3) into shadow, found, that libbsd's readpassphrase(3)
> never returns \n or \r
> <https://cgit.freedesktop.org/libbsd/tree/src/readpassphrase.c>
> line 122, while agetpass() uses a check for \n in agetpass.c line 108.
> I assume it always fails.

Indeed, it always failed.  I made a mistake when writing agetpass(), assuming that readpassphrase(3) would keep newlines.

>
> I propose a check of len == PASS_MAX - 1, with false positive error for
> exactly PASS_MAX - 1 long passwords.

Instead, I added an extra byte to the allocation to allow a maximum password length of PASS_MAX (which is the maximum for getpass(3), which we're replacing.

While doing that, I notice that my previous implementation also had another bug (minor): The maximum password length was PASS_MAX - 1 instead of PASS_MAX.  That's also fixed in this commit.

Reported-by: David Mudrich <dmudrich@gmx.de>

### DIFF
--- a/libmisc/agetpass.c
+++ b/libmisc/agetpass.c
@@ -19,7 +19,7 @@
 
 
 #if !defined(PASS_MAX)
-#define PASS_MAX  BUFSIZ
+#define PASS_MAX  BUFSIZ - 1
 #endif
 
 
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
Changes:

-  Edit commit message

```
$ git range-diff alx/master..gh/readpassphrase master..readpassphrase 
1:  0015e33f ! 1:  8668d889 libmisc: agetpass(): Fix bug detecting truncation
    @@ Commit message
         instead of PASS_MAX.  That's also fixed in this commit.
     
         Reported-by: David Mudrich <dmudrich@gmx.de>
    +    Fixes: 155c9421b935 ("libmisc: agetpass(), erase_pass(): Add functions for getting passwords safely")
     
```

alejandro-colomar:
I wrote that check for when I thought the buffer could contain \n.  In such
a scenario, I had code for removing the newline, but in the case of an
empty line (I was thinking of early ctrl-D), I couldn't remove the \n, so I
had to treat it specially.

Now, this case (empty password) is handled by the general case.

--

ikerexxe:
LGTM. Thanks for the patch!

alejandro-colomar:
We should probably add a regression test for this.  @hallyn ?  @ikerexxe ?

PR #660, opened 2023-02-20
Ci build

ikerexxe:


ikerexxe:
@alejandro-colomar do you know what would be the equivalent of `ULONG_WIDTH` for alpine linux (musl)?

cgzones:
`sizeof(unsigned long) * 8`?

alejandro-colomar:
We can add the following to defines.h:

#ifndef ULONG_WIDTH
#define ULONG_WIDTH (sizeof(unsigned long) * CHAR_BIT)
#endif


This macro will be defined by C23.

--

ikerexxe:
That's good. I've added a commit for it and I think I got the DCO right.

alejandro-colomar:
Yep, that LGTM.

--

alejandro-colomar:
Just a minor OCD thing: I'd put two blank lines between includes and
defines, and also two between defines and prototypes. :)

And I found a bad typo of mine while reviewing this: bit_ceil_wrapul() should
call bit_ceilul() rather than itself.  Can you please fix it here?  Thanks
for fixing my bugs :D

--

ikerexxe:
Fixed. I opened https://github.com/shadow-maint/shadow/pull/661 to track that work.

hallyn:
> @hallyn @alejandro-colomar do you know how I can fix the debian failure? `libbsd-dev` is already installed but it seems like something went wrong

I think you need pkgconf so autoconf can figure out that libbsd is there :)

ikerexxe:
Ok, I think this is ready for review.

PR #661, opened 2023-02-22
lib: two fixes in bit.h

ikerexxe:
* Define ULONG_WIDTH if non-existent
* bit_ceil_wrapul(): stop recursion and call bit_ceilul() instead


hallyn (approved):

alejandro-colomar:
Would you maybe add `Fixes: 0712b236c3bc ("Add bit manipulation functions")` to the second commit message?

ikerexxe:
Thanks!

PR #667, opened 2023-02-24
Validate fds created by the user

vinipsmaker:
write_mapping() will do the following:

openat(proc_dir_fd, map_file, O_WRONLY);

An attacker could create a directory containing a symlink named "uid_map" pointing to any file owned by root, and thus allow him to overwrite any root-owned file.

### DIFF
--- a/lib/get_pid.c
+++ b/lib/get_pid.c
@@ -41,6 +41,8 @@ int get_pidfd_from_fd(const char *pidfdstr)
 {
 	long long int val;
 	char *endptr;
+	struct stat st;
+	dev_t proc_st_dev, proc_st_rdev;
 
 	errno = 0;
 	val = strtoll (pidfdstr, &endptr, 10);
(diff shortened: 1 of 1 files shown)


src/newgidmap.c
```diff
@@ -196,6 +196,12 @@ int main(int argc, char **argv)
 		return EXIT_FAILURE;
 	}
 
+	/* Validate fds coming from get_pidfd_from_fd() */
```
hallyn, line 199: Would it, however, be better to do this in the get_pidfd_from_fd() fn itself?  I realize you have the struct stat here already, but any other callers of get_pidfd_from_fd() will need the same guarantee.

src/newgidmap.c
```diff
@@ -196,6 +196,12 @@ int main(int argc, char **argv)
...
+	if (st.st_dev != 22 || st.st_rdev != 0) {
```
hallyn, line 200: Hm, on my laptop it's 23.  That bodes ill.

src/newgidmap.c
vinipsmaker: I agree. Should I do the change?

hallyn: Yes, thanks.

src/newgidmap.c
vinipsmaker: There's another way to get these values. I'll update the patchset. One sec.

vinipsmaker: I pushed the new version. Please, review.

hallyn: Looks great, thanks.

I'll take another look in 2m and then merge.

I can't tell you how embarrassed I am.

hallyn (approved):

hallyn:
Thanks!

lib/get_pid.c
```diff
@@ -51,6 +53,21 @@ int get_pidfd_from_fd(const char *pidfdstr)
...
 	}
 
+	if (stat("/proc/self/uid_map", &st) < 0) {
+		return -1;
+	}
+
+	proc_st_dev = st.st_dev;
+	proc_st_rdev = st.st_rdev;
+
+	if (fstat(proc_dir_fd, &st) < 0) {
```
vinipsmaker, line 63: there's a copy'n'paste error here. It should be `fstat(val, &st)`

hallyn: Yup, updated, thanks.

PR #670, opened 2023-02-25
tests: newuidmap and newgidmap: update expected fail message

hallyn:
The failure message got changed, but the tests looking for it did not.


hallyn:
With this , ./run_some passes all tests on my ubuntu jammy vm anyway.

PR #672, opened 2023-02-25
Add .editorconfig

vinipsmaker:


.editorconfig
```diff
@@ -0,0 +1,4 @@
+root = true
+
+[*.{c,h}]
+indent_style = tab
```
ikerexxe, line 4: Are we using tabs? I don't even think this is defined and we are reusing whatever each file specifies as the default.

vinipsmaker: @ikerexxe the `.c` file for the last PR I did was using tabs. I was unaware that each file had its own standard. Is there any pattern? I could codify the pattern/globs to use spaces for a bunch of files, tabs for another bunch, and a default for new files.

hallyn: tabs for c.  spaces for shell.

hallyn: @ikerexxe yes we do try to maintain legacy style as much as possible.  And rewriting a file to convert to tabs can make backporting patches painful, so we have avoided it.  But in general for new files we want tabs.

vinipsmaker: >  spaces for shell

@hallyn, there aren't many shell files, are you sure you want rules for them? After a `find . -name '*.sh'`, I found out that there's no pattern on them (e.g. `contrib/adduser2.sh` uses 3-spaces level1/tabs level 2 for the first block, and all tabs on the rest of the file... except when there's 2-spaces for level 2 on random spots).

vinipsmaker: I manually verified verified all files `*.{c,h,y,h.in}` that had lines starting with spaces. It was mostly comment lines (e.g. a ` * rest of a multiline comment started at the previous lines`), function parameter declaration split among multiple lines, and goto labels. These files would be fine to be recognized as tab-indentation.

For other files, there were just a few obviously wrongly space-indented lines surrounded by tab-indented line. These files would also be fine to be recognized as tab-indentation.

And then there were two files with no clear indentation rule that are neither spaces nor tabs:

* `contrib/adduser.c`
* `libmisc/getdate.y`

Therefore, I think it's fine to just specify the glob `*.{c,h}` for tab-indentation. There are no files where space-indentation is consistently used anyway.

hallyn: > > spaces for shell
> 
> @hallyn, there aren't many shell files, are you sure you want rules for them?

But by the same logic, I have no idea what programs .editorconfig is read by or whether we want to include it in shadow at all :)

Anyway I see that many of the scripts do use tabs.  Which I'm also ok with.

vinipsmaker: > Anyway I see that many of the scripts do use tabs. Which I'm also ok with.

@hallyn, are you okay with the following rule then?

```ini
[*.sh]
indent_style = tab
```

I can add it to the commit once you ACK'it. Otherwise I'll add the rule for space-indented shell scripts.

hallyn: I'm ok with the rule, but I'm still curious what tools will actually use this thing, and, therefore, whether it's worth having in the codebase.

vinipsmaker: > but I'm still curious what tools will actually use this

Emacs, the editor I've used to send a previous PR, has a plugin for this. The repository didn't have an `.editorconfig` file, so I had to configure Emacs to use tabs once I've figured the rest of the file I was editing used tabs.

`.editorconfig` is **NOT** supposed to be a CI job that re-styles existing files. `.editorconfig` is supposed to just store the most commonly used settings by text editors that change per-repository (e.g. indentation style).

I used Emacs, but EditorConfig is not an Emacs project. Plugins for other text editors exist as well: https://editorconfig.org/#download

And some editors even have native support for EditorConfig: https://editorconfig.org/#pre-installed

I'm unaware of any other initiative to standardize this type of setting, so I just settled for EditorConfig a few years ago, and moved on.

hallyn:
... that's quite an invasive plugin to us it with vim, so i wouldn't be using it at any rate.

If a lot of emacs users use it, then maybe it's worth it.

But I'd like to hear from someone else that they would be using it...

But my git instinct is to listen to https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/coding-style.rst?h=v6.2#n1074 .

vinipsmaker:
> But my git instinct is to listen to https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/coding-style.rst?h=v6.2#n1074 .

I agree with Torvalds. We shouldn't pollute source code files with editor configuration. I never liked this solution. However `.editorconfig` is a separate file.

hallyn:
> > But my git instinct is to listen to https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/coding-style.rst?h=v6.2#n1074 .
> 
> I agree with Torvalds. We shouldn't pollute source code files with editor configuration. I never liked this solution. However `.editorconfig` is a separate file.

One thing I'm worired about is - does 'tab mode' in .editorconfig have additional implications?  I don't want to in the future be told "but you can't reject my patch making (some formatting change) because .editorconfig says this is the correct way".

Anyway I'll approve this - +1 for tabs not spaces in c and h files.  I'll leave it to someone else to implicitly give second +1 by merging.

hallyn (approved):

hallyn:
BTW - thanks for trying to make things better :)

vinipsmaker:
> One thing I'm worired about is - does 'tab mode' in .editorconfig have additional implications? One thing I'm worired about is - does 'tab mode' in .editorconfig have additional implications? I don't want to in the future be told "but you can't reject my patch making (some formatting change) because .editorconfig says this is the correct way".

It should only be a hint for the text editor. Nothing else.

vinipsmaker:
> If we are going to adopt this approach, I think we should write some documentation, even a markdown document would be sufficient, to indicate this information.

I can write such document, but first, I think there's a misconception here. `.editorconfig` is just a hint to text editors. Nobody should be forced to use editorconfig. If you don't want to install a plugin to your favorite text editor even when one exists, be free to ignore it.

Think of `editorconfig` like this: you open a random repo on the internet, notices an issue, and want to quickly clone the repo, edit the code to fix it, and submit the patch back. `.editorconfig` will contain just a minimal set of safe options (in a file format simple to parse... which is just a small INI) you can pass to the editor to minimize impedance to contribute to the project (imagine configuring indentation rules for every 2-lines or 10-lines 1-patch you submit to new projects... or worse, importing a whole Docker-based devenv created by the ones maintaining the project the project just to adhere to its conventions).

It should minimize the contribution ritual, not increase it. Now we have yet another repository policy document every contributor needs to read before making its contribution? It looks suspiciously like it's defeating the purpose already. Ideally the `.editorconfig` would just be dropped at the root of the repo and then forgotten forever (just like we do to `.gitignore` and license files).

However I don't want to extend the discussion much further either. Developer time is a precious resource, and you guys maintain an important piece of the Linux ecosystem. If you think this PR will just derail in a long discussion, I think it'd be best just to close it and everyone can move on to more important tasks (EditorConfig isn't that important, after all).

hallyn:
I do understand what the file does.

My concern is one about human nature.  Importing this file will be seen by some as making an authoritative statement about coding style, overruling maintainer decisions.

But it's ok.  I'll just tell them no :)

thanks.

PR #675, opened 2023-03-06
ch(g)passwd: Check selinux permissions upon startup

nertpinx:
The permission also need to be checked before process_root_flag() since that can chroot into non-selinux environment (unavailable selinux mount point for example).

### DIFF
--- a/src/chgpasswd.c
+++ b/src/chgpasswd.c
@@ -424,6 +424,12 @@ int main (int argc, char **argv)
 	(void) bindtextdomain (PACKAGE, LOCALEDIR);
 	(void) textdomain (PACKAGE);
 
+#ifdef WITH_SELINUX
+	if (check_selinux_permit ("passwd") != 0) {
+		return (E_NOPERM);
+	}
+#endif				/* WITH_SELINUX */
+
 	process_root_flag ("-R", argc, argv);
 
 	process_flags (argc, argv);
--- a/src/chpasswd.c
+++ b/src/chpasswd.c
@@ -452,6 +452,12 @@ int main (int argc, char **argv)
 	(void) bindtextdomain (PACKAGE, LOCALEDIR);
 	(void) textdomain (PACKAGE);
 
+#ifdef WITH_SELINUX
+	if (check_selinux_permit ("passwd") != 0) {
+		return (E_NOPERM);
+	}
+#endif				/* WITH_SELINUX */
+
 	process_flags (argc, argv);
 
 	salt = get_salt();


ikerexxe (approved):
LGTM, but as I've influenced the final solution I'd like to have a second reviewer

cgzones:
This seems quite questionable:

1. Currently ch(g)passwd do not perform any SELinux checks (to my knowledge), which could be blocked inside of a chroot.

2. All other programs performing SELinux checks (chage, chfn, chsh, passwd) are set(u|g)id binaries (unlike ch(g)passwd)(at least on Debian). This makes them target for SELinux checks in the first places due to their availability for unprivileged user.  The SELinux permission `passwd : passwd` is commonly not checked for standard programs accessing */etc/shadow* (which required root privileges).

nertpinx:
The thing is that ch(g)passwd must have a SELinux domain transition and in case root user is restricted (not unconfined_u), then it should not allow the change. The domain transition is something like an SUID in DAC, but for SELinux.  I found the related BZ for this which was fixed similarly: https://bugzilla.redhat.com/show_bug.cgi?id=1336902

cgzones:
Since ch(g)passwd operates directly on */etc/shadow* is requires read-write permissions on it (*shadow_t*).
So why not run the programs in the caller domain and then rely on the fact whether confined root has that access (e.g. *sysadm_t* has, *staff_t* does not)? Alternatively only allow the transition to the ch(g)passwd domain only for the wanted userdomains.

The `passwd { passwd }` is mainly to control password changes in set(g|u)id programs, like passwd, which can be run by unprivileged user and run in the same domain for privileged and unprivileged users, and thus have rw-access to *shadow_t*.

bachradsusi:
If I get it right, this would not a problem for user domains like `staff_t` where read-write permissions on shadow_t could be used.  But there are agents like `qemu-ga`  which run in their own domains and it's probably not a good idea to add read-write permissions on `shadow_t` to them. Also this check is already used in passwd.c - https://github.com/shadow-maint/shadow/blob/e047a3acc6c8d6ad959648361f8fca5d653a7735/src/passwd.c#L969

hallyn:
@cgzones given @bachradsusi 's most recent comment, do you still object to this PR?

cgzones:
> If I get it right, this would not a problem for user domains like staff_t where read-write permissions on shadow_t could be used. But there are agents like qemu-ga which run in their own domains and it's probably not a good idea to add read-write permissions on shadow_t to them.

This makes sense as a use case.

(Note: with this change the argument `--help` will also be covered by this SELinux check; I am not sure whether this is wanted or relevant.)

cgzones (approved):

src/chgpasswd.c
```diff
@@ -424,6 +424,12 @@ int main (int argc, char **argv)
 	(void) bindtextdomain (PACKAGE, LOCALEDIR);
 	(void) textdomain (PACKAGE);
 
+#ifdef WITH_SELINUX
+	if (check_selinux_permit ("passwd") != 0) {
+		return (E_NOPERM);
+	}
+#endif
```
cgzones, line 431: Considering the overall coding style, should there be closing comments like `/* WITH_SELINUX */`?

nertpinx: Sure, I can add those, some ifdefs had it and some not.

ikerexxe:
If nobody objects I'll proceed to merge this PR next week.

hallyn (approved):

hallyn:
So long as in 'check_selinux_permit ("passwd")', 'passwd' to class 'passw' is a meaningful thing.

PR #676, opened 2023-03-06
semanage: Do not set default SELinux range

nertpinx:
Both semanage and libsemanage actually set the user's mls range to the default of the seuser, which makes more sense and removes a bit of code for usermod and useradd.  More fine-grained details must always be set with some other tool
(semanage) anyway.

### DIFF
--- a/lib/semanage.c
+++ b/lib/semanage.c
@@ -22,10 +22,6 @@
 
 #include "shadowlog_internal.h"
 
-#ifndef DEFAULT_SERANGE
-#define DEFAULT_SERANGE "s0"
-#endif
-
 
 format_attr(printf, 3, 4)
 static void semanage_error_callback (unused void *varg,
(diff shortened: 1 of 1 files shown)


nertpinx:
This fixes the issue when the default range should include categories as well

cgzones:
An alternative would be to add a command line option `--selinux-range` to useradd and usermod, and pass it as third argument to `set_seuser()`.

nertpinx:
Good idea, but I don't think that is an alternative, more like a possible future feature. I can try to add it in another PR, but with this one I was mainly trying to fix the default behaviour.

hallyn:
> An alternative would be to add a command line option `--selinux-range` to useradd and usermod, and pass it as third argument to `set_seuser()`.

@cgzones ok to merge this as is in the meantime?

cgzones (approved):

ikerexxe (approved):
Thank you for posting the change and for the reviews!

PR #677, opened 2023-03-06
usermod: Small optimization using memmove for password unlock

nertpinx:
And one extra optimization using memmove instead of manually moving bytes.

### DIFF
--- a/src/usermod.c
+++ b/src/usermod.c
@@ -430,8 +430,6 @@ static char *new_pw_passwd (char *pw_pass)
 		strcat (buf, pw_pass);
 		pw_pass = buf;
 	} else if (Uflg && pw_pass[0] == '!') {
-		char *s;
-
 		if (pw_pass[1] == '\0') {
 			fprintf (stderr,
 			         _("%s: unlocking the user's password would result in a passwordless account.\n"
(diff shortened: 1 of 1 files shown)


ikerexxe:
@hallyn can you take a look at this PR?

src/usermod.c
```diff
@@ -443,11 +441,7 @@ static char *new_pw_passwd (char *pw_pass)
 		              "updating password", user_newname, user_newid, 0);
 #endif
 		SYSLOG ((LOG_INFO, "unlock user '%s' password", user_newname));
-		s = pw_pass;
-		while ('\0' != *s) {
-			*s = *(s + 1);
-			s++;
-		}
+		memmove(pw_pass, pw_pass + 1, strlen(pw_pass) + 1);
```
hallyn, line 444: hm, pw_pass+1 is not right for the length, is it?  You're starting at pw_pass + 1, so you'll be accessing one byte past the end of the \0 ending pw_pass.

src/usermod.c
```diff
@@ -443,6 +443,12 @@ static char *new_pw_passwd (char *pw_pass)
 		SYSLOG ((LOG_INFO, "unlock user '%s' password", user_newname));
 		memmove(pw_pass, pw_pass + 1, strlen(pw_pass) + 1);
 	} else if (pflg) {
+		if (strchr(user_pass, ':') != NULL) {
```
hallyn, line 446: Should valid_field() be called on user_pass?

src/usermod.c
nertpinx: You're right, due to omission of temporary variable I was counting the wrong length, thanks for noticing.

src/usermod.c
nertpinx: It actually is, in `lib/pwio.c` which is used to update the password entry anyway. I guess it was an issue elsewhere and I cannot reproduce it now.  I probably just completely missed that.

nertpinx: I-m dropping this patch then and changing the PR name to reflect the only change left in here, sorry for the noise.

ikerexxe:
@hallyn can you merge this PR if you agree with it?

hallyn (approved):

Issue #680, opened 2023-03-11
Argv[0] Issues


skyler-ferrante:
Shadow does not handle argc==0 correcly. In many binaries we end up causing a segfault through Basename. Basename calls `strrchr` on argv without checking argc. Although I don't see a security issue stemming from this, it might be a good idea to fix.

Example code:
```
#include<unistd.h>

int main(){
        char* argv[] = {NULL};
        execv("/bin/chfn", argv);
}
```

(I edited this issue, since I did not correctly understand the problem at first.)

hallyn:
I wouldn't mind making Basename a bit safer.

However - you say "In many binaries we end up causing a segfault".  Unless you can provide an example, I'll have to call this an exaggeration.  If you can show a way to manipulate argv in these tools before we do Basename(argv[0]), then we should protect against it.  But the code you show simply is not relevant.

skyler-ferrante:
Look at line 480 of src/chsh.c
```
int main (int argc, char **argv)
{
        char *user;             /* User name                         */
        const struct passwd *pw;        /* Password entry from /etc/passwd   */

        sanitize_env ();

        /*
         * Get the program name. The program name is used as a prefix to
         * most error messages.
         */
        Prog = Basename (argv[0]);
```
Before we check argc, we call Basename with argv[0]. Argv[0] can be null, if the caller doesn't set it. This is a common pattern, found in many of the shadows binaries. The example listed above, if compiled and ran causes a segfault inside of chfn.

Running `gdb a.out` with the example in my first comment leads to:
```
Program received signal SIGSEGV, Segmentation fault.
__strrchr_avx2 () at ../sysdeps/x86_64/multiarch/strrchr-avx2.S:54 
```

This is the strrchr in basename.

The same could be said of `su`, `chfn`, `login`, ...

Sorry if I didn't give enough details before. You can't make Basename safer, you need to check argc (if argc==0) before calling Basename.

hallyn:
You say "Argv[0] can be null".  How would that happen?  Is there a path in kernel fs/exec.c that would cause that?  I'm not aware of any.

hallyn:
And why would Basename checking for (str == NULL || str[0] == NULL) not guard against argv[0] being NULL?  Now if argv is NULL or simply invalid, then Basename is indeed too late.  But there needs to be a user-controllable way to make that happen for it to be a problem.

vapier:
[...]

PR #685, opened 2023-03-22
Fix null dereference in basename

skyler-ferrante:
On older kernels (<=linux-5.17), argv[0] can be null. Basename would call strrchr with null if argc==0.

### DIFF
--- a/libmisc/basename.c
+++ b/libmisc/basename.c
@@ -21,6 +21,10 @@
 #include "prototypes.h"
 /*@observer@*/const char *Basename (const char *str)
 {
+	if (str == NULL) {
+		abort ();
+	}
+
 	char *cp = strrchr (str, '/');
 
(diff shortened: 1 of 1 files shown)


jubalh:
> Fixes the issue from https://github.com/shadow-maint/shadow/issues/680

I would mention that in the commit message

skyler-ferrante:
I don't think argv can ever be NULL. Execve allows you to call with NULL, but even with the old kernel created an array with the first element pointing to NULL.

`execve("./a.out", NULL, NULL);`

- Older kernels set `argv[0]==NULL` and `argc==0`
- Newer kernels set `argv[0]==""`, `argv[1]==NULL`, and `argc==1`

The new kernel still allows you to call execve with NULL, so I think the man page is still right

alejandro-colomar:
Thanks; that clarifies my head.

hallyn (approved):

Issue #688, opened 2023-03-24
Password strength


alejandro-colomar:
https://github.com/shadow-maint/shadow/blob/53a4bda9ba560db52da475bb27b88b0291fe2835/man/passwd.1.xml#L97

Hi!

Today I had to register in an official administration website in my country.  To my surprise, when I tried to paste the result of my usual `makepasswd --char 64`, it didn't accept more than 20 characters (but they of course mandated the usual upper+lower+num+symbol, d'oh).  I think that advice in passwd(1) is actively harmful to society, and is one of the reasons MFA is being enforced in some places: most passwords are crap.

Ignoring the possibility of bugs in password handling when it includes special characters, which is more than likely (not talking of shadow, but of software in general, especially in random websites), there's the issue that perceived strength of passwords is bogus.

My keyboard has 27 letters, so counting upper, lower, and nums, I have 62 different characters in non-weird passwords.  For a reasonable length of 64 chars, that's a strength of `log2(62^64) = 381 bits of entropy`.  Now, apart from those 62 different chars, my keyboard can write 32 extra different characters, which amounts to a total of 94 different chars; for a maximum length of 20 chars, I'm limited to a strength of `log2(94^20) = 131 bits of entropy`.  For a length of 8, as suggested in passwd(1), the strength is `log2(94^8) = 52 bits of entropy`, which is the same as a password of only lowercase letters of length 11: `log2(27^11) = 52`.  Can we please remove that bogus advice?  The exponent is the key in password strength, not the base.  Even a school child should know that exponential.

Can we please completely rewrite that paragraph to only recommend password length?

Of course, I can't submit this without referencing the relevant: <https://xkcd.com/936/>

And to all that, sum the chance that I may need to use a different keyboard at some point, and maybe some symbols have no key.

vapier:
that's quite the monstrous leap of faith to claim this one bit of obscure advice is a significant factor in the rise of MFAs.  i can guarantee you that the vast majority of the internet population has never read this text, and MFAs make sense even if people had better password hygiene.

alejandro-colomar:
> i can guarantee you that the vast majority of the internet population has never read this text

"this" as in read passwd(1), probably not.  "this" as in any other repetition of this message, which seems to be everywhere (including most sign up pages), I'd say even my mom has read or at least heard about it (after asking, she confirms: from sign up pages).

PR #689, opened 2023-03-24
libmisc, man: Drop old check and advice for complex character sets in passwords

alejandro-colomar:
I kept the minimum length as 8, since it was what we had for the simplest passwords already.  8 is probably not what you really want from a password, but I believe it should be up to users to chose what they want.  If we raise this too much, we may upset some people, so let's be cautious.

Add the relevant XKCD to the passwd(1) manual page.  It already explains most of the rationale behind this patch.

The strength of a password is an exponential formula, where the base is the size of the character set, and the exponent is the length of the password.  That already shows why long passwords of just lowercase letters are better than short Pa$sw0rdZ3.  But an even more important point is that humans, when forced to use symbols in a password, are more likely to do trivial substitutions on simple passwords, which doesn't increase strength, and can instead give a false sense of strength, which is dangerous.

Closes: <https://github.com/shadow-maint/shadow/issues/688>
Link: <https://xkcd.com/936/>

### DIFF
--- a/libmisc/obscure.c
+++ b/libmisc/obscure.c
@@ -75,57 +75,6 @@ static bool similar (/*@notnull@*/const char *old, /*@notnull@*/const char *new)
 	return true;
 }
 
-/*
- * a nice mix of characters.
- */
-
-static bool simple (unused const char *old, const char *new)
-{
-	bool digits = false;
-	bool uppers = false;
-	bool lowers = false;
-	bool others = false;
-	int size;
-	int i;
-
-	for (i = 0; '\0' != new[i]; i++) {
-		if (isdigit (new[i])) {
-			digits = true;
-		} else if (isupper (new[i])) {
-			uppers = true;
-		} else if (islower (new[i])) {
-			lowers = true;
-		} else {
-			others = true;
-		}
-	}
-
-	/*
-	 * The scam is this - a password of only one character type
-	 * must be 8 letters long.  Two types, 7, and so on.
-	 */
-
-	size = 9;
-	if (digits) {
-		size--;
-	}
-	if (uppers) {
-		size--;
-	}
-	if (lowers) {
-		size--;
-	}
-	if (others) {
-		size--;
-	}
-
-	if (size <= i) {
-		return false;
-	}
-
-	return true;
-}
-
 static char *str_lower (/*@returned@*/char *string)
 {
 	char *cp;
@@ -170,8 +119,6 @@ static /*@observer@*//*@null@*/const char *password_check (
 		msg = _("case changes only");
 	} else if (similar (oldmono, newmono)) {
 		msg = _("too similar");
-	} else if (simple (old, new)) {
-		msg = _("too simple");
 	} else if (strstr (wrapped, newmono) != NULL) {
 		msg = _("rotated");
 	} else {
--- a/man/passwd.1.xml
+++ b/man/passwd.1.xml
@@ -94,27 +94,10 @@
       </para>
 
       <para>
-	Then, the password is tested for complexity. As a general guideline,
-	passwords should consist of 6 to 8 characters including one or more
-	characters from each of the following sets:
-      </para>
-
-      <itemizedlist mark='bullet'>
-	<listitem>
-	  <para>lower case alphabetics</para>
-	</listitem>
-	<listitem>
-	  <para>digits 0 thru 9</para>
-	</listitem>
-	<listitem>
-	  <para>punctuation marks</para>
-	</listitem>
-      </itemizedlist>
-
-      <para>
-	Care must be taken not to include the system default erase or kill
-	characters. <command>passwd</command> will reject any password which
-	is not suitably complex.
+	Then, the password is tested for complexity.
+	<command>passwd</command> will reject any password which is not
+	suitably complex.  Care must be taken not to include the system
+	default erase or kill characters.
       </para>
 
     </refsect2>
@@ -137,6 +120,17 @@
 	written down. The password should also not be a proper name, your
 	license number, birth date, or street address. Any of these may be
 	used as guesses to violate system security.
+      </para>
+
+      <para>
+	As a general guideline, passwords should be long and random.  It's
+	fine to use simple character sets, such as passwords consisting
+	only of lowercase letters, if that helps memorizing longer
+	passwords.  For a password consisting only of lowercase English
+	letters randomly chosen, and a length of 32, there are 26^32
+	(approximately 2^150) different possible combinations.  Being an
+	exponential equation, it's apparent that the exponent (the length)
+	is more important than the base (the size of the character set).
       </para>
 
       <para>
@@ -474,6 +468,9 @@
 	<refentrytitle>chpasswd</refentrytitle><manvolnum>8</manvolnum>
       </citerefentry>,
       <citerefentry>
+	<refentrytitle>makepasswd</refentrytitle><manvolnum>1</manvolnum>
+      </citerefentry>,
+      <citerefentry>
 	<refentrytitle>passwd</refentrytitle><manvolnum>5</manvolnum>
       </citerefentry>,
       <citerefentry>
@@ -487,6 +484,12 @@
       <citerefentry>
 	<refentrytitle>usermod</refentrytitle><manvolnum>8</manvolnum>
       </citerefentry>.
+    </para>
+
+    <para>
+	The following web page comically (yet correctly) compares the
+	strength of two different methods for choosing a password:
+	"https://xkcd.com/936/"
     </para>
   </refsect1>
 </refentry>


alejandro-colomar:
Changes:

- wfix

alejandro-colomar:
Add reference to makepasswd(1) in passwd(1)'s SEE ALSO.

man/passwd.1.xml
```diff
@@ -488,5 +484,9 @@
 	<refentrytitle>usermod</refentrytitle><manvolnum>8</manvolnum>
       </citerefentry>.
     </para>
+
+    <para>
+	https://xkcd.com/936/
```
vapier, line 489: this link lacks any sort of introduction or explanation

libmisc/obscure.c
```diff
@@ -170,8 +119,8 @@ static /*@observer@*//*@null@*/const char *password_check (
 		msg = _("case changes only");
 	} else if (similar (oldmono, newmono)) {
 		msg = _("too similar");
-	} else if (simple (old, new)) {
-		msg = _("too simple");
+	} else if (strlen (new) < 8) {
```
vapier, line 122: this is more restrictive than the existing code.  having it be dynamic based on the inputs was aligned with the intention of this code.  but you're making it static, and we already have `PASS_MIN_LEN` as a configurable value that is checked earlier.  so if you want to delete the dynamic strength checks, you should delete this length check entirely.

not that i have a problem with the dynamic strength check as-is.

man/passwd.1.xml
```diff
@@ -94,23 +94,16 @@
       </para>
 
       <para>
-	Then, the password is tested for complexity. As a general guideline,
-	passwords should consist of 6 to 8 characters including one or more
-	characters from each of the following sets:
+	Then, the password is tested for length.  As a general guideline,
+	passwords should be long.  It's fine to use simple character sets,
+	such as passwords consisting only of lowercase letters.  For a
```
vapier, line 99: this statement is much much too weak.  yes, it follows a "general guideline that passwords should be long", but this sentence feels disconnected from that and reads as independent guidance.

man/passwd.1.xml
```diff
@@ -94,23 +94,16 @@
...
+	password consisting only of lowercase English letters randomly
+	chosen, and a length of 64, there are 27^64 (approximately 2^304)
```
vapier, line 101: ok, but no one is generating & memorizing 64 randomly generated alphas.  it's one thing to use a password manager where long random generated values are never typed by users, but shadow is specifically geared towards users typing in their password interactively.

vapier, line 101: you say "only lowercase English letters", but then say 27.  presumably you're including space in the count.

man/passwd.1.xml
alejandro-colomar: Hmm.  Usually references in SEE ALSO, such as references to other pages, are not introduced.  If a user goes to that site, they'll know why it was referenced.  But I'll think of adding some intro to it, especially for those who don't know what XKCD is, and foreseeing a possible future in which it doesn't exist anymore and people forgot what it was.

libmisc/obscure.c
alejandro-colomar: > this is more restrictive than the existing code. having it be dynamic based on the inputs was aligned with the intention of this code. but you're making it static, and we already have `PASS_MIN_LEN` as a configurable value that is checked earlier. so if you want to delete the dynamic strength checks, you should delete this length check entirely.

Yep, I didn't see the separate `PASS_MIN_LEN` test.  Then I'll just remove this test.

> 
> not that i have a problem with the dynamic strength check as-is.

It rewards passwords using wider character sets by allowing them to be shorter, which can reinforce the idea that it's a good thing to do in general.  Again, if one can remember long passwords using symbols and uppers, that's great, but enforcing that is problematic, and even this kind of reinforcement I believe it to be counterproductive in that regard.

man/passwd.1.xml
alejandro-colomar: True.  Maybe rephrased as: `It's fine to use simple character sets,
	such as passwords consisting only of lowercase letters` + ` if it helps memorize longer passwords`.

man/passwd.1.xml
alejandro-colomar: I do remember one, for my master PGP key.  Considering I only need a handful of passwords in my life (all the rest, are pass(1)'s task), I can memorize just a few very strong ones.  The trick I use to remember them is to break them into shorter chunks (say around 10 chars each), and treat them as words in a foreign language; I can certainly remember a few dozen "words" that are originally gibberish, with some repetition; then those words are connected to make sentences which are the actual password.

However, I admit my login, my SSH and my PGP signing passwords are half of that; not because I don't remember long ones, but because typing them dozens of times an hour would be cumbersome.  But 32 random `[:lower:][:digit:]` is definitely easy, and strong enough for almost anything.

alejandro-colomar: Oops, miscount here :)  For some reason, `27 - 1 = 27` yesterday.

alejandro-colomar:
Changes:

-  Completely drop test; don't replace it by a fixed length check.
-  Reword advice on passwd(1) to make the advice on simple char sets dependent on making longer passwords.
-  Introduce XKCD properly.

```
$ git range-diff master gh/strength strength 
1:  581b923a ! 1:  d101612b libmisc, man: Drop old check and advice for complex character sets in passwords
    @@ Metadata
      ## Commit message ##
         libmisc, man: Drop old check and advice for complex character sets in passwords
     
    -    I kept the minimum length as 8, since it was what we had for the
    -    simplest passwords already.  8 is probably not what you really want from
    -    a password, but I believe it should be up to users to chose what they
    -    want.  If we raise this too much, we may upset some people, so let's be
    -    cautious.
    -
         Add the relevant XKCD to the passwd(1) manual page.  It already explains
         most of the rationale behind this patch.
     
         Add also reference to makepasswd(1), which is a good way to generate
         strong passwords.  Personally, I commonly run `makepasswd --chars 64` to
    -    create my passwords.
    +    create my passwords, or 32 for passwords I need to type interactively
    +    often.
     
         The strength of a password is an exponential formula, where the base is
         the size of the character set, and the exponent is the length of the
    @@ libmisc/obscure.c: static /*@observer@*//*@null@*/const char *password_check (
                msg = _("too similar");
     -  } else if (simple (old, new)) {
     -          msg = _("too simple");
    -+  } else if (strlen (new) < 8) {
    -+          msg = _("too short");
        } else if (strstr (wrapped, newmono) != NULL) {
                msg = _("rotated");
        } else {
    @@ man/passwd.1.xml
     -  Then, the password is tested for complexity. As a general guideline,
     -  passwords should consist of 6 to 8 characters including one or more
     -  characters from each of the following sets:
    -+  Then, the password is tested for length.  As a general guideline,
    -+  passwords should be long.  It's fine to use simple character sets,
    -+  such as passwords consisting only of lowercase letters.  For a
    -+  password consisting only of lowercase English letters randomly
    -+  chosen, and a length of 64, there are 27^64 (approximately 2^304)
    -+  different possible combinations.  Being an exponential equation,
    -+  it's apparent that the exponent (the length) is more important than
    -+  the base (the size of the character set).
    -       </para>
    - 
    +-      </para>
    +-
     -      <itemizedlist mark='bullet'>
     -  <listitem>
     -    <para>lower case alphabetics</para>
    @@ man/passwd.1.xml
     -  </listitem>
     -      </itemizedlist>
     -
    +-      <para>
    +-  Care must be taken not to include the system default erase or kill
    +-  characters. <command>passwd</command> will reject any password which
    +-  is not suitably complex.
    ++  Then, the password is tested for complexity.
    ++  <command>passwd</command> will reject any password which is not
    ++  suitably complex.  Care must be taken not to include the system
    ++  default erase or kill characters.
    +       </para>
    + 
    +     </refsect2>
    +@@
    +   used as guesses to violate system security.
    +       </para>
    + 
    ++      <para>
    ++  As a general guideline, passwords should be long and random.  It's
    ++  fine to use simple character sets, such as passwords consisting
    ++  only of lowercase letters, if that helps memorizing longer
    ++  passwords.  For a password consisting only of lowercase English
    ++  letters randomly chosen, and a length of 32, there are 26^32
    ++  (approximately 2^150) different possible combinations.  Being an
    ++  exponential equation, it's apparent that the exponent (the length)
    ++  is more important than the base (the size of the character set).
    ++      </para>
    ++
            <para>
    -   Care must be taken not to include the system default erase or kill
    -   characters. <command>passwd</command> will reject any password which
    +   You can find advice on how to choose a strong password on
    +   http://en.wikipedia.org/wiki/Password_strength
     @@
            <citerefentry>
        <refentrytitle>chpasswd</refentrytitle><manvolnum>8</manvolnum>
    @@ man/passwd.1.xml
          </para>
     +
     +    <para>
    -+  https://xkcd.com/936/
    ++  The following web page comically (yet correctly) compares the
    ++  strength of two different methods for choosing a password:
    ++  "https://xkcd.com/936/"
     +    </para>
        </refsect1>
      </refentry>

```

alejandro-colomar:
```
$ git range-diff master gh/strength strength 
1:  d101612b ! 1:  f9443f13 libmisc, man: Drop old check and advice for complex character sets in passwords
    @@ Commit message
     
         Closes: <https://github.com/shadow-maint/shadow/issues/688>
         Link: <https://xkcd.com/936/>
    +    Cc: Mike Frysinger <vapier@gentoo.org>
     
      ## libmisc/obscure.c ##
```

alejandro-colomar:
Changes:

- Rebased to master
- Sign commit with GPG

```
$ git range-diff f9443f13^..f9443f13 shadow/master..strength 
1:  f9443f13 = 1:  3fe545b2 libmisc, man: Drop old check and advice for complex character sets in passwords
```

hallyn (approved):

hallyn:
@ikerexxe look good to you?

ikerexxe (approved):
Yes, it looks good to me. Merging!

PR #691, opened 2023-03-26
usermod: respect --prefix for --gid option

floppym:
The --gid option accepts a group name or id. When a name is provided, it is resolved to an id by looking up the name in the group database (/etc/group).

The --prefix option overides the location of the passwd and group databases. I suspect the --gid option was overlooked when wiring up the --prefix option.

useradd --gid already respects --prefix; this change makes usermod behave the same way.

Fixes: https://github.com/shadow-maint/shadow/commit/b6b2c756c91806b1c3e150ea0ee4721c6cdaf9d0

### DIFF
--- a/src/usermod.c
+++ b/src/usermod.c
@@ -1072,7 +1072,7 @@ static void process_flags (int argc, char **argv)
 				fflg = true;
 				break;
 			case 'g':
-				grp = getgr_nam_gid (optarg);
+				grp = prefix_getgr_nam_gid (optarg);
 				if (NULL == grp) {
 					fprintf (stderr,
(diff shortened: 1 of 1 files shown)


ikerexxe (changes requested):
Do you mind providing more information in the commit message as to why this change is necessary? I guess it's because we'd like to change the user's primary group by providing the GID value instead of its name, but it isn't very clear right now.

floppym:
I've updated the commit message.

hallyn (approved):

ikerexxe (approved):
Thanks for the explanation. Merging.

PR #702, opened 2023-04-05
CI: Make build logs more readable

alejandro-colomar:
If make fails in a multi-process invocation, the log is pretty much unreadable.  To make it readable, build as much as can be built without failing.  If it succeeds, then we're done.  If there was a failure, run a single-process make again, which we know should fail at its first target; that will print the few lines we're really interested in, and they will be at the very end of the log.

This has some side effects:  Now we build as much as we can, instead of failing as early as possible; this may make CI a bit slower.  However, it also has the benefit that you see _all_ the error messages that could be given, instead of needing to fix the first error to see the next and so on.

### DIFF
--- a/.github/workflows/static-code-analysis.yml
+++ b/.github/workflows/static-code-analysis.yml
@@ -32,7 +32,10 @@ jobs:
     - name: Build shadow-utils
       run: |
         PROCESSORS=$(/usr/bin/getconf _NPROCESSORS_ONLN)
-        make -j$PROCESSORS
+        make -kj$PROCESSORS || true
+
+    - name: Check build errors
+      run: make
 
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@v2
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ addons:
     notification_email: christian.brauner@ubuntu.com,serge@hallyn.com
 
     build_command_prepend: "./autogen.sh --without-selinux --disable-man"
-    build_command: "make -j4"
+    build_command: "make -kj4 || make"
     branch_pattern: master
 
 script:
--- a/share/containers/alpine.dockerfile
+++ b/share/containers/alpine.dockerfile
@@ -9,7 +9,8 @@ COPY ./ /usr/local/src/shadow/
 WORKDIR /usr/local/src/shadow/
 
 RUN ./autogen.sh --without-selinux --disable-man --disable-nls --with-yescrypt
-RUN make -j4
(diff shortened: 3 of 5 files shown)


ikerexxe:
In my opinion the user experience would be more complicated for people that don't know that the building process is executed twice, and the same error would be reported twice. Moreover, probably other errors would appear, and only once.

ikerexxe:
From the proposed options the one that I like the most is to separate it in different steps. Still, that feels a little bit awkward to me.

ikerexxe:
I keep coming to this PR and I don't know if this is a good idea. In my opinion, the UX would become confusing for people that aren't aware that we are running the build process twice.

alejandro-colomar:
On 4/14/23 08:56, Iker Pedrosa wrote:
> I keep coming to this PR and I don't know if this is a good idea. In my opinion, the UX would become confusing for people that aren't aware that we are running the build process twice.

	make -kj || true
	make

I think the above can only be read with one meaning:

	Build as fast as possible, ignoring errors
	Check if there's any build errors.

The CI step names should help, and the commit message in which that
code was introduced would help for those who want some confirmation,
but the two lines of code, are quite explicit.  They are a bit
surprising, because it's not usual to see them, but I wouldn't say
they're difficult to read.

-- 
GPG key fingerprint: A9348594CE31283A826FBDD8D57633D441E25BB5

alejandro-colomar:
On 4/18/23 09:21, Iker Pedrosa wrote:
>> I think the above can only be read with one meaning: Build as fast as possible, ignoring errors Check if there's any build errors.
> 
> You are assuming that people read all the CI steps, and not just the failing one.
> 
> Although I have my doubts I will proceed to merge this change. I reserve the right to reverse it if too many people complain about it.

Fair enough :-)

Thanks!

-- 
GPG key fingerprint: A9348594CE31283A826FBDD8D57633D441E25BB5

alejandro-colomar:
@ikerexxe  Maybe we can use `make -j4 -Otarget` instead of `make -j4 || true`+`make`.  That would also keep the output synchronized, but without using two steps, and would also reduce the load of the CI, since it wouldn't need to build everything.

Cc: @thresheek, @ac000

ikerexxe:
> @ikerexxe Maybe we can use `make -j4 -Orecurse` instead of `make -j4 || true`+`make`. That would also keep the output synchronized, but without using two steps, and would also reduce the load of the CI, since it wouldn't need to build everything.
> 
> Cc: @thresheek, @ac000

I didn't like running `make` two times, so your proposal would be wonderful.

PR #703, opened 2023-04-06
fix typos

uniontech-lilinjie:
Signed-off-by: lilinjie lilinjie@uniontech.com
fix typos

### DIFF
--- a/man/po/da.po
+++ b/man/po/da.po
@@ -5471,7 +5471,7 @@ msgstr ""
 
 #: login.defs.5.xml:34(para) login.1.xml:34(para)
 msgid "Enable logging of successful logins."
-msgstr "Aktiver logning af succesfulde logind."
+msgstr "Aktiver logning af successfulde logind."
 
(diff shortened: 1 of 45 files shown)


man/po/da.po
```diff
@@ -5471,7 +5471,7 @@ msgstr ""
 
 #: login.defs.5.xml:34(para) login.1.xml:34(para)
 msgid "Enable logging of successful logins."
-msgstr "Aktiver logning af succesfulde logind."
+msgstr "Aktiver logning af successfulde logind."
```
ikerexxe, line 5474: Was this change intentional? I'm not a Danish speaker so I'm not sure, but it seems like you replaced all `succesful` instances by `successful`

hallyn: Looking at translate.google.com, it looks like succesfulde is 'success' while successfulde is successful.  So I'll take it these are intentional, and merge.

Thanks, @uniontech-lilinjie , and thanks for looking, @ikerexxe

PR #710, opened 2023-04-22
libmisc/yesno.c: Use getline(3) and rpmatch(3)

alejandro-colomar:
getline(3) is much more readable than manually looping.  It has some
overhead due to the allocation of a buffer, but that shouldn't be a
problem here.  If that was a problem, we could reuse the buffer (thus
making the function non-reentrant), but I don't think that's worth the
extra complexity.

Using getline(3) has the benefit that the function can then use
rpmatch(3) just by changing one line; note to downstream packagers:
distros that use glibc may be interested in patching that line for i18n
purposes.

While we're at it, apply some other minor improvements to this file:

-  Remove comment saying which files use this function.  That's likely
   to get outdated.  And anyway, it's just a grep(1) away, so it doesn't
   really add any value.

-  Remove unnecessary casts to (void) that were used to verbosely ignore
   errors from stdio calls.  They add clutter without really adding much
   value to the code (or I don't see it).

-  Add in a comment how this code could be implemented using rpmatch(3),
   so that if it becomes more portable in the future, we might just
   replace that line.  It will serve as a remainder of the intention.

-  Remove comments from the function body.  They make the function less
   readable.  Instead, centralize the description of the function into a
   man-page-like comment before the function definition.  This keeps the
   function body short and sweet.

-  Minor whitespace style changes (it doesn't hurt the diff at this
   point, since most of the affected lines were already touched by other
   changes, so I applied my preferred style :).

Cc: @ferivoz 
Cc: @hallyn 
Signed-off-by: @alejandro-colomar

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -50,6 +50,7 @@ AC_CHECK_FUNCS(arc4random_buf futimes \
 	initgroups lckpwdf lutimes mempcpy \
 	setgroups updwtmp updwtmpx innetgr \
 	getspnam_r \
+	rpmatch \
 	memset_explicit explicit_bzero stpecpy stpeprintf)
 AC_SYS_LARGEFILE
 
--- a/libmisc/yesno.c
+++ b/libmisc/yesno.c
@@ -1,59 +1,84 @@
 /*
  * SPDX-FileCopyrightText: 1992 - 1994, Julianne Frances Haugh
  * SPDX-FileCopyrightText: 2007 - 2008, Nicolas François
+ * SPDX-FileCopyrightText: 2023, Alejandro Colomar <alx@kernel.org>
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-/*
- * Common code for yes/no prompting
- *
- * Used by pwck.c and grpck.c
- */
 
 #include <config.h>
 
 #ident "$Id$"
 
+#include <stdbool.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include "prototypes.h"
 
-/*
- * yes_or_no - get answer to question from the user
- *
- *	It returns false if no.
- *
- *	If the read_only flag is set, it will print No, and will return
- *	false.
- */
-bool yes_or_no (bool read_only)
-{
-	int c;
-	bool result;
 
-	/*
-	 * In read-only mode all questions are answered "no".
-	 */
+/*
+ * Synopsis
+ *	bool yes_or_no(bool read_only);
+ *
+ * Arguments
+ *	read_only
+ *		In read-only mode, all questions are answered "no".  It
+ *		will print "No" to stdout.
+ *
+ * Description
+ *	After a yes/no question, this function gets the answer from the
+ *	user.
+ *
+ *	Calls to this function will normally be preceeded by a prompt on
+ *	stdout, so we should fflush(3).
+ *
+ * Return value
+ *	false	"no"
+ *	true	"yes"
+ *
+ * See also
+ *	rpmatch(3)
+ */
+
+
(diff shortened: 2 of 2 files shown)


alejandro-colomar:
Changes:

-  Add missing #include <stdbool.h>

alejandro-colomar:
Changes:

-  Remove unnecessary parentheses.  This makes the line shorter, so that
   it doesn't pass the 80-col limit so much (still, but it's readable).

ferivoz:
This changes the algorithm from O(1) memory complexity to O(n). I did not do this in my original pull request just because it wastes memory without benefit.

I would argue that it makes sense if rpmatch is in place though.

Also, the callers are not setuid/setgid, so it probably isn't harmful either. Just my two cents!

alejandro-colomar:
Hi Samanta!

@ferivoz wrote:

> This changes the algorithm from O(1) memory complexity to O(n).  I did
> not do this in my original pull request just because it wastes memory
> without benefit.                                                                           

True, but the lines we actually expect to parse are not that long.  I
mean, it's a yes/no question :)

> I would argue that it makes sense if rpmatch is in place though.                                                    

Yes; while we can't use it, for portability to musl, most GNU/Linux
distros use glibc, so they can patch upstream shadow to uncomment the
rpmatch(3) call.  This patch makes it easy for downstream to use
rpmatch(3).

> Also, the callers are not setuid/setgid, so it probably isn't harmful
> either.  Just my two cents!

Cheers,
Alex

alejandro-colomar:
Changes:

-  Use rpmatch(3).

I gon't know why I didn't consider this as an option.  Maybe it's my
tendence to avoid preprocessor conditionals.  But in this case it was
useful.  Now I have a justification for getline(3) more than "it's more
readable" :)

ferivoz:
With a fallback rpmatch solution this definitely makes sense!

alejandro-colomar:
Changes:

-  Squash commits.
-  Add Acked-by tag (@ferivoz).
-  Add myself to Copyright.

Here's the diff for the commit message:

```
    @@ Metadata
     
      ## Commit message ##
    -    libmisc/yesno.c: Use getline(3); and other cosmetic changes
    +    libmisc/yesno.c: Use getline(3) and rpmatch(3)
     
         getline(3) is much more readable than manually looping.  It has some
         overhead due to the allocation of a buffer, but that shouldn't be a
    @@ Commit message
         making the function non-reentrant), but I don't think that's worth the
         extra complexity.
     
    -    Using getline(3) has the benefit that the function can then use
    -    rpmatch(3) just by changing one line; note to downstream packagers:
    -    distros that use glibc may be interested in patching that line for i18n
    -    purposes.
    +    Using rpmatch(3) instead of a simple y/n test provides i18n to the
    +    response checking.  We have a fall-back minimalistic implementation for
    +    systems that lack this function (e.g., musl libc).
     
         While we're at it, apply some other minor improvements to this file:
     
    @@ Commit message
            errors from stdio calls.  They add clutter without really adding much
            value to the code (or I don't see it).
     
    -    -  Add in a comment how this code could be implemented using rpmatch(3),
    -       so that if it becomes more portable in the future, we might just
    -       replace that line.  It will serve as a remainder of the intention.
    -
         -  Remove comments from the function body.  They make the function less
            readable.  Instead, centralize the description of the function into a
            man-page-like comment before the function definition.  This keeps the
    @@ Commit message
            point, since most of the affected lines were already touched by other
            changes, so I applied my preferred style :).
     
    -    Cc: Samanta Navarro <ferivoz@riseup.net>
    +    Acked-by: Samanta Navarro <ferivoz@riseup.net>
```

PR #711, opened 2023-04-26
newgrp/useradd: always set SIGCHLD to default

ferivoz:
The tools newgrp and useradd expect waitpid to behave as described in its manual page. But the notes indicate that if SIGCHLD is ignored, waitpid behaves differently.

A user could set SIGCHLD to ignore before starting newgrp through exec. Children of newgrp would not become zombies and their PIDs could be reassigned before newgrp could call kill with the child pid and SIGCONT.

The useradd tool is not installed setuid, but I have added the default there as well (copied from vipw).


hallyn (approved):
Thanks, looks good.

PR #712, opened 2023-04-28
libmisc/yesno.c: Fix regression

ferivoz:
The getline function does not return a pointer but the amount of read characters. The error return value to check for is -1.

Set buf to NULL to avoid dereference of an uninitialized stack value.

The getline function returns -1 if size argument is NULL. Always use a valid pointer even if size is unimportant.

### DIFF
--- a/libmisc/yesno.c
+++ b/libmisc/yesno.c
@@ -50,8 +50,9 @@ static int rpmatch(const char *response);
 bool
 yes_or_no(bool read_only)
 {
-	bool  ret;
-	char  *buf;
+	bool   ret;
+	char   *buf;
+	size_t size;
 
 	if (read_only) {
 		puts(_("No"));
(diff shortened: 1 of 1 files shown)


hallyn (approved):
Thank you.

This is why I usually like to stick with one idiom in a codebase.  Though I do see there's exactly one other user of getline, which I added :)  Even so I failed to notice this.

alejandro-colomar:
@ferivoz wrote:
>  The getline function does not return a pointer but the amount of read       
>  characters.  The error return value to check for is -1.                      
>                                                                              
>  Set buf to NULL to avoid dereference of an uninitialized stack value.       
>                                                                              
>  The getline function returns -1 if size argument is NULL. Always use a valid
>  pointer even if size is unimportant.                                        

D'oh!  Thanks for fixing those bugs!  And sorry for them.  :)

PR #713, opened 2023-04-28
login_prompt: Fix implementation issues

ferivoz:
The `login_prompt` function used in `login` utility allows setting of environment variables when entered next to a user name. These variables still exist in memory even if login attempt failed. A subsequent login from another user takes these environment variables into account.

Also within the privileged code of login running as root a memory leak exists with these environment variables. With default values it would be very hard to trigger an OOM or do other malicious things though.

My suggestion is to remove this undocumented feature.

Also simplified API and fixed a signal unsafe usage while at it.

### DIFF
--- a/libmisc/loginprompt.c
+++ b/libmisc/loginprompt.c
@@ -22,7 +22,7 @@
 
 static void login_exit (unused int sig)
 {
-	exit (EXIT_FAILURE);
+	_exit (EXIT_FAILURE);
 }
 
 /*
--- a/libmisc/loginprompt.c
+++ b/libmisc/loginprompt.c
@@ -32,7 +32,7 @@ static void login_exit (unused int sig)
  * is set in login.defs, this file is displayed before the prompt.
  */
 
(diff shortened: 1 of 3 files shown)


libmisc/loginprompt.c
```diff
@@ -112,34 +112,6 @@ void login_prompt (const char *prompt, char *name, int namesize)
 
 	name[i] = '\0';
 
-	/*
```
hallyn, line 115: So long as you're doing this (which sounds good to me), two of the blocks above this (NOT including name[i] = '\0') can be dropped as well, right?
 
```
105         while (isgraph (*cp)) {
106                 cp++;
107         }
108
109         if ('\0' != *cp) {
110                 cp++;
111         }
```

ferivoz: Thank you for pointing it out. I have removed these lines as well. Now shadow login is compatible with dosbox login.

hallyn (approved):

hallyn:
Thanks.

PR #714, opened 2023-05-04
Add --prefix to passwd, chpasswd and chage

### LINKED ISSUE #657: Unifying --prefix behaviour
(I actually intended to open a PR for this, but since I didn't get around to it I at least wanted to create this issue now)

The following commands support the `--prefix` flat:
useradd, userdel, usermod, groupadd, groupdel and groupmod
Other commands don't, like:
passwd, chpasswd and chage

I think to be consistent all our commands should support the `--prefix` option.

I only took a quick look at the code. And it seems like in the first group of commands we use helper functions from `libmisc/prefix_flat.c` like `process_prefix_flag()` and wrappers like `prefix_getgrnam()` instead of `getgrnam()`.

Dzejrou:
I'm trying to implement the `--prefix`/`-P` options for `passwd`, `chpasswd` and `chage` to (partially) fix #657. Unfortunately, the code of `passwd`, when compiled with the USE_PAM macro defined (default), does not seem to be able to work with the `--prefix` option due to pam using services rather than database paths. This can be worked around by an approach already used in `chpasswd` (the `use_pam` variable, which can cause `chpasswd` to fall back to non-PAM code), so in my changes I tried to mimic that - is this the right approach?


Dzejrou:
Not sure why my changes to chage would add a dependency on dlclose on debian, currently investigating. However, I'd say those issues found by the bot are false positives (let me know if I'm wrong):

* The inclusion of a .c file is done in the same way as the other getXXbyYY files.
* The prefix_getpwnam_r function uses fopen in exactly the same way as prefix_getpwnam.

src/passwd.c
```diff
@@ -838,6 +843,8 @@ int main (int argc, char **argv)
 				break;
 			case 'R': /* no-op, handled in process_root_flag () */
 				break;
+            case 'P': /* no-op, handled in process_prefix_flag () */
```
hallyn, line 855: (please do use the same tabbing as the lines around it)

Dzejrou: Done, thanks.

libmisc/prefix_flag.c
```diff
@@ -229,6 +229,7 @@ extern struct passwd *prefix_getpwnam(const char* name)
 		return getpwnam(name);
 	}
 }
+#if HAVE_FGETPWENT_R
```
hallyn, line 232: The #if HAVE_FGETPWENT_R is also needed around the lib/prototypes.h definition of prefix_getpwnam_r.

ferivoz:
Please evaluate https://github.com/shadow-maint/shadow/pull/731 before progressing in adding the `--prefix` option to setuid binaries like passwd.

libmisc/prefix_flag.c
Dzejrou: Done, thank you.

hallyn:
Please squash-rebase the last two commits into the relevant earlier commits.

As a separate suggestion:

Overall, in src/passwd.c, if you did something like

```
#ifdef USE_PAM
static bool use_pam = true;
#else
static bool use_pam = false;
#endif
```

Then all the other instances of

```
#ifdef USE_PAM
if (use_pam)
#endif
{
```
Could simply become
```
if (use_pam)
{
```

Dzejrou:
> Please squash-rebase the last two commits into the relevant earlier commits.

Done, though this is the first time I'm doing this, so please let me know whether this is what you wanted.

> As a separate suggestion:
> ...

I can do that, but just to be sure I'd mention the reason why I did it this way (I originally wanted to do what you suggested) and that is that chpasswd already does this [0] and I wanted to be consistent. But if you'd still prefer that, let me know here and I will do it.

[0] https://github.com/shadow-maint/shadow/blob/master/src/chpasswd.c#L440

jubalh:
As discussed elsewhere. Commenting here for completeness:
This will be needed for https://github.com/yast/yast-users/pull/374

hallyn:
> As discussed elsewhere. Commenting here for completeness: This will be needed for [yast/yast-users#374](https://github.com/yast/yast-users/pull/374)

Thanks.

And, for these users, using the not-PAM when doing --prefix will be ok?  Or do they expect to use the host PAM modules, but with relocated passwd+shadow?

configure.ac
```diff
@@ -678,6 +678,8 @@ if test "$with_skey" = "yes"; then
 	]])],[AC_DEFINE(SKEY_BSD_STYLE, 1, [Define to support newer BSD S/Key API])],[])
 fi
 
+AC_CHECK_FUNC(fgetpwent_r, [AC_DEFINE(HAVE_FGETPWENT_R, 1, [Define to 1 if you have the declaration of 'fgetpwent_r'])])
```
hallyn, line 681: Please change this to "Defined".  Because "Define" implies that the user should do it.

hallyn:
With the one requested change, and assuming a positive answer to the question, +1 from me, thanks.

configure.ac
Dzejrou: Should be fixed now.

ancorgs:
> And, for these users, using the not-PAM when doing --prefix will be ok?

Yes. It is ok.

> Or do they expect to use the host PAM modules, but with relocated passwd+shadow?

The use-case is to update databases of users/groups that are in an alternative location and that are then exported via NIS. Those users are not used to authenticate locally. They are only used to feed NIS so they can be used by PAM to authenticate on remote systems.

ikerexxe:
I'll take a look at it tomorrow.

src/passwd.c
```diff
@@ -506,11 +501,14 @@ oom (void)
 
 static char *update_crypt_pw (char *cp)
 {
-#ifndef USE_PAM
-	if (do_update_pwd) {
-		cp = xstrdup (crypt_passwd);
+#ifdef USE_PAM
+	if (!use_pam)
+#endif				/* USE_PAM */
```
ikerexxe, line 506: If you already set `use_pam` when `USE_PAM` is defined, then why do you need to build this conditionally?

src/passwd.c
```diff
@@ -533,11 +531,14 @@ static char *update_crypt_pw (char *cp)
 
 		strcpy (newpw, "!");
 		strcat (newpw, cp);
-#ifndef USE_PAM
-		if (do_update_pwd) {
-			free (cp);
-		}
+#ifdef USE_PAM
+		if (!use_pam)
 #endif /* USE_PAM */
```
ikerexxe, line 536: Same here.

src/passwd.c
```diff
@@ -748,6 +750,13 @@ int main (int argc, char **argv)
 	process_root_flag ("-R", argc, argv);
 	prefix = process_prefix_flag ("-P", argc, argv);
 
+#ifdef USE_PAM
+	if (prefix[0]) {
+		use_pam = false;
+		do_update_age = true;
+	}
+#endif /* USE_PAM */
```
ikerexxe, line 758: Do you really need to build this conditionally?

ikerexxe (changes requested):
I have some concerns regarding the usage of `use_pam`  variable and `USE_PAM` macro in the code. If you are to use `use_pam`, then try to avoid conditionally building its code with `USE_PAM`. I noted some comments in this regard, but there might be other occurrences.

src/passwd.c
jubalh: I think https://github.com/shadow-maint/shadow/pull/714#issuecomment-1554849599 and https://github.com/shadow-maint/shadow/pull/714#issuecomment-1555167191 are about this as well.

Dzejrou: Fixed. I already had the changes stages since when Hallyn mentioned this, the original reason was that I tried to be consistent with chpasswd (which does this).

src/passwd.c
Dzejrou: Fixed.

src/passwd.c
Dzejrou: Fixed.

src/passwd.c
ikerexxe: I prefer to have a neater code than consistency with other tools.

src/passwd.c
```diff
@@ -1003,53 +1006,55 @@ int main (int argc, char **argv)
...
-		 * Otherwise, go ahead and set a new password.
+		 * The user name is valid, so let's get the shadow file entry.
 		 */
-		check_password (pw, sp);
+		sp = prefix_getspnam (name); /* !use_pam, no need for xprefix_getspnam */
+		if (NULL == sp) {
+			if (errno == EACCES) {
+				(void) fprintf (stderr,
+								_("%s: Permission denied.\n"),
+								Prog);
```
ikerexxe, line 1019: Indentation issue

src/passwd.c
```diff
@@ -1003,53 +1006,55 @@ int main (int argc, char **argv)
...
-		if (new_password (pw) != 0) {
-			(void) fprintf (stderr,
-			                _("The password for %s is unchanged.\n"),
-			                name);
-			closelog ();
-			exit (E_NOPERM);
+			if (new_password (pw) != 0) {
+				(void) fprintf (stderr,
+								_("The password for %s is unchanged.\n"),
+								name);
```
ikerexxe, line 1049: Indentation issue

ikerexxe (commented):
Just two minor comments. Once those are fixed we'll be ready to merge.

src/passwd.c
Dzejrou: Should be fixed now, thanks.

src/passwd.c
Dzejrou: Should be fixed now.

ikerexxe (approved):
Perfect! Thank you

PR #715, opened 2023-05-05
commonio.c: Simplification

ferivoz:
I have removed some obsolete code. Also, since POSIX.1-2008 is a requirement for shadow, O_CLOEXEC can be used which removes the need to call fcntl for a file descriptor.


ikerexxe (approved):
LGTM

hallyn (approved):

Issue #686, opened 2023-03-23
an error is reported when the grpconv command is executed


wyj611:
![image](https://user-images.githubusercontent.com/22022968/227149679-fda4b235-88f1-43e6-9287-0bbbb781678c.png)

wyj611:
Maybe there's something wrong with this PR:[changing lock mechanism](https://github.com/shadow-maint/shadow/commit/65470e5c7ddbc2885ad35fbd0218b91a75626c2f), should O_CREAT be added to the second parameter of function open in the function do_fcntl_lock?

jubalh:
@wyj611 did the changes fix your issue?
If yes we can close this.

wyj611:
> @wyj611 did the changes fix your issue? If yes we can close this.

yes, of course

PR #716, opened 2023-05-06
open with O_CREAT when lock path does not exist

edneville:
Reported in #686, by wyj611 when trying to lock a file that is not present

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -215,7 +215,7 @@ int do_fcntl_lock (const char *file, bool log, short type)
 		.l_len = 0,
 	};
 
-	fd = open (file, O_WRONLY, 0600);
+	fd = open (file, O_WRONLY | O_CREAT, 0600);
 	if (-1 == fd) {
 		if (log) {
 			(void) fprintf (shadow_logfd, "%s: %s: %s\n",
@@ -224,8 +224,7 @@ int do_fcntl_lock (const char *file, bool log, short type)
 		return 0;
 	}
 
-	fcntl (fd, F_OFD_SETLKW, &lck);
-	close(fd);
+	fcntl (fd, F_SETLKW, &lck);
 	return(1);
 }
 


cgzones:
Why check first and not use `O_WRONLY | O_CREAT` instead?
(Also the previous code used `O_TRUNC`, do not know whether it is necessary here.)

edneville:
Hey,

On 2023-05-07 08:14-0700, cgzones wrote:
> Why check first and not use `O_WRONLY | O_CREAT` instead?
> (Also the previous code used `O_TRUNC`, do not know whether it is 
> necessary here.)

Fair spot.

Because in a past life I truc'd in error, the stat was a guard around 
repeating that, but used creat.

It isn't needed, so long as we don't truc when locking files with data.

Ed

hallyn:
Wait, what is going on in this code?

It locks it, then immediately closes it.  Which releases the lock.

hallyn (approved):

hallyn:
(technically the removal of the close(fd) should be a separate commit or mentioned in the changelog, but let's rush this as not actually locking turns out to be more urgent to fix than exiting because we can't lock :) )

cgzones:
While leaking the file descriptor is acceptable for short running shadow binaries, it is definitely not appropriate for long running clients of libsubid.

hallyn:
> While leaking the file descriptor is acceptable for short running shadow binaries, it is definitely not appropriate for long running clients of libsubid.

What exactly is this in reference to?

Note that if we immediately close the fd here, we immediately lose the lock.  The close was wrong.

cgzones:
>>   While leaking the file descriptor is acceptable for short running shadow binaries, it is definitely not appropriate for long running clients of libsubid.
>
>What exactly is this in reference to?
>
>Note that if we immediately close the fd here, we immediately lose the lock. The close was wrong.

This is now obsolete with https://github.com/shadow-maint/shadow/commit/72290ede0e0d0cf960127bf42f11a7e227d4dac5.

PR #722, opened 2023-05-10
Fix broken docbook translations

hallyn:
its by default does not support xml tags inside translatable units.  Use custom its rules from

https://www.w3.org/TR/xml-i18n-bp/#relating-docbook-plus-its

to enable the tags which are in use by docbook.


ikerexxe (approved):
LGTM so merging

PR #725, opened 2023-05-10
Update German translations

hallyn:
find the attached German message catalogue proofread by the German language team.

Best regards
Markus


hallyn:
Interesting - looks like a github infrastructure error.

PR #727, opened 2023-05-11
Update French translations

hallyn:
Please find attached the french updated translation of shadow-man-page, proofread by the debian-l10n-french mailing list contributors.


po/fr.po
```diff
@@ -2722,12 +2707,14 @@ msgstr "Indiquez les permissions du fichier mailbox"
 
 #, c-format
 msgid "%s warning: %s's uid %d is greater than SYS_UID_MAX %d\n"
-msgstr ""
+msgstr "%s attention: l'uid %d de %s est plus grand que SYS_UID_MAX %d\n"
```
ikerexxe, line 2710: This line fails because the format changes from `%s %s %d %d` to `%s %d %s %d`.

po/fr.po
```diff
@@ -2722,12 +2707,14 @@ msgstr "Indiquez les permissions du fichier mailbox"
...
 
 #, c-format
 msgid ""
 "%s warning: %s's uid %d outside of the UID_MIN %d and UID_MAX %d range.\n"
 msgstr ""
+"%s attention: l'uid %d de %s est en dehors de la plage UID_MIN %d et UID_MAX "
```
ikerexxe, line 2716: There's a similar problem here.

po/fr.po
ikerexxe: I'd suggest to use the following message:
`%s attention: l'uid de %s, %d, est plus grand que SYS_UID_MAX %d\n`

po/fr.po
ikerexxe: I'd suggest to use the following message:
`%s attention: l'uid de %s, %d, est en dehors de la plage UID_MIN %d et UID_MAX `

ikerexxe (approved):
Fixing those two points made CI happy. So, I'm good with the PR.

By the way, do you mind replying to the original email with the modified patch?

PR #730, opened 2023-05-18
chsh: Verify that login shell path is absolute

ferivoz:
The getusershell implementation of musl returns every line within the /etc/shells file, which even includes comments. Only consider absolute paths for login shells.

Never allow a user to set a shell starting with `*` because that would lead to security issues (login and su would treat the home directory as a subsystem root).


hallyn (approved):

ikerexxe (approved):
LGTM. Thanks for the patch.

PR #732, opened 2023-05-18
Plug econf memory leaks

stoeckmann:
You can see the memory leaks with address sanitizer if shadow is compiled with `--enable-vendordir=/usr/etc`.

How to reproduce:

1. Prepare a custom shell file as root
```
mkdir -p /etc/shells.d
echo /bin/myshell > /etc/shells.d/custom
```

2. Run chsh as regular user
```
chsh
```


hallyn (approved):
Thank you.  I left some extra comments in here for my own future reference, but what you have here looks good.  There's just more to be done.

PR #733, opened 2023-05-19
Extend Subsystem Protection

ferivoz:
These commits do not address any real security issues and are merely defensive measurements against faulty configurations:

- login and su could run into infinite loops if subsystems address each other (easiest form: home directory is root)
- check_perms in su is recursive. Nested subsystems could trigger a stack overflow
- chroot/chdir should be called in that order to avoid races

### DIFF
--- a/libmisc/sub.c
+++ b/libmisc/sub.c
@@ -15,8 +15,10 @@
 #include <sys/types.h>
 #include "prototypes.h"
 #include "defines.h"
+#define	MAX_SUBROOT2	"maximum subsystem depth reached\n"
 #define	BAD_SUBROOT2	"invalid root `%s' for user `%s'\n"
 #define	NO_SUBROOT2	"no subsystem root `%s' for user `%s'\n"
+#define	MAX_DEPTH	1024
 /*
  * subsystem - change to subsystem root
  *
@@ -27,6 +29,18 @@
  */
 void subsystem (const struct passwd *pw)
 {
+	static int depth = 0;
+
+	/*
+	 * Prevent endless loop on misconfigured systems.
+	 */
+	if (++depth > MAX_DEPTH) {
+		printf (_("Maximum subsystem depth reached\n"));
+		SYSLOG ((LOG_WARN, MAX_SUBROOT2));
+		closelog ();
+		exit (EXIT_FAILURE);
+	}
+
 	/*
 	 * The new root directory must begin with a "/" character.
 	 */
--- a/libmisc/sub.c
+++ b/libmisc/sub.c
@@ -57,8 +57,8 @@ void subsystem (const struct passwd *pw)
 	 * must be able to change into it.
 	 */
 
-	if (   (chdir (pw->pw_dir) != 0)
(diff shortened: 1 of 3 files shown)


libmisc/sub.c
```diff
@@ -27,6 +29,18 @@
  */
 void subsystem (const struct passwd *pw)
 {
+	static int depth;
```
hallyn, line 32: Hm, but depth will only be initialized to 0 if it's BSS.  Does being 'static' but defined inside a function qualify?  I did not think it did.

ferivoz: I have updated the commit to initialize depth with 0. Even though I think that it will always be 0 initialized, it matches other situations in which static variables are used.

ikerexxe (changes requested):
Apart from the inline comments I'd like to mention that Debian's build is failing. Please take a look.

src/su.c
```diff
@@ -692,7 +693,7 @@ static /*@only@*/struct passwd * check_perms (void)
 		endpwent ();		/* close the old password databases */
 		endspent ();
 		pw_free (pw);
-		return check_perms ();	/* authenticate in the subsystem */
+		goto top; /* authenticate in the subsystem */
```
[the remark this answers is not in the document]
hallyn: Heh - I don't share the aversion to goto, however, since the only state that actually needs to be maintained between the calls is the tasks's chroot and curdir, an alternative implementation would be something like

```
static struct passwd *check_perms(void)
{
    struct passwd *pw = NULL;

    while (pw == NULL)
        pw = do_check_perms();
    return pw;
}
```
with do_check_perms() being the current code except return NULL instead of return check_perms() at the bottom.

libmisc/sub.c
```diff
@@ -15,8 +15,10 @@
 #include <sys/types.h>
 #include "prototypes.h"
 #include "defines.h"
+#define	MAX_SUBROOT2	"maximum subsystem depth reached\n"
 #define	BAD_SUBROOT2	"invalid root `%s' for user `%s'\n"
 #define	NO_SUBROOT2	"no subsystem root `%s' for user `%s'\n"
+#define MAX_DEPTH	1024
```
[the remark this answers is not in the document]
ferivoz: Thanks, adjusted!

src/su.c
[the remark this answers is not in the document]
ferivoz: I have integrated the do_check_perms approach. This should also fix the Debian build.

ikerexxe (approved):
LGTM! Thank you for the patches.

libmisc/sub.c
hallyn: Hm, but the patch doesn't show it being initialized.

Did that change get lost?

ferivoz: I'm sorry for that. The rework of goto to a function call was easier for me with a reset of the branch. And unfortunately I have reset this change as well. It is in the branch again. Thank you for carefully reviewing everything!

hallyn (approved):
Thanks.

PR #734, opened 2023-05-19
Fix possible NULL dereference in econf code.

hallyn:


lib/getdef.c
```diff
@@ -512,7 +505,12 @@ static void def_load (void)
 	for (size_t i = 0; i < key_number; i++) {
 		char *value;
 
-		econf_getStringValue(defs_file, NULL, keys[i], &value);
+		error = econf_getStringValue(defs_file, NULL, keys[i], &value);
+		if (error) {
+			SYSLOG ((LOG_CRIT, "failed reading key %d from econf [%s]",
```
stoeckmann, line 510: %zu but otherwise I agree. :)

hallyn: Thanks - pushed with %zu

lib/getdef.c
```diff
@@ -505,7 +505,12 @@ static void def_load (void)
 	for (size_t i = 0; i < key_number; i++) {
 		char *value;
 
-		econf_getStringValue(defs_file, NULL, keys[i], &value);
+		error = econf_getStringValue(defs_file, NULL, keys[i], &value);
+		if (error) {
+			SYSLOG ((LOG_CRIT, "failed reading key %zu from econf [%s]",
+				i, econf_errString(error)));
+			exit (EXIT_FAILURE);
```
ikerexxe, line 512: You are leaking `keys` and `defs_file`. It's everywhere in this function, but since you are modifying this path I think you should at least make sure that they don't leak in it.

PR #739, opened 2023-05-27
Typo fix, and while at it, simplify code

alejandro-colomar:
- xgetXXbyYY: tfix
- xgetXXbyYY: Centralize error handling
- xgetXXbyYY: Simplify elifs
- xgetXXbyYY: Reorganize code [dropped]

Cc: @ferivoz, @hallyn

### DIFF
--- a/libmisc/xgetXXbyYY.c
+++ b/libmisc/xgetXXbyYY.c
@@ -54,9 +54,7 @@
 
 	result = MALLOC(LOOKUP_TYPE);
 	if (NULL == result) {
-		fprintf (log_get_logfd(), _("%s: out of memory\n"),
-		         "x" STRINGIZE(FUNCTION_NAME));
-		exit (13);
+		goto oom;
 	}
 
 	while (true) {
@@ -70,10 +68,7 @@
 			 * the shadow *_free functions. */
 			LOOKUP_TYPE *ret_result = DUP_FUNCTION(result);
 			if (NULL == ret_result) {
-				fprintf (log_get_logfd(),
-				         _("%s: out of memory\n"),
-				         "x" STRINGIZE(FUNCTION_NAME));
-				exit (13);
+				goto oom;
 			}
 			free(buffer);
 			free(result);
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
Changes:

-  Reorganize in a way that I don't need to divide by 4 the initial
   length value.  I was never convinced by that.

```
$ git range-diff master..gh/tfix master..tfix 
1:  3a344a66 = 1:  3a344a66 xgetXXbyYY: tfix
2:  b87ce119 = 2:  b87ce119 xgetXXbyYY: Centralize error handling
3:  d0522b5c = 3:  d0522b5c xgetXXbyYY: Simplify elifs
4:  b4aeb604 ! 4:  35cb84b8 xgetXXbyYY: Reorganize code
    @@ Metadata
     
      ## Commit message ##
    -    xgetXXbyYY: Reorganize code
    +    xgetXXbyYY: Rewrite forever loop as a do while
     
    -    -  Move assignment of length to before use.  While it adds some extra
    -       complexity to reading the loop as a single iteration, because in
    -       a single iteration, we wouldn't even be reassigning length at all,
    -       I believe it will allow other refinements to the loop that compensate
    -       for that.  And in any case, the assignment is closer to its use,
    -       which is a good thing.
    +    -  Move assignment of length to before the breaks.  It will allow other
    +       refinements to the loop.
     
    -    -  Divide by 4 the initial length, to not modify behavior with the
    -       change mentioned in the paragraph above.
    -
    -    -  Now that the last things in this (not-)forever loop are conditional,
    -       we can actually encode them in the condition of the while, which is
    -       now rewritten as a do while.
    +    -  Now that the last things in this (not-)forever loop are conditional
    +       breaks, we can actually encode them in the condition of the while,
    +       which is now rewritten as a do while.
     
    @@ libmisc/xgetXXbyYY.c
     -  /* we have to start with something */
     -  size_t length = 0x100;
     +  int status;
    -+  size_t length = 0x40;  /* we have to start with something */
    ++  size_t length = 0x100;  /* we have to start with something */
      
        result = MALLOC(LOOKUP_TYPE);
        if (NULL == result) {
    @@ libmisc/xgetXXbyYY.c
     -          int status;
     +  do {
                LOOKUP_TYPE *resbuf = NULL;
    -+
    -+          length = (length <= SIZE_MAX / 4) ? length * 4 : SIZE_MAX;
     +
                buffer = XREALLOCARRAY (buffer, length, char);
                status = REENTRANT_NAME(ARG_NAME, result, buffer,
                                        length, &resbuf);
     @@
    -                   free(result);
                        return ret_result;
                }
    --
    + 
     -          if (ERANGE != status) {
     -                  break;
     -          }
    @@ libmisc/xgetXXbyYY.c
     -                  break;
     -          }
     -
    --          length = (length <= SIZE_MAX / 4) ? length * 4 : SIZE_MAX;
    +           length = (length <= SIZE_MAX / 4) ? length * 4 : SIZE_MAX;
     -  }
     +  } while (length < SIZE_MAX && status == ERANGE);
      
```

alejandro-colomar:
Not convinced by the 4th patch, I dropped it.

hallyn (approved):
Thanks.

PR #741, opened 2023-05-31
newusers: Improve error message

ikerexxe:


hallyn:
Thanks, it did seem a bit sparse.

PR #745, opened 2023-06-21
chgpasswd: fix segfault in command-line options

jeffbencteux:
Using the --sha-rounds option without first giving a crypt method via the --crypt-method option results in comparisons with a NULL pointer and thus make chgpasswd segfault:

$ chgpasswd -s 1
zsh: segmentation fault  chgpasswd -s 1

Current patch add a sanity check before these comparisons to ensure there is a defined encryption method.


hallyn (approved):
Thank you.

hallyn:
I don't know what's going on with the debian CI job.

PR #748, opened 2023-07-09
Fix yescrypt support

bkuhls:
Fixes build error:
newusers.c: In function 'update_passwd':
newusers.c:433:21: error: 'sflg' undeclared (first use in this function); did you mean 'rflg'?

introduced by
https://github.com/shadow-maint/shadow/commit/5cd04d03f94622c12220d4a6352824af081b8531 which forgot to define sflg for these configure options:

--without-sha-crypt --without-bcrypt --with-yescrypt


hallyn:
Thanks.

Issue #674, opened 2023-02-28
Y2038: utmp, utmpx and lastlog problem


ikerexxe:
The previous structures use `int32_t` instead of `time_t` to define its members, which will overflow them in Y2038. Anything that uses `utmp.h`, `utmpx.h`, `lastlog.h`, accesses [/run/utmp](https://manpages.opensuse.org/utmp.5), [/var/log/wtmp](https://manpages.opensuse.org/wtmp.5), [/var/log/btmp](https://manpages.opensuse.org/lastb.1) or uses the corresponding functions from glibc is affected. This project needs to move away from them. One of the solutions would be to logind instead.

This is just an informative note, and a placeholder for future work. It doesn't contain all the relevant information, for that please check [Thorsten Kukuk's note](https://github.com/thkukuk/utmpx/blob/main/Y2038.md).

thkukuk:
I looked at the source a little bit and at least reading from systemd-logind instead of utmp should be simple:
* limits.c/check_logins(): can be replaced with sd_get_sessions() and sd_session_get_user_name(). Requires systemd >= 254 (or current git)
* logoutd.c: utmp can also be replaced with sd_get_sessions() and sd_session_get_tty().
* login.c: I would keep writing utmp for now, but change reading from it.

Replacing the functionality is possible and shouldn't be a bigger efford, but may require bigger code refactors for functions using utmp as argument.

hallyn:
My general static linking desires probably don't apply here; I mainly need it for running in mostly empty containers.

However, this is a ...  venerable project, and it would feel like the height of hubris for me to tell users how they can and can't use it.

@thkukuk  thank you very much for all the thought you've put into this.  I will think about it more over the next few weeks.

thkukuk:
> My general static linking desires probably don't apply here; I mainly need it for running in mostly empty containers.

Isn't busybox used for this normally? At least the majority of my running containers is using busybox, either compiled against glibc or musl libc, and the real service as single binary.

I quickly checked the glibc base containers of the three big distros (Debian, Feodra, openSUSE) and they all contain libsystemd, because it is already used by many low level projects (util-linux, procps, ...).

And we speak here about three tools:
- login - which is not useable in a container and doesn't make sense
- logoutd - which is not useable in a container and doesn't make sense.
[...]

PR #749, opened 2023-07-13
Conditionally build lastlog funcionality and documentation

ikerexxe:
Partially resolves: https://github.com/shadow-maint/shadow/issues/674

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -68,23 +68,6 @@ AC_CHECK_MEMBERS([struct utmp.ut_type,
                   struct utmp.ut_xtime,
                   struct utmp.ut_tv],,,[[#include <utmp.h>]])
 
-if test "$ac_cv_header_lastlog_h" = "yes"; then
-	AC_CACHE_CHECK(for ll_host in struct lastlog,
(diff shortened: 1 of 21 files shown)


configure.ac
```diff
@@ -746,6 +758,7 @@ echo "	yescrypt passwords encryption:	$with_yescrypt"
 echo "	nscd support:			$with_nscd"
 echo "	sssd support:			$with_sssd"
 echo "	subordinate IDs support:	$enable_subids"
+echo "	enable lastlog:		$enable_lastlog"
```
hallyn, line 761: The spacing here doesn't look right (even when I wget the patch and remove the first lines).  I think you need to remove the space after the :, and add two tabs.

hallyn (changes requested):
Thanks - looks good to me except for one cosmetic comment.

configure.ac
ikerexxe: There were 2 tabs, and I think another one was missing. Please check it on your setup and if it works close this conversation.

jubalh: I checked out now and spacing is correct now.

PR #750, opened 2023-07-14
CI: update debian repos

ikerexxe:
Latest debian version changed the location and format for the repos file.


ikerexxe:
As all checks passed I'm merging it.

hallyn:
Thanks.  Not a fan of this debian change, but enough arm-chair critiquing from me :)

Issue #752, opened 2023-07-15
doc: typo in subuid.5: "newusers(1)" should be "newusers(8)"


salewski:
The "SEE ALSO" section in `subuid(5)` references "newusers(1)", but that reference should be to "newusers(8)", instead.

I grepped through the tree to see if there were any other occurrences, but that is the only reference to "newusers(1)" that I found. All other references are to "newusers(8)".

PR #753, opened 2023-07-15
subuid.5: reference newusers(8) rather than newusers(1)

salewski:
Resolves: https://github.com/shadow-maint/shadow/issues/752

Fixes typo in `subuid(5)` that referenced "newusers(1)" rather than "newusers(8)".


ikerexxe:
LGTM. Thank you for the patch!

PR #756, opened 2023-07-19
shadow userdel: add the adaptation to the busybox ps in 01-kill_user_procs.sh

hikari63:
shadow userdel: add the adaptation to the busybox ps in 01-kill_user_procs.sh

### DIFF
--- a/etc/shadow-maint/userdel-pre.d/01-kill_user_procs.sh
+++ b/etc/shadow-maint/userdel-pre.d/01-kill_user_procs.sh
@@ -4,13 +4,16 @@ PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 
 # Check user exists, and if so, send sigkill to processes that the user owns
 
-RUNNING=`ps -eo user | grep -Fx "$SUBJECT" | wc -l`
-
-# if the user does not exist, RUNNING will be 0
-
-if [ "${RUNNING}x" = "0x" ]; then
-  exit 0
+ps -eo user >/dev/null 2>&1
+if [ $? -eq 0 ]; then
+    RUNNING=`ps -eo user | grep -Fx "$SUBJECT" | wc -l`
+    # if the user does not exist, RUNNING will be 0
+    if [ "${RUNNING}x" = "0x" ]; then
+        exit 0
+    fi
 fi
+
+# If there is no ps -eo, traverse the process directly.
 
 ls -1 /proc | while IFS= read -r PROC; do
   echo "$PROC" | grep -E '^[0-9]+$' >/dev/null


src/useradd.c
```diff
@@ -385,6 +385,7 @@ static void get_defaults (void)
 			} else {
 				def_group = grp->gr_gid;
 				def_gname = xstrdup (grp->gr_name);
+				gr_free(grp);
```
hallyn, line 388: This should be a separate PR.

I'm not sure this is safe.  For instance, in the case where grp has the result of fgetgrent(), putgrent() should be used to free it.

Given that this is not a library, and will exit and free all its resources, I think it's safer to leave it be.  We've had patches not too  long ago which tried to free memory right before exiting, which caused crashes in some uncommon situations.

hikari63: Thank you for your reply. I updated the commit message Do you think I've got enough reasons in my commit message?
And have you checked the other commit?

hallyn: No, you are wrong here:

```
and the memory is malloced
through the prefix_getgr_nam_gid function ->
__gr_dup function.
```

It can also be getgrgid(), whose manpage says not to free.

This is not a long running program.  The dangers of an accidental double free or similar problem far outweigh the problems of a tiny bit of memory not being freed.

etc/shadow-maint/userdel-pre.d/01-kill_user_procs.sh
```diff
@@ -4,14 +4,17 @@ PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
...
 # Check user exists, and if so, send sigkill to processes that the user owns
 
-RUNNING=`ps -eo user | grep -Fx "$SUBJECT" | wc -l`
-
-# if the user does not exist, RUNNING will be 0
-
-if [ "${RUNNING}x" = "0x" ]; then
-  exit 0
+ps -eo user >/dev/null 2>&1
+if [ $? -eq 0 ]; then
```
hallyn, line 8: I looked at this a few days ago.  Thanks for pointing out the issue - we should do something, but I don't think that this is the best way to solve it.  I'd prefer there be one way to do it.

@edneville do you have any suggestions?

hikari63: I've tried to add a few pieces of code to adapt to the ps command with different options, but it doesn't cover all user scenarios and the code is redundant. So I think if ps -eo doesn't work, go straight to traversal. The userdel may take a little longer, but is acceptable as a low frequency operation.

edneville:
On 2023-07-25 19:35-0700, hikari63 wrote:
> I've tried to add a few pieces of code to adapt to the ps command with 
> different options, but it doesn't cover all user scenarios and the 
> code is redundant. So I think if ps -eo doesn't work, go straight to 
> traversal. The userdel may take a little longer, but is acceptable as 
> a low frequency operation.

Probably better to base it on this, which covered a lot more cases:

  https://github.com/shadow-maint/shadow/blob/master/etc/shadow-maint/groupdel-pre.d/01-kill_group_procs.sh

Cheers,
Ed

src/useradd.c
hikari63: OK, that makes sense

hikari63:
> On 2023-07-25 19:35-0700, hikari63 wrote: I've tried to add a few pieces of code to adapt to the ps command with different options, but it doesn't cover all user scenarios and the code is redundant. So I think if ps -eo doesn't work, go straight to traversal. The userdel may take a little longer, but is acceptable as a low frequency operation.
> Probably better to base it on this, which covered a lot more cases: https://github.com/shadow-maint/shadow/blob/master/etc/shadow-maint/groupdel-pre.d/01-kill_group_procs.sh Cheers, Ed

Thank you for your reply. The original process is similar. But I mainly want to solve the error problem when there is no ps -eo option.

hallyn:
> Thank you for your reply. The original process is similar. But I mainly want to solve the error problem when there is no ps -eo option.

Yes but I do not want two flows in there.  So I prefer @edneville 's approach of always using the non-'ps' path.

hallyn:
I see - I guess I"m ok with this.

hallyn (approved):

PR #757, opened 2023-07-21
Replace `utmp` interface

ikerexxe:
Conditionally compile code related to the utmp interface, as it is know to face a problem in the year 2038. Replace it by systemd-logind interface when possible.

By the way, intermediate commits may not compile with the two possible options.

Finally, I have been able to test the `login` binary and see that it works under both configuration options, but I have not been able to test it thoroughly. It would be nice if someone from the community could test it on a distribution that uses it, or at least point me to a distribution that does so I can test it.

Resolves: https://github.com/shadow-maint/shadow/issues/674


libmisc/utmp.c
```diff
@@ -358,13 +360,35 @@ static int setutmp (struct utmp *ut)
...
+	utent = get_current_utmp ();
+	if (utent == NULL) {
+		return -1;
+	}
+
+	ut = prepare_utmp  (user, tty, host, utent);
 
 	(void) setutmp  (ut);	/* make entry in the utmp & wtmp files */
 	free (ut);
+
```
hallyn, line 378: should utent also be freed?

libmisc/utmp.c
```diff
@@ -358,13 +360,35 @@ static int setutmp (struct utmp *ut)
...
+                    const char *tty,
+                    const char *hostname)
+{
+	struct utmp *utent, *failent;
+
+	if (getdef_str ("FTMP_FILE") != NULL) {
+		utent = get_current_utmp ();
+		failent = prepare_utmp (failent_user, tty, hostname, utent);
+		failtmp (failent_user, failent);
+		free (failent);
```
hallyn, line 392: Should utent be freed here?

libmisc/logind.c
```diff
@@ -34,3 +34,18 @@ int get_session_host (char **out)
...
+    struct passwd *pw;
+    unsigned long count = 0;
+
+    pw = prefix_getpwnam(name);
+    if (pw == NULL) {
+        return 0;
+    }
+
+    count = sd_uid_get_sessions(pw->pw_uid, 0, NULL);
+
```
hallyn, line 49: Do you need to free pw?

src/login.c
```diff
@@ -573,9 +573,9 @@ int main (int argc, char **argv)
 #ifdef RLOGIN
 	if (rflg) {
 		assert (NULL == username);
-		username = XMALLOC(USER_NAME_MAX_LENGTH + 1, char);
-		username[USER_NAME_MAX_LENGTH] = '\0';
-		if (do_rlogin (hostname, username, USER_NAME_MAX_LENGTH, term, sizeof term)) {
+		username = XMALLOC(sysconf(_SC_LOGIN_NAME_MAX) + 1, char);
```
hallyn, line 576: nitpick, but I'd prefer you use a temp size_t to store sysconf(_SC_LOGIN_NAME_MAX) and reuse in the below three lines.  (Same in the next block)

src/logoutd.c
```diff
@@ -15,6 +15,7 @@
 #include <stdio.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <utmp.h>
```
hallyn, line 18: Huh, weren't we going to get rid of logoutd? :)

hallyn:
Thanks, looks great, just a few comments.

libmisc/logind.c
ikerexxe: I'm not sure. Moreover, it is not clear to me that I have to use that interface and not another one (i.e. getpwnam) . I always get confused with these.

libmisc/utmp.c
ikerexxe: Yes, it should be done everywhere

src/login.c
ikerexxe: Done

src/logoutd.c
ikerexxe: Yes, but the compilation was failing because the `include` was missing and this seemed the easier and fastest approach to fixing it.

libmisc/logind.c
hallyn: Ok, yeah - you do not in fact want to free it :)  Sorry.

libmisc/utmp.c
```diff
@@ -53,6 +55,57 @@
...
+	/*
+	 * Open the file for append.  It must already exist for this
+	 * feature to be used.
+	 */
+
+	if (access (ftmp, F_OK) != 0) {
+		return;
+	}
+
+	fd = open (ftmp, O_WRONLY | O_APPEND);
```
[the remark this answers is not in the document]
hallyn: For posterity - the check which was previously done was just for it to already exist.

hallyn (approved):

Issue #535, opened 2022-06-06
chsh should check whether shell is listed in /etc/shells for root


CamberLoid:
Currently, `chsh` will check whether the shell is listed in `/etc/shells` for non-root users but not for root.

https://github.com/shadow-maint/shadow/blob/1f84142915809713c920c8c01ccaca6835bca179/src/chsh.c#L517-L530

This may lead to a login failure when, for example, user sets shell to `/bin/bash` instead of `/usr/bin/bash`, which the former is not listed in `/etc/shells`. `chsh` should warn user that the possible invalid shell is provided. 

An expected behavior is to warn the root user that `Warning: $SHELL is not present in /etc/shells`.

hallyn:
Yeah I suppose that would be polite.

hallyn:
Did you want to post a PR for this?

jubalh:
Honestly I feel like the current settings and warnings are fine.
However I created https://github.com/shadow-maint/shadow/pull/759 to change according to this PR.
I leave it up to the maintainers to decide which way they prefer and close this issue without merging the PR or with merging it.

PR #759, opened 2023-07-26
chsh: warn if root sets a shell not listed in /etc/shells

jubalh:
Print a warning even for the root user if the provided shell isn't listed in /etc/shells, but continue to execute the action. In case of non root user exit.

See https://github.com/shadow-maint/shadow/issues/535


hallyn (approved):
Thanks!

Issue #763, opened 2023-08-03
4.14.0-rc1: include alloc.h bit.h not found


jubalh:
```
[    2s] + export 'CFLAGS=-O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g -fpie'
[    2s] + CFLAGS='-O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g -fpie'
[    2s] + export LDFLAGS=-pie
[    2s] + LDFLAGS=-pie
[    2s] + autoreconf -fvi
[    2s] autoreconf: export WARNINGS=
[    2s] autoreconf: Entering directory '.'
[    2s] autoreconf: running: autopoint --force
[    3s] Copying file m4/codeset.m4
[    3s] Copying file m4/extern-inline.m4
[    3s] Copying file m4/fcntl-o.m4
[    3s] Copying file m4/glibc2.m4
[    3s] Copying file m4/glibc21.m4
[    3s] Copying file m4/intdiv0.m4
[    3s] Copying file m4/intl.m4
[    3s] Copying file m4/intldir.m4
[    3s] Copying file m4/intmax.m4
[    3s] Copying file m4/inttypes-pri.m4
[    3s] Copying file m4/inttypes_h.m4
[    3s] Copying file m4/lcmessage.m4
[    3s] Copying file m4/lock.m4
[    3s] Copying file m4/longlong.m4
[    3s] Copying file m4/printf-posix.m4
[    3s] Copying file m4/size_max.m4
[    3s] Copying file m4/stdint_h.m4
[    3s] Copying file m4/threadlib.m4
[    3s] Copying file m4/uintmax_t.m4
[    3s] Copying file m4/visibility.m4
[    3s] Copying file m4/wchar_t.m4
[    3s] Copying file m4/wint_t.m4
[    3s] Copying file m4/xsize.m4
[    3s] Copying file po/Makevars.template
[    5s] autoreconf: running: aclocal --force 
[    5s] autoreconf: configure.ac: tracing
[    6s] autoreconf: running: libtoolize --copy --force
[    6s] libtoolize: putting auxiliary files in '.'.
[...]

PR #776, opened 2023-08-04
Add alloc.h and bit.h to libmisc_la_SOURCES

jubalh:
See https://github.com/shadow-maint/shadow/issues/763

### DIFF
--- a/libmisc/Makefile.am
+++ b/libmisc/Makefile.am
@@ -17,9 +17,11 @@ libmisc_la_SOURCES = \
 	age.c \
 	agetpass.c \
 	alloc.c \
+	../lib/alloc.h \
 	audit_help.c \
 	basename.c \
 	bit.c \
+	../lib/bit.h \
 	chkname.c \
 	chkname.h \
 	chowndir.c \
@@ -70,7 +72,9 @@ libmisc_la_SOURCES = \
 	setupenv.c \
 	shell.c \
 	stpecpy.c \
+	../lib/stpecpy.h \
 	stpeprintf.c \
+	../lib/stpeprintf.h \
 	strtoday.c \
 	sub.c \
 	sulog.c \


jubalh:
@alejandro-colomar could you check this?
Also see https://github.com/shadow-maint/shadow/issues/763#issuecomment-1664383425 why they are in different dirs?

alejandro-colomar:
I didn't know the build system of this project very much.  I just built following the instructions in my system and it seemed to work, so I left it that way.  How did you build to notice those issues?  Maybe you can post that in the new <https://github.com/shadow-maint/shadow/blob/master/doc/contributions/build_install.md>, or point me to it if it's already documented and I missed it?

If you patch fixes this, I guess `Acked-by: Alejandro Colomar <alx@kernel.org>`

Please add `Fixes: hash ("commit message")` tags for the commits it fixes.

Thanks for fixing my bugs! :-)

alejandro-colomar:
As to why they are in different dirs, I don't know how these libs should be organized ideally.  Feel free to move them if that's better.  I agree that having the header in lib and the source in libmisc is weird.  But I think I checked before doing that and found some case where it already happened (or something similar) and imitated it; can't tell for sure.

ikerexxe:
One minor comment, I'd change the commit message from `See https://github.com/shadow-maint/shadow/issues/763`to `Resolves: https://github.com/shadow-maint/shadow/issues/763`

Apart from that I'd like @hallyn to check that this actually includes the missing files in the compressed file.

hallyn:
@alejandro-colomar  yes we should add a check to CI which does 'make dist' followed by an attempt to build from the resulting tarball, to catch things like this.

jubalh:
>  I just built following the instructions in my system and it seemed to work, so I left it that way. How did you build to notice those issues? 

I was testing it on our [build system](https://build.opensuse.org/package/show/home:jubalh:branches:Base:System/shadow).

> One minor comment, I'd change the commit message from See https://github.com/shadow-maint/shadow/issues/763to Resolves: https://github.com/shadow-maint/shadow/issues/763

I wasn't sure if it's the best fix and why the files were missing and thus opted for a related. But I'll change that :)

What's confusing me though is that  CI for PR 777 builds fine even without this change here. How comes the it doesn't need those header files in libmisc_la_SOURCES and why seem the files not to be missing there but apparently they are in the tarball mentioned above? Confusing :)

hallyn:
https://github.com/shadow-maint/shadow/pull/777  Catches this error.  Do you want to cherrypick that commit from my pr branch into yours and verify that it passes then? :)

PR #778, opened 2023-08-04
libmisc/write_full.c: Improve write_full()

alejandro-colomar:
Documentation:

-  Correct the comment documenting the function:

   write_full() doesn't write "up to" count bytes (which is write(2)'s behavior, and exactly what this function is designed to avoid), but rather exactly count bytes (on success).

-  While fixing the documentation, take the time to add a man-page-like comment as in other APIs.  Especially, since we'll have to document a few other changes from this patch, such as the modified return values.

-  Partial writes are still possible on error.  It's the caller's responsibility to handle that possibility.

API:

-  In write(2), it's useful to know how many bytes were transferred, since it can have short writes.  In this API, since it either writes it all or fails, that value is useless, and callers only want to know if it succeeded or not.  Thus, just return 0 or -1.

Implementation:

-  Use `== -1` instead of `< 0` to check for write(2) syscall errors. This is wisdom from Michael Kerrisk.  This convention is useful because it more explicitly tells maintainers that the only value which can lead to that path is -1.  Otherwise, a maintainer of the code might be confused to think that other negative values are possible.  Keep it simple.

-  The path under `if (res == 0)` was unreachable, since the loop condition `while (count > 0)` precludes that possibility.  Remove the dead code.

-  Use a temporary variable of type `const char *` to avoid a cast.

-  Rename `res`, which just holds the result from write(2), to `w`, which more clearly shows that it's just a very-short-lived variable (by it's one-letter name), and also relates itself more to write(2). I find it more readable.

-  Move the definition of `w` to the top of the function.  Now that the function is significantly shorter, the lifetime of the variable is clearer, and I find it more readable this way.

Use:

-  Also use `== -1` to check errors.

Cc: Christian Göttsche <cgzones@googlemail.com>

### DIFF
--- a/lib/write_full.c
+++ b/lib/write_full.c
@@ -1,7 +1,8 @@
 /*
- * SPDX-FileCopyrightText:  2023, Christian Göttsche <cgzones@googlemail.com>
+ * SPDX-FileCopyrightText: 2023, Christian Göttsche <cgzones@googlemail.com>
+ * SPDX-FileCopyrightText: 2023, Alejandro Colomar <alx@kernel.org>
  *
- * SPDX-License-Identifier:  BSD-3-Clause
+ * SPDX-License-Identifier: BSD-3-Clause
  */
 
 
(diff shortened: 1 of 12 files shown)


alejandro-colomar:
BTW, just to make it explicit:

This API still **can do short writes**.  It's just that they happen **together with a failure** in the middle of the loop, and so are not especially reported.  This means that it's responsibility of the caller to make sure that a short write is not problematic, and if it is, it should probably try to delete the file on error.

alejandro-colomar:
v2 changes:

-  CAVEATS: Document that partial writes are possible when this call fails.

```
$ git range-diff shadow/master gh/write_full write_full 
1:  abb94949 ! 1:  12589919 libmisc/write_full.c: Improve write_full()
    @@ Commit message
            a few other changes from this patch, such as the modified return
            values.
     
    +    -  Partial writes are still possible on error.  It's the caller's
    +       responsibility to handle that possibility.
    +
         API:
     
         -  In write(2), it's useful to know how many bytes were transferred,
    @@ lib/write_full.c
     + *
     + * ERRORS
     + *        See write(2).
    ++ *
    ++ * CAVEATS
    ++ *        This function can still perform partial writes: if the function
    ++ *        fails in the loop after one or more write(2) calls have
    ++ *        succeeded, it will report a failure, but some data may have been
    ++ *        written.  In such a case, it's the caller's responsibility to
    ++ *        make sure that the partial write is not problematic, and
    ++ *        remediate it if it is --maybe by trying to remove the file--.
       */
     -ssize_t write_full(int fd, const void *buf, size_t count) {
     -  ssize_t written = 0;
```

lib/write_full.c
```diff
@@ -1,52 +1,77 @@
...
-		res = write(fd, buf, count);
-		if (res < 0) {
-			if (errno == EINTR) {
-				continue;
-			}
+int
+write_full(int fd, const void *buf, size_t count)
+{
+	ssize_t     w;
+	const char  *p;
```
cgzones, line 59: My preference would be to use `unsigned char *, since we operate on bytes, not strings.

cgzones (approved):

lib/write_full.c
alejandro-colomar: Agree.  My bad.

alejandro-colomar:
v3 changes:

-  Use u_char, as we're copying bytes, not strings.  [reported by @cgzones ]

```
$ git range-diff shadow/master gh/write_full write_full 
1:  12589919 ! 1:  bd78273b libmisc/write_full.c: Improve write_full()
    @@ lib/write_full.c
     +int
     +write_full(int fd, const void *buf, size_t count)
     +{
    -+  ssize_t     w;
    -+  const char  *p;
    ++  ssize_t              w;
    ++  const unsigned char  *p;
     +
     +  p = buf;
      
```

hallyn (approved):
Thanks, very nice.

hallyn (approved):
Thanks!

PR #789, opened 2023-08-18
Replace __{BEGIN,END}_DECLS with #ifdef __cplusplus

heirecka:
Fixes the build with musl libc.


hallyn:
Thank you.  Per https://www.gnu.org/software/libtool/manual/html_node/C-header-files.html this looks correct.

hallyn (approved):

PR #798, opened 2023-09-01
groupadd: Improve error message when opening group file fails.

velichkov:
Both `gr_open` and `sgr_open` are using `commonio_open` function and when there is a failure this function sets `errno` accordingly.


ikerexxe:
It seems CI doesn't want to work today. In any case, the changes look good to me so I'm merging the PR.

PR #806, opened 2023-09-14
man/chsh.1: specify what is a restricted shell

MaxHearnden:
Update the documentation to reflect the the functionally within src/chsh.c
The functionality was in 45c6603cc86c5881b00ac40e0f9fe548c30ff6be so I am unsure why the documentation wasn't updated to reflect this change

### DIFF
--- a/man/chsh.1.xml
+++ b/man/chsh.1.xml
@@ -120,10 +120,11 @@
       name must be listed in <filename>/etc/shells</filename>, unless the
       invoker is the superuser, and then any value may be added. An
       account with a restricted login shell may not change her login shell.
+      A restricted shell is a shell which is not in
+      <filename>/etc/shells</filename>.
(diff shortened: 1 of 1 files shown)


hallyn:
Hm, well, is that really true?  I believe the intent was simply that a 'restricted shell' is a shell which only allows very limited commands.  Hence '/bin/rsh' used as an example.  And then, from there it follows that we do not want a user with a restricted shell to be able to change their shells.  So the guidance that you should not put a restricted shell in /etc/shells, and that we then do not allow you to change your shell if your current shell is not already listed in /etc/shells (and you are unprivileged), follows from what we want in a restricted shell.

But that doesn't mean that a restricted shell is defined as a shell not in /etc/shells.

hallyn:
Please re-open if you feel it appropriate.

Issue #790, opened 2023-08-21
Use CMocka as the unit testing framework


ikerexxe:
I have done a [little research](https://ikerexxe.github.io/idm/2023/08/21/shadow-unit-test.html) on which unit test framework we should use in this project. The result is that the best option is CMocka, so I would like to proceed to include it in the project but I would like to get the approval of the community first.

hallyn:
Sounds good to me, thanks for doing the research!

PR #807, opened 2023-09-15
Add cmocka as unit testing framework

ikerexxe:
Add two simple test cases for the logind functionality, update CI to run the unit tests and briefly explain how to develop tests with this framework.

Resolves: https://github.com/shadow-maint/shadow/issues/790

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -719,6 +719,10 @@ if test "$with_skey" = "yes"; then
 	]])],[AC_DEFINE(SKEY_BSD_STYLE, 1, [Define to support newer BSD S/Key API])],[])
 fi
 
+PKG_CHECK_MODULES([CMOCKA], [cmocka], [have_cmocka="yes"],
+	[AC_MSG_WARN([libcmocka not found, cmocka tests will not be built])])
+AM_CONDITIONAL([HAVE_CMOCKA], [test x$have_cmocka = xyes])
+
 AC_CHECK_FUNC(fgetpwent_r, [AC_DEFINE(HAVE_FGETPWENT_R, 1, [Defined to 1 if you have the declaration of 'fgetpwent_r'])])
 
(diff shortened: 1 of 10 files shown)


jubalh (approved):

maherme (approved):

configure.ac
```diff
@@ -719,6 +719,11 @@ if test "$with_skey" = "yes"; then
 	]])],[AC_DEFINE(SKEY_BSD_STYLE, 1, [Define to support newer BSD S/Key API])],[])
 fi
 
+AC_SUBST(CMOCKA_LIBS)
+AC_CHECK_LIB(cmocka, _will_return, [CMOCKA_LIBS=-lcmocka]; [have_cmocka="yes"],
+	[AC_MSG_WARN([libcmocka not found, cmocka tests will not be built])])
```
lslebodn, line 724: This will just check whether library is available for linking
* https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Libraries.html
And that would pass just with the **libcmocka0** installed on system
* https://packages.debian.org/sid/amd64/libcmocka0/filelist

One should check also availability of header files.

Or you can rely on pkgconfig which is already provided by cmocka and it makes configure time check much simpler.
* https://packages.debian.org/sid/amd64/libcmocka-dev/filelist

and it is already used in the same file
* https://github.com/shadow-maint/shadow/blob/5178f8c5afb612f6ddf5363823547e080e7f546b/configure.ac#L424

ikerexxe: I preferred the `PKG_CHECK_MODULES` way, so I've used it. Take a look at it to check that I got it right.

lslebodn: LGTM

ikerexxe:
If no one complains I will proceed to merge this one later this week.

PR #812, opened 2023-09-26
useradd: Set proper SELinux labels for def_usrtemplate

jsegitz:
currently content in this directory will have labels derived from /usr/etc, leading to strange errors for user operations later on

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2788,7 +2788,7 @@ int main (int argc, char **argv)
 		if (home_added) {
 			copy_tree (def_template, prefix_user_home, false, true,
 			           (uid_t)-1, user_id, (gid_t)-1, user_gid);
-			copy_tree (def_usrtemplate, prefix_user_home, false, false,
+			copy_tree (def_usrtemplate, prefix_user_home, false, true,
 			           (uid_t)-1, user_id, (gid_t)-1, user_gid);
 		} else {
 			fprintf (stderr,


jubalh:
Since we don't take the SELinux attributes for def_template I guess it makes sense for def_usrtemplate as well.
BTW this is the only place in the code where `reset_selinux` is set to `false` for `copy_tree()`.

AdamMajer:
https://github.com/shadow-maint/shadow/commit/06eb4e4d76ac7f1ac86e68a89b2dc9be7c7323a2

Seems same change for similar reasons.

jubalh:
So I believe the mistake was actually introduced by me in https://github.com/shadow-maint/shadow/commit/74c17c716. Where I took that patch without updating it to the change we got in https://github.com/shadow-maint/shadow/commit/06eb4e4d76ac7f1ac86e68a89b2dc9be7c7323a2.

jubalh (approved):

jubalh:
One could argue that copy_tree doesn't need the `reset_selinux` variable. But I guess its good to leave it in case later there will be a use case.

ikerexxe:
@jsegitz do you mind modifying your commit message? The first line of the commit message is too long. `useradd: Set proper SELinux labels for def_usrtemplate` should suffice.

While you're at it also reference the commit that it fixes. Example:
```
Fixes: 74c17c716 ("Add support for skeleton files from /usr/etc/skel")
```

PR #794, opened 2023-08-28
Gentoo2: Wipe out libmisc

alejandro-colomar:
Merge after the other fix for Gentoo.

Link: <https://github.com/shadow-maint/shadow/pull/792>
Reported-by: @SoapGenroo
Cc: @thesamesam 
Cc: @cbricart
Cc: @jubalh 
Signed-off-by: Alejandro Colomar <alx@kernel.org>

PR #792, opened 2023-08-25
lib, libmisc: Move source files to lib (where their headers were)

alejandro-colomar:
```
Scripted change:

$ find lib/ -type f \
| grep '\.h$' \
| sed 's,lib/,libmisc/,' \
| sed 's,\.h$,.c,' \
| xargs find 2>/dev/null \
| xargs mv -t lib/;

Plus updating the Makefiles.
```

Closes: <https://github.com/shadow-maint/shadow/issues/791>
Closes: <https://bugs.gentoo.org/912446>
Reported-by: Christian Bricart @cbricart
Reported-by: Robert Marmorstein
Cc: Sam James @thesamesam 
Cc: Michael Vetter @jubalh

Issue #684, opened 2023-03-21
bye github


alejandro-colomar:
Hi all,

While github has always had a nice interface (especially for git blame; but projects like cgit are catching up on this), I'm very concerned about the decisions it has taken lately, especially since one year ago around the npm issues (if somebody doesn't know them, I can try to find links to the discussions).

I personally prefer sending patches through a mailing list, but acknowledge that it's nice to have automated CI for patch incoming patch sets.  Hopefully, soon we'll be able to run all tests locally in a way that I can know that my patch sets are correct before sending.

Are there alternatives as good as github for this?  I don't know much about gitlab.  While I used it in the past in a job, I don't know if they host free runners for open-source projects, as github does.

This also reminds me that even if shadow is not migrated to gitlab or somewhere else, we should strongly consider not being locked in to github's CI, so the tests and build/CI system should be as generic as possible in that regard, in case we change our minds later.

I'd like to be able to continue contributing to the project, and at the same time I'd like to leave github :).

Cheers,
Alex

hallyn:
I have in the past lobbied for moving to sr.ht.  It has a good build infrastructure, and patches through mailing lists.

The main concern has been that something like https://git.sr.ht/~hallyn/shadow  is too tied to one person.

alejandro-colomar:
> I have in the past lobbied for moving to sr.ht. It has a good build infrastructure, and patches through mailing lists.

+1

> The main concern has been that something like https://git.sr.ht/~hallyn/shadow is too tied to one person.

The Linux man-pages project lived for more than a year at <http://www.alejandro-colomar.es/src/alx/linux/man-pages/man-pages.git> until I got access to <https://git.kernel.org>.  I think it was good as a remainder that free software is decentralized, and lives in the forks of the maintainers and contributors.

We could have several "official" repositories (so listing them all as official repositories in the README), one for each of you maintainers, and have one of them be the "current" main tree (to which the others sync their default branch), but the others could take over at any moment if the main one dies.  Kind of a Docker Swarm of repos :).

rbalint:
> Hi all,
> 
> While github has always had a nice interface (especially for git blame; but projects like cgit are catching up on this), I'm very concerned about the decisions it has taken lately, especially since one year ago around the npm issues (if somebody doesn't know them, I can try to find links to the discussions).

Please details the concerns. I have not noticed anything disturbing from this project's POV lately.

[...]

PR #813, opened 2023-09-27
README.md and STABLE.md: record the stable branch url(s).

hallyn:


STABLE.md
```diff
@@ -0,0 +1,5 @@
+# Supported stable branches
+
+The following stable branches are kindly maintained by trusted volunteers:
+
+* [4.14](https://www.alejandro-colomar.es/src/alx/shadow/stable/shadow.git/log/?h=4.14.x)
```
rbalint, line 5: If there is an official stable branch, please keep it in the main repository which is currently on GitHub.
I assume the official 1.14.1, etc. releases will still be released on GitHub.
What would be the point of having an external repository for a stable branch?

rbalint (changes requested):

STABLE.md
rbalint: > My reticence to use GitHub. :)

IMO it hurts the project.

rbalint: What is cheese?

rbalint: > Please detail your opinion. I don't agree, since I'm only adding cheese to the project. I provide support to branches that previously had no support.

I'm sorry, but I'm not going to be engaged in another endless thread, like https://github.com/shadow-maint/shadow/issues/684 .

I believe official stable branches should be maintained with even more care and with at least the same level of testing (infrastructure) as the master branch and having a single maintainer and the very basic external git hosting infrastructure provide none of those.

README.md
```diff
@@ -17,6 +17,13 @@ are used for managing group accounts.
 * [Issue tracker](https://github.com/shadow-maint/shadow/issues)
 * [Releases](https://github.com/shadow-maint/shadow/releases)
 
+## Code
+
+The main development branch is at 
+ [https://github.com/shadow-maint/shadow.git](https://github.com/shadow-maint/shadow)
+
+See STABLE.md for a list of supported stable branches.
```
ikerexxe, line 25: Add reference to STABLE file. Example: [STABLE.md](STABLE.md)

jubalh:
If I'm not mistaken this idea originated with:
https://github.com/shadow-maint/shadow/pull/792, https://github.com/shadow-maint/shadow/pull/794 and https://bugs.gentoo.org/912446.

The main point was that Gentoo devs preferred a point release to not have a patch that changes a lot in the build files.

I'm also under the impression that adding another git remote which will hold branches for stable releases and distributes tarballs will just confuse people and adds necessary complexity to the project.

It will also be weird for tools, like repology, when on GitHub the latest version is shown as 4.14.0 and in some distros will be 4.14.1 taken from somewhere else.

I think what should happen is:
* we should create a point release or
* distros should backport patches that they need from out master branch

And the latter actually happened with https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=75209a50623aa5538ecbf940c20e69d55b0693f1 for Gentoo.


So it looks like all problems are actually solved and I don't see a reason to introduce such a setup.

If we take the first route of creating point releases then I agree with @rbalint that volunteers to take care of that should also be willing to work on the upstream repo. Wherever that is hosted...In this case GitHub.

README.md
hallyn: I'd thought about it, but wasn't sure what to link to.  I guess just

```
https://github.com/shadow-maint/shadow/blob/master/STABLE.md
```
?

rbalint:
I don't feel the need for explaining that a project hosted on GitHub has the development branch on GitHub and it may have stable branches on GitHub as well. 
It was implied without any explanation and stating it in the project documentation would be redundant like this sentence. :-)

hallyn:
Created the stable branch and updated STABLE.md to point to it.

Tried to add alejandro-colomar to the list of ppl who can push to the stable branch, but github doesn't seem to find him.

jubalh:
> but github doesn't seem to find him

Maybe he needs to be added as member of the organization first?

hallyn:
I can't add him under 'collaborators and teams'.

When I look at https://github.com/alejandro-colomar , it says his "activity is private".  Perhaps there's some setting in his account - though then I'm a bit surprised he can still do PRs.

jubalh:
As a test I tried this in another organization.
Under https://github.com/orgs/lazarus-pkgs/people I can add him via "Invite members". Isn't is possible in https://github.com/orgs/shadow-maint/people ?

alejandro-colomar:
Maybe I can open up my activity for you to add me and then close it again?  These interfaces tend to be buggy...

hallyn:
> As a test I tried this in another organization. Under https://github.com/orgs/lazarus-pkgs/people I can add him via "Invite members". Isn't is possible in https://github.com/orgs/shadow-maint/people ?

Ah, yes, thanks.  This seems new to me - I didn't have to do this for other members (that I recall).  I do remember have gotten invitations for other projects, so I'm not sure what  makes it so that you do or don't need it...

jubalh:
>  I didn't have to do this for other members 

The first time I was also confused. There is a difference between contributors, members and people who have access for other certain things. Also one setting is per organization and another one is per repository. Maybe worth looking through all the tabs and settings in https://github.com/shadow-maint/shadow/settings and the above mentioned one.

hallyn:
> I guess I'll also need permissions to push tags. Can you please figure out the way to enable that?
> 
> ```shell
> $ git push shadow 4.14.1
> Enter passphrase for key '/home/alx/.ssh/id_rsa':
> ERROR: Permission to shadow-maint/shadow.git denied to alejandro-colomar.
> fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights
> and the repository exists.
> ```

You'll have to accept the github invite before I can do that.

hallyn:
Oh, maybe you have?  Now how many steps will it take before I can add you to the push list.

alejandro-colomar:
> I believe you should now have it.
> 
> https://i.kym-cdn.com/photos/images/original/000/234/765/b7e.jpg

Yup :D

```
$ git push shadow 4.14.1
Enter passphrase for key '/home/alx/.ssh/id_rsa': 
Enumerating objects: 1, done.
Counting objects: 100% (1/1), done.
Writing objects: 100% (1/1), 962 bytes | 962.00 KiB/s, done.
Total 1 (delta 0), reused 0 (delta 0), pack-reused 0
To ssh://github.com/shadow-maint/shadow.git
 * [new tag]           4.14.1 -> 4.14.1
```

jubalh:
Are we already mentioning somewhere (README.md / security.md?) the signing keys for our tarballs? I couldn't find it right now. Maybe it would make sense to add this.

alejandro-colomar:
We merged <https://github.com/shadow-maint/shadow/pull/847> instead,

PR #820, opened 2023-10-13
Singular badname

xnox:
Fix badname option to be singular just like useradd.

Fixes: [45d6746219](https://github.com/shadow-maint/shadow/commit/45d674621918664c8736f94f862e86bddf4c3fd4) ("src: correct "badname" option")


hallyn:
I suppose this was meant to be done in 45d6746219186 .

The option has been there since 2019.  Should we just allow both --badname and --badnames at this point?  Probably most people are like me and always use the shortname anyway, but it would be rude to break someone's script at this point.

xnox:
@hallyn updated to accept both.

hallyn (approved):
Thanks

Issue #822, opened 2023-10-20
passwd option `--prefix` is not documented in man page


ariel-miculas:
```
$ passwd --help | grep '\--prefix'
  -P, --prefix PREFIX_DIR       directory prefix
$ man passwd | grep '\--prefix'
```

PR #823, opened 2023-10-20
man: document --prefix option in chage, chpasswd and passwd

jubalh:
Support for `--prefix` was added in
https://github.com/shadow-maint/shadow/pull/714 and is available since shadow 4.14.0.

Close https://github.com/shadow-maint/shadow/issues/822

### DIFF
--- a/man/chpasswd.8.xml
+++ b/man/chpasswd.8.xml
@@ -173,6 +173,21 @@
 	  </para>
 	</listitem>
       </varlistentry>
+      <varlistentry>
+	<term>
+	  <option>-P</option>, <option>--prefix</option>&nbsp;<replaceable>PREFIX_DIR</replaceable>
+	</term>
+	<listitem>
+	  <para>
+	    Apply changes to configuration files under the root filesystem
+	    found under the directory <replaceable>PREFIX_DIR</replaceable>.
+	    This option does not chroot and is intended for preparing a cross-compilation
+	    target.  Some limitations: NIS and LDAP users/groups are
+	    not verified.  PAM authentication is using the host files.
+	    No SELINUX support.
+	  </para>
+	</listitem>
+      </varlistentry>
       <varlistentry condition="sha_crypt">
 	<term>
 	  <option>-s</option>, <option>--sha-rounds</option>&nbsp;<replaceable>ROUNDS</replaceable>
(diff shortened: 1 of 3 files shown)


man/chpasswd.8.xml
```diff
@@ -173,6 +173,21 @@
...
+	  <option>-P</option>, <option>--prefix</option>&nbsp;<replaceable>PREFIX_DIR</replaceable>
+	</term>
+	<listitem>
+	  <para>
+	    Apply changes to configuration files under the root filesystem
+	    found under the directory <replaceable>PREFIX_DIR</replaceable>.
+	    This option does not chroot and is intended for preparing a cross-compilation
+	    target.  Some limitations: NIS and LDAP users/groups are
+	    not verified.  PAM authentication is using the host files.
+	    No SELINUX support.
```
alejandro-colomar, line 187: You may want to use semantic newlines.

```
$ MANWIDTH=72 man man-pages | sed -n '/Use semantic newlines/,/^$/p'
<standard input>:846: warning: table wider than line length minus indentation
   Use semantic newlines
       In the source of a manual page, new sentences should be started
       on  new  lines,  long  sentences  should be split into lines at
       clause breaks (commas, semicolons, colons, and so on), and long
       clauses should be split at phrase boundaries.  This convention,
       sometimes known as "semantic newlines", makes it easier to  see
       the  effect of patches, which often operate at the level of in‐
       dividual sentences, clauses, or phrases.
```

alejandro-colomar: This would mean breaking at least after '.', and then maybe also after ':' and at meaningful places.

jubalh: Ok :) Will wait for another maintainer to reply and if I should do this change then I will do it for useradd, userdel, usermod, groupadd, groupdel, groupmod as well.

hallyn: Thanks - that sounds like it would be nice to have.  Please feel free to open an issue or just a pull request for that.  But of course that's not a part of this change.

hallyn: That is - please feel free to make the change to this delta here, but let's not add other manpages, or other chunks of these manpages, to this commit.

Please let me know if you want to squash such a change in this delta.  Otherwise I'll merge as is.

hallyn (approved):

man/chpasswd.8.xml
jubalh: @hallyn you can merge as is and I'll create a new PR with the spacing the next days :)

PR #825, opened 2023-10-26
Add keys/ directory with public keys for maintainers

hallyn:
These can be used to verify releases.


dvzrv:
Thanks, this should work! :tada: 

Hm, but are you not adding your own key and the key files?

hallyn:
Heh i guess i could and for convenience should, it just felt weird since I'm signing the commit with that key.

alejandro-colomar:
Be careful with the merge, since rebasing will remove the signature (and this repo is configured to rebase).

hallyn:
@alejandro-colomar yeah i'm aware of that stuff, as in some places we have mandatory commit signing, but then we are forced to do squash-and-merge, and as if that wasn't bad enough, the new commits end up just signed by github's key instead of the author's.  That people place that much trust in a company, an organization, and a huge deployment of cloud instances, hastily written code, scripts, and duct tape, when every day they're complaining about another company's data leak, is incredible to me, but here we are.

You should be able to push straight to the 4.14 branch.  If you're not, please send me an email.

hallyn:
Does what I have here now look ok?  If so, I'll push it.

ikerexxe (approved):
From my point of view it is fine

jubalh (approved):

alejandro-colomar:
LGTM.  And yes, the 4.14 branch isn't "protected".  :)

hallyn:
Thanks, everyone.

PR #834, opened 2023-11-02
Define SUBUID_FILE/SUBGID_FILE

joakim-tjernlund:
These where hard coded, make them definable like SHADOW_FILE

### DIFF
--- a/lib/prefix_flag.c
+++ b/lib/prefix_flag.c
@@ -125,10 +125,10 @@ extern const char* process_prefix_flag (const char* short_opt, int argc, char **
 		spw_setdbname(spw_db_file);
 
 #ifdef ENABLE_SUBIDS
-		xasprintf(&suid_db_file, "%s/%s", prefix, "/etc/subuid");
+		xasprintf(&suid_db_file, "%s/%s", prefix, SUBUID_FILE);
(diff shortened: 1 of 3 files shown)


joakim-tjernlund:
Ideally there should be a configure option to define a common new prefix such as /etc/pwdb

lib/prefix_flag.c
```diff
@@ -125,10 +125,10 @@ extern const char* process_prefix_flag (const char* short_opt, int argc, char **
 		spw_setdbname(spw_db_file);
 
 #ifdef ENABLE_SUBIDS
-		xasprintf(&suid_db_file, "%s/%s", prefix, "/etc/subuid");
+		xasprintf(&suid_db_file, "%s/%s", prefix, SUBUID_FILE");
```
ikerexxe, line 128: You forgot to remove `"` after `SUBUID_FILE`

joakim-tjernlund: ehh, so it seems. Wonder why I didn't see any mail with your comment

joakim-tjernlund: fixed now

ikerexxe (commented):
LGTM! Merging

ikerexxe (approved):

PR #837, opened 2023-11-11
chsh: limit acceptable shells to absolute paths

stoeckmann:
If an entry in /etc/shells is not an absolute path (comments or partial reads due to fgets), the line should not be considered as a valid login shell.

In general all systems should have getusershells, but let's better be safe than sorry.

### DIFF
--- a/src/chsh.c
+++ b/src/chsh.c
@@ -204,21 +204,17 @@ static bool shell_is_listed (const char *sh)
 	}
 	endusershell ();
 #else
-	char buf[BUFSIZ];
+	char *buf = NULL;
 	FILE *fp;
+	size_t n = 0;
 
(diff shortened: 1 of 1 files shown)


hallyn (approved):

hallyn:
Thanks, lgtm, let's see if the fedora failure was transient.

hallyn:
The error appears unrelated to this pr.

> 10.85 configure: error: readpassphrase() is missing, either from libc or libbsd

pinging @ikerexxe .  Please feel free to merge this and fix up elsewhere.

ikerexxe:
It was already fixed at https://github.com/shadow-maint/shadow/pull/840, the author would need to rebase on top of the master branch to make CI happy. Let's merge as we already know the reason for the failure.

PR #838, opened 2023-11-12
Replace strlcpy(3) with strtcpy(3)

alejandro-colomar:
There's been a very long and interesting discussion in linux-man@ and
    libc-alpha@, where we've discussed all the string-copying functions,
    their pros and cons, when should each be used and avoided, etc.
    
    Paul Eggert pointed out an important problem of strlcpy(3): it is
    vulnerable to DoS attacks if an attacker controls the length of the
    source string.  And even if it doesn't control it, the function is dead
    slow (because its API forces it to calculate strlen(src)).
    
    We've agreed that the general solution for a truncating string-copying
    function is to write a wrapper over strnlen(3)+memcpy(3), which is
    limited to strnlen(src, sizeof(dst)).  This is not vulnerable to DoS,
    and is very fast for all buffer sizes.  string_copying(7) has been
    updated to reflect this, and provides a reference implementation for this
    wrapper function.
    
    This strtcpy(3) (t for truncation) wrapper happens to have the same API
    that our strlcpy_() function had, so replace it with the faster/safer
    implementation.  We don't need to update callers nor tests, since the
    API is the same.
    
    A future commit will rename STRLCPY() to STRTCPY(), and replace
    remaining calls to strlcpy(3) by calls to this strtcpy(3).
    
Link: <https://lore.kernel.org/linux-man/ZU4SDh-Se5gjPny5@debian/T/#mfb5a3fdeb35487dec6f8d9e3d8548bd0d92c4975/>


alejandro-colomar:
The build failure seems unrelated to this PR.  I can reproduce it on master locally.

ikerexxe:
> We've agreed that the general solution for a truncating string-copying
> function is to write a wrapper over strnlen(3)+memcpy(3)

Does every project need to implement this wrapper, or will it be provided in some other project such as the kernel or libc?

alejandro-colomar:
On Tue, Nov 14, 2023 at 03:39:51AM -0800, Iker Pedrosa wrote:
> > We've agreed that the general solution for a truncating string-copying
> > function is to write a wrapper over strnlen(3)+memcpy(3)
> 
> Does every project need to implement this wrapper, or will it be
> provided in some other project such as the kernel or libc?

Every project will need to write its own wrapper.

Glibc doesn't give much support to truncating strings; they consider
it's usually a bug to truncate (and they're right), so they want to
disuade programmers from truncating in the first place.  That's why
they've refused so many years to provide a truncating function, and only
added strlcpy(3) when POSIX added it.

In our case, I think many cases of truncation are wrong, and I have
plans to check every place where we truncate, to see if we need to
truncate or instead report an error.  For now, I'm in the process of
using wrappers that clearly tell what the code is doing, without really
changing the behavior (unless I see it's blatantly wrong).

-- 

ikerexxe (approved):
Looks good and very neat. If I don't hear nobody objecting I'll merge it next week.

PR #845, opened 2023-11-15
src/logoutd.c: Fix theoretical buffer overrun

alejandro-colomar:
ut_line doesn't hold a string.  It is a null-padded fixed-width array. Luckily, I don't think there has ever existed a ut_line ("/dev/tty*") that was 32 bytes long.  That would have resulted in a buffer overrun. Anyway, do the right thing, which is copying into a temporary string.


ikerexxe (approved):
LGTM. Thanks for the improvement!

PR #849, opened 2023-11-23
lib/: Use NITEMS() instead of SIZEOF_ARRAY() where number of elements is meant

alejandro-colomar:
For arrays of char, both NITEMS() and SIZEOF_ARRAY() return the same value.  However, NITEMS() is more appropriate.  Think of wide-character equivalents of the same code; with NITEMS(), they would continue to be valid, while with SIZEOF_ARRAY(), they would be wrong.

### DIFF
--- a/lib/strtcpy.h
+++ b/lib/strtcpy.h
@@ -31,7 +31,7 @@
  *	at the buffer pointed to by dst.  If the destination buffer,
  *	isn't large enough to hold the copy, the resulting string is
  *	truncated.  The size of the buffer is calculated internally via
- *	SIZEOF_ARRAY().
+ *	NITEMS().
  *
  * RETURN VALUE
  *	-1	If this call truncated the resulting string.
@@ -44,7 +44,7 @@
  */
 
 
-#define STRTCPY(dst, src)  strtcpy(dst, src, SIZEOF_ARRAY(dst))
+#define STRTCPY(dst, src)  strtcpy(dst, src, NITEMS(dst))
 
 
 inline ssize_t strtcpy(char *restrict dst, const char *restrict src,
--- a/lib/utmp.c
+++ b/lib/utmp.c
@@ -246,7 +246,7 @@ static
 #ifdef HAVE_STRUCT_UTMP_UT_HOST
 	} else if (   (NULL != ut)
 	           && ('\0' != ut->ut_host[0])) {
-		hostname = XMALLOC(SIZEOF_ARRAY(ut->ut_host) + 1, char);
+		hostname = XMALLOC(NITEMS(ut->ut_host) + 1, char);
 		ZUSTR2STP(hostname, ut->ut_host);
 #endif				/* HAVE_STRUCT_UTMP_UT_HOST */
 	}
--- a/lib/zustr2stp.h
+++ b/lib/zustr2stp.h
@@ -22,7 +22,7 @@
 ({                                                                            \
 	static_assert(!is_array(dst) || sizeof(dst) > SIZEOF_ARRAY(src), ""); \
                                                                               \
-	zustr2stp(dst, src, SIZEOF_ARRAY(src));                               \
+	zustr2stp(dst, src, NITEMS(src));                                     \
 })
 
 
@@ -62,9 +62,9 @@ inline char *zustr2stp(char *restrict dst, const char *restrict src, size_t sz);
  *
  * EXAMPLES
  *	char  src[13] = "Hello, world!"  // No '\0' in this buffer!
- *	char  dst[SIZEOF_ARRAY(src) + 1];
+ *	char  dst[NITEMS(src) + 1];
  *
- *	zustr2stp(dst, src, SIZEOF_ARRAY(src));
+ *	zustr2stp(dst, src, NITEMS(src));
  *	puts(dst);
  */
 


alejandro-colomar:
v2 changes:

-  Revert a change in the implementation of ZUSTR2STP().

```
$ git range-diff shadow/master gh/nitems nitems 
1:  ca51a134 ! 1:  38d1b779 lib/: Use NITEMS() instead of SIZEOF_ARRAY() where number of elements is meant
    @@ Commit message
         equivalents of the same code; with NITEMS(), they would continue to be
         valid, while with SIZEOF_ARRAY(), they would be wrong.
     
    +    In the implementation of ZUSTR2STP(), we want SIZEOF_ARRAY() within the
    +    static assert, because we're just comparing the sizes of the source and
    +    destination buffers, and we don't care if we compare sizes of numbers of
    +    elements, and using sizes is just simpler.  But we want NITEMS() in the
    +    zustr2stp() call, where we want to copy a specific number of characters.
    +
     
      ## lib/strtcpy.h ##
    @@ lib/utmp.c: static
     
      ## lib/zustr2stp.h ##
     @@
    - 
    - #define ZUSTR2STP(dst, src)                                                   \
      ({                                                                            \
    --  static_assert(!is_array(dst) || sizeof(dst) > SIZEOF_ARRAY(src), ""); \
    -+  static_assert(!is_array(dst) || sizeof(dst) > NITEMS(src), "");       \
    +   static_assert(!is_array(dst) || sizeof(dst) > SIZEOF_ARRAY(src), ""); \
                                                                                    \
     -  zustr2stp(dst, src, SIZEOF_ARRAY(src));                               \
     +  zustr2stp(dst, src, NITEMS(src));                                     \
```

alejandro-colomar:
v2b changes:

-  tfix

```
$ git range-diff shadow/master gh/nitems nitems 
1:  38d1b779 ! 1:  6c72182d lib/: Use NITEMS() instead of SIZEOF_ARRAY() where number of elements is meant
    @@ Commit message
     
         In the implementation of ZUSTR2STP(), we want SIZEOF_ARRAY() within the
         static assert, because we're just comparing the sizes of the source and
    -    destination buffers, and we don't care if we compare sizes of numbers of
    +    destination buffers, and we don't care if we compare sizes or numbers of
         elements, and using sizes is just simpler.  But we want NITEMS() in the
         zustr2stp() call, where we want to copy a specific number of characters.
     
```

alejandro-colomar:
v2c changes:

-  Rebase to master

```
$ git range-diff master..gh/nitems shadow/master..nitems 
1:  6c72182d = 1:  f760d404 lib/: Use NITEMS() instead of SIZEOF_ARRAY() where number of elements is meant
```

lib/utmp.c
```diff
@@ -246,7 +246,7 @@ static
 #ifdef HAVE_STRUCT_UTMP_UT_HOST
 	} else if (   (NULL != ut)
 	           && ('\0' != ut->ut_host[0])) {
-		hostname = XMALLOC(SIZEOF_ARRAY(ut->ut_host) + 1, char);
+		hostname = XMALLOC(NITEMS(ut->ut_host) + 1, char);
```
hallyn, line 249: This doesn't seem right.  If ut_host were to become something other than char, the result would be wrong, whereas with the current use of SIZEOF_ARRAY it would continue to work.  So auto-detecting the base type, to pass as second arg to XMALLOC, would be ok.

alejandro-colomar: If ut_host would become something other than char, then `hostname` should also become something other than char.

---
Regarding auto-detecting the base type:

The second argument to [X]MALLOC() is coupled with the asignee, thanks to the cast of the return type:

```c
$ grepc XMALLOC .
./lib/alloc.h:#define XMALLOC(n, type)  ((type *) xmallocarray(n, sizeof(type)))
```

That cast ensures that if `hostname` changed to `wchar_t`, the compiler would complain about implicit conversion from `char *` to `wchar_t *` at the assignment.

---

Regarding SIZEOF_ARRAY() vs NITEMS():

[X]MALLOC()'s arguments are 1) number of elements 2) type of each element.  Then, the [x]mallocarray() function performs the multiplication nelem * sizeof_elem, to pass the size to the actual malloc(3).

NITEMS() returns a number of elements, which is why it's the correct call.  SIZEOF_ARRAY() returns total size of the array, that is nelem * sizeof_elem.  If we pass SIZEOF_ARRAY() to MALLOC(), mallocarray() will multiply again by sizeof_elem, resulting in nelem * sizeof_elem * sizeof_elem.  That's conceptually wrong, even if sizeof_elem happens to be 1 in the present case.

alejandro-colomar: Let me show an example:


```c
wchar_t  src[4] = L"foo";
char     *c;
wchar_t  *w;

c = MALLOC(4, wchar_t); // Warning: implicit conversion for char* to wchar_t*
p = MALLOC(SIZEOF_ARRAY(src), wchar_t);  // We're allocating too much!  4 * sizeof(wchar_t) * sizeof(wchar_t)
p = MALLOC(NITEMS(src), wchar_t);  // Ok

**Comment by alejandro-colomar** [comment]
v2d changes:

-  Rebase to master
```
$ git range-diff master..gh/nitems shadow/master..nitems 
1:  f760d404 = 1:  fd051a19 lib/: Use NITEMS() instead of SIZEOF_ARRAY() where number of elements is meant
```

File: lib/utmp.c (hunk 1 above)
Comment (hallyn) [reply lib/utmp.c:249]: yeah but too much is ok, too little is not :)

but the tying to dest type is enough for me, thanks.

File: lib/strtcpy.h [hunk 2]
```diff
@@ -44,7 +44,7 @@
  */
 
 
-#define STRTCPY(dst, src)  strtcpy(dst, src, SIZEOF_ARRAY(dst))
+#define STRTCPY(dst, src)  strtcpy(dst, src, NITEMS(dst))
```
hallyn, line 47: But this should only ever be used on arrays of byte-sized elements, right?

strtcpy won't account for the sizeof(*dst).

lib/utmp.c
alejandro-colomar: > yeah but too much is ok, too little is not :)

Except when that may cause confusion.  :)

> but the tying to dest type is enough for me, thanks.

Good!

lib/strtcpy.h
alejandro-colomar: Yes and no.

STRTCPY() is designed to be used for copying both single-byte strings and multi-byte strings (UTF8).  So, there are multi-byte entities.  But yeah, they're both stored in arrays of `char`, so in the sense of the C language, these handle bytes.

But it would be conceivable that one may copy this source code to develop a WCSTCPY() macro (and the corresponding wcstcpy() function) to copy wide strings with truncation.  I prefer to give good example in our implementation, so that it'll be straight-forward to implement them as:


```c
#define WCSTCPY(dst, src)  wcstcpy(dst, src, NITEMS(dst))

inline ssize_t
wcstcpy(wchar_t *restrict dst, const wchar_t *restrict src, size_t dsize)
{
	bool     trunc;
	size_t   dlen, slen;
	wchar_t  *p;

	if (dsize == 0)
		return -1;

	slen = wcsnlen(src, dsize);
	trunc = (slen == dsize);
	dlen = slen - trunc;

	p = wmempcpy(dst, src, dlen);
	*p = '\0';

	return trunc ? -1 : slen;
}
```

Symmetry is a good thing.  :)

alejandro-colomar: So, yes, in practice, both SIZEOF_ARRAY() and NITEMS() will give the same result.  But one is more appropriate than the other.

alejandro-colomar: Oh, and a curious --and a bit unrelated-- fact I learnt a few days ago:

swprintf(3) was designed correctly, unlike snprintf(3).  It shares API with this wcstcpy().  It is not limited by an `int`, but rather `size_t`; and returns -1 on truncation instead of the length of the would-be string.

hallyn: FWIW I do not consider "length of the total string it would have tried to create" to be a bad return value.  In fact I quite like it.  But you do have to know to look for it.

alejandro-colomar: It has several problems:

-  Requires that snprintf(3) reads until the end, which makes it vulnerable to DoS attacks, very much like strlcpy(3).  Okay, one should probably validate an untrusted string before passing it to snprintf(3); but then it encourages programmers to design the functions that should be used for validating such strings with a compatible API by symmetry, as if it were a good thing to return (does it sound like OpenBSD's strlcpy(3)? :D)  (and symmetry strikes again! :).

   -  swprintf(3) can stop reading when it knows it has truncated.  Anyway, it'll return -1.

-  It is prone to off-by-one bugs.  One may accidentally write `if (snprintf(dst, ...) > sizeof(dst))` instead of using `>=`.  Not only off-by-ones; it can also happen an off-by-many, using accidentally `==`.  You may say it's so interiorized that it doesn't really happen.  I've seen this bug happen a couple of weeks ago by a libc maintainer (in strlcpy(3), not snprintf(3), but it's the same bug): <https://lore.kernel.org/linux-man/ZU4SDh-Se5gjPny5@debian/T/#m7c213beac8ddc019ec2fa0922a0cd500eb6dc65a>.  See the end of that email.

   -  swprintf() is instead checked as `if (swprintf(dst, ...) == -1)`, which is hard to get wrong.

In case it doesn't fail, it still returns the length, so in useful cases you still have it.

PR #853, opened 2023-11-27
lib/defines.h: Remove condition on __STRICT_ANSI__

alejandro-colomar:
We require C11 since a few releases ago.  It seems I missed this reminder of ANSI C (C89) back then.


ikerexxe (approved):
LGTM!

PR #855, opened 2023-11-28
lib/, src/: Say 'long' instead of 'long int'

alejandro-colomar:
We were using 'long' in most places, so be consistent and use it everywhere.  It's also shorter.


alejandro-colomar:
v1b changes:

-  Rebase to master

```
$ git range-diff master..gh/int shadow/master..int 
1:  09ff6e3d = 1:  666aba99 lib/, src/: Say 'long' instead of 'long int'
```

Issue #839, opened 2023-11-12
`./share/containers/fedora.dockerfile` doesn't build


alejandro-colomar:
```sh
$ sudo docker build -f ./share/containers/fedora.dockerfile .
[...]
checking for library containing readpassphrase... no
configure: error: readpassphrase() is missing, either from libc or libbsd
The command '/bin/sh -c ./autogen.sh --enable-shadowgrp --enable-man --with-audit         --with-sha-crypt --with-bcrypt --with-yescrypt --with-selinux         --without-libcrack --without-libpam --enable-shared         --with-group-name-max-length=32 --enable-lastlog --enable-logind=no' returned a non-zero code: 1
```

That is on the `master` branch.

I removed all the cached images from docker so that it starts from scratch, and I still see the failure.  Also, CI started failing for a PR that doesn't touch anything related to readpassphrase(3bsd) or libbsd.

ikerexxe:
It's failing because the latest stable Fedora has been switched to 39 recently, and this one doesn't containe libbsd dependency.

alejandro-colomar:
But Fedora 39 still has libbsd, right?  I see it in this page: <https://packages.fedoraproject.org/pkgs/libbsd/libbsd/>

Did they remove it from the docker container or something?

ikerexxe:
Yes, libbsd is still there but it's [not recommended](https://github.com/minimization/content-resolver-input/blob/95d4087bdf7a78a0b026d753e5c4b3b22b0df7ce/configs/sst_pt_gcc-unwanted.yaml#L51) to link base packages against libbsd.

The package is available for download, but it has to be installed manually, or via a dependency. I dropped the shadow dependency for this package some time ago for Fedora 39+, but I forgot to update the build configuration options.

Last week Fedora 39 was released and I guess at some point during this time the containers tag for `stable` was updated from 38 to 39. That's why we can see the error now, and not before.

alejandro-colomar:
I think we should be prefer linking against libbsd:

-  Our programs are not threaded, so being MT-Unsafe (Multi-Thread Unsafe) is not a bug here.
-  We already made sure we don't use libbsd's arc4random(3bsd): <https://github.com/shadow-maint/shadow/commit/ac8b81c2b7ed378528e5cc60754e10be1e369e40>
-  We don't use functions that are incompatible with glibc (fgetln, setproctitle).

[...]

PR #860, opened 2023-12-02
configure.ac: Move header and function checks to the end of the file

alejandro-colomar:
This allows checking functions that come from libraries like libbsd.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -35,25 +35,6 @@ LT_INIT
 
 dnl Checks for libraries.
 
-dnl Checks for header files.
-AC_CHECK_HEADERS(crypt.h utmp.h \
-	termio.h sgtty.h sys/ioctl.h paths.h \
-	sys/capability.h sys/random.h \
-	gshadow.h lastlog.h rpc/key_prot.h acl/libacl.h \
-	attr/libattr.h attr/error_context.h)
-
-dnl shadow now uses the libc's shadow implementation
-AC_CHECK_HEADER([shadow.h],,[AC_MSG_ERROR([You need a libc with shadow.h])])
-
-AC_CHECK_FUNCS(arc4random_buf futimes \
-	getentropy getrandom getspnam getusershell \
-	initgroups lckpwdf lutimes \
-	setgroups updwtmp updwtmpx innetgr \
-	getspnam_r \
-	rpmatch \
-	memset_explicit explicit_bzero stpecpy stpeprintf)
-AC_SYS_LARGEFILE
-
 dnl Checks for typedefs, structures, and compiler characteristics.
 
 AC_CHECK_MEMBERS([struct utmp.ut_type,
@@ -703,6 +684,25 @@ if test "$with_skey" = "yes"; then
 		skeychallenge((void*)0, (void*)0, (void*)0, 0);
 	]])],[AC_DEFINE(SKEY_BSD_STYLE, 1, [Define to support newer BSD S/Key API])],[])
 fi
+
+dnl Checks for header files.
+AC_CHECK_HEADERS(crypt.h utmp.h \
+	termio.h sgtty.h sys/ioctl.h paths.h \
+	sys/capability.h sys/random.h \
+	gshadow.h lastlog.h rpc/key_prot.h acl/libacl.h \
+	attr/libattr.h attr/error_context.h)
+
+dnl shadow now uses the libc's shadow implementation
+AC_CHECK_HEADER([shadow.h],,[AC_MSG_ERROR([You need a libc with shadow.h])])
+
+AC_CHECK_FUNCS(arc4random_buf futimes \
+	getentropy getrandom getspnam getusershell \
+	initgroups lckpwdf lutimes \
+	setgroups updwtmp updwtmpx innetgr \
+	getspnam_r \
+	rpmatch \
+	memset_explicit explicit_bzero stpecpy stpeprintf)
+AC_SYS_LARGEFILE
 
 PKG_CHECK_MODULES([CMOCKA], [cmocka], [have_cmocka="yes"],
 	[AC_MSG_WARN([libcmocka not found, cmocka tests will not be built])])


alejandro-colomar:
I'm not sure if I moved too much, or if I did anything wrong.  Please review thoroughly.

ikerexxe:
> I'm not sure if I moved too much, or if I did anything wrong if that has other consequences. Please review carefully.

I'd like to, but this is beyond my autoconf knowledge. I don't know which would be the consequences for such a change.

alejandro-colomar:
@thesamesam Can you review, please?  :)

configure.ac
```diff
@@ -35,25 +35,6 @@ LT_INIT
...
-AC_CHECK_HEADER([shadow.h],,[AC_MSG_ERROR([You need a libc with shadow.h])])
-
-AC_CHECK_FUNCS(arc4random_buf futimes \
-	getentropy getrandom getspnam getusershell \
-	initgroups lckpwdf lutimes \
-	setgroups updwtmp updwtmpx innetgr \
-	getspnam_r \
-	rpmatch \
-	memset_explicit explicit_bzero stpecpy stpeprintf)
-AC_SYS_LARGEFILE
```
vapier, line 55: `AC_SYS_LARGEFILE` should come asap.  it should already be called before header probes.  it will change `CPPFLAGS` which could, at least in theory, affect header inclusion behavior.

configure.ac
```diff
@@ -35,25 +35,6 @@ LT_INIT
 
 dnl Checks for libraries.
 
-dnl Checks for header files.
-AC_CHECK_HEADERS(crypt.h utmp.h \
-	termio.h sgtty.h sys/ioctl.h paths.h \
-	sys/capability.h sys/random.h \
-	gshadow.h lastlog.h rpc/key_prot.h acl/libacl.h \
```
vapier, line 42: `acl/libacl.h` shouldn't be probed like this.  plus, it's already probed below as part of the larger libacl detection.  ideally all of that would be deleted in favor of PKG_CHECK_MODULES calls ... doing that would also help cut down on noise by turning ~30 lines into like <5.

configure.ac
```diff
@@ -35,25 +35,6 @@ LT_INIT
...
-	attr/libattr.h attr/error_context.h)
```
vapier, line 43: same here -- these attr/ probes should be deleted since libattr is detected below, including these headers

configure.ac
```diff
@@ -35,25 +35,6 @@ LT_INIT
...
-	initgroups lckpwdf lutimes \
-	setgroups updwtmp updwtmpx innetgr \
-	getspnam_r \
-	rpmatch \
-	memset_explicit explicit_bzero stpecpy stpeprintf)
-AC_SYS_LARGEFILE
-
 dnl Checks for typedefs, structures, and compiler characteristics.
 
 AC_CHECK_MEMBERS([struct utmp.ut_type,
```
vapier, line 40: i think some of these tests below could be affected by which headers are detected.  worth double checking config.log to see what the generated tests are with the default set of header includes.

configure.ac
```diff
@@ -704,6 +685,25 @@ if test "$with_skey" = "yes"; then
 	]])],[AC_DEFINE(SKEY_BSD_STYLE, 1, [Define to support newer BSD S/Key API])],[])
 fi
 
+dnl Checks for header files.
+AC_CHECK_HEADERS(crypt.h utmp.h \
```
vapier, line 689: not exactly a new issue, but this (and funcs below) should ideally be using the `_ONCE` variant to avoid double-checking for the headers

vapier (changes requested):
if the goal is to allow 3rd party libs (like libbsd) to provide coverage, i think it'd make sense to organize those at the top.  then again, having a large flat configure.ac like this (~800 lines) makes it hard to organize and keep organized.  feels like it should be split up into some m4 files so that the configure.ac is easier to follow.  all of the shadow configure options that are independent of OS checks for example would be easy to pull out, and then make sure that they only do things like AC_ARG_ENABLE/AC_ARG_WITH, and don't call any CHECK funcs.

ikerexxe:
I read vapier's message and then I noticed

> This allows checking functions that come from libraries like libbsd.

I just want to remind you that Fedora and its derivatives don't allow any base OS package to have libbsd dependency. We already discussed this topic at https://github.com/shadow-maint/shadow/issues/839. So, unless the functions that you will use are provided by some other library (i.e. glibc) this way of working in general is a no-go for me.

vapier:
i don't see the problem.  the configure script has `--with-libbsd` / `--without-libbsd` options already.  if a distro (e.g. Fedora) doesn't want a package like shadow to depend on a specific library like libbsd, then they need to explicitly pass the corresponding `--without-libbsd` option.  if they aren't doing that, it's a bug in their packaging scripts.

plus, shadow is already in this situation: libbsd is probed by default and automatically used if it provides readpassphrase and the host C library does not (which glibc doesn't).  so if Fedora wanted to avoid depending on libbsd linkage, it already needs to pass the flag.

somewhat related, the default for libbsd should be "maybe", not "yes".

alejandro-colomar:
> I read vapier's message and then I noticed
> 
> > This allows checking functions that come from libraries like libbsd.
> 
> I just want to remind you that Fedora and its derivatives don't allow any base OS package to have libbsd dependency. We already discussed this topic at #839. So, unless the functions that you will use are provided by some other library (i.e. glibc) this way of working in general is a no-go for me.

It was precisely for that: I was going to provide an alternative definition for some libbsd functions if they were not available.

I could use the current test for libbsd, but I thought of testing the functions instead, which would be more precise.

alejandro-colomar:
I'll convert to draft, to address the issues raised by @vapier .  BTW, thanks for the review!

ikerexxe:
> It was precisely for that: I was going to provide an alternative definition for some libbsd functions if they were not available.

Perfect, perfect. A big kudos to you!

alejandro-colomar:
I'll close for now.  I'm not comfortable with it.

PR #864, opened 2023-12-04
lib/, src/: snprintf(3) already terminates strings with NUL

alejandro-colomar:
We don't need to terminate them manually after the call.  Remove all that paranoid code, which in some cases was even wrong.  While at it, let's do a few more things:

-  Use sizeof(buf) for the size of the buffer.  I found that a few cases were passing one less byte (probably because the last one was manually zeroed later).  This caused a double NUL.  snprintf(3) wants the size of the entire buffer to properly terminate it.  Passing the exact value hardcoded is brittle, so use sizeof().

-  Align and improve style of variable declarations.  This makes them appear in this diff, which will help review the patch.

### DIFF
--- a/lib/shell.c
+++ b/lib/shell.c
@@ -30,8 +30,8 @@ extern size_t newenvc;
 
 int shell (const char *file, /*@null@*/const char *arg, char *const envp[])
 {
-	char arg0[1024];
-	int err;
+	int   err;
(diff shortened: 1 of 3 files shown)


ikerexxe (approved):
LGTM. Thanks for the patch!

PR #866, opened 2023-12-04
src/login.c: Group preprocessor conditionals

alejandro-colomar:
Group them at the end of the list of variable definitions, and use '#if defined()' instead of '#if[n]def'.  Also indent nested ones.

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -478,21 +478,9 @@ int main (int argc, char **argv)
 {
 	const char *tmptty;
 	char tty[BUFSIZ];
-
-#ifdef RLOGIN
-	char term[128] = "";
-#endif				/* RLOGIN */
-#if !defined(USE_PAM)
-#ifdef ENABLE_LASTLOG
-	char ptime[80];
-#endif /* ENABLE_LASTLOG */
-#endif
 	unsigned int delay;
 	unsigned int retries;
 	bool subroot = false;
-#ifndef USE_PAM
-	bool is_console;
-#endif
 	int err;
 	unsigned int timeout;
 	const char *cp;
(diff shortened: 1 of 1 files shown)


src/login.c
```diff
@@ -503,13 +491,21 @@ int main (int argc, char **argv)
...
 
-#ifdef USE_PAM
+#if defined(USE_PAM)
 	int retcode;
 	pid_t child;
 	char *pam_user = NULL;
 #else
+# if defined(ENABLE_LASTLOG)
+	char ptime[80];
+# endif
```
ikerexxe, line 501: Move this after the declaration of `struct spwd *spwd = NULL;`. I think it makes it easier to read.

ikerexxe (commented):
General question, what is the purpose of this patch? Is it just a cosmetic change?

alejandro-colomar:
Hi Iker,

On Wed, Dec 13, 2023 at 03:39:49AM -0800, Iker Pedrosa wrote:
> @ikerexxe commented on this pull request.
> 
> General question, what is the purpose of this patch? Is it just a cosmetic change?

Yes, it's a cosmetic change.  I have some other patches that apply after
this one, because I had first to understand what was going on.  The
declarations were very confusing before this patch set.

> 
> > +# if defined(ENABLE_LASTLOG)
> +	char ptime[80];
> +# endif
> 
> Move this after the declaration of `struct spwd *spwd = NULL;`. I think it makes it easier to read.

Agree; I had that feeling a few days ago.  :)

Have a lovely day,
Alex

> -- 
> https://github.com/shadow-maint/shadow/pull/866#pullrequestreview-1779446272
> 
> Message ID: ***@***.***>

-- 
Looking for a remote C programming job at the moment.

alejandro-colomar:
v2 changes:

-  Move code around

```
$ git range-diff shadow/master gh/cpp cpp 
1:  9f3c3b14 ! 1:  a8358d3b src/login.c: Group preprocessor conditionals
    @@ src/login.c: int main (int argc, char **argv)
        pid_t child;
        char *pam_user = NULL;
      #else
    ++  bool is_console;
    +   struct spwd *spwd = NULL;
     +# if defined(ENABLE_LASTLOG)
     +  char ptime[80];
     +# endif
    -+  bool is_console;
    -   struct spwd *spwd = NULL;
      #endif
     +#if defined(RLOGIN)
     +  char term[128] = "";
```

hallyn (approved):

PR #869, opened 2023-12-10
lib/, src/: The only useful errno after strto[u][l]l(3) is ERANGE

alejandro-colomar:
```
EINVAL should be tested in a call where the string is known to contain a valid number and nothing else; otherwise, it's very easy to trigger Undefined Behavior.

Also, we're using 0 and 10 as bases, which cannot EINVAL.

Fixes: 1c464d9a2df7 ("lib/, src/: Fix error handling after strto[u]l[l](3)")
```
Link: <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/commit/?id=628d27c3c1d8362a7e73198460b8c6a739be4973>

Please merge, not rebase, so it's right after the commit that it fixes, which will help see the overall diff of both commits (`git diff HEAD^^`).

### DIFF
--- a/lib/get_gid.c
+++ b/lib/get_gid.c
@@ -20,7 +20,7 @@ int get_gid (const char *gidstr, gid_t *gid)
 	val = strtoll(gidstr, &endptr, 10);
 	if (   ('\0' == *gidstr)
 	    || ('\0' != *endptr)
-	    || (0 != errno)
+	    || (ERANGE == errno)
 	    || (/*@+longintegral@*/val != (gid_t)val)/*@=longintegral@*/) {
 		return 0;
 	}
--- a/lib/get_pid.c
+++ b/lib/get_pid.c
@@ -23,7 +23,7 @@ int get_pid (const char *pidstr, pid_t *pid)
 	val = strtoll(pidstr, &endptr, 10);
 	if (   ('\0' == *pidstr)
 	    || ('\0' != *endptr)
-	    || (0 != errno)
+	    || (ERANGE == errno)
 	    || (val < 1)
 	    || (/*@+longintegral@*/val != (pid_t)val)/*@=longintegral@*/) {
 		return 0;
@@ -49,7 +49,7 @@ int get_pidfd_from_fd(const char *pidfdstr)
 	val = strtoll(pidfdstr, &endptr, 10);
 	if (   ('\0' == *pidfdstr)
 	    || ('\0' != *endptr)
-	    || (0 != errno)
+	    || (ERANGE == errno)
 	    || (val < 0)
 	    || (/*@+longintegral@*/val != (int)val)/*@=longintegral@*/) {
 		return -1;
--- a/lib/get_uid.c
+++ b/lib/get_uid.c
@@ -20,7 +20,7 @@ int get_uid (const char *uidstr, uid_t *uid)
 	val = strtoll(uidstr, &endptr, 10);
 	if (   ('\0' == *uidstr)
 	    || ('\0' != *endptr)
-	    || (0 != errno)
+	    || (ERANGE == errno)
 	    || (/*@+longintegral@*/val != (uid_t)val)/*@=longintegral@*/) {
 		return 0;
 	}
--- a/lib/getgr_nam_gid.c
+++ b/lib/getgr_nam_gid.c
@@ -34,7 +34,7 @@ extern /*@only@*//*@null@*/struct group *getgr_nam_gid (/*@null@*/const char *gr
 	gid = strtoll(grname, &endptr, 10);
 	if (   ('\0' != *grname)
 	    && ('\0' == *endptr)
-	    && (0 == errno)
+	    && (ERANGE != errno)
 	    && (/*@+longintegral@*/gid == (gid_t)gid)/*@=longintegral@*/) {
 		return xgetgrgid (gid);
 	}
--- a/lib/getlong.c
+++ b/lib/getlong.c
@@ -26,7 +26,7 @@ int getlong (const char *numstr, /*@out@*/long *result)
 
 	errno = 0;
 	val = strtol(numstr, &endptr, 0);
-	if (('\0' == *numstr) || ('\0' != *endptr) || (0 != errno)) {
+	if (('\0' == *numstr) || ('\0' != *endptr) || (ERANGE == errno)) {
 		return 0;
 	}
 
--- a/lib/getrange.c
+++ b/lib/getrange.c
@@ -43,7 +43,7 @@ int getrange (const char *range,
 		}
 		errno = 0;
 		n = strtoul(&range[1], &endptr, 10);
-		if (('\0' != *endptr) || (0 != errno)) {
+		if (('\0' != *endptr) || (ERANGE == errno)) {
 			/* invalid */
 			return 0;
 		}
@@ -54,7 +54,7 @@ int getrange (const char *range,
 	} else {
 		errno = 0;
 		n = strtoul(range, &endptr, 10);
-		if (endptr == range || 0 != errno) {
+		if (endptr == range || ERANGE == errno) {
 			/* invalid */
 			return 0;
 		}
@@ -82,7 +82,7 @@ int getrange (const char *range,
 					errno = 0;
 					n = strtoul(endptr, &endptr, 10);
 					if (   ('\0' != *endptr)
-					    || (0 != errno)) {
+					    || (ERANGE == errno)) {
 						/* invalid */
 						return 0;
 					}
--- a/lib/gettime.c
+++ b/lib/gettime.c
@@ -38,7 +38,7 @@
 
 	errno = 0;
 	epoch = strtoull(source_date_epoch, &endptr, 10);
-	if (errno != 0) {
+	if (errno == ERANGE) {
 		fprintf (shadow_logfd,
 			 _("Environment variable $SOURCE_DATE_EPOCH: strtoull: %s\n"),
 			 strerror(errno));
--- a/lib/getulong.c
+++ b/lib/getulong.c
@@ -28,7 +28,7 @@ int getulong (const char *numstr, /*@out@*/unsigned long *result)
 	val = strtoul(numstr, &endptr, 0);
 	if (    ('\0' == *numstr)
 	     || ('\0' != *endptr)
-	     || (0 != errno)
+	     || (ERANGE == errno)
 	   ) {
 		return 0;
 	}
--- a/lib/limits.c
+++ b/lib/limits.c
@@ -65,7 +65,7 @@ static int setrlimit_value (unsigned int resource,
 		errno = 0;
 		l = strtol(value, &endptr, 10);
 
-		if (value == endptr || errno != 0)
+		if (value == endptr || errno == ERANGE)
 			return 0;  // FIXME: We could instead throw an error, though.
 
 		if (__builtin_mul_overflow(l, multiplier, &limit)) {
--- a/lib/prefix_flag.c
+++ b/lib/prefix_flag.c
@@ -352,7 +352,7 @@ extern struct group *prefix_getgr_nam_gid(const char *grname)
 	gid = strtoll(grname, &endptr, 10);
 	if (   ('\0' != *grname)
 	    && ('\0' == *endptr)
-	    && (0 == errno)
+	    && (ERANGE != errno)
 	    && (gid == (gid_t)gid))
 	{
 		return prefix_getgrgid(gid);
--- a/src/check_subid_range.c
+++ b/src/check_subid_range.c
@@ -36,10 +36,10 @@ int main(int argc, char **argv)
 	check_uids = argv[2][0] == 'u';
 	errno = 0;
 	start = strtoul(argv[3], NULL, 10);
-	if (errno != 0)
+	if (errno == ERANGE)
 		exit(1);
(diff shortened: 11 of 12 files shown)


alejandro-colomar:
If anyone was curious:

```diff
$ git diff erange^^..erange 
diff --git a/lib/get_gid.c b/lib/get_gid.c
index a43a7d68..15fbef31 100644
--- a/lib/get_gid.c
+++ b/lib/get_gid.c
@@ -17,7 +17,7 @@ int get_gid (const char *gidstr, gid_t *gid)
        char *endptr;
 
        errno = 0;
-       val = strtoll (gidstr, &endptr, 10);
+       val = strtoll(gidstr, &endptr, 10);
        if (   ('\0' == *gidstr)
            || ('\0' != *endptr)
            || (ERANGE == errno)
diff --git a/lib/get_pid.c b/lib/get_pid.c
index 5eda2027..84cab6f3 100644
--- a/lib/get_pid.c
+++ b/lib/get_pid.c
@@ -20,7 +20,7 @@ int get_pid (const char *pidstr, pid_t *pid)
        char *endptr;
 
        errno = 0;
-       val = strtoll (pidstr, &endptr, 10);
+       val = strtoll(pidstr, &endptr, 10);
        if (   ('\0' == *pidstr)
            || ('\0' != *endptr)
            || (ERANGE == errno)
@@ -46,7 +46,7 @@ int get_pidfd_from_fd(const char *pidfdstr)
        dev_t proc_st_dev, proc_st_rdev;
 
        errno = 0;
-       val = strtoll (pidfdstr, &endptr, 10);
+       val = strtoll(pidfdstr, &endptr, 10);
        if (   ('\0' == *pidfdstr)
            || ('\0' != *endptr)
            || (ERANGE == errno)
diff --git a/lib/get_uid.c b/lib/get_uid.c
index 797b4bb9..169acd5a 100644
--- a/lib/get_uid.c
+++ b/lib/get_uid.c
@@ -17,7 +17,7 @@ int get_uid (const char *uidstr, uid_t *uid)
        char *endptr;
 
        errno = 0;
-       val = strtoll (uidstr, &endptr, 10);
+       val = strtoll(uidstr, &endptr, 10);
        if (   ('\0' == *uidstr)
            || ('\0' != *endptr)
            || (ERANGE == errno)
diff --git a/lib/getgr_nam_gid.c b/lib/getgr_nam_gid.c
index 53a9ee56..06284399 100644
--- a/lib/getgr_nam_gid.c
+++ b/lib/getgr_nam_gid.c
@@ -31,7 +31,7 @@ extern /*@only@*//*@null@*/struct group *getgr_nam_gid (/*@null@*/const char *gr
        }
 
        errno = 0;
-       gid = strtoll (grname, &endptr, 10);
+       gid = strtoll(grname, &endptr, 10);
        if (   ('\0' != *grname)
            && ('\0' == *endptr)
            && (ERANGE != errno)
diff --git a/lib/getlong.c b/lib/getlong.c
index 6317bf0f..13347f93 100644
--- a/lib/getlong.c
+++ b/lib/getlong.c
@@ -25,7 +25,7 @@ int getlong (const char *numstr, /*@out@*/long *result)
        char *endptr;
 
        errno = 0;
-       val = strtol (numstr, &endptr, 0);
+       val = strtol(numstr, &endptr, 0);
        if (('\0' == *numstr) || ('\0' != *endptr) || (ERANGE == errno)) {
                return 0;
        }
diff --git a/lib/getrange.c b/lib/getrange.c
index 82f2edfa..caa3aa11 100644
--- a/lib/getrange.c
+++ b/lib/getrange.c
@@ -42,7 +42,7 @@ int getrange (const char *range,
                        return 0;
                }
                errno = 0;
-               n = strtoul (&range[1], &endptr, 10);
+               n = strtoul(&range[1], &endptr, 10);
                if (('\0' != *endptr) || (ERANGE == errno)) {
                        /* invalid */
                        return 0;
@@ -53,8 +53,8 @@ int getrange (const char *range,
                *max = n;
        } else {
                errno = 0;
-               n = strtoul (range, &endptr, 10);
-               if (ERANGE == errno) {
+               n = strtoul(range, &endptr, 10);
+               if (endptr == range || ERANGE == errno) {
                        /* invalid */
                        return 0;
                }
@@ -80,7 +80,7 @@ int getrange (const char *range,
                                        *has_min = true;
                                        *min = n;
                                        errno = 0;
-                                       n = strtoul (endptr, &endptr, 10);
+                                       n = strtoul(endptr, &endptr, 10);
                                        if (   ('\0' != *endptr)
                                            || (ERANGE == errno)) {
                                                /* invalid */
diff --git a/lib/gettime.c b/lib/gettime.c
index bb6e8443..289bdf17 100644
--- a/lib/gettime.c
+++ b/lib/gettime.c
@@ -37,9 +37,8 @@
                return fallback;
 
        errno = 0;
-       epoch = strtoull (source_date_epoch, &endptr, 10);
-       if ((errno == ERANGE && (epoch == ULLONG_MAX || epoch == 0))
-                       || (errno != 0 && epoch == 0)) {
+       epoch = strtoull(source_date_epoch, &endptr, 10);
+       if (errno == ERANGE) {
                fprintf (shadow_logfd,
                         _("Environment variable $SOURCE_DATE_EPOCH: strtoull: %s\n"),
                         strerror(errno));
diff --git a/lib/getulong.c b/lib/getulong.c
index 543fcd47..ca0aa362 100644
--- a/lib/getulong.c
+++ b/lib/getulong.c
@@ -25,7 +25,7 @@ int getulong (const char *numstr, /*@out@*/unsigned long *result)
        char *endptr;
 
        errno = 0;
-       val = strtoul (numstr, &endptr, 0);
+       val = strtoul(numstr, &endptr, 0);
        if (    ('\0' == *numstr)
             || ('\0' != *endptr)
             || (ERANGE == errno)
diff --git a/lib/limits.c b/lib/limits.c
index 1da228ca..b45e87db 100644
--- a/lib/limits.c
+++ b/lib/limits.c
@@ -62,14 +62,12 @@ static int setrlimit_value (unsigned int resource,
                 * Also, we are limited to base 10 here (hex numbers will not
                 * work with the limit string parser as is anyway)
                 */
+               errno = 0;
                l = strtol(value, &endptr, 10);
 
-               if (value == endptr) {
-                       /* No argument at all. No-op.
-                        * FIXME: We could instead throw an error, though.
-                        */
-                       return 0;
-               }
+               if (value == endptr || errno == ERANGE)
+                       return 0;  // FIXME: We could instead throw an error, though.
+
                if (__builtin_mul_overflow(l, multiplier, &limit)) {
                        /* FIXME: Again, silent error handling...
                         * Wouldn't screaming make more sense?
diff --git a/src/check_subid_range.c b/src/check_subid_range.c
index 38703b60..1d1eb94d 100644
--- a/src/check_subid_range.c
+++ b/src/check_subid_range.c
@@ -34,11 +34,12 @@ int main(int argc, char **argv)
 
        owner = argv[1];
        check_uids = argv[2][0] == 'u';
+       errno = 0;
        start = strtoul(argv[3], NULL, 10);
-       if (start == ULONG_MAX && errno == ERANGE)
+       if (errno == ERANGE)
                exit(1);
        count = strtoul(argv[4], NULL, 10);
-       if (count == ULONG_MAX && errno == ERANGE)
+       if (errno == ERANGE)
                exit(1);
        if (check_uids) {
                if (have_sub_uids(owner, start, count))
```

alejandro-colomar:
BTW, that shows something interesting, which I'm preparing on a future patch.

See this:

```diff
diff --git a/lib/gettime.c b/lib/gettime.c
index bb6e8443..289bdf17 100644
--- a/lib/gettime.c
+++ b/lib/gettime.c
@@ -37,9 +37,8 @@
                return fallback;
 
        errno = 0;
-       epoch = strtoull (source_date_epoch, &endptr, 10);
-       if ((errno == ERANGE && (epoch == ULLONG_MAX || epoch == 0))
-                       || (errno != 0 && epoch == 0)) {
+       epoch = strtoull(source_date_epoch, &endptr, 10);
+       if (errno == ERANGE) {
                fprintf (shadow_logfd,
                         _("Environment variable $SOURCE_DATE_EPOCH: strtoull: %s\n"),
                         strerror(errno));
```

More specifically, in this line:
```diff
-       if ((errno == ERANGE && (epoch == ULLONG_MAX || epoch == 0))
```

This shows that the programmer didn't know (and I didn't either, until a few days ago) that strtoul(3) doesn't reject negative values.

The programmer, and anyone with common sense, would expect that -3 would underflow strtoul(3), and like with strtol(3), and underflow would be reported with `ERANGE` and the return value would be the lowest number it can return, which would be `0`.  But, for some reason, the designers of strtoul(3) decided to do the wrong thing:

<https://blog.habets.se/2022/10/No-way-to-parse-integers-in-C.html>
<https://softwareengineering.stackexchange.com/a/449060/332848>

strtoul(3) silently converts negative values to positive ones, and one needs to write strtoul_noneg() to do the right thing.

I'll send a PR when this one is merged (but please, remember not to rebase).

hallyn:
What is the danger in balking on (errno != 0 && errno != ERANGE) ?

Depending on the libc setting errno if there is an error seems safe enough, but depending on it setting it only to ERANGE seems risky.  If next week some libc decides that for security reasons it needs to return something else (EINVAL, E2BIG) for a case we're not currently considering, we'll need to patch shadow.

alejandro-colomar:
> What is the danger in balking on (errno != 0 && errno != ERANGE) ?
> 
> Depending on the libc setting errno if there is an error seems safe enough, but depending on it setting it only to ERANGE seems risky. If next week some libc decides that for security reasons it needs to return something else (EINVAL, E2BIG) for a case we're not currently considering, we'll need to patch shadow.

Another possibility is that in such a new error case, the error handling may need to be different.  If we check `ERANGE` only, a maintainer will understand that the code wasn't prepared for that error, and will update it accordingly.  However, if a maintainer sees `errno != 0`, they may think that it already took it into account, and do nothing at all, leaving a subtle problem.

alejandro-colomar:
One real case that I'm worried about is that with `errno != 0` a maintainer may think that the `base` is being checked (`EINVAL`) and so may change the code to accept a user-supplied base, because since it's already being checked against, there should be no danger, right?

There is, which is why I changed the manual page.  See this code:

```c
#include <errno.h>
#include <stdio.h>
#include <stdlib.h>

int
main(void)
{
	char  *end;
	char  *p = " 42foo";

	strtol(p, &end, -42);
	if (*end != '\0' || end == p || errno != 0)
		fprintf(stderr, "strtol() failed\n");
}
```

```sh
$ cc -Wall -Wextra strtol.c 
$ ./a.out 
Segmentation fault
```

alejandro-colomar:
In any case, I plan to reduce the uses of strto[u][l]l(3), in favor of other calls that are not so hard to test, so this shouldn't be a big problem after that.

hallyn:
> > What is the danger in balking on (errno != 0 && errno != ERANGE) ?
> 
> Only the danger of confusing the maintainer about what errors are actually being checked.
> 
> > Depending on the libc setting errno if there is an error seems safe enough, but depending on it setting it only to ERANGE seems risky. If next week some libc decides that for security reasons it needs to return something else (EINVAL, E2BIG) for a case we're not currently considering, we'll need to patch shadow.
> 
> Another possibility is that in such a new error case, the error handling may need to be different. If we check `ERANGE` only, a maintainer will understand that the code wasn't prepared for that error, and will update it accordingly. However, if a maintainer sees `errno != 0`, they may think that it already took it into account, and do nothing at all, leaving a subtle problem.

Ok, but things rarely go that way - much more likely is that glibc adds a new possible return code and - no matter how well they communicate it - we don't notice, or someone simply builds an old shadow against the new glibc.

What about adding just an ABORT_ON(errno < 0 && errno != ERANGE) ?

alejandro-colomar:
On Wed, Dec 27, 2023 at 08:38:09AM -0800, Serge Hallyn wrote:
> > Another possibility is that in such a new error case, the error handling may need to be different. If we check `ERANGE` only, a maintainer will understand that the code wasn't prepared for that error, and will update it accordingly. However, if a maintainer sees `errno != 0`, they may think that it already took it into account, and do nothing at all, leaving a subtle problem.
> 
> Ok, but things rarely go that way - much more likely is that glibc adds a new possible return code and - no matter how well they communicate it - we don't notice, or someone simply builds an old shadow against the new glibc.
> 
> What about adding just an ABORT_ON(errno < 0 && errno != ERANGE) ?

I have plans to significantly change this code anyway.  Please keep that
in mind for when that happens.

Since strtol(3) is so bad to handle errors, I plan to use strtoi(3bsd)
(writing a copy of it in our project, to keep Fedora happy); then
reimplement getlong() and friends in terms of strtoi(3bsd); and then
increasing the use of getlong() minimizing raw strto*() calls.  I like
getlong() (and things like get_pid() better, because they have more
type safety, thanks to using a pointer.

I'll fix in that process a few bugs of handling out-of-range bugs.

Does the plan sound good to you?

I still want this specific commit merged before that, just as a historic
warning that strtol(3) isn't so easy to use.  But I plan to nuke all
this code, so it isn't strictly necessary, if you prefer to oppose to it.

Have a lovely day,
Alex

-- 
Looking for a remote C programming job at the moment.

alejandro-colomar:
I'm reconsidering.  Since this change is controversial, I'll drop it, and do the nuking from the less controversial point.

PR #874, opened 2023-12-13
src/su.c: Fix type of variable

alejandro-colomar:
su.c:678:26: warning: format ‘%s’ expects argument of type ‘char *’, but argument 4 has type ‘const void *’ [-Wformat=]
su.c:681:44: warning: format ‘%s’ expects argument of type ‘char *’, but argument 3 has type ‘const void *’ [-Wformat=]
su.c:683:46: warning: format ‘%s’ expects argument of type ‘char *’, but argument 3 has type ‘const void *’ [-Wformat=]

Reported-by: Christian Göttsche <cgzones@googlemail.com>

### DIFF
--- a/src/su.c
+++ b/src/su.c
@@ -643,8 +643,8 @@ static /*@only@*/struct passwd * check_perms (void)
 static /*@only@*/struct passwd * do_check_perms (void)
 {
 #ifdef USE_PAM
-	const void *tmp_name;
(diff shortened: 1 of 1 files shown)


hallyn (approved):
I assume it was done like this because pam_get_item() says it expects a void**, but looks good, thanks.

PR #876, opened 2023-12-19
Day and time cleanups for shadow file entries

stoeckmann:
The first commit is about SCALE removal:
- SCALE is always DAY, so drop the definition
- With SCALE being gone, drop unneeded calculations

The other commits clean up calculations based on days per epoch and seconds per epoch. It is much easier to calculate with days (in a long) than with seconds (in a time_t) because the former is already the native encoding in shadow files and LONG_MAX exists, while TIME_T_MAX does not exist everywhere.

- Use proper casts which become relevant for 32 bit systems with 64 bit time_t, e.g. musl-based distributions
- Let `passwd` work with day precision instead of seconds since epoch to allow easier range checking
- Unify integer overflow checks in `chage`
- Add LONG_MAX range checks when working with sp_max etc.

Proof of Concept (for 64 bit systems):
1. Compile and definitely use an `sgetspent` implementation which gets long parsing right (glibc does not)
```
ac_cv_func_getspnam=no \
ac_cv_func_sgetspent=no \
CFLAGS="-fsanitize=undefined" \
./configure --without-libpam
```
2. Setup users with large shadow entries
```
# chage -m 10 user1
# chage -d 9223372036854775807 user1
```
```
# chage -M 10 user2
# chage -d 9223372036854775807 user2
```
```
# chage -M 10 user3
# chage -I 20 user3
# chage -d 9223372036854775807 user3
```
```
# chage -I 20 user4
# chage -W 1 user4
# chage -d 123 user4
# chage -M 9223372036854775807 user4
```
3. Change password as user1
```
$ passwd
../../src/passwd.c:393:6: runtime error: signed integer overflow: 9223372036854775807 * 86400 cannot be represented in type 'long int'
```
4. Run expiry as user2
```
$ expiry -c
../../lib/isexpired.c:97:28: runtime error: signed integer overflow: 9223372036854775807 + 10 cannot be represented in type 'long int'
```
5. Run expiry as user3
```
$ expiry -c
../../lib/age.c:165:25: runtime error: signed integer overflow: 9223372036854775807 + 99999 cannot be represented in type 'long int'
../../lib/isexpired.c:75:32: runtime error: signed integer overflow: 9223372036854775807 + 99999 cannot be represented in type 'long int'
```
6. Run expiry as user4
```
$ expiry -c
../../lib/age.c:165:25: runtime error: signed integer overflow: 123 + 9223372036854775807 cannot be represented in type 'long int'
../../lib/age.c:165:9: runtime error: signed integer overflow: -9223372036854775686 - 19709 cannot be represented in type 'long int'
../../lib/isexpired.c:75:32: runtime error: signed integer overflow: 123 + 9223372036854775807 cannot be represented in type 'long int'
```

The user1 password change should not be allowed because the minimum wait time since last change (in the future) is not reached yet.

As a solution I have capped these calculations at `LONG_MAX` since we are talking, even for 32 bit systems, about a time span of more than 6 million years before this capping becomes a problem. Rather... Theoretical... I hope. :)

### DIFF
--- a/lib/age.c
+++ b/lib/age.c
@@ -13,11 +13,14 @@
 #include <stdio.h>
 #include <time.h>
 #include <errno.h>
-#include "prototypes.h"
-#include "defines.h"
-#include "exitcodes.h"
 #include <pwd.h>
 #include <grp.h>
+
+#include "adds.h"
+#include "defines.h"
+#include "exitcodes.h"
+#include "prototypes.h"
+
 
 #ident "$Id$"
 
@@ -162,7 +165,8 @@ void agecheck (/*@null@*/const struct spwd *sp)
 		return;
 	}
 
-	remain = sp->sp_lstchg + sp->sp_max - now;
+	remain = addsl(sp->sp_lstchg, sp->sp_max, -now);
+
 	if (remain <= sp->sp_warn) {
 		if (remain > 1) {
 			(void) printf (_("Your password will expire in %ld days.\n"),
--- a/lib/isexpired.c
+++ b/lib/isexpired.c
@@ -15,10 +15,12 @@
 #include <config.h>
 
 #include <sys/types.h>
-#include "prototypes.h"
-#include "defines.h"
 #include <pwd.h>
 #include <time.h>
+
+#include "adds.h"
+#include "defines.h"
+#include "prototypes.h"
 
 #ident "$Id$"
 
@@ -38,7 +40,7 @@
  */
 int isexpired (const struct passwd *pw, /*@null@*/const struct spwd *sp)
 {
-	long now;
+	long  now;
 
 	now = time(NULL) / DAY;
 
@@ -72,7 +74,8 @@ int isexpired (const struct passwd *pw, /*@null@*/const struct spwd *sp)
 	if (   (sp->sp_lstchg > 0)
 	    && (sp->sp_max >= 0)
 	    && (sp->sp_inact >= 0)
-	    && (now >= (sp->sp_lstchg + sp->sp_max + sp->sp_inact))) {
+	    && (now >= addsl(sp->sp_lstchg, sp->sp_max, sp->sp_inact)))
+	{
 		return 2;
 	}
 
@@ -94,9 +97,9 @@ int isexpired (const struct passwd *pw, /*@null@*/const struct spwd *sp)
 	 * the password has expired.
 	 */
 
-	if (now >= (sp->sp_lstchg + sp->sp_max)) {
+	if (now >= addsl(sp->sp_lstchg, sp->sp_max))
 		return 1;
-	}
+
 	return 0;
 }
 
--- a/src/chage.c
+++ b/src/chage.c
@@ -76,7 +76,7 @@ static long expdate;
 /* local function prototypes */
 NORETURN static void usage (int status);
 static int new_fields (void);
-static void print_date (time_t date);
+static void print_day_as_date (long day);
 static void list_fields (void);
 static void process_flags (int argc, char **argv);
 static void check_flags (int argc, int opt_index);
@@ -231,10 +231,22 @@ static int new_fields (void)
 	return 1;
 }
 
-static void print_date (time_t date)
+
+static void
+print_day_as_date(long day)
 {
-	struct tm *tp;
-	char buf[80];
+	char       buf[80];
+	time_t     date;
+	struct tm  *tp;
+
+	if (day < 0) {
+		puts(_("never"));
+		return;
+	}
+	if (__builtin_mul_overflow(day, DAY, &date)) {
+		puts(_("future"));
+		return;
+	}
 
 	tp = gmtime (&date);
 	if (NULL == tp) {
@@ -245,6 +257,7 @@ static void print_date (time_t date)
 	}
 }
 
+
 /*
  * list_fields - display the current values of the expiration fields
  *
@@ -254,21 +267,15 @@ static void print_date (time_t date)
  */
 static void list_fields (void)
 {
-	long changed = 0;
-	long expires;
-
 	/*
 	 * The "last change" date is either "never" or the date the password
 	 * was last modified. The date is the number of days since 1/1/1970.
 	 */
 	(void) fputs (_("Last password change\t\t\t\t\t: "), stdout);
-	if (lstchgdate < 0 || lstchgdate > LONG_MAX / DAY) {
-		(void) puts (_("never"));
-	} else if (lstchgdate == 0) {
+	if (lstchgdate == 0) {
 		(void) puts (_("password must be changed"));
 	} else {
-		changed = lstchgdate * DAY;
-		print_date (changed);
+		print_day_as_date(lstchgdate);
 	}
 
 	/*
@@ -281,11 +288,11 @@ static void list_fields (void)
 	} else if (   (lstchgdate < 0)
 	           || (maxdays >= 10000)
 	           || (maxdays < 0)
-	           || ((LONG_MAX - changed) / DAY < maxdays)) {
+	           || (LONG_MAX - lstchgdate < maxdays))
+	{
 		(void) puts (_("never"));
 	} else {
-		expires = changed + maxdays * DAY;
-		print_date (expires);
+		print_day_as_date(lstchgdate + maxdays);
 	}
 
 	/*
@@ -301,12 +308,12 @@ static void list_fields (void)
 	           || (inactdays < 0)
 	           || (maxdays >= 10000)
 	           || (maxdays < 0)
-	           || (maxdays > LONG_MAX - inactdays)
-	           || ((LONG_MAX - changed) / DAY < maxdays + inactdays)) {
+	           || (LONG_MAX - inactdays < maxdays)
+	           || (LONG_MAX - lstchgdate < maxdays + inactdays))
+	{
 		(void) puts (_("never"));
 	} else {
-		expires = changed + (maxdays + inactdays) * DAY;
-		print_date (expires);
+		print_day_as_date(lstchgdate + maxdays + inactdays);
 	}
 
 	/*
@@ -314,12 +321,7 @@ static void list_fields (void)
 	 * password expiring or not.
 	 */
 	(void) fputs (_("Account expires\t\t\t\t\t\t: "), stdout);
-	if (expdate < 0 || LONG_MAX / DAY < expdate) {
-		(void) puts (_("never"));
-	} else {
-		expires = expdate * DAY;
-		print_date (expires);
-	}
+	print_day_as_date(expdate);
 
 	/*
 	 * Start with the easy numbers - the number of days before the


lib/isexpired.c
```diff
@@ -72,7 +90,7 @@ int isexpired (const struct passwd *pw, /*@null@*/const struct spwd *sp)
 	if (   (sp->sp_lstchg > 0)
 	    && (sp->sp_max >= 0)
 	    && (sp->sp_inact >= 0)
-	    && (now >= (sp->sp_lstchg + sp->sp_max + sp->sp_inact))) {
+	    && (now >= sum(sp->sp_lstchg, sp->sp_max, sp->sp_inact))) {
```
alejandro-colomar, line 93: I think I would prefer to reject overflows reporting an error, instead of silently truncating.

How about using something like `if (__builtin_add_overflow(...)) { goto error; }`?

<https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html#index-_005f_005fbuiltin_005fadd_005foverflow>,

src/chage.c
```diff
@@ -231,10 +231,17 @@ static int new_fields (void)
...
 {
 	struct tm *tp;
 	char buf[80];
+	time_t date;
+
+	if (day < 0 || LONG_MAX / DAY < day) {
+		(void) puts (_("never"));
+		return;
+	}
+	date = (time_t) day * DAY;
```
alejandro-colomar, line 244: I suggest the following:

```c
if (day < 0) {
    puts(_("never"));
    return;
}
if (__builtin_mul_overflow(day, DAY, &date)) {
    puts(_("future"));
    return;
}
```

It's a bit simpler and more readable about checking for overflow, and it's also more consistent in the handling of overflow in dates compared to other functions:

```c
$ grepc date_to_str .
./lib/prototypes.h:extern void date_to_str (size_t size, char buf[size], long date);
./lib/date_to_str.c:void
date_to_str(size_t size, char buf[size], long date)
{
	time_t           t;
	const struct tm  *tm;

	t = date;
	if (date < 0) {
		(void) strtcpy(buf, "never", size);
		return;
	}

	tm = gmtime(&t);
	if (tm == NULL) {
		(void) strtcpy(buf, "future", size);
		return;
	}

	if (strftime(buf, size, "%Y-%m-%d", tm) == 0)
		(void) strtcpy(buf, "future", size);
}
```

src/passwd.c
```diff
@@ -426,7 +426,7 @@ static void print_status (const struct passwd *pw)
 
 	sp = prefix_getspnam (pw->pw_name); /* local, no need for xprefix_getspnam */
 	if (NULL != sp) {
-		date_to_str (sizeof(date), date, sp->sp_lstchg * DAY),
+		date_to_str (sizeof(date), date, (time_t) sp->sp_lstchg * DAY),
```
alejandro-colomar, line 429: How about redefining DAY to use `long long` instead of adding casts?

<https://github.com/shadow-maint/shadow/pull/880>

stoeckmann: Removed the casts, so this PR basically depends on https://github.com/shadow-maint/shadow/pull/880

src/chage.c
stoeckmann: Oh, didn't see that "future" is actually used in other places. Adjusted.

lib/isexpired.c
stoeckmann: The isexpired function would have to be restructured a bit to actually return such error values. I'm not sure if it's worth it, considering the fact that we talk about millions of years before this is a real world scenario issue -- instead of an issue with broken files.

src/passwd.c
alejandro-colomar: :-)

lib/isexpired.c
alejandro-colomar: Hmm.

What's the behavior now with broken files or with huge dates, and what is the behavior after your patch?  Does the behavior change?

stoeckmann: If the overflow occurs, `now` is larger than this limit, at which an account shall be locked. At least if signed overflows just wraps around, as it normally should do.

So, right now accounts are locked too early. With this patch, they are locked "too early" in a few million years. So, I guess that's acceptable and resembles as close as possible what a system administrator expressed with such huge values.

alejandro-colomar: LGTM.

lib/age.c
```diff
@@ -162,9 +162,14 @@ void agecheck (/*@null@*/const struct spwd *sp)
...
 	}
 
-	remain = sp->sp_lstchg + sp->sp_max - now;
+	remain = sp->sp_lstchg - now;
+	if (remain < LONG_MAX - sp->sp_max) {
+		remain += sp->sp_max;
+	} else {
+		remain = LONG_MAX;
+	}
+
```
alejandro-colomar, line 171: Why didn't you use sum() here?

stoeckmann: sum is static in lib/isexpired.c

lib/isexpired.c
```diff
@@ -22,6 +22,24 @@
...
+
+	if (v1 > LONG_MAX - v2)
+		return LONG_MAX;
+	value = v1 + v2;
+
+	if (value > LONG_MAX - v3)
+		return LONG_MAX;
+
+	return value + v3;
+}
```
alejandro-colomar, line 42: I'd use something in the name that indicates saturation, even if just an 's'.

Also, an l for long.

For symmetry with __builtin_addcl(), how about addsl()?  Or maybe addl_saturate(), if you prefer a longer name.

lib/age.c
alejandro-colomar: Why not make it extern?

lib/isexpired.c
stoeckmann: Renamed to addsl

lib/age.c
stoeckmann: The if-statement was simple enough from my point of view. I've reused sum (now addsl) in the function just because adding three values takes one more check than this simple if-statement.

lib/isexpired.c
```diff
@@ -22,6 +22,24 @@
...
```
alejandro-colomar, line 42: We could also make it slightly more generic to accept negative numbers and saturate to LONG_MIN, although that would increase complexity a little bit.

lib/isexpired.c
```diff
@@ -22,6 +22,24 @@
 
 #ident "$Id$"
 
+/*
+ * addsl - adds three non-negative long values
+ *
+ *	the result is capped at LONG_MAX.
```
alejandro-colomar, line 28: I suggest using the word saturate instead of cap, since the function name uses an s.

lib/age.c
alejandro-colomar: You could have a 3-input function called addsl3(), and a 2-input addsl().

lib/isexpired.c
stoeckmann: Adjusted

lib/isexpired.c
stoeckmann: Since the input values are always positive, this adds complexity without benefit.

stoeckmann:
I've checked the box which allows edits by maintainers. Perhaps it's easier to consider my PR as a suggestion so wordings and other things can be changed directly without going through comments/reviews.

lib/age.c
alejandro-colomar: Here's a suggestion for that:

```c
$ cat addsl.c 
inline long
addsl(long a, long b)
{
	if (a > 0 && b > LONG_MAX - a) {
		errno = EOVERFLOW;
		return LONG_MAX;
	}
	if (a < 0 && b < LONG_MIN - a) {
		errno = EOVERFLOW;
		return LONG_MIN;
	}
	return a + b;
}


static
long cmpl(const void *p1, const void *p2)
{
	long  *l1 = p1;
	long  *l2 = p2;

	if (*l1 < *l2)
		return -1;
	if (*l1 > *l2)
		return +1;
	return 0;
}


inline long
addsl3(long a, long b, long c)
{
	int   e;
	long  n[3];

	n[0] = a;
	n[1] = b;
	n[2] = c;

	e = errno;
	qsort(n, 3, sizeof(long), cmpl);

	errno = 0;
	sum = addsl(n[0], n[2]);
	if (errno == EOVERFLOW)
		return sum;
	sum = addsl(sum, n[1]);
	if (errno == 0)
		errno = e;
	return sum;
}
```

alejandro-colomar:
> I've checked the box which allows edits by maintainers. Perhaps it's easier to consider my PR as a suggestion so wordings and other things can be changed directly without going through comments/reviews.

I find it more interesting to have a conversation.  I may say very dumb things, so if you do the changes after my suggestions, there's some chance of catching any mistakes I may say.  :)

lib/isexpired.c
alejandro-colomar: Agree.  After implementing it myself, it proves to be quite complex.

The benefit would be that if someone uses that function in a case where we don't know if the values are all non-negative, we may get surprises, but if the function is prepared to receive negative values, then it's all fine.

I'll try to add it myself in a separate patch, and provide tests for it.

lib/age.c
stoeckmann: Extended the if-conditions a bit so the builtin functions can be used directly. Tested with the test cases specified in PR.

stoeckmann:
Got the return value checks mixed up. Now it should be functional again.

alejandro-colomar:
<https://github.com/shadow-maint/shadow/pull/876/commits/2ce77ab920fd69804a402cd62e1cc3b166150a19>

```
```

src/passwd.c
```diff
@@ -387,10 +384,12 @@ static void check_password (const struct passwd *pw, const struct spwd *sp)
...
-		time_t ok;
-		ok = (time_t) sp->sp_lstchg * DAY;
-		if (sp->sp_min > 0) {
-			ok += (time_t) sp->sp_min * DAY;
+		long now, ok;
+		now = time (NULL) / DAY;
+		ok = sp->sp_lstchg;
+		if (   (sp->sp_min > 0)
+		    && __builtin_saddl_overflow(ok, sp->sp_min, &ok)) {
+			ok = LONG_MAX;
```
alejandro-colomar: I'd prefer the check for overflow in a separate commit.

stoeckmann: Separated

stoeckmann:
> Got the return value checks mixed up. Now it should be functional again.

And the limit had to be adjusted... AGAIN.

If this keeps going on like this, I'll change this PR into an issue and someone else can look at it. Maybe people can find this kind of interaction enjoyable, but I'm not part of that group.

alejandro-colomar:
<https://github.com/shadow-maint/shadow/pull/876/commits/747ca85c65031522b6b2ea88db822849e1d9f203>:

```
```

alejandro-colomar:
<https://github.com/shadow-maint/shadow/pull/876/commits/73d8e90a6b056ffed30eddfea02a23c437a9eb00>:

```
```

src/pwck.c
```diff
@@ -816,7 +816,7 @@ static void check_spw_file (int *errors, bool *changed)
 		if (!quiet) {
 			time_t t = time (NULL);
 			if (   (t != 0)
-			    && (spw->sp_lstchg > (long) t / DAY)) {
+			    && (spw->sp_lstchg > (long) (t / DAY))) {
```
alejandro-colomar: I think this patch should be dropped, right?  <https://github.com/shadow-maint/shadow/pull/876/commits/17f48ff54675924580da37cb7712df0fdd624d35>

alejandro-colomar: Or maybe the commit message changed.

stoeckmann: The cast has to disappear, otherwise `t` is capped at LONG_MAX for 32 bit systems with a 64 bit time_t. Adjusted.

alejandro-colomar:
<https://github.com/shadow-maint/shadow/pull/876/commits/918877fb89c13d3ed5519a206a73596023f12ea5>:

```
```

lib/age.c
```diff
@@ -162,7 +162,11 @@ void agecheck (/*@null@*/const struct spwd *sp)
 		return;
 	}
 
-	remain = sp->sp_lstchg + sp->sp_max - now;
+	remain = sp->sp_lstchg - now;
```
hallyn, line 165: Is there any way that the - now can lead to overflow?  Would probably require time(2) to return -1, and sp->sp_lstchg to already be max?

stoeckmann: Yes, I assume that time is positive, i.e. the system time is properly set. We can protect against it though.

stoeckmann: Updated the PR. This should do the trick. In this scenario, the difference is larger than LONG_MAX, i.e. the warn period cannot be reached. So we can directly call return.

lib/isexpired.c
```diff
@@ -69,10 +69,15 @@ int isexpired (const struct passwd *pw, /*@null@*/const struct spwd *sp)
...
+	if (   __builtin_saddl_overflow(sp->sp_lstchg, sp->sp_max, &limit)
+	    || __builtin_saddl_overflow(limit, sp->sp_inact, &limit)) {
+		limit = LONG_MAX;
+	}
+
 	if (   (sp->sp_lstchg > 0)
 	    && (sp->sp_max >= 0)
 	    && (sp->sp_inact >= 0)
-	    && (now >= (sp->sp_lstchg + sp->sp_max + sp->sp_inact))) {
+	    && (now >= limit)) {
```
hallyn, line 80: Hm, I think I'm missing something here: you don't initialize limit to anything, so it is only set in the new stanza above if one of the additions overflows.  But I don't think you only check for (now >= limit) only in those cases.

lib/age.c
hallyn: Thanks, yes, I think that's good.

lib/isexpired.c
stoeckmann: The result of __builtin_saddl_overflow is stored in its third argument, which is limit. So, limit is the result of adding sp->sp_lstchg, sp->sp_max and sp->sp_inact. If this calculation overflows, it is manually set to LONG_MAX.

alejandro-colomar:
> [918877f](https://github.com/shadow-maint/shadow/commit/918877fb89c13d3ed5519a206a73596023f12ea5):
> 
> ```
> ```

Please, include those in the commit messages that I reviewed (unless you modify a patch).  Otherwise, it's hard to keep track of what's reviewed and what not, and what changed.

stoeckmann:
As mentioned before, I close this PR and let others do it now.

alejandro-colomar:
I've opened&closed to get the patches.

alejandro-colomar:
v2 (by @alejandro-colomar ) changes:

-  Use untyped versions of GCC integer builtin functions.  They are safer.
-  Don't cast input to GCC integer builtin functions.  They work well on arbitrary input integer types.
-  Some style changes
-  Use addsl()

```
$ git range-diff 8ea6076^..8017c0f shadow/master..scale 
1:  8ea60764 < -:  -------- lib/, src/: Remove SCALE definition
2:  ecc15ebb < -:  -------- src/passwd.c: Switch to day precision
3:  e9df8fe4 < -:  -------- src/passwd.c: Add overflow check
4:  dc6c7d69 < -:  -------- src/: Fix long/time_t handling
5:  24181b0d ! 1:  7064949f src/chage.c: unify long overflow checks
    @@ Metadata
     Author: Tobias Stoeckmann <tobias@stoeckmann.org>
     
      ## Commit message ##
    -    src/chage.c: unify long overflow checks
    +    src/chage.c: Unify long overflow checks in print_day_as_date()
     
         The conversion from day to seconds can be done in print_date
    -    (renamed to print_day_as_date for clarification). This has the nice
    +    (renamed to print_day_as_date for clarification).  This has the nice
         benefit that DAY multiplication and long to time_t conversion are done
         at just one place.
     
    -    Signed-off-by: Tobias Stoeckmann <tobias@stoeckmann.org>
    +    Co-developed-by: Tobias Stoeckmann <tobias@stoeckmann.org>
    +    Co-developed-by: Alejandro Colomar <alx@kernel.org>
    +    Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
      ## src/chage.c ##
     @@ src/chage.c: static long expdate;
    @@ src/chage.c: static int new_fields (void)
      }
      
     -static void print_date (time_t date)
    -+static void print_day_as_date (long day)
    ++
    ++static void
    ++print_day_as_date(long day)
      {
    -   struct tm *tp;
    -   char buf[80];
    -+  time_t date;
    +-  struct tm *tp;
    +-  char buf[80];
    ++  char       buf[80];
    ++  time_t     date;
    ++  struct tm  *tp;
     +
     +  if (day < 0) {
     +          puts(_("never"));
     +          return;
     +  }
    -+  if (__builtin_mul_overflow((time_t) day, DAY, &date)) {
    ++  if (__builtin_mul_overflow(day, DAY, &date)) {
     +          puts(_("future"));
     +          return;
     +  }
      
        tp = gmtime (&date);
        if (NULL == tp) {
    +@@ src/chage.c: static void print_date (time_t date)
    +   }
    + }
    + 
    ++
    + /*
    +  * list_fields - display the current values of the expiration fields
    +  *
     @@ src/chage.c: static void print_date (time_t date)
       */
      static void list_fields (void)
    @@ src/chage.c: static void print_date (time_t date)
        } else {
     -          changed = lstchgdate * DAY;
     -          print_date (changed);
    -+          print_day_as_date (lstchgdate);
    ++          print_day_as_date(lstchgdate);
        }
      
        /*
    @@ src/chage.c: static void list_fields (void)
                   || (maxdays >= 10000)
                   || (maxdays < 0)
     -             || ((LONG_MAX - changed) / DAY < maxdays)) {
    -+             || (LONG_MAX - lstchgdate < maxdays)) {
    ++             || (LONG_MAX - lstchgdate < maxdays))
    ++  {
                (void) puts (_("never"));
        } else {
     -          expires = changed + maxdays * DAY;
     -          print_date (expires);
    -+          print_day_as_date (lstchgdate + maxdays);
    ++          print_day_as_date(lstchgdate + maxdays);
        }
      
        /*
    @@ src/chage.c: static void list_fields (void)
     -             || (maxdays > LONG_MAX - inactdays)
     -             || ((LONG_MAX - changed) / DAY < maxdays + inactdays)) {
     +             || (LONG_MAX - inactdays < maxdays)
    -+             || (LONG_MAX - lstchgdate < maxdays + inactdays)) {
    ++             || (LONG_MAX - lstchgdate < maxdays + inactdays))
    ++  {
                (void) puts (_("never"));
        } else {
     -          expires = changed + (maxdays + inactdays) * DAY;
     -          print_date (expires);
    -+          print_day_as_date (lstchgdate + maxdays + inactdays);
    ++          print_day_as_date(lstchgdate + maxdays + inactdays);
        }
      
        /*
    @@ src/chage.c: static void list_fields (void)
     -          expires = expdate * DAY;
     -          print_date (expires);
     -  }
    -+  print_day_as_date (expdate);
    ++  print_day_as_date(expdate);
      
        /*
         * Start with the easy numbers - the number of days before the
6:  8017c0f1 < -:  -------- lib/: add long overflow checks
-:  -------- > 2:  dc65aae7 lib/: Saturate addition to avoid overflow
```

@stoeckmann 

Please sign the patches, if you don't mind.

alejandro-colomar:
v2b (by @alejandro-colomar) changes:

-  Add missing include

```
$ git range-diff shadow/master ts/scale scale 
1:  7064949f = 1:  7064949f src/chage.c: Unify long overflow checks in print_day_as_date()
2:  dc65aae7 ! 2:  431c1370 lib/: Saturate addition to avoid overflow
    @@ lib/age.c: void agecheck (/*@null@*/const struct spwd *sp)
                        (void) printf (_("Your password will expire in %ld days.\n"),
     
      ## lib/isexpired.c ##
    +@@
    + #include <config.h>
    + 
    + #include <sys/types.h>
    +-#include "prototypes.h"
    +-#include "defines.h"
    + #include <pwd.h>
    + #include <time.h>
    + 
    ++#include "adds.h"
    ++#include "defines.h"
    ++#include "prototypes.h"
    ++
    + #ident "$Id$"
    + 
    + 
     @@
       */
      int isexpired (const struct passwd *pw, /*@null@*/const struct spwd *sp)
```

src/chage.c
```diff
@@ -231,10 +231,20 @@ static int new_fields (void)
...
 {
 	struct tm *tp;
 	char buf[80];
+	time_t date;
+
+	if (day < 0) {
+		puts(_("never"));
+		return;
+	}
+	if (__builtin_mul_overflow((time_t) day, DAY, &date)) {
```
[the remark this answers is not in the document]
alejandro-colomar: Fixed it myself.

alejandro-colomar:
v3 changes:

-  Add DAY_TO_STR, and replace date_to_str() by day_to_str().

```
$ git range-diff shadow/master ts/scale scale 
1:  7064949f = 1:  7064949f src/chage.c: Unify long overflow checks in print_day_as_date()
2:  431c1370 = 2:  431c1370 lib/: Saturate addition to avoid overflow
-:  -------- > 3:  b7d41838 lib/, src/, po/: date_to_str(): Move function to header, and make inline
-:  -------- > 4:  4dcd3740 lib/day_to_str.h: DAY_TO_STR(): Add macro
-:  -------- > 5:  0c98bcc7 src/: Use DAY_TO_STR() instead of its pattern
-:  -------- > 6:  0a6115f1 src/passwd.c: print_status(): Fix typo (bogus use of the comma operator)
-:  -------- > 7:  88fa4850 lib/day_to_str.[ch]: day_to_str(): Accept a day instead of a date, and rename function
```

alejandro-colomar:
v3b changes:

-  tfix

```
$ git range-diff shadow/master ts/scale scale 
1:  7064949f = 1:  7064949f src/chage.c: Unify long overflow checks in print_day_as_date()
2:  431c1370 = 2:  431c1370 lib/: Saturate addition to avoid overflow
3:  b7d41838 = 3:  b7d41838 lib/, src/, po/: date_to_str(): Move function to header, and make inline
4:  4dcd3740 = 4:  4dcd3740 lib/day_to_str.h: DAY_TO_STR(): Add macro
5:  0c98bcc7 = 5:  0c98bcc7 src/: Use DAY_TO_STR() instead of its pattern
6:  0a6115f1 = 6:  0a6115f1 src/passwd.c: print_status(): Fix typo (bogus use of the comma operator)
7:  88fa4850 ! 7:  dc76b4e2 lib/day_to_str.[ch]: day_to_str(): Accept a day instead of a date, and rename function
    @@ lib/day_to_str.h
      
      
     -#define DAY_TO_STR(str, day)   date_to_str(NITEMS(str), str, date * DAY)
    -+#define DAY_TO_STR(str, day)   day_to_str(NITEMS(str), str, date)
    ++#define DAY_TO_STR(str, day)   day_to_str(NITEMS(str), str, day)
      
      
     -inline void date_to_str(size_t size, char buf[size], long date);
```

alejandro-colomar:
v4 changes:

-  Use gmtime_r(3) to silence a CodeQL warning

```
$ git range-diff shadow/master ts/scale scale 
1:  7064949f = 1:  7064949f src/chage.c: Unify long overflow checks in print_day_as_date()
2:  431c1370 = 2:  431c1370 lib/: Saturate addition to avoid overflow
3:  b7d41838 = 3:  b7d41838 lib/, src/, po/: date_to_str(): Move function to header, and make inline
4:  728c4439 = 4:  728c4439 lib/day_to_str.h: DAY_TO_STR(): Add macro
5:  0db7c279 = 5:  0db7c279 src/: Use DAY_TO_STR() instead of its pattern
6:  d5d05811 = 6:  d5d05811 src/passwd.c: print_status(): Fix typo (bogus use of the comma operator)
7:  b762fa5c = 7:  b762fa5c lib/day_to_str.[ch]: day_to_str(): Accept a day instead of a date, and rename function
-:  -------- > 8:  cdf2b18d lib/, src/: Call gmtime_r(3) instead of gmtime(3)
```

alejandro-colomar:
v4b changes:

-  tfix

```
$ git range-diff shadow/master ts/scale scale 
1:  7064949f = 1:  7064949f src/chage.c: Unify long overflow checks in print_day_as_date()
2:  431c1370 = 2:  431c1370 lib/: Saturate addition to avoid overflow
3:  b7d41838 = 3:  b7d41838 lib/, src/, po/: date_to_str(): Move function to header, and make inline
4:  728c4439 = 4:  728c4439 lib/day_to_str.h: DAY_TO_STR(): Add macro
5:  0db7c279 = 5:  0db7c279 src/: Use DAY_TO_STR() instead of its pattern
6:  d5d05811 = 6:  d5d05811 src/passwd.c: print_status(): Fix typo (bogus use of the comma operator)
7:  b762fa5c = 7:  b762fa5c lib/day_to_str.[ch]: day_to_str(): Accept a day instead of a date, and rename function
8:  cdf2b18d ! 8:  5eca2425 lib/, src/: Call gmtime_r(3) instead of gmtime(3)
    @@ lib/day_to_str.h: inline void day_to_str(size_t size, char buf[size], long day);
     -  time_t           date;
     -  const struct tm  *tm;
     +  time_t     date;
    -+  struct tm  *tm;
    ++  struct tm  tm;
      
        if (day < 0) {
                strtcpy(buf, "never", size);
```

alejandro-colomar:
v4c changes:

-  Rebase to master

```
$ git range-diff master..ts/scale shadow/master..scale 
1:  7064949f = 1:  b87b00b8 src/chage.c: Unify long overflow checks in print_day_as_date()
2:  431c1370 = 2:  2e9c342e lib/: Saturate addition to avoid overflow
3:  b7d41838 = 3:  84ee257f lib/, src/, po/: date_to_str(): Move function to header, and make inline
4:  728c4439 = 4:  b35c17c4 lib/day_to_str.h: DAY_TO_STR(): Add macro
5:  0db7c279 = 5:  f618c183 src/: Use DAY_TO_STR() instead of its pattern
6:  d5d05811 = 6:  7a774808 src/passwd.c: print_status(): Fix typo (bogus use of the comma operator)
7:  b762fa5c = 7:  621d80db lib/day_to_str.[ch]: day_to_str(): Accept a day instead of a date, and rename function
8:  5eca2425 = 8:  3ddd2772 lib/, src/: Call gmtime_r(3) instead of gmtime(3)
```

alejandro-colomar:
v4e changes:

-  Add missing include

```
$ git range-diff shadow/master ts/scale scale 
1:  741fc096 = 1:  741fc096 src/chage.c: Unify long overflow checks in print_day_as_date()
2:  d5d9194e = 2:  d5d9194e lib/: Saturate addition to avoid overflow
3:  20bb258f = 3:  20bb258f lib/, src/, po/: date_to_str(): Move function to header, and make inline
4:  8461594e ! 4:  46362e76 lib/day_to_str.h: DAY_TO_STR(): Add macro
    @@ lib/day_to_str.h
      #include <time.h>
      
     +#include "defines.h"
    ++#include "sizeof.h"
      #include "string/strtcpy.h"
      
      
5:  b1f47ef6 = 5:  fd9ca675 src/: Use DAY_TO_STR() instead of its pattern
6:  3379b384 = 6:  ce086de9 src/passwd.c: print_status(): Fix typo (bogus use of the comma operator)
7:  fd3908fa = 7:  1a73b4d1 lib/day_to_str.[ch]: day_to_str(): Accept a day instead of a date, and rename function
8:  5f46fdf5 = 8:  d7b2729e lib/, src/: Call gmtime_r(3) instead of gmtime(3)
```

hallyn (approved):

PR #875, opened 2023-12-17
Reject negative numbers in strtoul(3)

alejandro-colomar:

PR #877, opened 2023-12-19
lib/getdef.c: Adjust getdef_ boundaries

stoeckmann:
Negative values in login.defs never make sense (except someone wants to forcefully disable a feature with `-1` instead of commenting out a line or start breaking things by using even smaller values, which will be fixed in later PRs).

As a side note, util-linux does not support negative values either. See https://github.com/util-linux/util-linux/blob/master/lib/logindefs.c

### DIFF
--- a/lib/getdef.c
+++ b/lib/getdef.c
@@ -247,7 +247,7 @@ int getdef_num (const char *item, int dflt)
 
 	if (   (getlong (d->value, &val) == 0)
 	    || (val > INT_MAX)
-	    || (val < INT_MIN)) {
+	    || (val < -1)) {
 		fprintf (shadow_logfd,
 		         _("configuration error - cannot parse %s value: '%s'"),
 		         item, d->value);
@@ -315,7 +315,8 @@ long getdef_long (const char *item, long dflt)
 		return dflt;
 	}
 
-	if (getlong (d->value, &val) == 0) {
+	if (   (getlong (d->value, &val) == 0)
+	    || (val < -1)) {
 		fprintf (shadow_logfd,
 		         _("configuration error - cannot parse %s value: '%s'"),
 		         item, d->value);
--- a/man/login.defs.d/PASS_WARN_AGE.xml
+++ b/man/login.defs.d/PASS_WARN_AGE.xml
@@ -9,8 +9,8 @@
   <listitem>
     <para>
       The number of days warning given before a password expires. A zero
-      means warning is given only upon the day of expiration, a negative
-      value means no warning is given. If not specified, no warning will
+      means warning is given only upon the day of expiration, a value of
+      -1 means no warning is given. If not specified, no warning will
       be provided.
     </para>
   </listitem>


lib/getdef.c
```diff
@@ -282,7 +282,7 @@ unsigned int getdef_unum (const char *item, unsigned int dflt)
 
 	if (   (getlong (d->value, &val) == 0)
 	    || (val < 0)
-	    || (val > INT_MAX)) {
+	    || (val > UINT_MAX)) {
```
alejandro-colomar, line 285: If the max is raised to UINT_MAX, then maybe we should call getulong()?

BTW, that would require <https://github.com/shadow-maint/shadow/pull/875>.

stoeckmann: True, that would be an issue on 32 bit systems. Dropped the commit.

alejandro-colomar:
The manual page seems to say, IIUC, that negative values are acceptable in one case:

```
     PASS_WARN_AGE (number)
         The number of days warning given before a password expires. A zero
         means warning is given only upon the day of expiration, a negative
         value means no warning is given. If not specified, no warning will be
         provided.
```

stoeckmann:
Yes, -1 is used as a default if it's not specified in login.defs. We can add -1 as a valid value.

stoeckmann:
> Is that included in
> 
> > ...  which will be fixed in later PRs)

No, the problem arises in https://github.com/shadow-maint/shadow/blob/master/lib/isexpired.c#L86 if there is indeed a "negative value" set which is not -1. Suddenly, the code sometimes checks for >= 0 and sometimes for != -1 but would happily accept -2, -3 or INT_MIN etc.

alejandro-colomar:
> > Is that included in
> > > ...  which will be fixed in later PRs)
> 
> No, the problem arises in https://github.com/shadow-maint/shadow/blob/master/lib/isexpired.c#L86 if there is indeed a "negative value" set which is not -1. Suddenly, the code sometimes checks for >= 0 and sometimes for != -1 but would happily accept -2, -3 or INT_MIN etc.

Then I guess we need [-1, INT_MAX].

alejandro-colomar (approved):

stoeckmann:
Clarified documentation that only -1 really works.

alejandro-colomar (approved):

hallyn:
Probably a good change, but it's likely to break some installations.

I'll merge it, but am wondering whether/how we should warn users.

stoeckmann:
> I'll merge it, but am wondering whether/how we should warn users.

If you are concerned, we could also enforce that every negative value is automatically returned as -1.

hallyn:
> > I'll merge it, but am wondering whether/how we should warn users.
> 
> If you are concerned, we could also enforce that every negative value is automatically returned as -1.

Magically massaging values read from config files risks confusing and antagonizing users ("but my files clearly say -3), but it might be the best thing here.

alejandro-colomar:
I prefer shouting users that their value is bogus, rather than massaging.  Unless users complain that their installations break with this change, in which case we could offer a patch for massaging bogus values in the stable branch, as a transition.

So, adding something to the release news should probably be the way, and maybe also add a shouty error message.

PR #878, opened 2023-12-19
man/: Support compiling in build directory

stoeckmann:
Having a dedicated build directory breaks manual page creation.

Failing command sequence:
```
mkdir build; cd build
../configure --enable-man
make -C man
```

### DIFF
--- a/man/generate_mans.mak
+++ b/man/generate_mans.mak
@@ -56,7 +56,7 @@ man1/% man3/% man5/% man8/%: %.xml-config Makefile config.xml
 	            --stringparam vendordir "$(VENDORDIR)" \
 	            --param "man.output.in.separate.dir" "1" \
 	            --path "$(srcdir)/login.defs.d" \
-	            -nonet $(top_builddir)/man/shadow-man.xsl $<
+	            -nonet $(top_srcdir)/man/shadow-man.xsl $<
 
 clean-local:
 	rm -rf man1 man3 man5 man8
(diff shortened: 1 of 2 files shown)


ikerexxe:
I don't understand the problem. Can you give additional details?

stoeckmann:
> I don't understand the problem. Can you give additional details?

I have multiple build directories, e.g. build32 and build64 located in shadow's source directory, so I can easily perform tests in different setups. Directory layout is like:

```
shadow
shadow/build32
shadow/build64
shadow/configure
shadow/lib
shadow/src
...
```
Only created files are stored in these build directories. Source files still reside in shadow directory structure but never in these build directories.

The advantage of such a setup is that I can easily use `git checkout` to switch between branches and both build directories will use the correct source files.

Now, the problem is that `$(top_builddir)/man/shadow-man.xsl` points into the build directory, but the shadow-man.xsl is a source file, i.e. it will never end up there. Using `$(top_srcdir)` instead fixes the problem, because it will ultimately point to something like `shadow/man/shadow-man.xsl` instead of `shadow/build64/man/shadow-man.xsl`.

The same is true for relative paths, which are resolved within the build directories. Using `$(srcdir)` in front of a relative path fixes the problem because `$(srcdir)` points to the directory in which the current source file is located.

ikerexxe (approved):
Thanks for clarifying.

The change looks good so I'm merging it.

alejandro-colomar (approved):

PR #879, opened 2023-12-19
man/: add BCRYPT and YESCRYPT information

stoeckmann:
The BCRYPT and YESCRYPT relevant items should be described in manual pages.

### DIFF
--- a/man/chgpasswd.8.xml
+++ b/man/chgpasswd.8.xml
@@ -6,10 +6,12 @@
 -->
 <!DOCTYPE refentry PUBLIC "-//OASIS//DTD DocBook V4.5//EN"
   "http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd" [
+<!ENTITY BCRYPT_MIN_ROUNDS     SYSTEM "login.defs.d/BCRYPT_MIN_ROUNDS.xml">
 <!ENTITY ENCRYPT_METHOD        SYSTEM "login.defs.d/ENCRYPT_METHOD.xml">
 <!ENTITY MAX_MEMBERS_PER_GROUP SYSTEM "login.defs.d/MAX_MEMBERS_PER_GROUP.xml">
 <!ENTITY MD5_CRYPT_ENAB        SYSTEM "login.defs.d/MD5_CRYPT_ENAB.xml">
 <!ENTITY SHA_CRYPT_MIN_ROUNDS  SYSTEM "login.defs.d/SHA_CRYPT_MIN_ROUNDS.xml">
+<!ENTITY YESCRYPT_COST_FACTOR  SYSTEM "login.defs.d/YESCRYPT_COST_FACTOR.xml">
 <!-- SHADOW-CONFIG-HERE -->
 ]>
 
@@ -88,12 +90,16 @@
 	<term><option>-c</option>, <option>--crypt-method</option></term>
 	<listitem>
 	  <para>Use the specified method to encrypt the passwords.</para>
-	  <para condition="no_sha_crypt">
-	    The available methods are DES, MD5, and NONE.
-	  </para>
-	  <para condition="sha_crypt">
-	    The available methods are DES, MD5, NONE, and SHA256 or SHA512
-	    if your libc support these methods.
+	  <para>
+	    The available methods are <phrase condition="bcrypt">
+	    <replaceable>BCRYPT</replaceable>,</phrase>
+	    <replaceable>DES</replaceable>,
+	    <replaceable>MD5</replaceable><phrase condition="sha_crypt">,
+	    <replaceable>SHA256</replaceable>,
+	    <replaceable>SHA512</replaceable></phrase><phrase condition="yescrypt">,
+	    <replaceable>YESCRYPT</replaceable></phrase> and
+	    <replaceable>NONE</replaceable>
+	    if your libc supports these methods.
 	  </para>
 	</listitem>
       </varlistentry>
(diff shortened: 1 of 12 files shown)


man/chgpasswd.8.xml
```diff
@@ -88,12 +90,16 @@
...
 	  <para>Use the specified method to encrypt the passwords.</para>
-	  <para condition="no_sha_crypt">
-	    The available methods are DES, MD5, and NONE.
-	  </para>
-	  <para condition="sha_crypt">
-	    The available methods are DES, MD5, NONE, and SHA256 or SHA512
-	    if your libc support these methods.
+	  <para>
+	    The available methods are <phrase condition="bcrypt">
+	    <replaceable>BCRYPT</replaceable>,</phrase>
```
ikerexxe, line 95: Can you move the condition beginning and end to the same line?

man/chgpasswd.8.xml
```diff
@@ -88,12 +90,16 @@
...
+	    <replaceable>DES</replaceable>,
+	    <replaceable>MD5</replaceable><phrase condition="sha_crypt">,
+	    <replaceable>SHA256</replaceable>,
```
ikerexxe, line 98: Similar here. Can you move the condition beginning to the same line as SHA256?

man/chgpasswd.8.xml
```diff
@@ -88,12 +90,16 @@
...
+	    <replaceable>SHA512</replaceable></phrase><phrase condition="yescrypt">,
+	    <replaceable>YESCRYPT</replaceable></phrase> and
```
ikerexxe, line 100: Same here

man/chgpasswd.8.xml
```diff
@@ -131,29 +137,48 @@
...
+	    <replaceable>SHA256</replaceable>
+	    <replaceable>SHA512</replaceable></phrase>
+	    <phrase condition="yescrypt">
+	    <replaceable>YESCRYPT</replaceable></phrase>
 	  </para>
-	  <para>
-	    A minimal value of 1000 and a maximal value of 999,999,999
-	    will be enforced.
+	  <para condition="bcrypt">
+	    By default, the number of rounds for BCRYPT is defined by the
```
ikerexxe, line 157: I don't think you need `By default,` here. The default is explained below and is 13.

man/chgpasswd.8.xml
```diff
@@ -131,29 +137,48 @@
...
-	    You can only use this option with the SHA256 or SHA512
-	    crypt method.
+	  <para condition="bcrypt">
+	    A minimal value of 4 and a maximal value of 31
+	    will be enforced for BCRYPT. The default is 13.
 	  </para>
-	  <para>
-	    By default, the number of rounds is defined by the
+	  <para condition="sha_crypt">
+	    By default, the number of rounds for SHA256 or SHA512 is defined by the
```
ikerexxe, line 166: Same here

man/chpasswd.8.xml
```diff
@@ -115,12 +117,16 @@
 	</term>
 	<listitem>
 	  <para>Use the specified method to encrypt the passwords.</para>
-	  <para condition="no_sha_crypt">
-	    The available methods are DES, MD5, and NONE.
-	  </para>
-	  <para condition="sha_crypt">
-	    The available methods are DES, MD5, NONE, and SHA256 or SHA512
-	    if your libc support these methods.
+	  <para>
```
ikerexxe, line 120: I think that all my previous comments also apply to this file. Please review it.

man/login.defs.d/ENCRYPT_METHOD.xml
```diff
@@ -10,11 +10,13 @@
       passwords (if no algorithm are specified on the command line).
     </para>
     <para>
-      It can take one of these values:
+      It can take one of these values: <phrase condition="bcrypt">
```
ikerexxe, line 13: Move condition to the next line

man/login.defs.d/ENCRYPT_METHOD.xml
```diff
@@ -10,11 +10,13 @@
...
+      <replaceable>BCRYPT</replaceable>,</phrase>
       <replaceable>DES</replaceable> (default),
       <replaceable>MD5</replaceable><phrase condition="sha_crypt">,
       <replaceable>SHA256</replaceable>,
-      <replaceable>SHA512</replaceable></phrase>.
+      <replaceable>SHA512</replaceable></phrase><phrase condition="yescrypt">,
```
ikerexxe, line 18: Move condition to the next line

ikerexxe (changes requested):
Comments inline

man/chgpasswd.8.xml
stoeckmann: If I do that, there will be a blank between `MD5` and the following comma, i.e. `MD5 ,` instead of `MD5,`. It is the same logic used as in ENCRYPT_METHOD.xml (see https://github.com/shadow-maint/shadow/blob/master/man/login.defs.d/ENCRYPT_METHOD.xml#L15).

So this comment applies to the other remarks regarding newlines as well.

man/chgpasswd.8.xml
stoeckmann: There are two different defaults explained. Just as it has been before for the whole sha_crypt case. By default in this case means that these fields are parsed from login.defs if no `-s` option is applied.

The other default states which value will be used if no such fields exist in login.defs AND no `-s` option is supplied.

man/chgpasswd.8.xml
ikerexxe: Ok, now I see why you decided to do it this way. Then, disregard this and other comments like this one.

man/chgpasswd.8.xml
ikerexxe: Yes, I know that but the text that you wrote may not be understood like that. Perhaps changing to `The default number of rounds is 13`?

man/chgpasswd.8.xml
stoeckmann: Adjusted now.

hallyn (approved):

PR #880, opened 2023-12-20
lib/defines.h: Use 'time_t' for DAY

alejandro-colomar:
Special care has to be taken for 32 bit systems with a 64 bit time_t, since their long data type is still 32 bit.

Reported-by: @stoeckmann

### DIFF
--- a/lib/defines.h
+++ b/lib/defines.h
@@ -144,7 +144,7 @@
 
 /* Solaris defines this in shadow.h */
 #ifndef DAY
-#define DAY (24L*3600L)
+#define DAY  ((time_t) 24 * 3600)
 #endif
 
 #define WEEK (7*DAY)


alejandro-colomar:
Or maybe we should `#define DAY ((time_t) 24 * (time_t) 3600)`.  This would be more appropriate, since it's seconds, and seconds are expressed in `time_t` units in C.

What do you prefer?

stoeckmann:
Let's use time_t because it's sometimes 32 bit and sometimes 64 bit. It won't hurt to go with long long, since on systems where shadow would be used, it's very very likely 64 bit and large enough.

But time_t sounds more reasonable. Wouldn't take two casts but it doesn't hurt either.

alejandro-colomar:
v2:

-  Use time_t, and use it only once

```
$ git range-diff shadow/master gh/DAY DAY 
1:  a3f8f445 ! 1:  34abbdf9 lib/defines.h: Use 'long long' for DAY
    @@ Metadata
     
      ## Commit message ##
    -    lib/defines.h: Use 'long long' for DAY
    +    lib/defines.h: Use appropriate types for DAY
     
         Special care has to be taken for 32 bit systems with a 64 bit time_t,
         since their long data type is still 32 bit.
     
    +    Since this macro expresses a number of seconds, and seconds are in units
    +    of 'time_t' in C, the appropriate type for the multiplication is
    +    'time_t'.
    +
         Reported-by: Tobias Stoeckmann <tobias@stoeckmann.org>
     
    @@ lib/defines.h
      /* Solaris defines this in shadow.h */
      #ifndef DAY
     -#define DAY (24L*3600L)
    -+#define DAY (24LL*3600LL)
    ++#define DAY  ((time_t) 24 * 3600)
      #endif
      
      #define WEEK (7*DAY)
```

alejandro-colomar:
> But time_t sounds more reasonable. Wouldn't take two casts but it doesn't hurt either.

You're right.  Originally, I was worried that default promotions would do weird things, but I checked and no, so one cast is enough.

alejandro-colomar:
v2b:

-  Slight reword of commit message

```
$ git range-diff shadow/master gh/DAY DAY 
1:  34abbdf9 ! 1:  82fa3e1e lib/defines.h: Use appropriate types for DAY
    @@ Metadata
     
      ## Commit message ##
    -    lib/defines.h: Use appropriate types for DAY
    +    lib/defines.h: Use 'time_t' for DAY
     
         Special care has to be taken for 32 bit systems with a 64 bit time_t,
         since their long data type is still 32 bit.
```

hallyn (approved):

PR #881, opened 2023-12-20
man/: Remove more login.defs items if PAM is in use

stoeckmann:
CONSOLE_GROUPS and ENV_HZ are not used if PAM is enabled. Remove them (and even whole paragraphs since they would be empty now) if PAM is enabled.

### DIFF
--- a/man/login.1.xml
+++ b/man/login.1.xml
@@ -277,7 +277,7 @@
       &CONSOLE;
       &CONSOLE_GROUPS;
       &DEFAULT_HOME;
-      <phrase condition="no_pam">&ENV_HZ;</phrase>
+      &ENV_HZ;
       <phrase>&ENV_PATH;</phrase>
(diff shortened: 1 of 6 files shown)


man/login.1.xml
```diff
@@ -277,7 +277,7 @@
       &CONSOLE;
       &CONSOLE_GROUPS;
       &DEFAULT_HOME;
-      <phrase condition="no_pam">&ENV_HZ;</phrase>
```
hallyn, line 280: what's the reason why you remove the condition here?

stoeckmann: The condition is now part of ENV_HZ.xml, i.e. automatically included. The same is true for the other PAM-only XML-files in login.conf.d.

hallyn: Oh, I see - thanks.

PR #884, opened 2023-12-27
lib/getdef.c: Introduce getlong_normalized

stoeckmann:
Stay compatible with current configuration files which allow all negative values, but convert them to -1 on the fly.

Print a warning so these values can be modified by an administrator for a future release which won't have this support anymore.

As discussed in https://github.com/shadow-maint/shadow/pull/877#issuecomment-1870427972

### DIFF
--- a/lib/getdef.c
+++ b/lib/getdef.c
@@ -181,6 +181,26 @@ static void def_load (void);
 
 
 /*
+ * getlong_normalized - get long value and set all negative values to -1.
+ *
+ * Return getlong result and normalize all negative values to -1 if
+ * configuration file contains an even smaller number.
+ */
+static int getlong_normalized(const char *item, const char *str, long *val)
+{
+	int result;
+
+	result = getlong (str, val);
+	if (result && *val < -1) {
+		fprintf (shadow_logfd,
+		         _("normalizing %s value to -1: '%s'"),
+		         item, str);
+		*val = -1;
+	}
+	return result;
+}
+
+/*
  * getdef_str - get string value from table of definitions.
  *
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
"normalizing" is not something intrinsic to `long` (what is normal for a long?).  I'm not convinced by this function.  Rejecting negative values is more appropriate to do in functions that handle specific values.  I don't like this change so much.

alejandro-colomar:
A getlong() variant that handles a range (similar to strtoi(3bsd)) is something I have in mind.

Here's what I have in mind:

```c
getlong(const char *s, long *num, char **endptr, int base, long min, long max)
{
	int  status;

	*num = strtoi(s, endptr, base, min, max, &status);
	if (status != 0) {
		errno = status;
		return -1;
	}

	return 0;
}

getl(const char *s, long *num)
{
	return getlong(s, NULL, 0, LONG_MIN, LONG_MAX);
}
```

The current getlong() would be (almost) my getl() suggestion, and getlong() would be the argumentful version.

You could call `getlong(..., -1, LONG_MAX)` for rejecting negative values, and then the caller could report any errors or do whatever is more appropriate.

alejandro-colomar:
This PR is superseeded by <https://github.com/shadow-maint/shadow/pull/899>.

The commits are:

-  <https://github.com/shadow-maint/shadow/pull/899/commits/e6289782279f74a1899570f3a6391dab905aec0a>
-  <https://github.com/shadow-maint/shadow/pull/899/commits/953fff23db10c9e63631bbf84aab7629b2af85d8>

PR #902, opened 2024-01-11
lib/idmapping.c: write_mapping(): Fixx off-by-one bug

alejandro-colomar:
Link: <https://github.com/shadow-maint/shadow/commit/673c2a6f9aa6c69588f4c1be08589b8d3475a520#r136830993>
Cc: @hallyn 
Cc: @ebiederm

### DIFF
--- a/lib/idmapping.c
+++ b/lib/idmapping.c
@@ -192,7 +192,7 @@ void write_mapping(int proc_dir_fd, int ranges, const struct map_range *mappings
 	}
 #endif
 
-	bufsize = ranges * ((ULONG_DIGITS + 1) * 3);
+	bufsize = (ULONG_DIGITS + 1) * 3 * ranges + 1;
 	pos = buf = XMALLOC(bufsize, char);
 	end = buf + bufsize;
 


hallyn (approved):
Looks right.  The last missing \n might not even upset the map_file parser, but we would exit with failure here.

hallyn:
Note also that ULONG_BITS is 21, while any valid subid has max 11 digits.  So the outcome would be failure here in any case.

alejandro-colomar:
> Note also that ULONG_BITS is 21, while any valid subid has max 11 digits. So the outcome would be failure here in any case.

I didn't understand this.  Can you please clarify?

hallyn:
If you cat /proc/self/uid_map in init_user_ns, you'll see that the top uid is 4294967295.  Any value higher than 4294967295 - longer than 11 digits - will be invalid.  So any range which has the lower, range, and upper as 21 digits will be an invalid range.  We would return an error in this case because we'd be truncating the write into the buffer which, as you noted, will be 1 byte too short.  So we'll error correctly but for the wrong reasons.  There is AFAICS not a valid case which we'll error on.

alejandro-colomar:
Ahhh, heh!  Makes sense.

PR #904, opened 2024-01-12
lib/sgetgrent.c: fix null pointer dereference

ferivoz:
If reallocation fails in function list, then reset the size to 0 again. Without the reset, the next call assumes that `members` points to a memory location with reserved space.

Also use size_t instead of int for size to prevent signed integer overflows. The length of group lines is not limited.

Fixes 45c0003e53ab671c63dcd530fd9f3245d3b29e76 (4.14 release series)

Proof of Concept:
1. Prepare a group file (one long group line and a shorter one, both with a list of users)
```
$ echo -n "root:x:0:" > /tmp/uwu
$ yes , | tr -d '\n' | dd of=/tmp/uwu bs=10 count=3145728 seek=1 conv=notrunc iflag=fullblock
$ echo -e "\nbin:x:1:," >> /tmp/uwu
```
2. Run grpck with tight memory constraints
```
$ ulimit -d 102400
$ grpck /tmp/uwu
Segmentation fault (core dumped)
```

### DIFF
--- a/lib/sgetgrent.c
+++ b/lib/sgetgrent.c
@@ -37,8 +37,8 @@
 static char **list (char *s)
 {
 	static char **members = NULL;
-	static int size = 0;	/* max members + 1 */
-	int i;
+	static size_t size = 0;	/* max members + 1 */
+	size_t i;
 
 	i = 0;
 	for (;;) {
@@ -47,8 +47,10 @@ static char **list (char *s)
 		if (i >= size) {
 			size = i + 100;	/* at least: i + 1 */
 			members = REALLOCF(members, size, char *);
-			if (!members)
+			if (!members) {
+				size = 0;
 				return NULL;
+			}
 		}
 		if (!s || s[0] == '\0')
 			break;


alejandro-colomar:
```
```

Thanks!

alejandro-colomar:
Did you check if I forgot to do the same thing in any other reallocf(3) calls?

alejandro-colomar (approved):

alejandro-colomar:
> Also use size_t instead of int for size to prevent signed integer overflows. The length of group lines is not limited.

I like this change too, as size_t is more appropriate for a size.  However, can you please clarify how it prevents signed integer overflow?

Integer overflow could only happen in the line `size = i + 100;`, and it could still happen if we switch to `size_t`, I think (it will be harder, but still possible).  If that's a concern, maybe the addition should be checked for overflow with a builtin?

In the call to REALLOCF(), the number is converted to a size_t, and the multiplication should be performed with checks inside reallocarray(3), so in this call there's no overflow.

```c
alx@debian:~/src/shadow/shadow/4.14.x$ grepc REALLOCF .
./lib/alloc.h:#define REALLOCF(ptr, n, type)                                                \
({                                                                            \
	__auto_type  p_ = (ptr);                                              \
                                                                              \
	static_assert(__builtin_types_compatible_p(typeof(p_), type *), "");  \
                                                                              \
	(type *) reallocarrayf(p_, n, sizeof(type));                          \
})
alx@debian:~/src/shadow/shadow/4.14.x$ grepc reallocarrayf .
./lib/alloc.c:extern inline void *reallocarrayf(void *p, size_t nmemb, size_t size);
./lib/alloc.h:ATTR_MALLOC(free)
inline void *reallocarrayf(void *p, size_t nmemb, size_t size);
./lib/alloc.h:inline void *
reallocarrayf(void *p, size_t nmemb, size_t size)
{
	void  *q;

	q = reallocarray(p, nmemb, size);

	/* realloc(p, 0) is equivalent to free(p);  avoid double free.  */
	if (q == NULL && nmemb != 0 && size != 0)
		free(p);
	return q;
}
```

alejandro-colomar:
> Did you check if I forgot to do the same thing (set/reset static variables) after any other reallocf(3) calls?

I've checked with the following, and it seems this is the only mistake of this kind.

```sh
$ git log -G 'REALLOC[ARY]*F' -p
$ git log -G 'realloc[ary]*f' -p
```

ferivoz:
> I like this change too, as size_t is more appropriate for a size. However, can you please clarify how it prevents signed integer overflow?

The `size_t` data type is unsigned, so it cannot trigger a signed integer overflow. :)

It is not realistically possible to trigger an overflow with `size_t` with the constant 100 steps. At least on systems where shadow is supposed to run, i.e. on Linux systems. The 32 bit architectures use 1 GB of virtual address space for kernel, so only 3 GB are left. Even if it would be possible to reallocate in 100 steps up to this, it would not overflow the counter.

But let's entertain the thought for a bit for 64 bit systems. Even if all of that is possible, this huge group line would already reside twice in memory due to previous string duplications and memory copy operations. It is save to assume that this cannot overflow. Again, at least if we do not talk about embedded systems where `size_t` might be smaller than the whole address space, but these would most likely not be even POSIX systems.

ferivoz:
> Integer overflow could only happen in the line `size = i + 100;`, and it could still happen if we switch to `size_t`, I think (it will be harder, but still possible).

With sufficient amount of RAM, and my previous explanation, a very large line can fit into heap space on 64 bit systems and let this integer calculation overflow. With a negative integer, the cast to `size_t` in reallocfarray turns it huge and this will let the function call fail.

And this failure in turn leads to the null pointer dereference if a subsequent line is parsed.

alejandro-colomar:
> > I like this change too, as size_t is more appropriate for a size. However, can you please clarify how it prevents signed integer overflow?
> 
> The `size_t` data type is unsigned, so it cannot trigger a signed integer overflow. :)

:)

> 
> It is not realistically possible to trigger an overflow with `size_t` with the constant 100 steps. At least on systems where shadow is supposed to run, i.e. on Linux systems. The 32 bit architectures use 1 GB of virtual address space for kernel, so only 3 GB are left. Even if it would be possible to reallocate in 100 steps up to this, it would not overflow the counter.
> 
> But let's entertain the thought for a bit for 64 bit systems. Even if all of that is possible, this huge group line would already reside twice in memory due to previous string duplications and memory copy operations. It is save to assume that this cannot overflow. Again, at least if we do not talk about embedded systems where `size_t` might be smaller than the whole address space, but these would most likely not be even POSIX systems.

That's convincing.  Thanks!

alejandro-colomar:
@ikerexxe When you apply this one, I'll cherry-pick it to the stable branch.

alejandro-colomar:
@ferivoz Small typo in commit message:

`s/fucntion/function/`

ferivoz:
> `s/fucntion/function/`

Fixed, thanks!

alejandro-colomar:
Maybe it would be good to include the proof of concept in the commit message.  That's an interesting thing to keep there, I think.

hallyn (approved):

hallyn:
Thanks.

PR #908, opened 2024-01-17
src/sulogin.c: Free previously allocated memory

ferivoz:
The sulogin program calls pw_entry in a loop while incorrect root passwords are entered.

Free the previously allocated memory to avoid memory exhaustion.

Proof of Concept (compile with -fsanitize=address):

Run `sulogin` and enter wrong root passwords multiple times. Then press CTRL+D. You can see as many direct leaks as wrong passwords have been entered.

### DIFF
--- a/src/sulogin.c
+++ b/src/sulogin.c
@@ -199,22 +199,29 @@ pw_entry(const char *name, struct passwd *pwent)
 	struct passwd  *passwd;
 
 	if (!(passwd = getpwnam(name))) {  /* local, no need for xgetpwnam */
+		free(pwent->pw_name);
 		pwent->pw_name = NULL;
 		return;
 	}
 
+	free(pwent->pw_name);
 	pwent->pw_name = xstrdup(passwd->pw_name);
 	pwent->pw_uid = passwd->pw_uid;
 	pwent->pw_gid = passwd->pw_gid;
+	free(pwent->pw_gecos);
 	pwent->pw_gecos = xstrdup(passwd->pw_gecos);
+	free(pwent->pw_dir);
 	pwent->pw_dir = xstrdup(passwd->pw_dir);
+	free(pwent->pw_shell);
 	pwent->pw_shell = xstrdup(passwd->pw_shell);
 #if !defined(AUTOSHADOW)
 	/* local, no need for xgetspnam */
 	if ((spwd = getspnam(name))) {
+		free(pwent->pw_passwd);
 		pwent->pw_passwd = xstrdup(spwd->sp_pwdp);
 		return;
 	}
 #endif
+	free(pwent->pw_passwd);
 	pwent->pw_passwd = xstrdup(passwd->pw_passwd);
 }


ferivoz:
It would be nice if `pw_entry` could be moved into `sulogin.c` because no other program uses it. Also, `sulogin.c` could be refactored to remove its global variables.

Is there interest in this for another pull request?

alejandro-colomar:
```
```

alejandro-colomar (approved):

lib/entry.c
```diff
@@ -29,19 +29,25 @@ void pw_entry (const char *name, struct passwd *pwent)
 		pwent->pw_name = NULL;
```
alejandro-colomar, line 29: Isn't this a leak too?

hallyn:
> It would be nice if `pw_entry` could be moved into `sulogin.c` because no other program uses it.

That would be good to do if we're doing this patch, because here we are depending
on the pwent being zeroed out at the start, else we'll start freeing random values.
So best not to make it seem like "just anyone" can use it.

Alternatively, we could introduce a pw_init() function which zeroes the pwent at the
top of the loop.

> Also, `sulogin.c` could be refactored to remove its global variables.

This patch is depending on pwent being in BSS so it starts out zeroed.  So don't just
move its definition into main or into the while loop without accounting for that.

Anyway, I'll note that while this 'out of memory' is not a security issue (since this
is only in sulogin, not in general library code), risking freeing uninitialized memory
is.  So let's be careful with this.  We've had unexpected side effects in the past from
over-zealous attempts at freeing memory in programs where it just didn't matter.

lib/entry.c
alejandro-colomar: After refactoring sulogic.c and pw_entry(), I see that the caller immediately exits after that, so it's a short-lived leak, but still a leak.

I'm thinking that I'll resolve this by refactoring pw_entry() to return a -1 on error, so you can just ignore this.  Or you can refactor it yourself if you want.  :)

ferivoz: Yes, it is a leak. And it is short-lived, but still a leak. I have adjusted my commit and added you as a co-author. Thank you for pointing it out!

ferivoz:
Rebased on current master. Thank you for refactoring the mentioned parts @alejandro-colomar!

lib/entry.c
alejandro-colomar: Thanks!

alejandro-colomar:
Thank you!

```
```

alejandro-colomar:
Oh, you should probably keep the Co-authored-by (or Co-developed-by, as you prefer) line next to the signed-off-by.  :)

<https://www.kernel.org/doc/html/latest/process/submitting-patches.html?highlight=signed%20off#when-to-use-acked-by-cc-and-co-developed-by>

hallyn (approved):

hallyn:
It would be nice, if anyone feels included to do more with this code, to replace void pw_entry() with int pw_entry() which returns error if we run out of memory.

Furthermore, given doc/HOWTO note on AUTOSHADOW, which is referenced in this hunk, it seems like we ought to also drop that, and just refuse to compile if AUTOSHADOW is set.

alejandro-colomar:
> It would be nice, if anyone feels included to do more with this code, to replace void pw_entry() with int pw_entry() which returns error if we run out of memory.

Yup, I had this in my plans.

> Furthermore, given doc/HOWTO note on AUTOSHADOW, which is referenced in this hunk, it seems like we ought to also drop that, and just refuse to compile if AUTOSHADOW is set.

I don't even know what AUTOSHADOW is.  I've been seeing that thing and trying to ignore it.  :D
I'll have a look.

PR #913, opened 2024-01-18
src/sulogin.c: Simpler password handling

ferivoz:
The `pass` array can be removed. It is possible to directly work with the result of `agetpass`, which in turn removes the need to prepare and clean up the array.

And there was one more external variable which was not in use.

### DIFF
--- a/src/sulogin.c
+++ b/src/sulogin.c
@@ -64,6 +64,7 @@ main(int argc, char **argv)
 	char           **envp = environ;
 	TERMIO         termio;
 	struct passwd  pwent = {};
+	bool           done;
 #ifndef USE_PAM
 	const char     *env;
 #endif
@@ -133,7 +134,7 @@ main(int argc, char **argv)
 	(void) signal (SIGALRM, catch_signals);	/* exit if the timer expires */
 	(void) alarm (ALARM);		/* only wait so long ... */
 
-	while (true) {		/* repeatedly get login/password pairs */
+	do {			/* repeatedly get login/password pairs */
 		char *cp;
 		if (pw_entry("root", &pwent) == -1) {	/* get entry from password file */
 			/*
@@ -170,13 +171,13 @@ main(int argc, char **argv)
 		STRTCPY(pass, cp);
 		erase_pass (cp);
 
-		if (valid (pass, &pwent)) {	/* check encrypted passwords ... */
-			break;	/* ... encrypted passwords matched */
+		done = valid(pass, &pwent);
+		if (!done) {	/* check encrypted passwords ... */
+			/* ... encrypted passwords did not match */
+			sleep (2);
+			(void) puts (_("Login incorrect"));
 		}
-
-		sleep (2);
-		(void) puts (_("Login incorrect"));
-	}
+	} while (!done);
 	MEMZERO(pass);
 	(void) alarm (0);
 	(void) signal (SIGALRM, SIG_DFL);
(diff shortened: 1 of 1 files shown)


src/sulogin.c
```diff
@@ -133,8 +131,8 @@ main(int argc, char **argv)
 	(void) signal (SIGALRM, catch_signals);	/* exit if the timer expires */
 	(void) alarm (ALARM);		/* only wait so long ... */
 
-	while (true) {		/* repeatedly get login/password pairs */
-		char *cp;
+	while (!done) {		/* repeatedly get login/password pairs */
```
alejandro-colomar, line 133: This would read better as a do-while loop, and that would itself allow removing the initialization of `bool done = false;`

src/sulogin.c
```diff
@@ -160,25 +158,25 @@ main(int argc, char **argv)
...
+		done = valid(pass, &pwent);
+		if (!done) {
+			/* ... encrypted passwords did not match */
+			sleep (2);
+			(void) puts (_("Login incorrect"));
 		}
 
-		sleep (2);
-		(void) puts (_("Login incorrect"));
+		erase_pass (pass);
```
alejandro-colomar, line 177: This would be more appropriate before the conditional from above.  Otherwise we have the password in memory for 2 seconds unnecessarily.

src/sulogin.c
alejandro-colomar: Also, you could maybe refactor to a do-while loop in a separate commit before the one that removes the `pass` array.

ferivoz: I have applied both suggestions.

src/sulogin.c
ferivoz: Good point, thanks! Applied.

alejandro-colomar:
Thanks!

```
```

alejandro-colomar (approved):

hallyn (approved):

alejandro-colomar:
Thanks @hallyn !  :)

alejandro-colomar:
@ferivoz Sorry, I introduced some conflict, due to changing the way pw_entry() returns.  Can you please rebase?  (I would do a manual merge to resolve the conflict, but I can't :| )

ferivoz:
When became shadow so overly bureaucratic and complex? Especially after pushing a change into this before this PR was merged, while knowing exactly that it will create conflicts?

It's up to you how to handle this project. Since this change is not that important, we can just close the pull request.

alejandro-colomar:
Hi @ferivoz ,

> Especially after pushing a change into this before this PR was merged, while knowing exactly that it will create conflicts?

Honestly, I didn't know it.  If I knew, I wouldn't have done it, and would have waited for this to be merged before doing the other change.  I just forgot that these two touch code that is close to each other.

> When became shadow so overly bureaucratic and complex? 

I'm very sorry if that's the impression.

I asked for the range-diff, because it makes it a lot easier to review the changes from v1 to v2, and presented you the command to make it easier for you, in case you don't use it often.

Github's interface shows a diff when you force-push, but if you force-push after a rebase, it includes all the changes that have been introduced to master in the meantime, which makes it difficult to see what changes correspond actually to the PR.  This is a limitation of GH's interface.

Anyway, seeing that you're not the only one that complained about my reviews, I assume it's my fault, so I'll try to be less bureaucratic with reviews, and do less reviews, in general).

> 
> It's up to you how to handle this project. Since this change is not that important, we can just close the pull request.

No, please.  Again, I'm sorry.

Have a lovely day,
Alex

alejandro-colomar:
LGTM.  Thanks for your work!

PR #924, opened 2024-01-22
W

alejandro-colomar:
@thesamesam 

I've prepared it already, per your suggestion.


lib/cast.h
```diff
@@ -0,0 +1,21 @@
+// SPDX-FileCopyrightText: 2022-2024, Alejandro Colomar <alx@kernel.org>
```
ikerexxe, line 1: Since this is a new file I guess this should include only 2024, and not 2022

autogen.sh
```diff
@@ -5,8 +5,18 @@ autoreconf -v -f --install "$(dirname "$0")" || exit 1
...
 CFLAGS="$CFLAGS -Wextra"
+#CFLAGS="$CFLAGS -Werror=declaration-missing-parameter-type"
+#CFLAGS="$CFLAGS -Werror=deprecated-non-prototype"
 CFLAGS="$CFLAGS -Werror=implicit-function-declaration"
+CFLAGS="$CFLAGS -Werror=implicit-int"
+#CFLAGS="$CFLAGS -Werror=incompatible-function-pointer-types"
+CFLAGS="$CFLAGS -Werror=incompatible-pointer-types"
+CFLAGS="$CFLAGS -Werror=int-conversion"
+#CFLAGS="$CFLAGS -Werror=return-mismatch"
+#CFLAGS="$CFLAGS -Werror=strict-prototypes"
```
ikerexxe, line 16: Can you add a comment at the end of the line explaining why these are commented?

ikerexxe (commented):
Two minor comments inline

lib/cast.h
alejandro-colomar: It's new for this project, but I copied from some other projects of mine that were older.  Since I've tweaked it a little bit for this project, I updated the copyright, but without removing the original one.

autogen.sh
alejandro-colomar: It's commented in the commit message already.  Do you want it there too?  Or maybe I should just remove those lines?

alejandro-colomar:
v2 changes:

-  Remove `-Wno-unknown-warning`.  It was a red herring.
-  Remove warnings that are new in GCC 14, or exclusive to Clang, as GCC doesn't have a way to deal with them.  [@ikerexxe ]

```
$ git range-diff master gh/W W 
1:  baf1f916 = 1:  baf1f916 lib/must_be.h: is_same_typeof(): Rename macro
2:  fe5774b5 = 2:  fe5774b5 lib/must_be.h: is_same_type(): Add macro
3:  df8ee946 = 3:  df8ee946 lib/cast.h: const_cast(): Add macro for dropping 'const'
4:  6461c51e = 4:  6461c51e src/su.c: Use const_cast() to silence -Wincompatible-pointer-types-discards-qualifiers
5:  08825432 = 5:  08825432 src/useradd.c: get_defaults(): Use const temporary pointer to fix const correctness
6:  469fbd13 = 6:  469fbd13 src/usermod.c: grp_update(): Remove scope of variable, and fix const correctness
7:  31c5ed1c < -:  -------- autogen.sh: CFLAGS: Add some -Werror=... flags that will be default soon
8:  b3f22d1e < -:  -------- autogen.sh: CFLAGS: Use -Wno-unknown-attributes; Clang doesn't know [[gnu::access()]]
-:  -------- > 7:  f0f1d548 autogen.sh: CFLAGS: Add some -Werror=... flags that will be default soon
-:  -------- > 8:  a99b3a50 autogen.sh: CFLAGS: Use -Wno-unknown-attributes; Clang doesn't know [[gnu::access()]]
```

autogen.sh
ikerexxe: Thanks!

ikerexxe (approved):
LGTM! Merging...

PR #925, opened 2024-01-23
lib/pam_pass_non_interactive.c: use strzero/free

ferivoz:
The combination of bzero and free could be optimized away.

### DIFF
--- a/lib/pam_pass_non_interactive.c
+++ b/lib/pam_pass_non_interactive.c
@@ -19,6 +19,7 @@
 
 #include "alloc.h"
 #include "attr.h"
+#include "memzero.h"
 #include "prototypes.h"
 #include "shadowlog.h"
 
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
```
```

alejandro-colomar (approved):

lib/pam_pass_non_interactive.c
```diff
@@ -97,9 +98,8 @@ static int ni_conv (int num_msg,
 failed_conversation:
 	for (count=0; count < num_msg; count++) {
 		if (NULL != responses[count].resp) {
-			bzero(responses[count].resp,
-			      strlen(responses[count].resp));
-			free (responses[count].resp);
+			freezero(responses[count].resp,
+			         strlen(responses[count].resp));
```
alejandro-colomar, line 102: Alternatively, you may want to

```c
strzero(s);
free(s);
```

alejandro-colomar: I have my internal conflict, because it's simpler, but your approach is more portable.  I think simplicity tends to win, but it's your proposal, so whatever you prefer.  ;-)

ferivoz: The combination of strzero and free is more often available in shadow than freezero. And no other example of freezero with strlen.

Thank you for the suggestion. I have adjusted the PR accordingly.

alejandro-colomar: =)

hallyn (approved):

Issue #886, opened 2023-12-30
Insufficient password length management


stoeckmann:
The login.defs file contains the two fields `PASS_MIN_LEN` and `PASS_MAX_LEN`.

### PASS_MAX_LEN

The example etc/login.defs states:
```
#
# Number of significant characters in the password for crypt().
# Default is 8, don't change unless your crypt() is better.
# Ignored if MD5_CRYPT_ENAB set to "yes".
#
#PASS_MAX_LEN           8
```

Contrary to its description, `PASS_MAX_LEN` is only considered if the encryption method is DES. You can easily test this by using e.g. YESCRYPT and change the password with `passwd` and enter a password which is longer than the configured maximum length.

Yet, the maximum length is actually capped. The `agetpass` function supplied by shadow allows up to PASS_MAX characters (definition in C file, not configurable). Yet passwords are already silently truncated at 199 characters due to internal limit in `passwd`.

### PASS_MIN_LEN

If `PASS_MIN_LEN` is not set login.defs, then an invalid message is printed in `passwd`:

Proof of Concept even if PAM support is compiled in (run as root):
```
# passwd -P /.
Changing password for root
Enter the new password (minimum of 5 characters)
Please use a combination of upper and lower case letters and numbers.
New password:
```

According to this message, a minimum of 5 characters is in place. But in fact, empty passwords are allowed as well. This happens because src/ and lib/ use different default values if `PASS_MIN_LEN` does not exist.

alejandro-colomar:
I believe at least some of these issues are resolved (the 199 limit within passwd(1) has been removed).

Would you mind checking what's the status of this?  Also, could you propose some patches?

stoeckmann:
Reported issue is fixed. I think that it can be closed, even though an open merge request exists to remove defaults from login.defs. But that would have no impact on this very issue.

alejandro-colomar:
Thanks!

PR #927, opened 2024-01-25
Implement reading password from pipe/stdin

thalman:
New option --stdin/-s is available for root user. It is useful
for automation/setup and it makes shadow utils passwd more versatile.

### DIFF
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -68,7 +68,8 @@ static bool
     Sflg = false,			/* -S - show password status */
     uflg = false,			/* -u - unlock the user's password */
     wflg = false,			/* -w - set warning days */
-    xflg = false;			/* -x - set maximum days */
+    xflg = false,			/* -x - set maximum days */
+    sflg = false;			/* -s - read passwd from stdin */
 
 /*
  * set to 1 if there are any flags which require root privileges,
@@ -158,6 +159,7 @@ usage (int status)
 	(void) fputs (_("  -w, --warndays WARN_DAYS      set expiration warning days to WARN_DAYS\n"), usageout);
 	(void) fputs (_("  -x, --maxdays MAX_DAYS        set maximum number of days before password\n"
 	                "                                change to MAX_DAYS\n"), usageout);
+	(void) fputs (_("  -s, --stdin                   read new token from stdin\n"), usageout);
 	(void) fputs ("\n", usageout);
 	exit (status);
 }
@@ -252,7 +254,7 @@ static int new_password (const struct passwd *pw)
 			pass_max_len = getdef_num ("PASS_MAX_LEN", 8);
 		}
 	}
-	if (!qflg) {
+	if (!qflg && !sflg) {
 		if (pass_max_len == -1) {
 			(void) printf (_(
 "Enter the new password (minimum of %d characters)\n"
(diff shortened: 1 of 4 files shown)


src/passwd.c
```diff
@@ -266,56 +268,69 @@ static int new_password (const struct passwd *pw)
...
 		if (NULL == cp) {
-			MEMZERO(orig);
-			MEMZERO(pass);
 			return -1;
 		}
-		if (warned && (strcmp (pass, cp) != 0)) {
-			warned = false;
-		}
-		STRTCPY(pass, cp);
+		STRTCPY (pass, cp);
```
alejandro-colomar, line 279: Hmmmm, I need to come back some day to string hardening, and check what to do when STRTCPY() doesn't fit.

From what I can see, this is silently clamping the password to the first 199 bytes +'\0'.  That doesn't look like a good thing to do.

src/passwd.c
```diff
@@ -682,11 +697,11 @@ static void update_shadow (void)
  *	-l	lock the password of the named account (*)
  *	-n #	set sp_min to # days (*)
  *	-r #	change password in # repository
- *	-s	execute chsh command to interpret flags
```
alejandro-colomar, line 685: What heppened to -s?

src/passwd.c
```diff
@@ -850,6 +866,16 @@ int main (int argc, char **argv)
 				xflg = true;
 				anyflag = true;
 				break;
+			case 't':
+				if (!amroot) {
```
alejandro-colomar, line 870: Ahh, okay, you have the check here.  Forget what I said above.

stoeckmann:
> It is useful for automation/setup and it makes shadow utils passwd more versatile.

Which use case does this cover compared to what `newusers` has to offer? The advantage of `newusers` is that it's not setuid. Pretty much "root only" already.

src/passwd.c
stoeckmann: Yes, that's what's happening: https://github.com/shadow-maint/shadow/issues/886

alejandro-colomar: Thanks!

thalman:
Hmm, I found that my PR does not work correctly with PAM. Let me investigate it first.

src/passwd.c
alejandro-colomar: Please don't remove that line.

alejandro-colomar (dismissed):
Please don't remove the `-s` comment in this PR.

src/passwd.c
thalman: It seems that -s was never there or it is long gone. I hesitated if I can usurp -s short flag for --stdin. Now another PR has been merged that cleaned this comment so I updated my PR to
- not remove old comment
- use -s as a short option for --stdin

thalman:
> Please don't remove the `-s` comment in this PR. (Never mind; fixed in master.)

Removed from PR

src/passwd.c
alejandro-colomar: Ahh, makes sense.  I checked, and none of FreeBSD, NetBSD, nor OpenBSD use -s, so that LGTM.

ikerexxe (approved):
LGTM. Thanks for the patch and your hard work here!

stoeckmann:
> > Which use case does this cover compared to what `newusers` has to offer? The advantage of `newusers` is that it's not setuid. Pretty much "root only" already.
> 
> But yes you are right, this can be achieved in many different ways.

Just to add to my initial remark: The solution with `newusers` has a critical issue: Passwords cannot contain colons because they are field separators. So I'm not a fan of newusers anymore at all.

Thank you for clarification!

PR #935, opened 2024-02-03
lib/chkname.c: Fix off by one validation of _SC_LOGIN_NAME_MAX

stoeckmann:
According to sysconf manual page, _SC_LOGIN_NAME_MAX is the maximum name length, **including** the terminating null byte.
Take this into account while validating user names.

Also support systems which might have no upper limit or no _SC_LOGIN_NAME_MAX at all.

### DIFF
--- a/lib/chkname.c
+++ b/lib/chkname.c
@@ -80,7 +80,7 @@ bool is_valid_user_name (const char *name)
 	 * User names length are limited by the kernel
 	 */
 	maxlen = sysconf(_SC_LOGIN_NAME_MAX);
-	if (strlen(name) > maxlen)
+	if (strlen(name) >= maxlen)
 		return false;
 
 	return is_valid_name (name);
--- a/lib/chkname.c
+++ b/lib/chkname.c
@@ -20,6 +20,8 @@
 #ident "$Id$"
 
 #include <ctype.h>
+#include <errno.h>
+#include <limits.h>
 #include "defines.h"
 #include "chkname.h"
 
@@ -74,13 +76,16 @@ static bool is_valid_name (const char *name)
 
 bool is_valid_user_name (const char *name)
 {
-	size_t  maxlen;
+	long  maxlen;
 
 	/*
 	 * User names length are limited by the kernel
 	 */
+	errno = 0;
 	maxlen = sysconf(_SC_LOGIN_NAME_MAX);
-	if (strlen(name) >= maxlen)
+	if (maxlen == -1 && errno != 0)
+		maxlen = LOGIN_NAME_MAX;
+	if (maxlen != -1 && strlen(name) >= (size_t)maxlen)
 		return false;
 
 	return is_valid_name (name);
--- a/lib/chkname.c
+++ b/lib/chkname.c
@@ -79,7 +79,7 @@ bool is_valid_user_name (const char *name)
 	long  maxlen;
 
 	/*
-	 * User names length are limited by the kernel
+	 * User name length should be limited by the kernel
 	 */
 	errno = 0;
 	maxlen = sysconf(_SC_LOGIN_NAME_MAX);
--- a/tests/unit/test_chkname.c
+++ b/tests/unit/test_chkname.c
@@ -134,15 +134,15 @@ test_is_valid_user_name_long(void **state)
 	char    *name;
 
 	max = sysconf(_SC_LOGIN_NAME_MAX);
-	name = MALLOC(max + 2, char);
+	name = MALLOC(max + 1, char);
 	assert_true(name != NULL);
 
-	memset(name, '_', max + 1);
-
-	name[max + 1] = '\0';
-	assert_true(false == is_valid_user_name(name));
+	memset(name, '_', max);
 
 	name[max] = '\0';
+	assert_true(false == is_valid_user_name(name));
+
+	name[max - 1] = '\0';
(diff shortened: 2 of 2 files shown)


lib/chkname.c
```diff
@@ -74,13 +76,16 @@ static bool is_valid_name (const char *name)
...
-	size_t  maxlen;
+	long  maxlen;
 
 	/*
 	 * User names length are limited by the kernel
 	 */
+	errno = 0;
 	maxlen = sysconf(_SC_LOGIN_NAME_MAX);
-	if (strlen(name) >= maxlen)
+	if (maxlen == -1 && errno != 0)
```
hallyn, line 86: Your patch description said:

```
If the system has no _SC_LOGIN_NAME_MAX, use
LOGIN_NAME_MAX constant instead.
```

But the sysconf manpage says that if errno is unchanged (so, here it is 0) that means the value is unset.  You are, in that case, not doing any max length checking at all, which is not what the patch description seems to be saying.

Am I misunderstanding, or did you mean to also set maxlen to LOGIN_NAME_MAX if (maxlen == -1 && errno == 0) ?

alejandro-colomar: I understand the commit message as:

- if (_SC_LOGIN_NAME_MAX == infinite)  // ret -1; errno 0
    - don't check length
- elif (sysconf didn't understand _SC_LOGIN_NAME_MAX)  // ret -1; errno != 0
    - use LOGIN_NAME_MAX
- else
    - use _SC_LOGIN_NAME_MAX

alejandro-colomar: Which seems to match the code

lib/chkname.c
```diff
@@ -79,7 +79,7 @@ bool is_valid_user_name (const char *name)
 	long  maxlen;
 
 	/*
-	 * User names length are limited by the kernel
+	 * User name length should be limited by the kernel
```
alejandro-colomar, line 82: I would rather remove the comment.  Calling sysconf(3) already tells that it's limited by the kernel, doesn't it?  The comment is just a chance of being confusing.

alejandro-colomar (approved):

lib/chkname.c
hallyn: Yeah, the manpage is confusing, because it uses the term 'indeterminate' without defining it.  I disagree with your reading.  It appears to me that errno != 0 will only happen on an *invalid* name.  _SC_LOGIN_NAME_MAX is listed in the sysconf manpage, and is, I would assume, always valid.  So when the commit message for patch 2 says:

| If the system has no _SC_LOGIN_NAME_MAX, use LOGIN_NAME_MAX constant instead

I would assume that means 'is indeterminate'.

I'm happy with what the code does, but when I look through patch history in 3 years I don't want to think this patch did something differnet from what it does.  Please let's update the commit message.

alejandro-colomar: Maybe if you run an old-enough kernel you may trigger EINVAL with a known-valid name.

Or maybe a future kernel removes that name and starts reporting EINVAL (I doubt this).

But yeah, the wording of the commit can be improved.
And the manual page probably too.

hallyn: Ok - I've changed my mind, let's merge as is.  Thanks.

lib/chkname.c
hallyn: The comment is probably redundant, but let's not spend too much time thinking about it :)

lib/chkname.c
alejandro-colomar:
```
lib/chkname.c: Support unlimited user name lengths

If the system doesn't set a user-name limit, support it accordingly, by
allowing any user-name length.  If sysconf(3) fails (but it shouldn't),
use the LOGIN_NAME_MAX constant instead.
```

I think this might be more readable.

hallyn: Sorry - I think you're right, but I merged as is :)

alejandro-colomar: I was looking at cherry-picking it to stable, and thanks to some conflict, I had a look at the commit that introduced the conflict:

```diff
commit 6be85b0bafb5c2c544dd6328c599369d47334429
Author: Alejandro Colomar <alx@kernel.org>
Date:   Sat Sep 2 14:15:43 2023 +0200

    lib/chkname.c: Use tmp variable to avoid a -Wsign-compare warning
    
    I used size_t because:
    
    sysconf(3) can return -1 if the value is not supported, but then it can
    only mean that there's no limit.  Having no limit is the same as having
    a limit of SIZE_MAX (to which -1 is converted).
    
    Signed-off-by: Alejandro Colomar <alx@kernel.org>

diff --git a/lib/chkname.c b/lib/chkname.c
index 0a18af2b..cbfbff51 100644
--- a/lib/chkname.c
+++ b/lib/chkname.c
@@ -74,12 +74,14 @@ static bool is_valid_name (const char *name)
 
 bool is_valid_user_name (const char *name)
 {
+       size_t  maxlen;
+
        /*
         * User names length are limited by the kernel
         */
-       if (strlen (name) > sysconf(_SC_LOGIN_NAME_MAX)) {
+       maxlen = sysconf(_SC_LOGIN_NAME_MAX);
+       if (strlen(name) > maxlen)
                return false;
-       }
 
        return is_valid_name (name);
 }
```

It seems I had taken into account infinite login names and I forgot.  :)

At least this commit prevents the EINVAL case, even if it's unlikely.

PR #209, opened 2020-01-12
add new HOME_MODE login.defs(5) option

Duncaen:
@hallyn sorry, forgot to add the file after rebasing the PR.
 I rebased again and added the missing file and added references to the new variable to the man pages of useradd(8) and newusers(8) which I forgot before.

Fixed version of #208

PR #946, opened 2024-02-11
Set default HOME_MODE to 0700

weseven:
Since the introduction of https://github.com/shadow-maint/shadow/pull/209, many distros started to set HOME_MODE by default:
 -  RedHat & RH based distro like [Fedora](https://src.fedoraproject.org/rpms/shadow-utils/blob/rawhide/f/shadow-utils.login.defs#_122), but also [Arch linux](https://gitlab.archlinux.org/archlinux/packaging/packages/shadow/-/blob/main/0003-Add-Arch-Linux-defaults-for-login.defs.patch?ref_type=heads#L40), [OpenSuse](https://build.opensuse.org/package/view_file/openSUSE:Factory/shadow/shadow-login_defs-suse.patch?expand=1) all set it to 0700
 - [Ubuntu](https://git.launchpad.net/ubuntu/+source/shadow/tree/debian/login.defs#n156) sets it to 0750
 - notable exceptions are [Debian](https://salsa.debian.org/debian/shadow/-/blob/master/debian/login.defs?ref_type=heads#L156) and Gentoo that keep the variable unset.

I propose to set it to 0700 or at least 0750 by default, since it is a security best practice and it's something often tested by security benchmarks (e.g., CIS Benchmarks).

### DIFF
--- a/etc/login.defs
+++ b/etc/login.defs
@@ -204,7 +204,7 @@ UMASK		022
 # HOME_MODE is used by useradd(8) and newusers(8) to set the mode for new
 # home directories.
 # If HOME_MODE is not set, the value of UMASK is used to create the mode.
-#HOME_MODE	0700
+HOME_MODE	0700
 
 #
 # Password aging controls:


hallyn:
No - distros like debian may get caught off guard.  There's nothing wrong with downstreams patching their values in their deltas. We do not lightly make changes which change defaults.

weseven:
Understood, thanks for taking a look.
Is there a possibility this change will be applied here in the future, and in the meantime give maintainers a notice that this will happen after some releases?

I still think it's a better default value than leaving it unset, and its impact is limited to useradd and newusers. 
I understand your concern in possibly changing historic behavior and respecting downstream maintainers, but it's also a downstream maintainer responsibility to read upstream changes and adapt the defaults to the distro expected use case.

PR #893, opened 2024-01-08
String-to-numeric hardening

alejandro-colomar:
---

<details>
<summary>v3</summary>
v3 changes:

-  Rebase

```
$ git range-diff gh/getuh..gh/atoi getuh..atoi 
 1:  62e21a24 =  1:  14ef2653 lib/limits.c: check_logins(): Report LOGIN_ERROR_LOGIN if str2ul() ERANGE
 2:  c993320f =  2:  c4ad6a5d lib/idmapping.c: get_map_ranges(): Rename local variable
 3:  d7186a5a =  3:  212a171b lib/idmapping.c: get_map_ranges(): Move range check to a2ul() calls
 4:  3efa3a4d =  4:  8c0b20ed lib/idmapping.c: get_map_ranges(): Move range check to a2ul() call
 5:  06d88123 =  5:  2b1b6da9 lib/idmapping.c: get_map_ranges(): Move range check to a2ul() call
 6:  962b4e45 =  6:  7c20cbef lib/idmapping.c: get_map_ranges(): Remove dead code
 7:  41adf242 =  7:  1c6e2418 lib/idmapping.c: get_map_ranges(): Simplify iterator variables
 8:  36ff918f =  8:  686d8de8 lib/idmapping.c: get_map_ranges(): Don't exit() from a library function
 9:  9f049eef !  9:  fa5581d1 lib/atoi/: Use libbsd's strto[iu](3bsd) when available
    @@ lib/atoi/a2i.h
      
      #include "atoi/strtoi.h"
      #include "atoi/strtou_noneg.h"
    -@@ lib/atoi/a2i.h: a2sh_nc(short *restrict n, const char *s,
    +@@ lib/atoi/a2i.h: a2sh_nc(short *restrict n, char *s,
      {
        int  status;
      
    @@ lib/atoi/a2i.h: a2sh_nc(short *restrict n, const char *s,
        if (status != 0) {
                errno = status;
                return -1;
    -@@ lib/atoi/a2i.h: a2si_nc(int *restrict n, const char *s,
    +@@ lib/atoi/a2i.h: a2si_nc(int *restrict n, char *s,
      {
        int  status;
      
    @@ lib/atoi/a2i.h: a2si_nc(int *restrict n, const char *s,
        if (status != 0) {
                errno = status;
                return -1;
    -@@ lib/atoi/a2i.h: a2sl_nc(long *restrict n, const char *s,
    +@@ lib/atoi/a2i.h: a2sl_nc(long *restrict n, char *s,
[...]

PR #951, opened 2024-02-13
usermod: refuse invalid uidmaps during --add-sub{u,g}ids

tych0:
It is slightly confusing to allow adding these only to later refuse them.

Here is a (lightly tested :) patch to also refuse them when adding.

### DIFF
--- a/src/usermod.c
+++ b/src/usermod.c
@@ -331,6 +331,13 @@ static struct ulong_range getulong_range(const char *str)
 	if (first > last)
 		goto out;
 
+	/*
+	 * uid_t in linux is an unsigned int, anything over this is an invalid
+	 * range will be later refused anyway by get_map_ranges().
+	 */
+	if (first > UINT_MAX || last > UINT_MAX)
+		goto out;
+
 	result.first = (unsigned long)first;
 	result.last = (unsigned long)last;
 out:


hallyn (approved):
If it were to somehow happen that a kernel supported wider uid range, then this would become hairy to fix, but it does seem the right thing at the moment, thanks.

alejandro-colomar:
I've assigned myself because I want to investigate this in the future, regarding <https://github.com/shadow-maint/shadow/pull/893>.

hallyn:
> I've assigned myself because I want to investigate this in the future, regarding #893.

I thought you might :)

And generally speaking I'm a bit torn on this - on the one hand it's good to catch bad values as early as possible, on the other hand the kernel protects itself already, and we don't want to hardcode limitations which might be removed in the future.  So that's just to say I'll want to discuss any other improvements in this area :)

Thanks!

alejandro-colomar:
> > I've assigned myself because I want to investigate this in the future, regarding #893.
> 
> I thought you might :)

=)

> And generally speaking I'm a bit torn on this - on the one hand it's good to catch bad values as early as possible, on the other hand the kernel protects itself already, and we don't want to hardcode limitations which might be removed in the future. So that's just to say I'll want to discuss any other improvements in this area :)

Agree.  I don't really like this patch either, although it was very helpful, to understand the purpose of that function.  So, thanks for the patch, @tych0 !  :)

When we add liba2i functions, we can do something like:

```c
if (str2i(id_t, &num, str) == -1)
    goto err;
```

That would do the appropriate range checks for the type `id_t`, without hard-coding what `id_t` actually is.

> 
> Thanks!

Have a lovely day!
Alex

tych0:
Maybe it would look better to put all this "uid range check" logic in one place, instead of trying to duplicate it across functions? Or even better: fork and try to create a ns, see if the kernel allows it? Then it is future proof.

hallyn:
Certainly yes to commonizing the code in one place.

But I think just checking /proc/self/uid_map to look for the available range is probably safer than actually trying the mapping.  All of it, however, becomes a tradeoff between the fragility of the new tests and the risk of not doing the tests at all.

alejandro-colomar:
> Maybe it would look better to put all this "uid range check" logic in one place, instead of trying to duplicate it across functions? Or even better: fork and try to create a ns, see if the kernel allows it? Then it is future proof.

That's more or less my plan.  The code base is full of range checks after strtol(3) calls.  What I did is create a library that parses numbers into arbitrary types, such as `uid_t`, or `id_t`, implicitly performing the range checks for the type.

So, just by changing this code to use `id_t` instead of `unsigned long`, and calling these functions of mine, the code would be doing the range checks in a centralized place.

See for example:

-  <https://github.com/shadow-maint/shadow/pull/893/commits/b8967bea124bb4eccbd3d9589afbd05cc44e8734>
-  <https://github.com/shadow-maint/shadow/pull/893/commits/77430a71ab04db9fccab0fe260432eef708e476e>
-  <https://github.com/shadow-maint/shadow/pull/893/commits/ef0f17feeb30a7ecc0992b9dbb4977851c23cc1c>

BTW, looking at those, we already have a get_uid() function.  I'll probably use that.

<https://www.alejandro-colomar.es/src/alx/alx/lib/liba2i.git/>

PR #953, opened 2024-02-16
Passwd length

thalman:
This PR addresses two issues:

* The passwd utility has limit 200 characters. Some admins wants uses longer passwords for technical accounts
* The passwd utility cuts the provided password silently

First commit unifies the password length limit between `agetpass.c` and `passwd.c`.

The second one introduces checking the length of provided password.

### DIFF
--- a/lib/agetpass.c
+++ b/lib/agetpass.c
@@ -11,7 +11,6 @@
 
 #include <limits.h>
 #include <readpassphrase.h>
-#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -22,11 +21,6 @@
 #if WITH_LIBBSD == 0
 #include "freezero.h"
 #endif /* WITH_LIBBSD */
-
-
-#if !defined(PASS_MAX)
-#define PASS_MAX  BUFSIZ - 1
-#endif
 
 
 /*
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -175,8 +175,8 @@ static int new_password (const struct passwd *pw)
 	char *cipher;		/* Pointer to cipher text */
 	const char *salt;	/* Pointer to new salt */
 	char *cp;		/* Pointer to agetpass() response */
-	char orig[200];		/* Original password */
-	char pass[200];		/* New password */
+	char orig[PASS_MAX + 1];	/* Original password */
+	char pass[PASS_MAX + 1];	/* New password */
 	int i;			/* Counter for retries */
 	bool warned;
 	int pass_max_len = -1;
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -178,6 +178,7 @@ static int new_password (const struct passwd *pw)
 	char orig[PASS_MAX + 1];	/* Original password */
 	char pass[PASS_MAX + 1];	/* New password */
 	int i;			/* Counter for retries */
+	int ret;
 	bool warned;
 	int pass_max_len = -1;
 	const char *method;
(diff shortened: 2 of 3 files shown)


lib/agetpass.c
```diff
@@ -24,11 +24,6 @@
 #endif /* WITH_LIBBSD */
 
 
-#if !defined(PASS_MAX)
-#define PASS_MAX  BUFSIZ - 1
```
alejandro-colomar, line 28: You can remove `#include <stdio.h>` from this file now, I think.

alejandro-colomar:
Thanks for doing this!

I see we have several macros that seem very related:

```
$ grep -rho 'PASS[A-Z_]*MAX[A-Z_]*' src/ lib* | sort | uniq | grep -v PASS_MAX_DAYS
PASSWD_ENTRY_MAX_LENGTH
PASS_MAX
PASS_MAX_LEN
```

Do any of those mean the same thing?  I suspect we have different maximum lengths in different places, and we should unify them.

alejandro-colomar:
https://github.com/shadow-maint/shadow/pull/953/commits/1a36ac7121dbf665414f7b42df621b8e69da0916 `("src/passwd.c: inconsistent password length limit ")`

```
```

src/passwd.c
```diff
@@ -290,8 +296,14 @@ static int new_password (const struct passwd *pw)
...
 				warned = false;
 			}
-			STRTCPY(pass, cp);
+			ret = STRTCPY (pass, cp);
 			erase_pass (cp);
+			if (ret == -1) {
+				(void) fputs (_("Password is too long.\n"), stderr);
+				MEMZERO(orig);
+				MEMZERO(pass);
+				return -1;
```
alejandro-colomar, line 305: Here you should suggest to try again, for consistency with the rest of the loop.

alejandro-colomar:
https://github.com/shadow-maint/shadow/pull/953/commits/6a0c380705f02d9d85e156ad52f8bf61d815dd45 `("src/passwd.c: check password length upper limit")`

```
```

alejandro-colomar:
> Thanks for doing this!
> 
> I see we have several macros that seem very related:
> 
> ```
> $ grep -rho 'PASS[A-Z_]*MAX[A-Z_]*' src/ lib* | sort | uniq | grep -v PASS_MAX_DAYS
> PASSWD_ENTRY_MAX_LENGTH
> PASS_MAX
> PASS_MAX_LEN
> ```
> 
> Do any of those mean the same thing? I suspect we have different maximum lengths in different places, and we should unify them.

Let's keep this for a different PR.  I think this one is good as is.

lib/agetpass.c
thalman: Yes, I can do that. I thought about it.

But in fact moving `#include <stdio.h>` to defines.h means that this include can be removed from about 70 files.

I'm not sure how to proceed with that. Extra commit? PR? What do you suggest?

thalman:
> > Thanks for doing this!
> > I see we have several macros that seem very related:
> > ```
> > $ grep -rho 'PASS[A-Z_]*MAX[A-Z_]*' src/ lib* | sort | uniq | grep -v PASS_MAX_DAYS
> > PASSWD_ENTRY_MAX_LENGTH
> > PASS_MAX
> > PASS_MAX_LEN
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do any of those mean the same thing? I suspect we have different maximum lengths in different places, and we should unify them.
> 
> Let's keep this for a different PR. I think this one is good as is.

I think that they have different meaning. If I understand correctly `PASSWD_ENTRY_MAX_LENGTH` is the whole record from /etc/passwd. `PASS_MAX_LEN` is from system configuration and `PASS_MAX` is our internal limit. But I might be wrong.

lib/agetpass.c
thalman: So far I added `stdio.h` removal from this file. For the others I seek a recommendation.

alejandro-colomar: > But in fact moving `#include <stdio.h>` to defines.h means that this include can be removed from about 70 files.

Don't do that, please.  :)

See:

-  <https://include-what-you-use.org/>
-  <https://google.github.io/styleguide/cppguide.html#Include_What_You_Use>

   > Do not rely on transitive inclusions.

alejandro-colomar (approved):
LGTM.  Thanks!

PR #957, opened 2024-02-21
Use trap(1) to show the testsuite log

alejandro-colomar:
Otherwise, 'cat testsuite.log' isn't run, since 'set -e' aborts the script earlier.

### DIFF
--- a/.github/workflows/runner.yml
+++ b/.github/workflows/runner.yml
@@ -49,8 +49,9 @@ jobs:
       run: |
         set -e
         cd tests
+        trap 'cat testsuite.log' ERR
         sudo ./run_some
-        cat testsuite.log
+        trap - ERR
 
   # Make sure that 'make dist' makes a usable tarball with no missing files
   dist-build:
--- a/share/containers/alpine.dockerfile
+++ b/share/containers/alpine.dockerfile
@@ -5,6 +5,7 @@ FROM "${OS_IMAGE}" AS build
 RUN apk add \
 	autoconf \
 	automake \
+	bash \
 	build-base \
 	byacc \
 	cmocka-dev \
@@ -27,7 +28,7 @@ RUN ./autogen.sh \
 	--with-yescrypt
 RUN make -kj4 || true
 RUN make
-RUN make check
+RUN bash -c "trap 'cat <tests/unit/test-suite.log >&2' ERR; make check;"
 RUN make install
 
 FROM scratch AS export
--- a/share/containers/alpine.dockerfile
+++ b/share/containers/alpine.dockerfile
@@ -2,13 +2,29 @@ ARG OS_IMAGE="alpine:latest"
 
 FROM "${OS_IMAGE}" AS build
 
-RUN apk add autoconf automake build-base byacc cmocka-dev expect gettext-dev \
-        git libbsd-dev libeconf-dev libtool libxslt pkgconf
+RUN apk add \
+	autoconf \
+	automake \
+	build-base \
+	byacc \
+	cmocka-dev \
+	expect \
+	gettext-dev \
+        git \
+	libbsd-dev \
+	libeconf-dev \
+	libtool \
+	libxslt \
+	pkgconf
 
 COPY ./ /usr/local/src/shadow/
 WORKDIR /usr/local/src/shadow/
 
-RUN ./autogen.sh --without-selinux --disable-man --disable-nls --with-yescrypt
+RUN ./autogen.sh \
+	--without-selinux \
+	--disable-man \
+	--disable-nls \
+	--with-yescrypt
 RUN make -kj4 || true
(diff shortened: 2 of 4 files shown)


alejandro-colomar:
I've tested it manually, by running these same commands in a docker container.

ikerexxe:
Reading the logs directly in Github WebUI is very painful. Can you upload them? You have an example at https://github.com/shadow-maint/shadow/blob/master/.github/workflows/runner.yml#L102. This will provide all the logs in a compressed archive in the checks section of the PR.

By the way, I think we are already using v4 for this action, so use that one instead of v3.

alejandro-colomar:
> Reading the logs directly in Github WebUI is very painful. Can you upload them? You have an example at https://github.com/shadow-maint/shadow/blob/master/.github/workflows/runner.yml#L102. This will provide all the logs in a compressed archive in the checks section of the PR.
> 
> By the way, I think we are already using v4 for this action, so use that one instead of v3.

Ahh, so there's a way to see cmocka logs?  I was quite frustrated in the past when cmocka reported a test failure because I didn't know how to see the logs, and had to manually reproduce the entire build inside a docker container and then cat the log file.  :D

But, I don't think that's working as intended.  See for example:

-  <https://github.com/shadow-maint/shadow/actions/runs/7966498776/job/21747847052#step:3:2424>
Since the previous step, <https://github.com/shadow-maint/shadow/blob/1af6b68cbeb95f99d096ac4490bd98bca73328b0/.github/workflows/runner.yml#L98>, has failed, the following step isn't run.  So, the logs seem to be stored only when they succeed, which... well.  :)

(BTW, that PR is expected to fail.  The reason is the broken strtoi(3bsd) from libbsd.  So, if you want to test a patch for a fix, that would be a good PR to test it.)

ikerexxe:
> Ahh, so there's a way to see cmocka logs? I was quite frustrated in the past when cmocka reported a test failure because I didn't know how to see the logs, and had to manually reproduce the entire build inside a docker container and then cat the log file. :D

I think I never added cmocka logs to the reporting, so that kind of makes sense.

> But, I don't think that's working as intended. See for example:

Something must have changed because it was working even when something failed. I'll take a look at it.

ikerexxe:
Ok, so it seems it's not possible to get the logs when a command in docker build fails. I know there have been some changes around it with BuildKit, but maybe it never worked.

So, as I see it there are two options: either we move all the steps to scripts, or we move them to the Github workflow. Any other idea?

alejandro-colomar:
v1b changes:

-  Rebase to master
```
$ git range-diff gh/trap^..gh/trap shadow/master..trap
1:  17573002 = 1:  f9f8dc8e .github/workflows/runner.yml: Use trap(1) to show the testsuite log
```

alejandro-colomar:
It isn't the most beautiful thing in the world, but it works:

```dockerfile
$ cat Dockerfile 
FROM debian:bullseye
RUN bash -c "trap 'echo foo >&2' ERR; false; trap - ERR;"
```
```
$ sudo docker build -f Dockerfile .
Sending build context to Docker daemon  28.67kB
Step 1/2 : FROM debian:bullseye
 ---> 873181a41b99
Step 2/2 : RUN bash -c "trap 'echo foo >&2' ERR; false; trap - ERR;"
 ---> Running in af0f3ee96fc2
foo
Removing intermediate container af0f3ee96fc2
 ---> 418ff96da49a
Successfully built 418ff96da49a
```

alejandro-colomar:
v2 changes:

-  trap(1) to see the cmocka logs in the Docker builds  [Cc: @ikerexxe ]

```
$ git range-diff shadow/master gh/trap trap 
1:  f9f8dc8e = 1:  f9f8dc8e .github/workflows/runner.yml: Use trap(1) to show the testsuite log
-:  -------- > 2:  eff32030 share/containers/: Specify one argument per line
-:  -------- > 3:  0dbd1b77 share/containers/: trap(1) to see the cmocka logs
```

I've also pushed those two commits to

- <https://github.com/shadow-maint/shadow/pull/893>

to test them.

hallyn (approved):

alejandro-colomar:
I'll merge it already.  It shouldn't be risky, given it only changes code running tests.  And it might be useful to have this working in stable branches, in case we need to debug any problems.  Thanks!

alejandro-colomar:
v2b changes:

-  Rebase to master
```
$ git range-diff master..gh/trap shadow/master..trap 
1:  f9f8dc8e = 1:  fe8caf4c .github/workflows/runner.yml: Use trap(1) to show the testsuite log
2:  eff32030 = 2:  ce2486f3 share/containers/: Specify one argument per line
3:  0dbd1b77 = 3:  ef77b580 share/containers/: trap(1) to see the cmocka logs
```

Issue #959, opened 2024-03-02
4.14.7  oldstable release


alejandro-colomar:
And it seems we already have cheese for this.  Let's release .7 on the next full moon after the first patches are ready; so possibly on 2024-03-25.

Cc: @skyler-ferrante

skyler-ferrante:
I tested a few other commands, and it seems like su was the only one that allowed messages to go to auth.log. To summarize the issue I sent in my email, su was allowing an attacker message to go to /var/log/auth.log without checking if it had newlines or escape sequences.

A quick PoC:
```
#include<stdio.h>
#include<unistd.h>
int main(){
        char* prog = "/usr/bin/su";
        char* argv[] = {"\033[33mYellow", "root", NULL};
        char* envp[] = {NULL};
        execve(prog, argv, envp);
        printf("Failed to exec\n");
}
```

alejandro-colomar:
False alarm.  I'll close the issue while there's nothing to fix.

alejandro-colomar:
The plans for the next moon are alive again.  Maybe even earlier, if possible.

alejandro-colomar:
-  <https://github.com/shadow-maint/shadow/pull/962>

PR #960, opened 2024-03-02
Fix escape sequence injection in /var/log/auth.log

skyler-ferrante:
See issue #959. su was sending argv[0] to auth.log, which can contain newlines or escape sequences, which can be used to hide messages.

### DIFF
--- a/src/chage.c
+++ b/src/chage.c
@@ -45,7 +45,7 @@
 /*
  * Global variables
  */
-const char *Prog;
+static const char Prog[] = "chage";
 
 static bool
     dflg = false,		/* set last password change date */
@@ -517,7 +517,7 @@ static void check_perms (void)
 		exit (E_NOPERM);
 	}
 
-	retval = pam_start ("chage", pampw->pw_name, &conv, &pamh);
+	retval = pam_start (Prog, pampw->pw_name, &conv, &pamh);
 
 	if (PAM_SUCCESS == retval) {
 		retval = pam_authenticate (pamh, 0);
@@ -771,7 +771,6 @@ int main (int argc, char **argv)
 	/*
 	 * Get the program name so that error messages can use it.
 	 */
-	Prog = Basename (argv[0]);
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
 
@@ -786,7 +785,7 @@ int main (int argc, char **argv)
 #ifdef WITH_AUDIT
 	audit_help_open ();
 #endif
-	OPENLOG ("chage");
+	OPENLOG (Prog);
 
 	ruid = getuid ();
 	rgid = getgid ();
--- a/src/su.c
+++ b/src/su.c
@@ -66,7 +66,7 @@
 /*
  * Global variables
  */
-const char *Prog;
+static const char Prog[] = "su";
 static /*@observer@*/const char *caller_tty = NULL;	/* Name of tty SU is run from */
 static bool caller_is_root = false;
 static uid_t caller_uid;
@@ -738,11 +738,6 @@ static void save_caller_context (char **argv)
 	const char *password = NULL;
 #endif				/* SU_ACCESS */
 #endif				/* !USE_PAM */
-	/*
-	 * Get the program name. The program name is used as a prefix to
-	 * most error messages.
-	 */
-	Prog = Basename (argv[0]);
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
 
@@ -1018,14 +1013,14 @@ int main (int argc, char **argv)
 
 	save_caller_context (argv);
 
-	OPENLOG ("su");
+	OPENLOG (Prog);
 
 	process_flags (argc, argv);
 
 	initenv ();
 
 #ifdef USE_PAM
-	ret = pam_start ("su", name, &conv, &pamh);
+	ret = pam_start (Prog, name, &conv, &pamh);
 	if (PAM_SUCCESS != ret) {
 		SYSLOG ((LOG_ERR, "pam_start: error %d", ret);
 		fprintf (stderr,
--- a/src/check_subid_range.c
+++ b/src/check_subid_range.c
@@ -20,14 +20,13 @@
 #include "idmapping.h"
 #include "shadowlog.h"
 
-const char *Prog;
+static const char Prog[] = "check_subid_range";
 
 int main(int argc, char **argv)
 {
 	char *owner;
 	unsigned long start, count;
 	bool check_uids;
-	Prog = Basename (argv[0]);
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
 
--- a/src/chfn.c
+++ b/src/chfn.c
@@ -39,7 +39,7 @@
 /*
  * Global variables.
  */
-const char *Prog;
+static const char Prog[] = "chfn";
 static char fullnm[BUFSIZ];
 static char roomno[BUFSIZ];
 static char workph[BUFSIZ];
@@ -365,7 +365,7 @@ static void check_perms (const struct passwd *pw)
 	 * check if the change is allowed by SELinux policy.
 	 */
 	if ((pw->pw_uid != getuid ())
-	    && (check_selinux_permit ("chfn") != 0)) {
+	    && (check_selinux_permit (Prog) != 0)) {
 		fprintf (stderr, _("%s: Permission denied.\n"), Prog);
 		closelog ();
 		exit (E_NOPERM);
@@ -380,7 +380,7 @@ static void check_perms (const struct passwd *pw)
 	 * --marekm
 	 */
 	if (!amroot && getdef_bool ("CHFN_AUTH")) {
-		passwd_check (pw->pw_name, pw->pw_passwd, "chfn");
+		passwd_check (pw->pw_name, pw->pw_passwd, Prog);
 	}
 
 #else				/* !USE_PAM */
@@ -392,7 +392,7 @@ static void check_perms (const struct passwd *pw)
 		exit (E_NOPERM);
 	}
 
-	retval = pam_start ("chfn", pampw->pw_name, &conv, &pamh);
+	retval = pam_start (Prog, pampw->pw_name, &conv, &pamh);
 
 	if (PAM_SUCCESS == retval) {
 		retval = pam_authenticate (pamh, 0);
@@ -620,11 +620,6 @@ int main (int argc, char **argv)
 	char                 *user;
 	const struct passwd  *pw;
 
-	/*
-	 * Get the program name. The program name is used as a
-	 * prefix to most error messages.
-	 */
-	Prog = Basename (argv[0]);
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
 
@@ -641,7 +636,7 @@ int main (int argc, char **argv)
 	 */
 	amroot = (getuid () == 0);
 
-	OPENLOG ("chfn");
+	OPENLOG (Prog);
 
 	/* parse the command line options */
 	process_flags (argc, argv);
--- a/src/chgpasswd.c
+++ b/src/chgpasswd.c
@@ -36,7 +36,7 @@
 /*
  * Global variables
  */
-const char *Prog;
+static const char Prog[] = "chgpasswd";
 static bool eflg   = false;
 static bool md5flg = false;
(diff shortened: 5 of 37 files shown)


alejandro-colomar:
Other programs may also write to auth.log.  For example, try `passwd root` from your user.  Maybe we should hardcode all calls to log_set_progname()?

```c
$ grep -rn log_set_progname -B1
lib/shadowlog.h-34-
lib/shadowlog.h:35:extern void log_set_progname(const char *);
--
lib/shadowlog.c-7-
lib/shadowlog.c:8:void log_set_progname(const char *progname)
--
libsubid/api.c-26-			return false;
libsubid/api.c:27:		log_set_progname(progname);
libsubid/api.c-28-	} else {
libsubid/api.c:29:		log_set_progname("(libsubid)");
--
src/vipw.c-474-	Prog = Basename (argv[0]);
src/vipw.c:475:	log_set_progname(Prog);
--
src/sulogin.c-75-	Prog = Basename (argv[0]);
src/sulogin.c:76:	log_set_progname(Prog);
--
src/pwunconv.c-118-	Prog = Basename (argv[0]);
src/pwunconv.c:119:	log_set_progname(Prog);
--
src/chage.c-774-	Prog = Basename (argv[0]);
src/chage.c:775:	log_set_progname(Prog);
--
src/chfn.c-627-	Prog = Basename (argv[0]);
src/chfn.c:628:	log_set_progname(Prog);
--
src/check_subid_range.c-30-	Prog = Basename (argv[0]);
src/check_subid_range.c:31:	log_set_progname(Prog);
--
src/gpasswd.c-945-	Prog = Basename (argv[0]);
src/gpasswd.c:946:	log_set_progname(Prog);
--
src/newgidmap.c-154-	Prog = Basename (argv[0]);
src/newgidmap.c:155:	log_set_progname(Prog);
--
src/lastlog.c-294-	Prog = Basename (argv[0]);
src/lastlog.c:295:	log_set_progname(Prog);
--
src/chsh.c-480-	Prog = Basename (argv[0]);
src/chsh.c:481:	log_set_progname(Prog);
--
src/chpasswd.c-453-	Prog = Basename (argv[0]);
src/chpasswd.c:454:	log_set_progname(Prog);
--
src/groupmod.c-757-	Prog = Basename (argv[0]);
src/groupmod.c:758:	log_set_progname(Prog);
--
src/newusers.c-1062-	Prog = Basename (argv[0]);
src/newusers.c:1063:	log_set_progname(Prog);
--
src/su.c-745-	Prog = Basename (argv[0]);
src/su.c:746:	log_set_progname(Prog);
--
src/groupmems.c-579-	Prog = Basename (argv[0]);
src/groupmems.c:580:	log_set_progname(Prog);
--
src/faillog.c-517-	Prog = Basename (argv[0]);
src/faillog.c:518:	log_set_progname(Prog);
--
src/groupdel.c-355-	Prog = Basename (argv[0]);
src/groupdel.c:356:	log_set_progname(Prog);
--
src/pwconv.c-158-	Prog = Basename (argv[0]);
src/pwconv.c:159:	log_set_progname(Prog);
--
src/logoutd.c-160-	Prog = Basename (argv[0]);
src/logoutd.c:161:	log_set_progname(Prog);
--
src/groupadd.c-578-	Prog = Basename (argv[0]);
src/groupadd.c:579:	log_set_progname(Prog);
--
src/getsubids.c-26-	Prog = Basename (argv[0]);
src/getsubids.c:27:	log_set_progname(Prog);
--
src/userdel.c-965-	Prog = Basename (argv[0]);
src/userdel.c:966:	log_set_progname(Prog);
--
src/new_subid_range.c-31-	Prog = Basename (argv[0]);
src/new_subid_range.c:32:	log_set_progname(Prog);
--
src/chgpasswd.c-426-	Prog = Basename (argv[0]);
src/chgpasswd.c:427:	log_set_progname(Prog);
--
src/usermod.c-2164-	Prog = Basename (argv[0]);
src/usermod.c:2165:	log_set_progname(Prog);
--
src/grpconv.c-129-	Prog = Basename (argv[0]);
src/grpconv.c:130:	log_set_progname(Prog);
--
src/newuidmap.c-83-	Prog = Basename (argv[0]);
src/newuidmap.c:84:	log_set_progname(Prog);
--
src/pwck.c-839-	Prog = Basename (argv[0]);
src/pwck.c:840:	log_set_progname(Prog);
--
src/newgrp.c-428-
src/newgrp.c:429:	log_set_progname(Prog);
--
src/grpunconv.c-127-	Prog = Basename (argv[0]);
src/grpunconv.c:128:	log_set_progname(Prog);
--
src/expiry.c-128-	Prog = Basename (argv[0]);
src/expiry.c:129:	log_set_progname(Prog);
--
src/passwd.c-736-	Prog = Basename (argv[0]);
src/passwd.c:737:	log_set_progname(Prog);
--
src/grpck.c-822-	Prog = Basename (argv[0]);
src/grpck.c:823:	log_set_progname(Prog);
--
src/login.c-523-	Prog = Basename (argv[0]);
src/login.c:524:	log_set_progname(Prog);
--
src/get_subid_owners.c-24-	Prog = Basename (argv[0]);
src/get_subid_owners.c:25:	log_set_progname(Prog);
--
src/free_subid_range.c-28-	Prog = Basename (argv[0]);
src/free_subid_range.c:29:	log_set_progname(Prog);
--
src/groups.c-103-	Prog = Basename (argv[0]);
src/groups.c:104:	log_set_progname(Prog);
--
src/useradd.c-2491-	Prog = Basename (argv[0]);
src/useradd.c:2492:	log_set_progname(Prog);
```

alejandro-colomar:
Hmmm, something seems to be escaping control characters in the case of passwd(1).  I see `#033` in the log, and if I pass newlines, I see `#012` in the log.  For some reason, that escape is working in passwd(1), but not in su(1).  Maybe there's a bug somewhere...

alejandro-colomar:
Hmmm, the escaped version (e.g., passwd(1)) comes from SYSLOG(), which calls syslog(3), and seems to be escaping control characters.  The problematic one (e.g., su(1)), I'm not sure what's calling, but it's likely fprintf(3).

alejandro-colomar:
I think we can hard-code all program names, just to be safe.  I don't see why one would want a different name in the logs.

alejandro-colomar:
And I see that we're initializing openlog(3) (via OPENLOG()) with hard-coded names, so I don't see why we later pass `argv[0]` to syslog(3).

src/su.c
```diff
@@ -743,7 +743,7 @@ static void save_caller_context (char **argv)
 	 * most error messages.
 	 */
 	Prog = Basename (argv[0]);
-	log_set_progname(Prog);
+	log_set_progname("su");
```
alejandro-colomar, line 746: You should probably `Prog = "su"`.  Otherwise, other uses of `Prog` in fprintf(3) calls will have similar issues.  Is it useful to read argv[0] at all?

skyler-ferrante:
> Other programs may also write to auth.log. For example, try `passwd root` from your user (for some reason, the escape sequence didn't work for me in the case of passwd(1), though).

Were you able to get anything other than `su` to send escape sequences? That's all I was able to. I was trying to modify behavior in the least amount possible, while still fixing the issue.

> You should probably Prog = "su"; in the previous line. Otherwise, other uses of Prog below in fprintf(3) calls will have similar issues. Is it useful to read argv[0] at all?

It seems that fprintf is not used to print to log files in shadow. If we consider an attacker changing argv, printing ansi escape sequences to stdout/stderr doesn't seem like a threat surface. Printing to /var/log/auth.log is a threat surface, since a low privilege user can cause this to happen, and a high privilege user might look at them in a dangerous way (e.g. tail).

I agree that all program names should probably just be hard-coded. I only changed su since it seemed only su was affected. But I can send in a patch that hardcodes everything since that's probably better.

skyler-ferrante:
Okay, I created a new patchset that changes every program in shadow to use a hard coded value for Prog. If these changes are considered too large, or we don't want to change the default Prog behavior, we can go back to just changing su.

alejandro-colomar:
> > Other programs may also write to auth.log. For example, try `passwd root` from your user (for some reason, the escape sequence didn't work for me in the case of passwd(1), though).
> 
> Were you able to get anything other than `su` to send escape sequences? That's all I was able to. I was trying to modify behavior in the least amount possible, while still fixing the issue.
> 
> > You should probably Prog = "su"; in the previous line. Otherwise, other uses of Prog below in fprintf(3) calls will have similar issues. Is it useful to read argv[0] at all?
> 
> It seems that fprintf is not used to print to log files in shadow. If we consider an attacker changing argv, printing ansi escape sequences to stdout/stderr doesn't seem like a threat surface. Printing to /var/log/auth.log is a threat surface, since a low privilege user can cause this to happen, and a high privilege user might look at them in a dangerous way (e.g. tail).
> 
> I agree that all program names should probably just be hard-coded. I only changed su since it seemed only su was affected. But I can send in a patch that hardcodes everything since that's probably better.

I've been trying to debug this, and again I couldn't find the source code line that prints the log that su(1) is producing.

And then I've noticed that Debian uses `--without-su` when building shadow:

<https://salsa.debian.org/debian/shadow/-/blob/master/debian/rules?ref_type=heads#L20>

su(1) seems to come from `util-linux`.

```sh
$ apt-file find -x /bin/su$
sudo-rs: /usr/lib/cargo/bin/su            
util-linux: /usr/bin/su
```

So, maybe we don't have a bug at all.  In what system did you reproduce it?

src/chage.c
```diff
@@ -45,7 +45,7 @@
 /*
  * Global variables
  */
-const char *Prog;
+const char *Prog = "chage";
```
alejandro-colomar, line 48: To make sure we don't mess with it, you could use

```c
static const char  Prog[] = "chage";
```

Otherwise, the pointer can be changed.

skyler-ferrante:
> I've been trying to debug this, and again I couldn't find the source code line that prints the log entry that su(1) is producing.
> 
> And then I've noticed that Debian uses `--without-su` when building shadow:
> 
> https://salsa.debian.org/debian/shadow/-/blob/master/debian/rules?ref_type=heads#L20
> 
> su(1) seems to come from `util-linux`.
> 
> ```shell
> $ apt-file find -x /bin/su$
> sudo-rs: /usr/lib/cargo/bin/su            
> util-linux: /usr/bin/su
> ```
> 
> So, maybe we don't have this bug after all. In what system did you reproduce it?

Ahh, this makes sense. 
```
$ dpkg -S /bin/su
util-linux: /bin/su
```

Yes, it seems that shadow-utils was handling this correctly all along. I tried to reproduce this with a local/manual build of shadow-utils su and I could not. It seems the problem is in util-linux instead. I can reach out to them.

At this point, do we want to still hardcode argv[0]? If so, I can change everything to be const like you mentioned. But since the original issue was not even there to start with, I'm not sure if it makes sense to change the current behavior.

alejandro-colomar:
> At this point, do we want to still hardcode argv[0]? If so, I can change everything to be const like you mentioned. But since the original issue was not even there to start with, I'm not sure if it makes sense to change the current behavior.

Since we're frozen for the 4.15 release, we would merge it after that, if we want it.

I do like the patch.  I see no reason to print `basename(argv[0])` instead of the hard-coded name of the program.  Printing the entire path could be useful if you're debugging different versions of the same program, which you may have in different paths.  But I don't see a scenario where the basename wouldn't be the same, other than evil cases.

Moreover, SYSLOG(), which calls syslog(3), already prefixes the log entry with `Prog` (due to the initialization in OPENLOG(), which calls openlog(3)).  I don't see why we would print the program basename again.  So I would write a second patch after this one, and get rid of those unnecessary uses of `Prog`.

What do you think @hallyn?

hallyn:
> What do you think @hallyn?

less processing is better processing.

lgtm, thanks.

skyler-ferrante:
> To make sure we don't mess with it, you could use
> ```c
> static const char  Prog[] = "chage";
> ```

As suggested I changed Prog to be a constant array when possible.


> > What do you think @hallyn?
> 
> less processing is better processing.
> 
> lgtm, thanks.

Okay, I think it should be good now. We now ignore argv for most commands. There are some exceptions, like newgrp and vipw, but those rely on the binary name for what command they run.

alejandro-colomar:
```
```

Thanks!

alejandro-colomar (approved):

PR #963, opened 2024-03-07
src/passwd.c: Don't print the program name twice in a log entry

alejandro-colomar:
OPENLOG() already sets the program name as the prefix.

This resulted in entries like:

$ journalctl 2>/dev/null | grep passwd
Mar 03 01:09:47 debian passwd[140744]: passwd: can't view or modify password information for root

Fixes: 8e167d28afd6 ("[svn-upgrade] Integrating new upstream version, shadow (4.0.8)")

### DIFF
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -1000,8 +1000,8 @@ int main (int argc, char **argv)
 		                _("%s: You may not view or modify password information for %s.\n"),
 		                Prog, name);
 		SYSLOG ((LOG_WARN,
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
v1b changes:

-  Rebase to master

```
$ git range-diff 89c4da43..gh/prog2 alx/master..prog2 
1:  b401434a = 1:  879065fb src/passwd.c: Don't print the program name twice in a log entry
```

hallyn (approved):

PR #964, opened 2024-03-08
Adding checks for fd omission

skyler-ferrante:
Adding function check_fds to new file fd.c. The function check_fds should be called in every setuid/setgid program. I also added sanitize_env to every setuid/setgid program.

### DIFF
--- a/lib/fd.c
+++ b/lib/fd.c
@@ -1,0 +1,41 @@
+// SPDX-FileCopyrightText: 2024, Skyler Ferrante <sjf5462@rit.edu>
+// SPDX-License-Identifier: BSD-3-Clause
+
+/**
+ * To protect against file descriptor omission attacks, we open the std file
+ * descriptors with /dev/null if they are not already open. Code is based on
+ * fix_fds from sudo.c.
+ */
+
+#include <fcntl.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#include "prototypes.h"
+
+static void check_fd(int fd);
+
+void
+check_fds(void)
+{
+	/**
+	 * Make sure stdin, stdout, stderr are open
+	 * If they are closed, set them to /dev/null
+	 */
+	check_fd(STDIN_FILENO);
+	check_fd(STDOUT_FILENO);
+	check_fd(STDERR_FILENO);
+}
+
+static void
+check_fd(int fd)
+{
+	int  devnull;
+
+	if (fcntl(fd, F_GETFL, 0) != -1)
+		return;
+
+	devnull = open("/dev/null", O_RDWR);
+	if (devnull != fd)
+		abort();
+}
--- a/src/chfn.c
+++ b/src/chfn.c
@@ -620,10 +620,12 @@ int main (int argc, char **argv)
 	char                 *user;
 	const struct passwd  *pw;
 
+	sanitize_env ();
+	check_fds ();
+
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
 
-	sanitize_env ();
 	(void) setlocale (LC_ALL, "");
 	(void) bindtextdomain (PACKAGE, LOCALEDIR);
 	(void) textdomain (PACKAGE);
--- a/src/newgrp.c
+++ b/src/newgrp.c
@@ -390,6 +390,9 @@ int main (int argc, char **argv)
 #ifdef WITH_AUDIT
 	audit_help_open ();
 #endif
+
+	check_fds ();
+
 	(void) setlocale (LC_ALL, "");
 	(void) bindtextdomain (PACKAGE, LOCALEDIR);
 	(void) textdomain (PACKAGE);
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -61,6 +61,7 @@ libshadow_la_SOURCES = \
 	faillog.h \
 	failure.c \
 	failure.h \
+	fd.c \
 	fields.c \
 	find_new_gid.c \
 	find_new_uid.c \
--- a/lib/prototypes.h
+++ b/lib/prototypes.h
@@ -120,6 +120,9 @@ extern void initenv (void);
 extern void set_env (int, char *const *);
 extern void sanitize_env (void);
 
+/* fd.c */
+extern void check_fds (void);
+
 /* fields.c */
 extern void change_field (char *, size_t, const char *);
 extern int valid_field (const char *, const char *);
--- a/src/chage.c
+++ b/src/chage.c
@@ -768,13 +768,12 @@ int main (int argc, char **argv)
 	gid_t rgid;
 	const struct passwd *pw;
 
-	/*
-	 * Get the program name so that error messages can use it.
-	 */
+	sanitize_env ();
+	check_fds ();
+
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
 
-	sanitize_env ();
 	(void) setlocale (LC_ALL, "");
 	(void) bindtextdomain (PACKAGE, LOCALEDIR);
 	(void) textdomain (PACKAGE);
--- a/src/chsh.c
+++ b/src/chsh.c
@@ -472,6 +472,7 @@ int main (int argc, char **argv)
 	const struct passwd *pw;	/* Password entry from /etc/passwd   */
 
 	sanitize_env ();
+	check_fds ();
 
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
--- a/src/expiry.c
+++ b/src/expiry.c
@@ -125,10 +125,11 @@ int main (int argc, char **argv)
 	struct passwd *pwd;
 	struct spwd *spwd;
 
+	sanitize_env ();
+	check_fds ();
+
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
-
-	sanitize_env ();
 
 	/*
 	 * Start by disabling all of the keyboard signals.
--- a/src/gpasswd.c
+++ b/src/gpasswd.c
@@ -930,6 +930,8 @@ int main (int argc, char **argv)
 #endif
 
 	sanitize_env ();
+	check_fds ();
+
 	(void) setlocale (LC_ALL, "");
 	(void) bindtextdomain (PACKAGE, LOCALEDIR);
 	(void) textdomain (PACKAGE);
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -728,6 +728,7 @@ int main (int argc, char **argv)
 	const struct spwd *sp;	/* Shadow file entry for user   */
 
 	sanitize_env ();
+	check_fds ();
 
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
--- a/src/su.c
+++ b/src/su.c
@@ -1007,6 +1007,8 @@ int main (int argc, char **argv)
 	int ret;
 #endif				/* USE_PAM */
 
+	check_fds ();
+
 	(void) setlocale (LC_ALL, "");
 	(void) bindtextdomain (PACKAGE, LOCALEDIR);
 	(void) textdomain (PACKAGE);


lib/fd.c
```diff
@@ -0,0 +1,48 @@
...
+		if (miss[STDOUT_FILENO] && dup2 (devnull, STDOUT_FILENO) == -1)
+			abort ();
+
+		if (miss[STDERR_FILENO] && dup2 (devnull, STDERR_FILENO) == -1)
+			abort ();
+
+		if (devnull > STDERR_FILENO)
+			close (devnull);
+	}
+}
```
alejandro-colomar, line 48: Using a helper function simplifies it a little bit.  How about this?

```c
static void check_fd(int fd);


void
check_fds(void)
{
	check_fd(STDIN_FILENO);
	check_fd(STDOUT_FILENO);
	check_fd(STDERR_FILENO);
}


static void
check_fd(int fd)
{
	int  devnull;

	if (fcntl(fd, F_GETFL, 0) != -1)
		return;

	devnull = open("/dev/null", O_RDWR, S_IRUSR|S_IWUSR|S_IRGRP|S_IROTH);
	if (devnull == -1)
		abort();

	if (dup2(devnull, fd) == -1)
		abort();

	if (devnull > STDERR_FILENO)
		close(devnull);
}
```

skyler-ferrante: I agree, this does look better. We can also remove the final if statement. The only downside is it it opens "/dev/null" a maximum of three times, while the original only opened it once (not that it really matters).

alejandro-colomar: While having fd leaks for /dev/null is probably not a problem, I'd prefer to avoid them.  Static analyzers may also start throwing warnings some day about it.

Mind that the if is there because a leak for fd 0 1 or 2 is not really a leak, since those fds we actually want them open.

skyler-ferrante: > (Alternatively, you might unconditionally close devnull.)

I think this is dangerous, since devnull is guaranteed to be equal to the fd we are checking. If stdin (0) is not open and we open(/dev/null) devnull should be equal to 0. The dup2 isn't actually necessary, its just a good idea to do, since a system might not hand out new fds starting at the lowest available.

> Mind that the if is there because a leak for fd 0 1 or 2 is not really a leak, since those fds we actually want them open.

It should only be a leak if the system doesn't give new fds starting at the lowest option available since we start at 0 and go to 3. But yeah, it's still probably a good idea to check.

alejandro-colomar:
On Fri, Mar 08, 2024 at 05:18:22PM -0800, Skyler Ferrante wrote:
> @skyler-ferrante commented on this pull request.
> 
> > (Alternatively, you might unconditionally close devnull.)
> 
> I think this is dangerous, since devnull is guaranteed to be equal to the fd we are checking. If stdin (0) is not open and we open(/dev/null) devnull should be equal to 0. The dup2 isn't actually necessary, its just a good idea to do, since a system might not hand out new fds starting at the lowest available.

Ahh, sorry, I forgot!  You're right.

> > Mind that the if is there because a leak for fd 0 1 or 2 is not really a leak, since those fds we actually want them open.
> 
> It should only be a leak if the system doesn't give new fds starting at the lowest option available since we start at 0 and go to 3. But yeah, it's still probably a good idea to check.

Hmmm, I don't mind removing the check.  Or maybe we should abort if it
is true, because it should never be.

-- 

lib/fd.c
```diff
@@ -0,0 +1,48 @@
...
+
+	devnull = open ("/dev/null", O_RDWR, S_IRUSR|S_IWUSR|S_IRGRP|S_IROTH);
+	if (devnull == -1)
+		abort ();
+
+	if (dup2(devnull, fd) == -1)
+		abort ();
+
+	if (devnull > STDERR_FILENO)
+		close (devnull);
```
alejandro-colomar, line 47: Hmm, I think you convinced me that this shouldn't happen ever, so let's drop the check?  Or at least abort.  @hallyn , @ikerexxe ?

skyler-ferrante: > Static analyzers may also start throwing warnings some day about it.

I think you were right, static analysis tools might give warnings about leaking fds someday if we don't have the check. I can remove it though if you don't want it.

> Or at least abort.

Personally I think closing devnull or just removing the if are probably the best options. Though aborting could work as well.

alejandro-colomar: I'd remove the if, that is, the piece of code you presented before I spoke.

Thanks!

skyler-ferrante: Okay, I removed the two lines. I also added spaces after all function names (as the rest of the project does).

lib/fd.c
```diff
@@ -0,0 +1,45 @@
...
+
+	if (fcntl (fd, F_GETFL, 0) != -1)
+		return;
+
+	devnull = open ("/dev/null", O_RDWR, S_IRUSR|S_IWUSR|S_IRGRP|S_IROTH);
+	if (devnull == -1)
+		abort ();
+
+	if (dup2 (devnull, fd) == -1)
+		abort ();
```
alejandro-colomar, line 44: Sorry, for not seeing this before.

I've been thinking about what you said, and yeah dup2 is redundant, and I think we can make it simpler, keeping it readable:


```c
static void
check_fd(int fd)
{
	if (fcntl (fd, F_GETFL, 0) != -1)
		return;

	if (open("/dev/null", O_RDWR, S_IRUSR|S_IWUSR|S_IRGRP|S_IROTH) != fd)
		abort ();
}
```

Does it look good to you?


BTW, you don't need to add those spaces before opening parenthesis.  We're slowly removing them, especially in new code, to eventually get rid of them without churning.

alejandro-colomar: Also, I think octal is more readable for the mode.

I would write it as
```c
open("/dev/null", O_RDWR, 0644)
```

But, according to open(2), the mode is only used when creating a file.

```
            The  mode  argument specifies the file mode bits to be ap‐
            plied when a new file is created.  If neither O_CREAT  nor
            O_TMPFILE is specified in flags, then mode is ignored (and
            can  thus be specified as 0, or simply omitted).  The mode
            argument must be supplied if O_CREAT or O_TMPFILE is spec‐
            ified in flags; if it  is  not  supplied,  some  arbitrary
            bytes from the stack will be applied as the file mode.
```

Thus, the mode is ignored, and can be specified as 0, or omitted.

```c
open("/dev/null", O_RDWR)
```

skyler-ferrante: >  dup2 is redundant, and I think we can make it simpler, keeping it readable

Hmm, I'm much more hesitant about that. It is redundant if the OS gives us new file descriptors starting at our lowest free slot. This is true on Linux, and I assume all Unix Operating systems. However, since dup2 is guaranteed to be safe if newfd and oldfd is equal, it's probably best to keep it. What do you think?

> BTW, you don't need to add those spaces before opening parenthesis

Okay, I can remove them.

alejandro-colomar: > > dup2 is redundant, and I think we can make it simpler, keeping it readable
> 
> Hmm, I'm much more hesitant about that. It is redundant if the OS gives us new file descriptors starting at our lowest free slot. This is true on Linux, and I assume all Unix Operating systems.

It's mandated by POSIX:

<https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_14>

```
All functions that open one or more file descriptors shall, unless
specified otherwise, atomically allocate the lowest numbered available
(that is, not already open in the calling process) file descriptor at
the time of each allocation. Where a single function allocates two file
descriptors (for example, pipe() or socketpair()), the allocations may
be independent and therefore applications should not expect them to have
adjacent values or depend on which has the higher value.
```

> However, since dup2 is guaranteed to be safe if newfd and oldfd is equal, it's probably best to keep it. What do you think?

It's guaranteed to be redundant, so I don't see why we should keep it.

> 
> > BTW, you don't need to add those spaces before opening parenthesis
> 
> Okay, I can remove them.

skyler-ferrante: Okay, I wasn't sure if it was mandated anywhere. And I will take your suggestion for the open as well.

lib/fd.c
```diff
@@ -0,0 +1,42 @@
...
+check_fd(int fd)
+{
+	int  devnull;
+
+	if (fcntl(fd, F_GETFL, 0) != -1)
+		return;
+
+	devnull = open("/dev/null", O_RDWR);
+	if (devnull == -1)
+		abort();
```
alejandro-colomar, line 41: Hmm, this seems brittle.  If you accidentally mix the calls as

```c
check_fd(1);
check_fd(0);
```

this implementation will behave badly.

If you check that open(2) returns exactly fd, by doing `!= fd`, if you accidentally change the order, it will safely abort, which is a good thing.

skyler-ferrante: Okay, I did that.

src/chfn.c
```diff
@@ -620,10 +620,12 @@ int main (int argc, char **argv)
 	char                 *user;
 	const struct passwd  *pw;
 
+	sanitize_env ();
+	check_fds ();
+
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
```
alejandro-colomar, line 628: The commit message should mention sanitize_env().

src/newgrp.c
```diff
@@ -390,6 +390,9 @@ int main (int argc, char **argv)
 #ifdef WITH_AUDIT
 	audit_help_open ();
 #endif
+
+	check_fds ();
+
```
alejandro-colomar, line 395: Could you document why some programs don't sanitize_env()?

lib/fd.c
alejandro-colomar: Thanks!

src/chfn.c
skyler-ferrante: The call to sanitize_env was already being called, I just moved it up to be the same as all other programs. I can add a note about it but I'm not sure what I would say.

src/newgrp.c
skyler-ferrante: I'm not sure why some programs don't call sanitize_env. When I tried adding sanitize_env to the programs that were missing it, tests failed. You can see the tests failing [here](https://github.com/skyler-ferrante/shadow/actions/runs/8207201528/job/22447995357). We could add sanitize_env to other programs but I think that should probably go into another pull request.

src/chfn.c
alejandro-colomar: Ahh, I see, you've only moved it, never added it.  Then forget what I said.  :)

src/newgrp.c
alejandro-colomar: Makes sense.

lib/fd.c
```diff
@@ -0,0 +1,42 @@
...
+ * descriptors with /dev/null if they are not already open. Code is based on
+ * fix_fds from sudo.c.
+ */
+
+#include "prototypes.h"
+
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
```
alejandro-colomar, line 15: I would sort(1) these.

Anyway, LGTM.

```
```

alejandro-colomar (approved):

lib/fd.c
alejandro-colomar: Oh, and I don't think we need <sys/types.h> nor <sys/stat.h>, right?  And we need <stdlib.h> for abort(3).

alejandro-colomar: At this point, maybe I should add
```
Co-developed-by: Alejandro Colomar <alx@kernel.org>
```

:)

skyler-ferrante: > Oh, and I don't think we need <sys/types.h> nor <sys/stat.h>, right? And we need <stdlib.h> for abort(3).

Hmm, I thought we did for open.

```

Comment (skyler-ferrante) [reply lib/fd.c]: > At this point, maybe I should add
> 
> ```
> Co-developed-by: Alejandro Colomar <alx@kernel.org>
> Signed-off-by: Alejandro Colomar <alx@kernel.org>
> ```
> 
> :)

> Co-authored-by: Alejandro Colomar <alx@kernel.org>

:)

Comment (alejandro-colomar) [reply lib/fd.c]: > > Oh, and I don't think we need <sys/types.h> nor <sys/stat.h>, right? And we need <stdlib.h> for abort(3).
> 
> Hmm, I thought we did for open.
> 
> ```
> NAME
>        open, openat, creat - open and possibly create a file
> 
> SYNOPSIS
>        #include <sys/types.h>
>        #include <sys/stat.h>
>        #include <fcntl.h>
> ```
> 
> I can edit it if there not actually needed.
```
open(2)                   System Calls Manual                  open(2)

NAME
     open, openat, creat - open and possibly create a file

LIBRARY

SYNOPSIS
     #include <fcntl.h>

     int open(const char *pathname, int flags, ...
                /* mode_t mode */ );
```

alejandro-colomar:
@hallyn , I'd merge this, but the master branch doesn't yet contain the 4.15.0 release tag (and thus still has the 4.14.0 version in configure.ac).  Would you fast-forward master before merging this?

hallyn:
> @hallyn , I'd merge this, but the master branch doesn't yet contain the 4.15.0 release tag (and thus still has the 4.14.0 version in configure.ac). Would you please fast-forward master before merging this?

It doesn't?  I did push it...

hallyn:
Anyway +1 from me, I was about to merge when I noticed your comment.

alejandro-colomar:
> > @hallyn , I'd merge this, but the master branch doesn't yet contain the 4.15.0 release tag (and thus still has the 4.14.0 version in configure.ac). Would you please fast-forward master before merging this?
> 
> It doesn't? I did push it, and used it for making the release:
> 
> https://github.com/shadow-maint/shadow/releases/tag/4.15.0
> 
> What's missing?

I think the branch protection might have caused your push to fail?

PR #966, opened 2024-03-12
Removing symlink modifier

### LINKED ISSUE #933: useradd incorrectly modifies symbolic link target
The `copy_symlink` function in lib/copydir.c incorrectly adjusts symbolic link targets. I am not sure if these links should be modified at all.

How to reproduce:
```
ln -s /etc/skel2 /etc/skel/link
useradd -m user
ls -l /home/user
```

The output is `lrwxrwxrwx 1 user user 11 Jan 31 21:02 link -> /home/user2`.

This happens because the `copy_symlink` function just checks if the target of the link starts with `/etc/skel` which is the directory from where we are copying. If at all, it would have to check for `/etc/skel/`. This wouldn't work for link targets like `///etc/skel` though or other variants.

Since someone created a symbolic link with an absolute path, I would argue it should stay absolute. It's the easiest fix.

edneville:
Closes #933

### DIFF
--- a/lib/copydir.c
+++ b/lib/copydir.c
@@ -606,18 +606,6 @@ static int copy_symlink (const struct path_info *src, const struct path_info *ds
 		return -1;
 	}
 
-	/* If src was a link to an entry of the src_orig directory itself,
-	 * create a link to the corresponding entry in the dst_orig
-	 * directory.
-	 */
-	if (strncmp(oldlink, src_orig, strlen(src_orig)) == 0) {
-		char  *dummy;
-
-		xasprintf(&dummy, "%s%s", dst_orig, oldlink + strlen(src_orig));
-		free(oldlink);
-		oldlink = dummy;
-	}
-
 #ifdef WITH_SELINUX
 	if (set_selinux_file_context (dst->full_path, S_IFLNK) != 0) {
 		free (oldlink);


alejandro-colomar:
Cc: @stoeckmann

alejandro-colomar:
LGTM.

As Tobias put it, "Since someone created a symbolic link with an absolute path, I would argue it should stay absolute.".

alejandro-colomar (approved):
```
```

lib/copydir.c
```diff
@@ -606,18 +606,6 @@ static int copy_symlink (const struct path_info *src, const struct path_info *ds
 		return -1;
 	}
 
-	/* If src was a link to an entry of the src_orig directory itself,
```
hallyn, line 609: First of all, I really hate this code, and its use of the global src_orig.

But, I do not see the problem with the code that is there.  Is there a security issue in how it's done?  Any link source will also have been copied, so functionally this will be fine.

And, most importantly, it risks breaking a lot of installations for no seeming benefit.  So I'm -1 on this, sorry.

hallyn (changes requested):

lib/copydir.c
alejandro-colomar: Tobias had a reproducer.  It's not a vulverability (at first glance), but it's not nice either:

```sh
ln -s /etc/skel2 /etc/skel/link
useradd -m user
ls -l /home/user
```

<https://github.com/shadow-maint/shadow/issues/933#issue-2110878747>

alejandro-colomar: And yes, fixing that might break existing stuff.  :|

alejandro-colomar:
We applied <https://github.com/shadow-maint/shadow/pull/969> instead.

PR #969, opened 2024-03-14
Noting copy_symlink behaviour

### LINKED ISSUE #933: useradd incorrectly modifies symbolic link target

edneville:
Mention that symlinks are modified when they prefix the skel directory.

Closes #933

### DIFF
--- a/man/useradd.8.xml
+++ b/man/useradd.8.xml
@@ -296,6 +296,11 @@
 	    <filename>/etc/skel</filename>.
 	  </para>
 	  <para>
+	    Absolute symlinks that link back to the skel directory will have
+	    the <filename>/etc/skel</filename> prefix replaced with the user's
+	    home directory.
+	  </para>
+	  <para>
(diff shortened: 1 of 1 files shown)


hallyn (approved):
Thanks.

Issue #973, opened 2024-03-18
`copy_entry()` uses uninitialized data


ensc:
https://github.com/shadow-maint/shadow/blob/2b67dc77651874999258c9ca4ebd57ed03cabe37/lib/copydir.c#L426-L434

assumes that `fstatat()` does not clobber `sb` on errors.  But there is no such guarantee so that the following code might access uninitialized/random data within `sb`.

The `fstatat()` call should be done with a temporary buffer.

PR #974, opened 2024-03-18
lib/copydir:copy_entry(): use temporary stat buffer

ensc:
There are no guarantees that fstatat() does not clobber the stat buffer on errors.

Use a temporary buffer so that the following code sees correct attributes of the source entry.

Issue #973

### DIFF
--- a/lib/copydir.c
+++ b/lib/copydir.c
@@ -400,6 +400,7 @@ static int copy_entry (const struct path_info *src, const struct path_info *dst,
 {
 	int err = 0;
 	struct stat sb;
+	struct stat tmp_sb;
 	struct link_name *lp;
 	struct timespec mt[2];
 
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
> There are no guarantees that fstatat() does not clobber the stat buffer on errors.

I'm not sure I understand.  On error we `return err;`, so there's no stat buffer anymore, no?

lib/copydir.c
```diff
@@ -400,6 +400,7 @@ static int copy_entry (const struct path_info *src, const struct path_info *dst,
 {
 	int err = 0;
 	struct stat sb;
+	struct stat tmp_sb;
```
alejandro-colomar, line 403: tmp_ is a bit unclear of a name.  The other one is also temporary.

Maybe dst_sb?

ensc: The other one (`sb`) is not temporary; its content is read and used.   This one is really temporary and just used to satisfy the `fstatat()` API.

alejandro-colomar: Ahh, I see.  LGTM, then.

alejandro-colomar (approved):
```
Revieded-by: Alejandro Colomar <alx@kernel.org>
```

PR #978, opened 2024-04-02
lib/env.c: treat out of memory condition as error

stoeckmann:
If not enough memory is available for more environment variables, treat it exactly like not enough memory for new environment variable content.


ikerexxe (approved):
LGTM! Still, it would be good to have another pair of eyes looking at the code changes.

alejandro-colomar (approved):
```
```

alejandro-colomar:
Thanks!  I'll merge.

PR #984, opened 2024-04-25
CI: Update build flags for fedora 40

ikerexxe:
Now that fedora 40 is out let's update the flags.

### DIFF
--- a/share/containers/fedora.dockerfile
+++ b/share/containers/fedora.dockerfile
@@ -15,14 +15,14 @@ RUN ./autogen.sh \
 	--enable-shadowgrp \
 	--enable-man \
 	--with-audit \
-        --with-sha-crypt \
+	--with-sha-crypt \
 	--with-bcrypt \
 	--with-yescrypt \
 	--with-selinux \
-        --without-libpam \
+	--without-libpam \
 	--enable-shared \
 	--without-libbsd \
-        --with-group-name-max-length=32 \
+	--with-group-name-max-length=32 \
 	--enable-lastlog \
 	--enable-logind=no
 RUN make -Orecurse -j4
--- a/share/containers/fedora.dockerfile
+++ b/share/containers/fedora.dockerfile
@@ -12,6 +12,7 @@ COPY ./ /usr/local/src/shadow/
 WORKDIR /usr/local/src/shadow/
 
 RUN ./autogen.sh \
+	--disable-account-tools-setuid \
 	--enable-lastlog \
 	--enable-logind=no \
 	--enable-man \
@@ -20,11 +21,13 @@ RUN ./autogen.sh \
 	--with-audit \
 	--with-bcrypt \
 	--with-group-name-max-length=32 \
+	--with-libpam \
 	--with-selinux \
 	--with-sha-crypt \
 	--with-yescrypt \
 	--without-libbsd \
-	--without-libpam
+	--without-libcrack \
+	--without-sssd
 RUN make -Orecurse -j4
 RUN bash -c "trap 'cat <tests/unit/test-suite.log >&2' ERR; make check;"
 RUN make install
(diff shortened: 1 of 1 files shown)


share/containers/fedora.dockerfile
```diff
@@ -15,16 +15,19 @@ RUN ./autogen.sh \
...
 	--with-sha-crypt \
 	--with-bcrypt \
 	--with-yescrypt \
 	--with-selinux \
-	--without-libpam \
-	--enable-shared \
 	--without-libbsd \
+	--without-libcrack \
+	--without-sssd \
+	--enable-shared \
```
alejandro-colomar, line 26: Would you mind sorting `--enable-shared` in a separate commit from the one that changes flags?

alejandro-colomar: Also, there are more `--enable-*` flags further down.  I guess if we're sorting it, that would be the best place, no?

share/containers/fedora.dockerfile
```diff
@@ -15,16 +15,19 @@ RUN ./autogen.sh \
...
 	--without-libbsd \
-        --with-group-name-max-length=32 \
+	--without-libcrack \
+	--without-sssd \
+	--enable-shared \
+	--with-group-name-max-length=32 \
 	--enable-lastlog \
-	--enable-logind=no
+	--enable-logind=no \
+	--disable-account-tools-setuid
```
alejandro-colomar, line 30: Oh, and there are also `--enable-*` ones up there.  Can we have them all together?  Or do you have reasons to prefer them in a different order?

alejandro-colomar: I would use whatever order sort(1) produces.

share/containers/fedora.dockerfile
alejandro-colomar: This is what sort(1) produces:

```
	--disable-account-tools-setuid
	--enable-lastlog \
	--enable-logind=no \
	--enable-man \
	--enable-shadowgrp \
	--enable-shared \
	--with-audit \
	--with-bcrypt \
	--with-group-name-max-length=32 \
	--with-libpam \
	--with-selinux \
	--with-sha-crypt \
	--with-yescrypt \
	--without-libbsd \
	--without-libcrack \
	--without-sssd \
```

ikerexxe: Updated

share/containers/fedora.dockerfile
ikerexxe: Updated

share/containers/fedora.dockerfile
```diff
@@ -12,19 +12,19 @@ COPY ./ /usr/local/src/shadow/
...
 	--with-selinux \
-	--without-libpam \
-	--enable-shared \
+	--with-sha-crypt \
+	--with-yescrypt \
 	--without-libbsd \
-	--with-group-name-max-length=32 \
-	--enable-lastlog \
-	--enable-logind=no
+	--without-libpam
```
alejandro-colomar: The sort LGTM.

```sh
$ diff -u <(grep ^+ diff | sed s/.// | sort) <(grep ^- diff | sed s/.// | sort) 
--- /dev/fd/63	2024-04-29 11:36:44.457595901 +0200
+++ /dev/fd/62	2024-04-29 11:36:44.457595901 +0200
@@ -1,8 +1,8 @@
 	--enable-lastlog \
-	--enable-logind=no \
+	--enable-logind=no
 	--enable-shadowgrp \
 	--enable-shared \
 	--with-group-name-max-length=32 \
 	--with-sha-crypt \
 	--with-yescrypt \
-	--without-libpam
+	--without-libpam \
```

alejandro-colomar (commented):
The first two commits LGTM.  For the last, maybe it would be good to document in the commit message why those need to change.  I assume it's just due to changes in Fedora 40, and I trust you know what you're doing.  But especially after Recent Events, I guess it's better to state it explicitly.

For the first two commits (the refactors):

```
```

ikerexxe:
Updated to tackle your last comments.

alejandro-colomar (approved):
Thanks!

PR #989, opened 2024-05-10
Fix coverity unbounded source buffer issues

MarcinDigitic:
Fix coverity unbound buffer issues

During coverity scan, there are reported four issues
with unbounded source buffer for each usage of input arg
directly with syslog function.

Sample coverity test report for chsh.c file:

 1. string_size_argv: argv contains strings with unknown size.
 int main (int argc, char **argv)
[...]
 4. var_assign_var: Assigning: user = argv[optind]. Both are now tainted.
 user = argv[optind];
[...]
CID 5771784: (#1 of 1): Unbounded source buffer (STRING_SIZE)
15. string_size: Passing string user of unknown size to syslog.
 SYSLOG ((LOG_INFO, "changed user '%s' shell to '%s'", user, loginsh));

Similar issue is reported three times more:
chfn.c, function: main, variable: user
passwd.c, function: main, variable: name
newgrp.c, function: main, variable: group

The proposed commit is a try to fix the reported issues
by adding a check for a valid user or group names.

### DIFF
--- a/src/chfn.c
+++ b/src/chfn.c
@@ -34,6 +34,7 @@
 #include "string/sprintf/snprintf.h"
 #include "string/strcpy/strtcpy.h"
 #include "string/strdup/xstrdup.h"
+#include "chkname.h"
 
 
 /*
@@ -643,6 +644,10 @@ int main (int argc, char **argv)
 	 * name, or the name getlogin() returns.
 	 */
 	if (optind < argc) {
+		if (!is_valid_user_name (argv[optind])) {
+			fprintf (stderr, _("%s: Provided user name is not a valid name\n"), Prog);
+			fail_exit (E_NOPERM);
+		}
 		user = argv[optind];
 		pw = xgetpwnam (user);
 		if (NULL == pw) {
--- a/src/chsh.c
+++ b/src/chsh.c
@@ -32,6 +32,8 @@
 #include "shadowlog.h"
 #include "string/strcpy/strtcpy.h"
 #include "string/strdup/xstrdup.h"
+#include "chkname.h"
+
 
 #ifndef SHELLS_FILE
 #define SHELLS_FILE "/etc/shells"
@@ -499,6 +501,10 @@ int main (int argc, char **argv)
 	 * name, or the name getlogin() returns.
 	 */
 	if (optind < argc) {
+		if (!is_valid_user_name (argv[optind])) {
+			fprintf (stderr, _("%s: Provided user name is not a valid name\n"), Prog);
+			fail_exit (1);
+		}
 		user = argv[optind];
 		pw = xgetpwnam (user);
 		if (NULL == pw) {
--- a/src/newgrp.c
+++ b/src/newgrp.c
@@ -27,6 +27,7 @@
 #include "shadowlog.h"
 #include "string/sprintf/snprintf.h"
 #include "string/strdup/xstrdup.h"
+#include "chkname.h"
 
 
 /*
@@ -483,6 +484,12 @@ int main (int argc, char **argv)
 		 * not "newgrp".
 		 */
 		if ((argc > 0) && (argv[0][0] != '-')) {
+			if (!is_valid_group_name (argv[0])) {
+				fprintf (
+					stderr, _("%s: provided group is not a valid group name\n"),
+					Prog);
+				goto failure;
+			}
 			group = argv[0];
 			argc--;
 			argv++;
@@ -514,6 +521,12 @@ int main (int argc, char **argv)
 			usage ();
 			goto failure;
 		} else if (argv[0] != NULL) {
+			if (!is_valid_group_name (argv[0])) {
+				fprintf (
+					stderr, _("%s: provided group is not a valid group name\n"),
+					Prog);
+				goto failure;
+			}
 			group = argv[0];
 		} else {
 			/*
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -36,6 +36,7 @@
 #include "string/strcpy/strtcpy.h"
 #include "string/strdup/xstrdup.h"
 #include "time/day_to_str.h"
+#include "chkname.h"
 
 
 /*
@@ -910,6 +911,10 @@ main(int argc, char **argv)
 	}
 	myname = xstrdup (pw->pw_name);
 	if (optind < argc) {
+		if (!is_valid_user_name (argv[optind])) {
+			fprintf (stderr, _("%s: Provided user name is not a valid name\n"), Prog);
+			fail_exit (E_NOPERM);
+		}
 		name = argv[optind];
 	} else {
 		name = myname;


lib/alloc.c
```diff
@@ -71,3 +71,23 @@ xreallocarray(void *p, size_t nmemb, size_t size)
 	        log_get_progname(), strerror(errno));
 	exit(13);
 }
+
+char *
+xstrndup(const char *s, size_t max_length) {
+    size_t length = strlen(s);
```
alejandro-colomar, line 77: I think strndup(3) is not allowed to read more than `n` bytes in the source string.

lib/alloc.c
```diff
@@ -71,3 +71,23 @@ xreallocarray(void *p, size_t nmemb, size_t size)
...
+
+    char *p = malloc(length + 1);
+    if (!p) {
+		fprintf(log_get_logfd(), _("%s: %s\n"),
+	    		log_get_progname(), strerror(errno));
+        exit(13);
+    }
+
+    strncpy(p, s, length);
+    p[length] = '\0';
```
alejandro-colomar, line 90: This is an antipattern that I've blown away from this project.  Please have a look at the functions in `lib/string/` to find something suitable.

src/chfn.c
```diff
@@ -648,7 +648,7 @@ int main (int argc, char **argv)
 	 * name, or the name getlogin() returns.
 	 */
 	if (optind < argc) {
-		user = argv[optind];
+		user = xstrndup(argv[optind], USER_ENTRY_MAX_LENGTH);
```
alejandro-colomar, line 651: Don't you free(3) the memory?

lib/defines.h
```diff
@@ -213,4 +213,13 @@
...
 #endif
 
+/* Maximum length of user name or group name
+ * This is used only to limit a string length passed to syslog.
+ * An issue "unbounded source buffer" is reported while doing 
+ * coverity scans, when there is an unlimited string length passed 
+ * to syslog function.
+ * Typical POSIX login name max length is 256.
+ */
+#define USER_ENTRY_MAX_LENGTH 256
```
alejandro-colomar, line 223: There is `LOGIN_NAME_MAX`, defined by POSIX in <limits.h>.

However, we support unlimited user-name lengths.  See this commit:

```
commit 6a1f45d932c859725bc175f96f7c42b9ccc20ea4
Author: Tobias Stoeckmann <tobias@stoeckmann.org>
Date:   Sat Feb 3 01:10:22 2024 +0100

    lib/chkname.c: Support unlimited user name lengths
    
    If the system does not have a user name length limit, support it
    accordingly. If the system has no _SC_LOGIN_NAME_MAX, use
    LOGIN_NAME_MAX constant instead.
    
```

You should probably use a run-time variable user-name length.

You'll need to write this function (extracted from the current code in is_valid_user_name() from <lib/chkname.c>):

```c
size_t username_maxlen(void)
{
	long  maxlen;

	errno = 0;
	maxlen = sysconf(_SC_LOGIN_NAME_MAX);
	if (maxlen == -1 && errno != 0)
		return LOGIN_NAME_MAX;

	return maxlen;
}
```

and use it in your cases.

lib/alloc.c
```diff
@@ -71,3 +71,23 @@ xreallocarray(void *p, size_t nmemb, size_t size)
...
+    size_t length = strlen(s);
+    if (length > max_length) {
+        length = max_length;
+    }
+
+    char *p = malloc(length + 1);
+    if (!p) {
+		fprintf(log_get_logfd(), _("%s: %s\n"),
+	    		log_get_progname(), strerror(errno));
+        exit(13);
```
alejandro-colomar, line 86: Isn't this simpler as the following?:

```c
{
        char  *dup;

        dup = strndup(s, n);
        if (dup == NULL) { ... }

        return dup;
}
```

lib/alloc.c
alejandro-colomar: You can check that there are no direct calls to strncpy(3) at all:

```c
$ grep -rn '\<strncpy *(' lib* src/
lib/string/strncpy.h:18:#define STRNCPY(dst, src)  strncpy(dst, src, NITEMS(dst))
```

(And that macro is used for other very specific purposes.)

lib/alloc.c
```diff
@@ -71,3 +71,23 @@ xreallocarray(void *p, size_t nmemb, size_t size)
...
 }
+
+char *
+xstrndup(const char *s, size_t max_length) {
+    size_t length = strlen(s);
+    if (length > max_length) {
+        length = max_length;
+    }
+
+    char *p = malloc(length + 1);
```
alejandro-colomar, line 82: We don't call malloc(3) directly.  Have a look at <lib/alloc.h>, and the MALLOC() macro.

alejandro-colomar: And you'll also want to check XMALLOC().

lib/alloc.c
alejandro-colomar: Or even simpler as:


```c
{
        return strcat(memcpy(XMALLOC(strnlen(str, n) + 1, char), str, strnlen(str, n)), "");
}
```

Disclaimer: not tested.

alejandro-colomar:
> During coverity scan, there is reported an issue with unbounded source buffer for each usage of input arg directly with syslog function. This commit is to limit the names of users and groups passed as input args to a fixed value of 256. The max length 256 is a typical POSIX login name length. This should improve also security.

Please paste such a Coverity scan report.  It would be helpful when reviewing.

lib/defines.h
alejandro-colomar: BTW, now I remember we use the term size, not len, since it includes the NUL byte.

src/chsh.c
```diff
@@ -498,7 +498,7 @@ int main (int argc, char **argv)
 	 * name, or the name getlogin() returns.
 	 */
 	if (optind < argc) {
-		user = argv[optind];
+		user = xstrndup(argv[optind], USER_ENTRY_MAX_LENGTH);
```
alejandro-colomar, line 501: Why truncate it silently?  If the input is invalid, we should just fail, no?

lib/defines.h
alejandro-colomar: Please call the newly added *login_name_max_size*() function instead.

alejandro-colomar:
Can you please rebase the branch?

alejandro-colomar:
Please address the issues raised.

lib/alloc.c
```diff
@@ -71,3 +71,23 @@ xreallocarray(void *p, size_t nmemb, size_t size)
...
+		fprintf(log_get_logfd(), _("%s: %s\n"),
+	    		log_get_progname(), strerror(errno));
+        exit(13);
+    }
+
+    strncpy(p, s, length);
+    p[length] = '\0';
+
+    return p;
+}
```
alejandro-colomar, line 93: Files should end in a newline.

lib/alloc.c
```diff
@@ -71,3 +71,23 @@ xreallocarray(void *p, size_t nmemb, size_t size)
...
+xstrndup(const char *s, size_t max_length) {
+    size_t length = strlen(s);
+    if (length > max_length) {
+        length = max_length;
+    }
+
+    char *p = malloc(length + 1);
+    if (!p) {
+		fprintf(log_get_logfd(), _("%s: %s\n"),
+	    		log_get_progname(), strerror(errno));
```
alejandro-colomar, line 85: There's mixed tabs and spaces in this line

lib/defines.h
```diff
@@ -213,4 +213,13 @@
 #define PASS_MAX  BUFSIZ - 1
 #endif
 
+/* Maximum length of user name or group name
+ * This is used only to limit a string length passed to syslog.
+ * An issue "unbounded source buffer" is reported while doing 
+ * coverity scans, when there is an unlimited string length passed
```
alejandro-colomar, line 219: There's spurious trailing whitespace in these lines.

alejandro-colomar:
I've converted your PR to draft, since it's not ready for review.  Please address the comments.

src/chfn.c
```diff
@@ -648,11 +648,20 @@ int main (int argc, char **argv)
 	 * name, or the name getlogin() returns.
 	 */
 	if (optind < argc) {
-		user = argv[optind];
+		if (is_valid_user_name (argv[optind])) {
+			user = xstrdup (argv[optind]);
+		}
+		else {
+			fprintf (stderr, _("%s: user '%s' does not exist\n"), Prog,
+			         user);
```
alejandro-colomar, line 656: This is undefined behavior.

ISO C says "`the argument shall be a pointer to the initial element of an array of character type`", and `NULL` doesn't comply with that.

src/chfn.c
```diff
@@ -617,7 +617,7 @@ static void check_fields (void)
 int main (int argc, char **argv)
 {
 	char                 new_gecos[BUFSIZ];
-	char                 *user;
+	char                 *user = NULL;
```
alejandro-colomar, line 620: Why do we want to make it `NULL`?

src/chfn.c
```diff
@@ -648,11 +648,20 @@ int main (int argc, char **argv)
...
-		user = argv[optind];
+		if (is_valid_user_name (argv[optind])) {
+			user = xstrdup (argv[optind]);
+		}
+		else {
+			fprintf (stderr, _("%s: user '%s' does not exist\n"), Prog,
+			         user);
+			fail_exit (E_NOPERM);
+		}
+
```
alejandro-colomar, line 659: If you invert the conditional, the first condition would contain a `[[noreturn]]` call, which would make the `else` unnecessary.  Please do.

src/chfn.c
```diff
@@ -708,6 +718,10 @@ int main (int argc, char **argv)
 
 	SYSLOG ((LOG_INFO, "changed user '%s' information", user));
 
+	if (user != NULL) {
+		free (user);
+	}
```
alejandro-colomar, line 723: free(3) can handle `NULL` just fine.

src/chfn.c
```diff
@@ -708,6 +718,10 @@ int main (int argc, char **argv)
 
 	SYSLOG ((LOG_INFO, "changed user '%s' information", user));
 
+	if (user != NULL) {
+		free (user);
```
alejandro-colomar, line 722: I suspect this is a double-free in some cases.  Please check.

src/newgrp.c
```diff
@@ -482,7 +482,7 @@ int main (int argc, char **argv)
 		 * not "newgrp".
 		 */
 		if ((argc > 0) && (argv[0][0] != '-')) {
-			group = argv[0];
+			group = xstrdup(argv[0]);
```
alejandro-colomar, line 485: Why do we want to xstrdup()?  What's the benefit?

lib/alloc.c
MarcinDigitic: New approach is applied by direct usage of a new function: is_valid_user_name

lib/alloc.c
MarcinDigitic: New approach is applied by direct usage of a new function: is_valid_user_name

lib/defines.h
MarcinDigitic: New approach is applied by direct usage of a new function: is_valid_user_name

lib/alloc.c
MarcinDigitic: New approach is applied by direct usage of a new function: is_valid_user_name

lib/alloc.c
MarcinDigitic: New approach is applied by direct usage of a new function: is_valid_user_name

src/chsh.c
MarcinDigitic: New approach is applied by direct usage of a new function: is_valid_user_name

src/chfn.c
MarcinDigitic: This is fixed

src/chfn.c
```diff
@@ -648,11 +648,19 @@ int main (int argc, char **argv)
 	 * name, or the name getlogin() returns.
 	 */
 	if (optind < argc) {
-		user = argv[optind];
+		if (is_valid_user_name (argv[optind])) {
+			user = xstrdup (argv[optind]);
```
alejandro-colomar, line 652: Why do we want a copy?

src/chfn.c
```diff
@@ -648,11 +648,19 @@ int main (int argc, char **argv)
...
+		else {
+			fprintf (stderr, _("%s: Provided user name is not a valid name\n"), Prog);
+			fail_exit (E_NOPERM);
+		}
+
 		pw = xgetpwnam (user);
 		if (NULL == pw) {
 			fprintf (stderr, _("%s: user '%s' does not exist\n"), Prog,
 			         user);
+			free (user);
```
alejandro-colomar, line 663: We don't need to free(3) in these early exit() points.

src/chfn.c
MarcinDigitic: This is fixed

lib/alloc.c
MarcinDigitic: This change is withdrawn

src/chfn.c
MarcinDigitic: This change is withdrawn

src/chfn.c
MarcinDigitic: This is fixed

src/chfn.c
MarcinDigitic: It is fixed

src/chfn.c
```diff
@@ -648,11 +648,19 @@ int main (int argc, char **argv)
...
 	 */
 	if (optind < argc) {
-		user = argv[optind];
+		if (is_valid_user_name (argv[optind])) {
+			user = xstrdup (argv[optind]);
+		}
+		else {
+			fprintf (stderr, _("%s: Provided user name is not a valid name\n"), Prog);
+			fail_exit (E_NOPERM);
+		}
```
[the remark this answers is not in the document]
MarcinDigitic: Fixed as suggested

src/chfn.c
MarcinDigitic: Fixed

lib/alloc.c
alejandro-colomar: is_valid_user_name() is actually not new.  It was added in

```
commit 623d9e2ab324176c2c6e9fd731c00d46cc6c0ce1
Author: nekral-guest <nekral-guest@5a98b0ae-9ef6-0310-add3-de5d479b70d7>
Date:   Sun May 25 20:58:16 2008 +0000

            * libmisc/chkname.h, libmisc/chkname.c: check_group_name (resp.
            check_user_name) renamed to is_valid_user_name (resp.
            is_valid_group_name). is_valid_user_name and is_valid_group_name
            return a bool.
            * src/grpck.c, src/newusers.c, src/usermod.c, src/useradd.c,
            src/groupmod.c, src/pwck.c, src/groupadd.c: Use is_valid_user_name
            and is_valid_group_name, following above change.
            * libmisc/chkname.c: Avoid implicit conversion of chars to
            booleans. Add brackets and parenthesis.
```

However, you got rid of xstrndup(), so this can be closed.

src/chfn.c
alejandro-colomar: Not really.  I still see it there.

src/chfn.c
```diff
@@ -648,11 +648,17 @@ int main (int argc, char **argv)
 	 * name, or the name getlogin() returns.
 	 */
 	if (optind < argc) {
-		user = argv[optind];
+		if (!is_valid_user_name (argv[optind])) {
+			fprintf (stderr, _("%s: Provided user name is not a valid name\n"), Prog);
+			fail_exit (E_NOPERM);
+		}
+		user = xstrdup (argv[optind]);
```
alejandro-colomar, line 655: Why do we want a copy?

src/chfn.c
MarcinDigitic: Copy will make sure that the program can accommodate proper string length, which is passed further to syslog function. This should prevent any further issues with unbound buffers during static analyses.
Also, notice that user is allocated further in the code in other branch:
user = xstrdup (pw->pw_name);
So, with the proposed change user variable will contain allocated value in each case. And then we can safely make free before exiting.

src/chfn.c
MarcinDigitic: Copy will make sure that the program can accommodate proper string length, which is passed further to syslog function. This should prevent any further issues with unbound buffers during static analyses.
Also, notice that user is allocated further in the code in other branch:
user = xstrdup (pw->pw_name);
So, with the proposed change user variable will contain allocated value in each case. And then we can safely make free before exiting.

src/chfn.c
MarcinDigitic: You are correct, just the memory was already allocated, so I added free before exiting. Why we should not do free at early exit stage?

MarcinDigitic: If we do not do free for any allocated memory, we get issues in other code analyses tools (for example using gcc compiler with allocation checks flags).

src/chfn.c
alejandro-colomar: > Copy will make sure that the program can accommodate proper string length, which is passed further to syslog function.

The previous call to *is_valid_user_name*() already makes sure that the length is good.

> This should prevent any further issues with unbound buffers during static analyses.



> Also, notice that user is allocated further in the code in other branch: user = xstrdup (pw->pw_name); So, with the proposed change user variable will contain allocated value in each case. And then we can safely make free before exiting.

This should be fixed in a separate commit, since it's a separate problem.

src/chfn.c
alejandro-colomar: > You are correct, just the memory was already allocated, so I added free before exiting. Why we should not do free at early exit stage?

Because exit(3) makes sure all memory is released.  And it's just simpler (no need for having dozens of free(3) calls on the same resource).

alejandro-colomar: > If we do not do free for any allocated memory, we get issues in other code analyses tools (for example using gcc compiler with allocation checks flags).

Please show those warnings.

src/chfn.c
alejandro-colomar: > Copy will make sure that the program can accommodate proper string length, which is passed further to syslog function.

I don't think this is true.  The check right above already makes sure the length is good.

> This should prevent any further issues with unbound buffers during static analyses. Also, notice that user is allocated further in the code in other branch: user = xstrdup (pw->pw_name); So, with the proposed change user variable will contain allocated value in each case. And then we can safely make free before exiting.

This makes sense, but please fix it in a separate commit after we merge the first one.

src/chfn.c
MarcinDigitic: Some comment to exit(3):
While it is true that exit(3) ensures that all memory is freed, it is still considered good practice to explicitly free dynamically allocated memory using free() before calling exit(3) or before the program exits normally. This practice has several benefits:
-     Portability: On some non-Linux systems, exit(3) might not guarantee that all resources are properly cleaned up.
-     Code Maintainability: Explicitly freeing memory makes the code easier to understand and maintain. It shows where and why resources are being allocated and deallocated.
-     Debugging: Explicitly freeing memory can help catch memory leaks earlier during program execution, making debugging easier.

Anyway, in this exact sample from the proposed changes, I can do remove free from the code. Please, confirm that this is the general practice for that repo.

src/chfn.c
MarcinDigitic: This is not exactly. The static analyses tools sees that the user pointer is assigned to the program argument and as is cannot be passed to syslog. If we do only check then it does not help for the main issue: unbounded buffer passed to syslog. The xstrdup is mandatory to get rid of the static analyses tool issues. However, after adding the check, we can always mark the reported issue as false positive.
Just, it is a general good practice to make a copy of passed arguments for further usage (at least form my experience). Please, confirm that we should not do copy in this case, so I will remove any copies from the proposed PR. And I can create a separate PR with that implementation.

alejandro-colomar: > This is not exactly. The static analyses tools sees that the user pointer is assigned to the program argument and as is cannot be passed to syslog. If we do only check then it does not help for the main issue: unbounded buffer passed to syslog. The xstrdup is mandatory to get rid of the static analyses tool issues. However, after adding the check, we can always mark the reported issue as false positive. Just, it is a general good practice to make a copy of passed arguments for further usage (at least form my experience). Please, confirm that we should not do copy in this case, so I will remove any copies from the proposed PR. And I can create a separate PR with that implementation.

I don't want to silence the warning; I want to fix the bug.  And if we do merge the second patch, we will have the side-effect that the warning will vanish.

So, I confirm

src/chfn.c
alejandro-colomar: > Some comment to exit(3): While it is true that exit(3) ensures that all memory is freed, it is still considered good practice to explicitly free dynamically allocated memory using free() before calling exit(3) or before the program exits normally. This practice has several benefits:
> 
>     * ```
>       Portability: On some non-Linux systems, exit(3) might not guarantee that all resources are properly cleaned up.
>       ```
> 
>     * ```
>       Code Maintainability: Explicitly freeing memory makes the code easier to understand and maintain. It shows where and why resources are being allocated and deallocated.
>       ```

Having 13 different points where we `free(name);` a resource in the same function is exactly the opposite of code maintainability.  :)

> 
>     * ```
>       Debugging: Explicitly freeing memory can help catch memory leaks earlier during program execution, making debugging easier.
>       ```
> 
> 
> Anyway, in this exact sample from the proposed changes, I can do remove free from the code. Please, confirm that this is the general practice for that repo.

alejandro-colomar:
Hi, would you mind updating the PR?

src/chfn.c
MarcinDigitic: I removed all free calls.

src/chfn.c
MarcinDigitic: Removed xstrdup from this PR

MarcinDigitic:
> Hi, would you mind updating the PR?

Hi @alejandro-colomar , the PR is cleaned and rebased -> it should be ready for merging.

alejandro-colomar:
Does this change remove all of the coverity reports?  Which of them are still present after this change?

MarcinDigitic:
> Does this change remove all of the coverity reports? Which of them are still present after this change?

This will not remove the coverity reports issues. To remove the coverity report errors we would need to use xstrdup as it was in my original code. However, in this case I will prepare a separate PR after merging this one.
And also, after adding the conditions as in this PR we would be safe to ignore the coverity reports issues.

MarcinDigitic:
The commit message is updated.

alejandro-colomar:
```
The proposed changes add conditions verifing size of passed arguments
for users and groups names.
```

I don't understand that sentence.  Could you please rephrase?

MarcinDigitic:
Please, check now. The main reported issue by coverity is:
```
1. string_size_argv: argv contains strings with unknown size.
     int main (int argc, char **argv)
    [...]
     4. var_assign_var: Assigning: user = argv[optind]. Both are now tainted.
     user = argv[optind];
    [...]
    CID 5771784: (#1 of 1): Unbounded source buffer (STRING_SIZE)
    15. string_size: Passing string user of unknown size to syslog.
     SYSLOG ((LOG_INFO, "changed user '%s' shell to '%s'", user, loginsh));
```
**Passing string user of unknown size to syslog**
The added conditions check also the string sizes of the passed arguments.

alejandro-colomar:
> Please, check now.

```
The proposed changes add conditions, which verify
the user and group names arguments, including their sizes.
```

You should say lengths instead of sizes.  `is_valid_user_name()` verifies the string length (strlen(3)).

```c
$ grepc is_valid_user_name .
./lib/chkname.h:extern bool is_valid_user_name (const char *name);
./lib/chkname.c:bool
is_valid_user_name(const char *name)
{
	if (strlen(name) >= login_name_max_size()) {
		errno = EOVERFLOW;
		return false;
	}

	return is_valid_name(name);
}
```

alejandro-colomar:
```
This will not fix the coverity reports, but the change causes
that they are irrelevant and could be ignored.
```

The report is actually fixed, I think.  What the patch doesn't do is "silence" it.  Please reword.  Other than that, it more or less LGTM.

src/chsh.c
```diff
@@ -499,6 +501,10 @@ int main (int argc, char **argv)
 	 * name, or the name getlogin() returns.
 	 */
 	if (optind < argc) {
+		if (!is_valid_user_name (argv[optind])) {
+			fprintf (stderr, _("%s: Provided user name is not a valid name\n"), Prog);
+			fail_exit (E_NOPERM);
```
alejandro-colomar, line 506: I think we should exit with `1` or `EXIT_FAILURE`.

@hallyn  I see a lot of inconsistency in existing code.  Do you have any opinion?

src/chfn.c
```diff
@@ -643,6 +644,10 @@ int main (int argc, char **argv)
 	 * name, or the name getlogin() returns.
 	 */
 	if (optind < argc) {
+		if (!is_valid_user_name (argv[optind])) {
```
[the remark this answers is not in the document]
MarcinDigitic: I used the same convention as in the current files. For changing formatting of the code I would suggest a separate PR and using clang format functionality, instead of manual formatting.

src/chsh.c
MarcinDigitic: To keep some consistency I changed the exit code only in this one place. Other places are consistent with the files, where the conditions were added.

alejandro-colomar (approved):
LGTM.  Thanks!

alejandro-colomar:
> LGTM!

Thanks!

> Alejandro, do you also want hallyn's feedback? If not, you can proceed to merge it.

Nah, I just wanted another pair of eyes to make sure I wasn't forgetting something.  Thank you!

PR #993, opened 2024-05-15
Simplifications using _Generic(3)

alejandro-colomar:
And they also add portability, since it uses ISO C11 instead of GNU extensions.


ikerexxe (approved):
LGTM!

PR #994, opened 2024-05-15
Fix issues detected by static analyzer

ikerexxe:
```
RESOURCE_LEAK (CWE-772)
shadow-4.15.0/src/usermod.c:864:3: alloc_fn: Storage is returned from allocation function "__sgr_dup".
shadow-4.15.0/src/usermod.c:864:3: var_assign: Assigning: "nsgrp" = storage returned from "__sgr_dup(sgrp)".
shadow-4.15.0/src/usermod.c:964:1: leaked_storage: Variable "nsgrp" going out of scope leaks the storage it points to.
962|                   free (nsgrp);
963|           }
964|-> }
965|   #endif                                /* SHADOWGRP */
966|

RESOURCE_LEAK (CWE-772)
shadow-4.15.0/src/usermod.c:734:3: alloc_fn: Storage is returned from allocation function "__gr_dup".
shadow-4.15.0/src/usermod.c:734:3: var_assign: Assigning: "ngrp" = storage returned from "__gr_dup(grp)".
shadow-4.15.0/src/usermod.c:815:1: leaked_storage: Variable "ngrp" going out of scope leaks the storage it points to.
813|                   gr_free(ngrp);
814|           }
815|-> }
816|
817|   #ifdef SHADOWGRP

Error: OVERRUN (CWE-119):
shadow-4.15.0/lib/port.c:154:2: alias: Assigning: "port.pt_names" = "ttys". "port.pt_names" now points to element 0 of "ttys" (which consists of 65 8-byte elements).
shadow-4.15.0/lib/port.c:155:2: cond_const: Checking "j < 64" implies that "j" is 64 on the false branch.
shadow-4.15.0/lib/port.c:175:2: overrun-local: Overrunning array of 65 8-byte elements at element index 65 (byte offset 527) by dereferencing pointer "port.pt_names + (j + 1)".
173|           *cp = '\0';
174|           cp++;
175|->         port.pt_names[j + 1] = NULL;
176|
177|           /*

Error: RESOURCE_LEAK (CWE-772):
shadow-4.15.0/src/useradd.c:575:2: alloc_fn: Storage is returned from allocation function "fdopen".
shadow-4.15.0/src/useradd.c:575:2: var_assign: Assigning: "ofp" = storage returned from "fdopen(ofd, "w")".
shadow-4.15.0/src/useradd.c:734:2: leaked_storage: Variable "ofp" going out of scope leaks the storage it points to.
732|           }
733|
734|->         return ret;
735|   }
736|

UNINIT (CWE-457)
shadow-4.15.0/src/chpasswd.c:566:3: var_decl: Declaring variable "newsp" without initializer.
shadow-4.15.0/src/chpasswd.c:617:5: assign: Assigning: "sp" = "&newsp", which points to uninitialized data.
shadow-4.15.0/src/chpasswd.c:629:4: uninit_use: Using uninitialized value "*sp". Field "sp->sp_pwdp" is uninitialized.
627|                    */
628|                   if (NULL != sp) {
629|->                         newsp = *sp;
630|                           newsp.sp_pwdp = cp;
631|                           newsp.sp_lstchg = gettime () / DAY;
```

Resolves: https://issues.redhat.com/browse/RHEL-35383

### DIFF
--- a/lib/port.c
+++ b/lib/port.c
@@ -168,7 +168,7 @@ again:
 	}
 	*cp = '\0';
 	cp++;
-	port.pt_names[j + 1] = NULL;
+	port.pt_names[j] = NULL;
 
 	/*
 	 * Get the list of user names.  It is the second colon


src/usermod.c
```diff
@@ -804,7 +804,7 @@ static void update_gshadow (void)
 	bool was_admin;
 	bool changed;
 	const struct sgrp *sgrp;
-	struct sgrp *nsgrp;
+	struct sgrp *nsgrp = NULL;
 
 	changed = false;
```
alejandro-colomar, line 810:
```

shadow-4.15.0/src/usermod.c:864:3: alloc_fn: Storage is returned from allocation function "__sgr_dup".
shadow-4.15.0/src/usermod.c:864:3: var_assign: Assigning: "nsgrp" = storage returned from "__sgr_dup(sgrp)".
shadow-4.15.0/src/usermod.c:964:1: leaked_storage: Variable "nsgrp" going out of scope leaks the storage it points to.
962|                   free (nsgrp);
963|           }
964|-> }
965|   #endif                                /* SHADOWGRP */
966|
```

Where do those line numbers come from?  Did you apply any patches downstream to this file?

Here's what I see in 4.15.0:

```c
alx@debian:~/src/shadow/stable/shadow/4.15.0$ sed -n '864p' src/usermod.c

alx@debian:~/src/shadow/stable/shadow/4.15.0$ sed -n '962,966p' src/usermod.c
static void grp_update (void)
{
	update_group ();
#ifdef SHADOWGRP
	if (is_shadow_grp) {
```

Please say what line numbers those correspond here, and if relevant, the patches applied downstream.

src/usermod.c
```diff
@@ -791,9 +791,9 @@ static void update_group (void)
 			SYSLOG ((LOG_WARN, "failed to prepare the new %s entry '%s'", gr_dbname (), ngrp->gr_name));
 			fail_exit (E_GRP_UPDATE);
 		}
-
-		gr_free(ngrp);
 	}
+
+	gr_free(ngrp);
 }
```
alejandro-colomar, line 798: This seems to have the same issue as in the other patch for this file.

lib/port.c
```diff
@@ -168,7 +168,7 @@ static struct port *getportent (void)
 	}
 	*cp = '\0';
 	cp++;
-	port.pt_names[j + 1] = NULL;
+	port.pt_names[j] = NULL;
```
alejandro-colomar, line 171:
```
```

BTW, this reminds me of something I had pending.  I'll save that refactor for after the current one is merged.  :-)

src/useradd.c
```diff
@@ -542,7 +542,7 @@ static int set_defaults (void)
 	char  *default_file = USER_DEFAULTS_FILE;
 	char  *cp;
 	FILE  *ifp;
-	FILE  *ofp;
+	FILE  *ofp = NULL;
```
alejandro-colomar, line 545: Why is this necessary?

alejandro-colomar, line 545: `NULL` is not valid input to *fclose*(3).

```c
$ grepc fclose /usr/include/
/usr/include/stdio.h:extern int fclose (FILE *__stream) __nonnull ((1));
```

src/useradd.c
```diff
@@ -749,6 +749,7 @@ static int set_defaults (void)
 	if (prefix[0]) {
 		free(default_file);
 	}
+	fclose (ofp);
 
 	return ret;
```
alejandro-colomar, line 754: This is double-closing in some cases.  The first close is in line 705.

src/chpasswd.c
```diff
@@ -557,7 +557,7 @@ int main (int argc, char **argv)
 		} else
 #endif				/* USE_PAM */
 		{
-		const struct spwd *sp;
+		const struct spwd *sp = NULL;
```
alejandro-colomar, line 560: This is wrong.  I'm checking if the report is valid, which I'm not sure, but this initialization doesn't seem to do anything good, AFAICS.

alejandro-colomar: I confirm: the static analyzer seems to be wrong.  From the information I can see in the report, there's no use of uninitialized data (or I'm missing it).  It looks like a false positive.

src/usermod.c
ikerexxe: Yes, there are downstream only patches that change the line numbering. The patch that changes the line numbering is https://gitlab.com/redhat/centos-stream/rpms/shadow-utils/-/blob/c10s/shadow-4.15.0-audit-update.patch

src/chpasswd.c
ikerexxe: The report is valid, but I misunderstood it. The variable contains some members that are uninitialized, and that's what it's reporting. I must say that the report is not very accurate.

src/usermod.c
```diff
@@ -936,9 +936,9 @@ static void update_gshadow (void)
 			         sgr_dbname (), nsgrp->sg_name));
 			fail_exit (E_GRP_UPDATE);
 		}
-
-		free (nsgrp);
 	}
+
+	free (nsgrp);
 }
 #endif				/* SHADOWGRP */
```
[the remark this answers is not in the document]
ikerexxe: I don't like having this goto to a section that is on a loop. I've provided another solution.

src/usermod.c
ikerexxe: Right, so I offered a similar solution.

src/useradd.c
ikerexxe: I wasn't aware of this. Thanks for pointing me in the right direction.

src/useradd.c
ikerexxe: Surrounded with a condition to make sure that it's NULL.

src/usermod.c
alejandro-colomar: Ahh, ok, so 864 in your report is line 857 in 4.15.0 (`nsgrp = __sgr_dup (sgrp);`).  Thanks!

src/chpasswd.c
alejandro-colomar: Yeah, that's what I suspected, but since the members are not being used, it's not something we should care.  (At least, I didn't see any use of a member being uninitialized.)

src/usermod.c
alejandro-colomar, line 812: I honestly don't like this patch.  Initializing stuff is usually bad, since it disables the ability of static analyzers to diagnose uninitialized uses.  It also makes it more complex to handle it correctly, which in this specific case can be seen by the need to have two free(3) calls, and set to NULL after free(3).

Regarding your feedback about using goto: That use of goto in my patch is just forward jumping to right before the continue point, so it doesn't do any magic.

If you still don't want to use goto in the loop, I can prepare some other refactor, adding a helper function.

I'll reject this patch, and prepare some alternative patch in a new PR.

src/chpasswd.c
```diff
@@ -601,8 +601,8 @@ int main (int argc, char **argv)
 				 * shadow, create one.
 				 */
 				newsp.sp_namp  = name;
-				/* newsp.sp_pwdp  = NULL; will be set later */
-				/* newsp.sp_lstchg= 0;    will be set later */
+				newsp.sp_pwdp  = NULL;
+				newsp.sp_lstchg= 0;
```
alejandro-colomar, line 605: Who uses those values?  I didn't find any code that reads them (although it goes through many layers, and I may have missed something).  I'd like to be sure that this is necessary.  Otherwise, we may be silencing some static analyzer by doing this initialization.

Please show a path that makes this initialization necessary.  Otherwise, I feel safer dropping this patch.

alejandro-colomar: Here's my opinion on initializing stuff to silence analyzers: <https://github.com/neomutt/neomutt/pull/4310#issuecomment-2110072727>.

ikerexxe: I didn't find them either. I'm fine with dropping them, but in general terms I prefer to initialize variables as they make the coder think what value they want in the variable instead of relying on the compiler to set some random value.

src/usermod.c
ikerexxe: Dropped in favour of https://github.com/shadow-maint/shadow/pull/996

src/chpasswd.c
alejandro-colomar: The benefit of leaving the value undefined is that a static analyzer is able to analyze some bugs, but if you initialize the value, it's impossible.

See for example this program:
```c
$ cat bug2.c 
#include <stdlib.h>

void
g(int *x)
{
	if (arc4random() % 2)
		*x = 42;
}

int
main(void)
{
	int x;

	g(&x);
	return x;  // maybe uninitialized use
}
```

It has a bug, because there's one path where I'm reading the value without having assigned to the variable, so the contents are undefined.

A static analyzer is able to detect that there's a path that leads to Undefined Behavior, and warns about it:

```c
$ clang-tidy ./bug2.c 
Error while trying to load a compilation database:
Could not auto-detect compilation database for file "./bug2.c"
No compilation database found in /home/alx/tmp/analyzer or any parent directory
fixed-compilation-database: Error while opening fixed database: No such file or directory
json-compilation-database: Error while opening JSON database: No such file or directory
Running without flags.
1 warning generated.
/home/alx/tmp/analyzer/bug2.c:16:2: warning: Undefined or garbage value returned to caller [clang-analyzer-core.uninitialized.UndefReturn]
        return x;  // maybe uninitialized use
        ^      ~
/home/alx/tmp/analyzer/bug2.c:13:2: note: 'x' declared without an initial value
        int x;
        ^~~~~
/home/alx/tmp/analyzer/bug2.c:15:2: note: Calling 'g'
        g(&x);
        ^~~~~
/home/alx/tmp/analyzer/bug2.c:6:6: note: Assuming the condition is false
        if (arc4random() % 2)
            ^~~~~~~~~~~~~~~~
/home/alx/tmp/analyzer/bug2.c:6:2: note: Taking false branch
        if (arc4random() % 2)
        ^
/home/alx/tmp/analyzer/bug2.c:8:1: note: Returning without writing to '*x'
}
^
/home/alx/tmp/analyzer/bug2.c:15:2: note: Returning from 'g'
        g(&x);
        ^~~~~
/home/alx/tmp/analyzer/bug2.c:16:2: note: Undefined or garbage value returned to caller
        return x;  // maybe uninitialized use
        ^      ~
```

However, if you chose a value, the static analyzer has no way to know that you had a bug.  The bug is still there; just that it's different now.

```c
$ cat bug3.c 
#include <stdlib.h>

void
g(int *x)
{
	if (arc4random() % 2)
		*x = 42;
}

int
main(void)
{
	int x = 0;

	g(&x);
	return x;  // let's say a 0 should never reach here
}
```

```sh
$ clang-tidy bug3.c 
Error while trying to load a compilation database:
Could not auto-detect compilation database for file "bug3.c"
No compilation database found in /home/alx/tmp/gcc or any parent directory
fixed-compilation-database: Error while opening fixed database: No such file or directory
json-compilation-database: Error while opening JSON database: No such file or directory
Running without flags.
```

By initializing the value to 0, we've transformed an obvious bug into a silent bug.  Moreover, a maintainer 10 years from now, who doesn't know what the code is doing, will not know if that 0 is necessary, or if it was just a random value that the programmer put to silence a bad static analyzer.

But did anyone check if having a 0 results in good behavior?  Probably not, because it's a path that in theory should never happen.

src/useradd.c
```diff
@@ -749,6 +750,9 @@ static int set_defaults (void)
 	if (prefix[0]) {
 		free(default_file);
 	}
+	if (ofp != NULL) {
+		fclose (ofp);
+	}
```
alejandro-colomar, line 755: I'm still not convinced by this patch.  I've opened an alternative here: <https://github.com/shadow-maint/shadow/pull/997>

ikerexxe: I prefer your proposal so I'm dropping the patch. Thanks for the help!

alejandro-colomar:
LGTM
Thanks!

alejandro-colomar (approved):

PR #995, opened 2024-05-15
configure: change cmocka library detection

ikerexxe:
For some reason cmocka library isn't detected in fedora, thus the unit
tests aren't run on this platform. Changing the detection process to be
able to run the unit tests.

I'll be honest, I don't know why the previous way of detecting the library
didn't work. I tried several changes and neither of them seemed to detect
the library on my system even though it was installed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -274,6 +274,7 @@ AC_DEFINE_UNQUOTED(GROUP_NAME_MAX_LENGTH, $with_group_name_max_length, [max grou
 AC_SUBST(GROUP_NAME_MAX_LENGTH)
 GROUP_NAME_MAX_LENGTH="$with_group_name_max_length"
 
+
 AM_CONDITIONAL(USE_SHA_CRYPT, test "x$with_sha_crypt" = "xyes")
 if test "$with_sha_crypt" = "yes"; then
 	AC_DEFINE(USE_SHA_CRYPT, 1, [Define to allow the SHA256 and SHA512 password encryption algorithms])
@@ -309,6 +310,10 @@ dnl other libraries.  This should prevent linking libnsl if not really
 dnl needed (Linux glibc, Irix), but still link it if needed (Solaris).
 
 AC_SEARCH_LIBS(gethostbyname, nsl)
+
+PKG_CHECK_MODULES([CMOCKA], [cmocka], [have_cmocka="yes"],
+	[AC_MSG_WARN([libcmocka not found, cmocka tests will not be built])])
+AM_CONDITIONAL([HAVE_CMOCKA], [test x$have_cmocka = xyes])
 
 AC_CHECK_LIB([econf],[econf_readDirs],[LIBECONF="-leconf"],[LIBECONF=""])
 if test -n "$LIBECONF"; then
@@ -700,10 +705,6 @@ if test "$with_skey" = "yes"; then
 		skeychallenge((void*)0, (void*)0, (void*)0, 0);
 	]])],[AC_DEFINE(SKEY_BSD_STYLE, 1, [Define to support newer BSD S/Key API])],[])
 fi
-
-PKG_CHECK_MODULES([CMOCKA], [cmocka], [have_cmocka="yes"],
-	[AC_MSG_WARN([libcmocka not found, cmocka tests will not be built])])
-AM_CONDITIONAL([HAVE_CMOCKA], [test x$have_cmocka = xyes])
 
 AC_CHECK_FUNC(fgetpwent_r, [AC_DEFINE(HAVE_FGETPWENT_R, 1, [Defined to 1 if you have the declaration of 'fgetpwent_r'])])
 


ikerexxe:
As an example of CI check where unit tests aren't run on fedora: https://github.com/shadow-maint/shadow/actions/runs/9096539741/job/25002179766?pr=994. Look for `Testsuite summary for shadow 4.15.1` and you'll see that nothing is run.

configure.ac
```diff
@@ -701,9 +701,28 @@ if test "$with_skey" = "yes"; then
...
+                  ])],
+                  [
+                      [have_cmocka="yes"]
+                      [CMOCKA_LIBS="-lcmocka"]
+                  ],
+                  [
+                      AC_MSG_WARN([libcmocka not found, cmocka tests will not be built])
+                  ])
 AM_CONDITIONAL([HAVE_CMOCKA], [test x$have_cmocka = xyes])
+AC_SUBST(CMOCKA_LIBS)
```
alejandro-colomar, line 725: I can't say I reviewed it, because I don't understand it either.  :-)

But I trust you know what you're doing, so if nobody complains in a reasonable time, feel free to merge.  As long as you test it and it works, it's ok for me.

Maybe @hallyn knows.

configure.ac
```diff
@@ -701,9 +701,28 @@ if test "$with_skey" = "yes"; then
 	]])],[AC_DEFINE(SKEY_BSD_STYLE, 1, [Define to support newer BSD S/Key API])],[])
 fi
 
-PKG_CHECK_MODULES([CMOCKA], [cmocka], [have_cmocka="yes"],
```
lslebodn, line 704: Previous macro is OK and works as expected.
There is not any reason for the change

cmocka was not detected by the configure script on the fedora(koji)
* https://kojipkgs.fedoraproject.org//packages/shadow-utils/4.15.1/5.fc41/data/logs/x86_64/build.log
because it was not installed
* https://kojipkgs.fedoraproject.org//packages/shadow-utils/4.15.1/5.fc41/data/logs/x86_64/root.log

And I cannot see it in the spec among BuildRequires
* https://src.fedoraproject.org/rpms/shadow-utils/blob/rawhide/f/shadow-utils.spec

ikerexxe: That's a problem but it's not what I'm trying to solve right now.

cmocka is installed in the CI runners from the [dockerfile](https://github.com/shadow-maint/shadow/blob/master/share/containers/fedora.dockerfile#L7), but the configure script is unable to detect it:
```
configure:18079: checking for cmocka
configure:18139: result: no
configure:18141: WARNING: libcmocka not found, cmocka tests will not be built
```
See full [logs](https://github.com/shadow-maint/shadow/actions/runs/9095171581/artifacts/1504929473).

I have the same problem in my local system, where I can assure you the library is installed but the configure script is unable to detect it.

lslebodn: OK found the problem.
If you want to use the macro `PKG_CHECK_MODULES` https://autotools.info/pkgconfig/pkg_check_modules.html
the configure script needs to detect the pkg-config binary

One can see that even in macro definition (line 140)
```
sh$ grep -n -C8 AC_DEFUN.*PKG_CHECK_MODULES /usr/share/aclocal/pkg.m4
131-dnl PKG_CHECK_MODULES(VARIABLE-PREFIX, MODULES, [ACTION-IF-FOUND],
132-dnl   [ACTION-IF-NOT-FOUND])
133-dnl --------------------------------------------------------------
134-dnl Since: 0.4.0
135-dnl
136-dnl Note that if there is a possibility the first call to
137-dnl PKG_CHECK_MODULES might not happen, you should be sure to include an
138-dnl explicit call to PKG_PROG_PKG_CONFIG in your configure.ac
139:AC_DEFUN([PKG_CHECK_MODULES],
140-[AC_REQUIRE([PKG_PROG_PKG_CONFIG])dnl
141-AC_ARG_VAR([$1][_CFLAGS], [C compiler flags for $1, overriding pkg-config])dnl
142-AC_ARG_VAR([$1][_LIBS], [linker flags for $1, overriding pkg-config])dnl
143-
144-pkg_failed=no
145-AC_MSG_CHECKING([for $2])
146-
147-_PKG_CONFIG([$1][_CFLAGS], [cflags], [$2])
```

However shell variables are initialized just once for the `PKG_PROG_PKG_CONFIG`
and not for each "call" of `PKG_CHECK_MODULES`.

The problem is the macro is used conditionally for the 1st time in configure.ac
https://github.com/shadow-maint/shadow/blob/9dddcd29f131ac4eeeb47e8c4c7a7ae45d4e6b72/configure.ac#L412-L432
and it was not set/executed on fedora because of  `--without-libbsd`

Therefore 2nd PKG_CHECK_MODULES automatically failed due to empty PKG_CONFIG variable.

Simple workaround without patch for fedora is to explicitly set env PKG_CONFIG before executing configure
`export PKG_CONFIG=/usr/bin/pkg-config; ./configure ...`

Much better solution is to use the macro `PKG_CHECK_MODULES` unconditionally for the first time.
e.g. move check for CMOCKA before the check for libbsd-overlay
https://github.com/shadow-maint/shadow/blob/9dddcd29f131ac4eeeb47e8c4c7a7ae45d4e6b72/configure.ac#L704

Or a bit hackish. Run internal macro for detecting the patch for pkg-config `PKG_PROG_PKG_CONFIG`
before 1st conditional usage `if test "$with_libbsd" != "no"; then`
* https://github.com/shadow-maint/shadow/blob/9dddcd29f131ac4eeeb47e8c4c7a7ae45d4e6b72/configure.ac#L413C1-L413C37

ikerexxe: IMPRESSIVE! A splendid analysis.

I've decided to move the `PKG_CHECK_MODULES` for cmocka to the top to avoid this problem. Thanks!

alejandro-colomar (approved):
LGTM.  Thank you both!
```
```

configure.ac
```diff
@@ -274,6 +274,7 @@ AC_DEFINE_UNQUOTED(GROUP_NAME_MAX_LENGTH, $with_group_name_max_length, [max grou
 AC_SUBST(GROUP_NAME_MAX_LENGTH)
 GROUP_NAME_MAX_LENGTH="$with_group_name_max_length"
 
+
```
lslebodn, line 277: Looks to me like unnecessary change otherwise LGTM

PR #996, opened 2024-05-16
Fix 2 memory leaks

alejandro-colomar:
Hi @ikerexxe ,

This is my implementation with a goto, plus additional refactors to address your concerns about goto, and also simplify some weirdness.

### DIFF
--- a/src/usermod.c
+++ b/src/usermod.c
@@ -183,6 +183,7 @@ static void update_group(const struct group *grp);
 
 #ifdef SHADOWGRP
 static void update_gshadow_file(void);
+static void update_gshadow(const struct sgrp *sgrp);
 #endif
 static void grp_update (void);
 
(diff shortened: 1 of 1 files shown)


src/usermod.c
```diff
@@ -685,263 +687,277 @@ fail_exit (int code)
...
+	struct group  *ngrp;
 
-		ngrp = __gr_dup (grp);
-		if (NULL == ngrp) {
-			fprintf (stderr,
-			         _("%s: Out of memory. Cannot update %s.\n"),
-			         Prog, gr_dbname ());
-			fail_exit (E_GRP_UPDATE);
-		}
+	changed = false;
```
ikerexxe, line 712: You can set the initial value  when defining the variable

src/usermod.c
```diff
@@ -685,263 +687,277 @@ fail_exit (int code)
...
+	bool         was_admin;
+	struct sgrp  *nsgrp;
 
-		/*
-		 * See if the user specified this group as one of their
-		 * concurrent groups.
-		 */
-		is_member = Gflg && (   (was_member && aflg)
-		                     || is_on_list (user_groups, sgrp->sg_name));
+	changed = false;
```
ikerexxe, line 834: You can set the initial value  when defining the variable

ikerexxe (commented):
Very good improvement. Thank you man!

I added two inline comments, and I'd also like you to change two commits messages: 14aa5a8ab70326702f22c99c5292f9d40b19c272 and 1ea4f91b6cc444614ef77549ce5bf5d1c5246902. The code explanation is too much, I think it's enough with the first two lines.

src/usermod.c
alejandro-colomar: I consciously avoid that.  I find it more readable if the declarations don't initialize, and in general, a safer practice (not in this case, though).

I prefer to have a strong separation between declarations and statements.

src/usermod.c
alejandro-colomar: Same as the other case.

ikerexxe (approved):
I don't see anything blocking from accepting these changes, so I proceed to merge them.

Issue #956, opened 2024-02-21
Heisenbug: `tests/su/10_su_sulog_success/su.test  tests/su/10_su_sulog_success/su.test`


alejandro-colomar:
I've seen that test fail in a PR, in which I wouldn't expect it to fail.  When trying to reproduce it, I wasn't able, not even in the same PR.  Let's just document that it has happened, in case we see it again.

The only thing we know until now, is that the following failed once:

<tests/su/10_su_sulog_success/su.test 
tests/su/10_su_sulog_success/su.test>

And it was in

-  <https://github.com/shadow-maint/shadow/pull/955>

alejandro-colomar:
Here's the action that failed:

-  <https://github.com/shadow-maint/shadow/actions/runs/7988856130>

alejandro-colomar:
I've experienced it again, spuriously in a branch that has been working for a long time, and without any changes (just reorganizing the commits internally), it failed in one rebase, and worked again in the next.  Here are the logs:

```
###############################################################################
#
# Test: ./su/10_su_sulog_success/su.test
#
###############################################################################
#
# Rationale: su can be used to switch to a non-root user
#
###############################################################################
# switch to the passwordless 'testsuite' user
# and expect a '$ ' prompt
spawn /bin/su testsuite

<r/work/shadow/shadow/tests/su/10_su_sulog_success$ 
# make sure we are now 'testsuite'
# id should return 'uid=424243(testsuite) gid=424243 groups=424243'

<r/work/shadow/shadow/tests/su/10_su_sulog_success$ id

uid=424243(testsuite) gid=424243 groups=424243

<r/work/shadow/shadow/tests/su/10_su_sulog_success$ 

# now switch to user 'myuser'
# and expect a password prompt

<r/work/shadow/shadow/tests/su/10_su_sulog_success$ su myuser

Password: 
# password 'myuserF00barbaz' sent

# expect prompt '$ '

myuser@fv-az665-93:~/runner/work/shadow/shadow/tests/su/10_su_sulog_success$ 

# make sure we are 'myuser'
# id should return '(myuser).*(myuser).*(myuser)

myuser@fv-az665-93:~/runner/work/shadow/shadow/tests/su/10_su_sulog_success$ 
<work/shadow/shadow/tests/su/10_su_sulog_success$ id                         

uid=424242(myuser) gid=424242(myuser) groups=424242(myuser)

[...]

Issue #835, opened 2023-11-10
New system test framework proposal


ikerexxe:
# Introduction

Currently, shadow has a very limited system test suite written in bash that helps us sanitize the new code. The framework is difficult to use, and obliges the person writing the test to ensure that the system under test is left in the same state at the end of the test. That’s dangerous and it involves having a knowledge that many people don’t have. In addition, the test suite only runs in Ubuntu. All of this makes test development hard work, when it should be an easy task.

Thus, I’d like to propose the usage of a new testing framework. One that keeps us up-to-date with the latest developments in this area.

# Requirements

I have identified a series of requirements that the new framework must fulfill:

- It must work on all the distributions that are part of our CI (i.e. Alpine, Debian and Fedora).
- The system under test can’t be the one where the tests are run. We can user VMs or containers to deploy the system under test.
- We must be able to run it locally.
- The tests must have a proper setup and teardown, where the environment is cleaned and left as before it was run. This includes the test cases that have failed, as we don’t want to pollute the environment.
- Automatically collect logs and artifacts when a test is finished.
- Build rich, high level test API to access common functionalities (i.e. adding a user, modifying it, checking the logs, etc.)
- The APIs must be documented, so that test developers can check and understand their behaviour.

# Proposal

The proposal would be to use python, [pytest](https://docs.pytest.org/en/7.4.x/ "‌") and at least two plugins. The first plugin is [pytest-mh](https://github.com/next-actions/pytest-mh/ "‌"), a pytest plugin to run shell commands and scripts over SSH on remote hosts. The second plugin will be new and it will be implemented to meet our needs. It will be in change of handling the common functionalities (i.e. useradd, chage…) that are required for this project, similar to [sssd-test-framework](https://github.com/SSSD/sssd-test-framework/ "‌") but for shadow. These plugins are written in python, so the tests will be written in this language and we will depend on it. I plan to use the python virtual environment to avoid having to install all these dependencies directly in the test runner (a.k.a. our systems).

In addition, I’d like to add other python plugins to make test engineers life easier. An example of those would be [pytest-ticket](https://github.com/next-actions/pytest-ticket "‌"), which adds the ability to filter test cases by an associated ticket of a tracker of your choice. Another example, [pytest-tier](https://github.com/next-actions/pytest-tier "‌"), adds the ability to filter test cases by associated tier.

## Dependencies

The following dependencies will be incorporated into this project:

- Python and its virtual environment
- pytest
- pytest-mh
[...]

PR #1009, opened 2024-05-29
Move container build infrastructure to Ansible

ikerexxe:
Using dockerfiles to build the containers in the CI prevented us from getting the logs when any of the build steps failed. Using Ansible will help us tackle this problem while still maintaining a certain independence from Github Actions.


alejandro-colomar:
These ansible playbooks still build Docker images, so I guess under the hood they still do the same thing that a Dockerfile and docker(1) could do.  Can't we achieve the same with just docker(1) and a Dockerfile?

That is, how is ansible necessary (or better than docker(1)) to achieve this?

ikerexxe:
Yes, they do the same, and if I could I would have stick with the docker and dockerfile architecture, but it isn't possible. At least if we want to obtain the logs without having to do some nasty things, like [1](https://github.com/shadow-maint/shadow/blob/master/share/containers/alpine.dockerfile#L30) and [2](https://github.com/shadow-maint/shadow/blob/master/share/containers/alpine.dockerfile#L34) (which in reality doesn't work because when something fails the container build never arrives to this point).

In addition, when dockerfile build fails the container is destroyed, which prevents us from accessing it to see what has happened. Which brings us to the previous point, getting the logs.

Using Ansible allows us to maintain a similar approach to what we had until now: build this project for different distributions, run the workflow locally, maintain some independence from Github actions, etc. On top of that, we'll be able to obtain all the logs, and if we are running the workflow locally and anything fails we'll be able to access the containers to debug what's happening.

alejandro-colomar:
> Yes, they do the same, and if I could I would have stick with the docker and dockerfile architecture, but it isn't possible. At least if we want to obtain the logs without having to do some nasty things, like [1](https://github.com/shadow-maint/shadow/blob/master/share/containers/alpine.dockerfile#L30) and [2](https://github.com/shadow-maint/shadow/blob/master/share/containers/alpine.dockerfile#L34) (which in reality doesn't work because when something fails the container build never arrives to this point).

This does actually work (I could see the logs).  I've seen it fail once (not exactly this; it was the same trick, but in a github actions), and documented it here:
<https://github.com/shadow-maint/shadow/issues/956#issuecomment-2056464361>

And I wrote some small Dockerfile to test this with a docker build:

```docker
$ cat Dockerfile 
FROM debian

RUN bash -c "trap 'cat </etc/os-release >&2' ERR; false;"
```
```sh
$ sudo docker build .
Sending build context to Docker daemon  23.55kB
Step 1/2 : FROM debian
 ---> 2a033a8c6371
Step 2/2 : RUN bash -c "trap 'cat </etc/os-release >&2' ERR; false;"
 ---> Running in 3c6f20ae7c6b
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
The command '/bin/sh -c bash -c "trap 'cat </etc/os-release >&2' ERR; false;"' returned a non-zero code: 1
```

Or do you have any case where it failed and it didn't work?

alejandro-colomar:
As an alternative, we could consider that `make check` could directly print to stderr, instead of a log.  Let the user redirect the logs if it wants to.  This is very much preferred behavior, IMO.  Then we wouldn't need to do trap(1) magic to print the logs on failure.

ikerexxe:
The `trap` works, but it feels nasty and it forces users to use the Github scroll, which doesn't work very well.

Regarding `COPY`, that doesn't work and we do not only want to see the build logs but also the configuration files. This would feel like an improvement to me considering our current situation.

Finally, this would also give us the opportunity of capturing all the test logs when https://github.com/shadow-maint/shadow/issues/835 is implemented and running.

share/container-build.sh
```diff
@@ -2,17 +2,13 @@
...
-	IFS=''
-	if ! docker build -f $FILE . --output build-out/${ADDR[0]}; then
-		exit
-	fi
-done
+pushd share/ansible/ || exit
+ansible-playbook playbook.yml -i inventory.ini -e 'distribution=alpine'
+ansible-playbook playbook.yml -i inventory.ini -e 'distribution=debian'
+ansible-playbook playbook.yml -i inventory.ini -e 'distribution=fedora'
+popd || exit
```
alejandro-colomar, line 14: Nice!  I didn't know pushd(1) and popd(1) existed.  :)

alejandro-colomar, line 14: Do we want to `set -eE` instead of `|| exit`?  Also, what if ansible fails?

share/ansible/roles/ci_run/tasks/fedora.yml
```diff
@@ -0,0 +1,69 @@
...
+  args:
+    chdir: /usr/local/src/shadow/
+  ignore_errors: true
+
+- name: Run unit-tests
+  ansible.builtin.command:
+    make check
+  args:
+    chdir: /usr/local/src/shadow/
+  ignore_errors: true
```
alejandro-colomar, line 52: What does this mean?  Why do we want to ignore errors?

share/ansible/roles/ci_run/README.md
```diff
@@ -0,0 +1,25 @@
...
+Example Playbook
+----------------
+
+Usage example:
+
+    - hosts: builder
+      connection: podman
+      become: true
+      roles:
+        - role: ci_run
```
alejandro-colomar, line 15: I don't understand this usage.  Where is this YAML code supposed to be used?

share/container-build.sh
alejandro-colomar, line 14: Actually, being the last line of the script, we don't need to popd(1), since it doesn't affect the caller.

```sh
alx@debian:~$ sh -c 'cd /tmp'
alx@debian:~$ 
```

share/ansible/roles/ci_run/tasks/fedora.yml
ikerexxe: By default if an ansible action fails, then the ansible execution is stopped. I want to ignore the errors and continue the execution to run the last action where the logs are copied from the container to the host system. This way we can gather them for inspection.

share/container-build.sh
ikerexxe: I discovered them two months ago or so. I don't think they are very popular

ikerexxe: Yes, that would make more sense

ikerexxe: Right. I've removed it.

alejandro-colomar: I was wondering the other day if there would be a way to keep more than one level of directory stack.  `cd -` is useful, but when you need to jump back and forth with more than two dirs, it doesn't help.  pushd(1) is what I was looking for.  :)

.github/workflows/runner.yml
```diff
@@ -96,15 +96,24 @@ jobs:
...
+      run: |
+        sudo apt-get update
+        sudo apt-get -y install ansible
+
     - name: Build container
       run: |
-        docker buildx build -f ./share/containers/${{ matrix.os }}.dockerfile . --output build-out
+        pushd share/ansible/
+        ansible-playbook playbook.yml -i inventory.ini -e 'distribution=${{ matrix.os }}'
+        popd
```
alejandro-colomar, line 108: If this runs as a child script (I don't know how Ansible works), popd(1) is again unnecessary.

share/container-build.sh
```diff
@@ -2,17 +2,13 @@
...
-	IFS='/'
-	read -ra ADDR <<< "$FILE"
-	IFS='.'
-	read -ra ADDR <<< "${ADDR[2]}"
-	IFS=''
-	if ! docker build -f $FILE . --output build-out/${ADDR[0]}; then
-		exit
-	fi
-done
+pushd ansible/
```
alejandro-colomar, line 10: pushd(1) without popd(1) is just cd(1)

.github/workflows/runner.yml
ikerexxe: This is a Github action and popd is necessary in this context

share/container-build.sh
ikerexxe: Not really. This is a script and everything that runs inside it it's affected by `pushd`, but as soon as the script finishes the state is recovered. If I want to do the same with `cd` I have to run it again at the end to recover the state.

share/ansible/roles/ci_run/tasks/fedora.yml
alejandro-colomar: Thanks!

share/container-build.sh
```diff
@@ -2,17 +2,13 @@
...
+set -eE
```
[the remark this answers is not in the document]
ikerexxe: You are right, I've changed the order for pushd and set

share/ansible/roles/ci_run/tasks/fedora.yml
alejandro-colomar: BTW, I see in that run that the store artifacts step isn't being run?

share/ansible/roles/ci_run/README.md
ikerexxe: In the ansible playbook: https://github.com/shadow-maint/shadow/pull/1009/files#diff-1968d6dc5c6937169e73de5fa2e83bf0d9810afcec288aa45b6cfabb40a5bf15R12-R17

Roles are self-contained units in Ansible. They are used for grouping tasks and other resources in a known file structure.

share/container-build.sh
alejandro-colomar: Please show the difference with an example; I'm not seeing it.

```sh
$ pwd
/home/alx/tmp/shadow
```
```sh
$ ls
cd.bash  pd.bash
```
```sh
$ cat cd.bash 
	cd /tmp
	pwd
	ls
```
```sh
$ cat pd.bash 
	pushd /tmp
	pwd
	ls
```
```sh
$ bash cd.bash 
/tmp
Temp-988ac446-4a36-4ca3-a295-a819768048fd
grepc.patterns.Di6KCW
grepc.patterns.nvO5bb
ssh-mEQv6ojqkOKI
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-ModemManager.service-NfEV9r
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-colord.service-hnzWJu
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-exim4.service-vGmQKr
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-polkit.service-TdhxGF
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-switcheroo-control.service-N089gp
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-systemd-logind.service-ILr6Ix
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-systemd-timesyncd.service-GvIeWo
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-upower.service-wSe2ke
```
```sh
$ pwd
/home/alx/tmp/shadow
```
```sh
$ bash pd.bash 
/tmp ~/tmp/shadow
/tmp
Temp-988ac446-4a36-4ca3-a295-a819768048fd
grepc.patterns.Di6KCW
grepc.patterns.nvO5bb
ssh-mEQv6ojqkOKI
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-ModemManager.service-NfEV9r
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-colord.service-hnzWJu
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-exim4.service-vGmQKr
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-polkit.service-TdhxGF
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-switcheroo-control.service-N089gp
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-systemd-logind.service-ILr6Ix
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-systemd-timesyncd.service-GvIeWo
systemd-private-2c58461cd7dd4e95939ee9a5b8df6f39-upower.service-wSe2ke
```
```sh
$ pwd
/home/alx/tmp/shadow
```

share/ansible/roles/ci_run/tasks/fedora.yml
ikerexxe: My bad, I forgot to set [if: always()](https://github.com/shadow-maint/shadow/pull/1009/files#diff-5c8806e0119aa46657122645a786c2d9caaf3d30652775f2ae4dc45ffce209cdR111) in the Github Action. It's fine now.

share/container-build.sh
ikerexxe: You are right, I changed it to `cd`

share/ansible/roles/ci_run/tasks/fedora.yml
alejandro-colomar: Thanks!  It seems to work now.  BTW, dumb question: how do I find and read the logs (artifacts)?

.github/workflows/runner.yml
```diff
@@ -96,15 +96,25 @@ jobs:
...
       uses: actions/checkout@v3
 
+    - name: Install Ansible
+      run: |
+        sudo apt-get update
+        sudo apt-get -y install ansible
+
     - name: Build container
       run: |
-        docker buildx build -f ./share/containers/${{ matrix.os }}.dockerfile . --output build-out
```
alejandro-colomar, line 101: BTW, what if we would map some volume to write the log files to it with docker(1)?

```docker
RUN bash -c "trap 'cat <tests/unit/test-suite.log | tee /some/mapped/volume/test-suite.log >&2' ERR; make check;"
```

alejandro-colomar: That would allow storing the artifacts without needing ansible.

share/ansible/roles/ci_run/tasks/fedora.yml
ikerexxe: You need to open the [action that failed](https://github.com/shadow-maint/shadow/actions/runs/9385996635/job/25845492826?pr=1009), then click on [summary](https://github.com/shadow-maint/shadow/actions/runs/9385996635?pr=1009), and finally scroll down to find all the artifacts

.github/workflows/runner.yml
ikerexxe: You can't mount volumes while building a container image, and that's exactly what we are doing with dockerfile. I have tried to solve this problem in various ways and with the technology we use it is not possible :sweat:

ikerexxe:
I think this is ready for review, so I'm moving it out of draft state.

share/ansible/roles/ci_run/tasks/fedora.yml
alejandro-colomar: Ughhh, and then download a .zip and extract it to find the logs.  Can we have a copy on stderr?  I very much prefer scrolling.  :)

.github/workflows/runner.yml
alejandro-colomar: How about something like this?

```sh
docker build ... | tee >(sed -n '/BEGIN MARKER/,/END MARKER/p' >/host/log/file)
```

We would only need to find some consistent markers in the log.

share/ansible/roles/ci_run/tasks/fedora.yml
ikerexxe: That will work until we start running the new tests and have several files to review

.github/workflows/runner.yml
ikerexxe: Same answer as in https://github.com/shadow-maint/shadow/pull/1009#discussion_r1638276746

alejandro-colomar: If each log file has a different marker, you can parse each one separately:

```sh
docker build ... | tee >(sed -n '/BEGIN A/,/END A/p' >/host/A.log) >(sed -n '/BEGIN B/,/END B/p' >/host/B.log)
```

PR #1016, opened 2024-06-05
man/lastlog: remove wrong use of keyword term

hallyn:
Per https://tdg.docbook.org/tdg/4.5/term, term is a word being defined in a varlistentry.  The 'high uid' description is not a varlistentry, so <term> and </term> show up in the processed manpage.  See debian Bug#1072297.

### DIFF
--- a/man/lastlog.8.xml
+++ b/man/lastlog.8.xml
@@ -211,8 +211,8 @@
       to hang as it processes entries with UIDs 171-799).
     </para>
     <para>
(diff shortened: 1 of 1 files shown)


ikerexxe (approved):
LGTM! I think it was me who made that mistake :facepalm:

alejandro-colomar (approved):
LGTM.  I guess it was pasted from line 197 (`<term><filename>/var/log/lastlog</filename></term>`).

PR #1043, opened 2024-07-07
tests/libsubid/04_nss: fix setting basedir

zeha:
Align on variable name BUILD_BASE_DIR for overriding the toplevel directory. This is the same name as accepted by tests/common/config.sh.

Without this, the test libsubid/04_nss fails in Debian's autopkgtests.


hallyn (approved):
Thank you.

Issue #1055, opened 2024-07-20
fputsx is not round-trip safe with backslash-ending usernames


zeha:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1076619 reports:

The following sequence breaks /etc/group:

```
$ useradd 'vijai\'
$ usermod -aG sudo 'vijai\'
$ useradd blafasel
$ usermod -aG sudo blafasel
```

Results in:
```
sudo:x:27:install,vijaiaudio:x:29:install
```

Should have been:
```
sudo:x:27:install,vijai\,blafasel
audio:x:29:install
```

I believe this is caused by `fgetsx` "supporting" backslash as line-continuation indicator, but
* fputsx does not escape backslashes before line end
* there's a comment in fputsx indicating this is unsupported by everything else anyway.

alejandro-colomar:
I had been looking at that code, and it seemed too brittle, and inconsistent.

I preferred to not change it at all while it worked, but your comment seems to be what I was waiting for: a report that it actually doesn't work.

Do you have any specific fix in mind?

Should we remove support for line continuations completely?  That would keep it simple, by calling standard fgets(3).  (But I was worried that it could break existing config files.)

zeha:
> But I was worried that it could break existing config files.

From what I can tell, line continuation doesn't work on glibc systems for lookup:

```
root@tiksta:~# tail -n2 /etc/group
tftp:x:112:\
ch
root@tiksta:~# getent group tftp
tftp:x:112:\

root@tiksta:~# tail -n1 /etc/group
tftp:x:112:ch
root@tiksta:~# getent group tftp
tftp:x:112:ch
```

Thus I cannot imagine it being used a lot, and wouldn't worry about breaking config files (which were already broken if used).

> Do you have any specific fix in mind?

I was thinking replace fgetsx/fputsx by fgets/fputs.

alejandro-colomar:
> > But I was worried that it could break existing config files.
> 
> From what I can tell, line continuation doesn't work on glibc systems for lookup:
> 
> ```
> root@tiksta:~# tail -n2 /etc/group
> tftp:x:112:\
> ch
> root@tiksta:~# getent group tftp
> tftp:x:112:\
> 
> root@tiksta:~# tail -n1 /etc/group
> tftp:x:112:ch
> root@tiksta:~# getent group tftp
> tftp:x:112:ch
> ```
> 
[...]

PR #1048, opened 2024-07-11
lib/, src/: Use consistent style using strchr(3) in conditionals

alejandro-colomar:
Among other things, remove most uses of strtok(3).  Replace its calls by strsep(3), which is stateless.

This probably silently fixes some bugs, where two separators were being merged as the same one.

Also, that's the reason why I didn't replace all strtok(3) calls.  In some cases (paths), it makes sense to merge adjacent `//`.

----

Revisions:

<details>
<summary>v1b</summary>

-  Rebase

```
$ git range-diff master..gh/string shadow/master..string 
 1:  5ba4cfca !  1:  e825c03e lib/, src/: Use stpspn() instead of its pattern
    @@ lib/getdate.y: yylex (void)
     
      ## lib/limits.c ##
     @@
    - #include "atoi/a2i.h"
    - #include "atoi/str2i.h"
    + #include "atoi/str2i/str2s.h"
    + #include "atoi/str2i/str2u.h"
      #include "memzero.h"
     +#include "string/strchr/stpspn.h"
      #include "typetraits.h"
    @@ lib/strtoday.c
     @@
      #ident "$Id$"
      
    - #include "atoi/str2i.h"
    + #include "atoi/str2i/str2s.h"
     -#include "prototypes.h"
      #include "getdate.h"
     +#include "prototypes.h"
 2:  f5cde8a7 =  2:  e7710fda lib/port.c: portcmp(): Use strcmp(3) instead of its pattern
 3:  009a6aaa =  3:  13ff7c5a lib/gshadow.c: endsgent(): Remove dead assignment
 4:  49317790 =  4:  f090d6c3 src/suauth.c: check_su_auth(): Use strspn(3) instead of its pattern
 5:  1e064044 =  5:  526c8f72 src/suauth.c: check_su_auth(): Use pointers to simplify
 6:  faae7f68 =  6:  30835d33 lib/, src/: Use strsep(3) instead of strtok(3)
 7:  edd0e442 =  7:  e4213fdf lib/loginprompt.c: login_prompt(): Use strtcpy() instead of its pattern
 8:  0cd40e34 =  8:  446541e1 contrib/, lib/, src/: Use strchr(3) to compact comparisons
 9:  c960c5f9 !  9:  946d07ca lib/, src/: Use str[n]cmp(3) instead of explicit byte comparisons
    @@ lib/getrange.c
      #include <stdlib.h>
     +#include <string.h>
      
    - #include "atoi/a2i.h"
[...]

PR #1056, opened 2024-07-21
Replace fgetsx() by fgets(3) and fputsx() by fputs(3)

alejandro-colomar:
It seems they never worked correctly.  Let's keep it simple and remove
support for escaped newlines.
    
Closes: <https://github.com/shadow-maint/shadow/issues/1055>
Reported-by: @zeha


lib/gshadow.c
```diff
@@ -155,29 +155,26 @@ void endsgent (void)
...
+		buf = cp;
+		buflen *= 2;
+
+		len = strlen (buf);
+		if (fgets(&buf[len], buflen - len, fp) == NULL)
+			return NULL;
 	}
-	return NULL;
+	stpsep(buf, "\n");
+	return (sgetsgent (buf));
```
alejandro-colomar, line 177: @thesamesam 

Hi Sam,

It might be worth commenting this here.

Here's an overview of the branch:
```
$ git log --oneline master..fgets 
cd64d530 (HEAD -> fgets, gh/fgets) lib/, po/: Remove fgetsx() and fputsx()
941c951c lib/, src/: Remove useless casts in fgets(3)
4d66df60 lib/, src/: Consistently use NULL with fgets(3)
d0cdc9c8 lib/gshadow.c: endsgent(): Invert logic to reduce indentation
a719bc39 (gh/string, string) lib/fields.c: Remove dead code
b71d2236 contrib/, lib/, src/: Use consistent style using strchr(3) in conditionals
44df1083 lib/, src/: Use str[n]cmp(3) instead of explicit byte comparisons
d3350b67 contrib/, lib/, src/: Use strchr(3) to compact comparisons
9a6e3284 lib/loginprompt.c: login_prompt(): Use strtcpy() instead of its pattern
db93a222 lib/, src/: Use strsep(3) instead of strtok(3)
376e4166 src/suauth.c: check_su_auth(): Use pointers to simplify
ae1c284e src/suauth.c: check_su_auth(): Use strspn(3) instead of its pattern
93362cc3 lib/gshadow.c: endsgent(): Remove dead assignment
5a46723a lib/port.c: portcmp(): Use strcmp(3) instead of its pattern
7442a127 lib/, src/: Use stpspn() instead of its pattern
```

This `fgets` branch is really only
```
$ git log --oneline string..fgets 
cd64d530 (HEAD -> fgets, gh/fgets) lib/, po/: Remove fgetsx() and fputsx()
941c951c lib/, src/: Remove useless casts in fgets(3)
4d66df60 lib/, src/: Consistently use NULL with fgets(3)
d0cdc9c8 lib/gshadow.c: endsgent(): Invert logic to reduce indentation
```
but I based it on `string`, which corresponds to PR <https://github.com/shadow-maint/shadow/pull/1048>.

Here's why.  Basically, according to the bug report that it closes, the PR is about replacing fgetsx => fgets.  Here are two of the calls that we replace:

```c
$ grep -rnC4 fgetsx lib/gshadow.c
154-	if (NULL == fp) {
155-		return NULL;
156-	}
157-
158:	if (fgetsx(buf, buflen, fp) == buf) {
159-		while (   ((cp = strrchr (buf, '\n')) == NULL)
160-		       && (feof (fp) == 0)) {
161-			size_t len;
162-
--
167-			buf = cp;
168-			buflen *= 2;
169-
170-			len = strlen (buf);
171:			if (fgetsx (&buf[len],
172-			            (int) (buflen - len),
173-			            fp) != &buf[len]) {
174-				return NULL;
175-			}
```

I want to fully understand what the code is doing before doing the change, which makes me more confident that the change is correct.  To that end, I need to reduce superfluous complexity.

In the code above, you may notice a dead assignment:
```c
159-		while (   ((cp = strrchr (buf, '\n')) == NULL)
```
I already had removed that dead assignment in <https://github.com/shadow-maint/shadow/pull/1048> in commit <https://github.com/shadow-maint/shadow/pull/1048/commits/93362cc39a78a12aa5732a83e28fddd5891bc63d>. That's the only reason why I decided to base it on that PR.

Now let's see why this PR has 4 new commits, instead of 1.

Line
```c
158:	if (fgetsx(buf, buflen, fp) == buf) {
```
Has a conditional that spans the entire rest of the function.  Commit 1 turns this into the usual thing, which is an early return on error.

Then we get to
```c
171:			if (fgetsx (&buf[len],
172-			            (int) (buflen - len),
173-			            fp) != &buf[len]) {
```

Why do we have that cast?  It's nonsense.  I'll remove it.
And also `&buf[len]` is very brittle, and a small typo could add a bug here.  Let's use NULL there.

While doing those two changes, I'll check any other places that do similarly bad stuff, and replace it in the entire code base.  Then commit 4 does what we originally wanted.

I don't reason too much the first 3 commits, which you might find frustrating while reviewing the changes a posteriori.  I didn't because they're not the main purpose of the PR; they're just cosmetic changes in preparation for the PR.

Do you have any suggestions for making it easier for you to review the changes later?

Maybe I could add an explicit `[cosmetic]` in the second line of the commit message, to mark those as unimportant commits.

Still, I need to apply those cosmetic changes, or it would be much more difficult to me to reason about the important changes of the PR.

alejandro-colomar: And these cosmetic patches combined with the change fgetsx=>fgets have allowed me to see a transformation fgets=>getline too.  I probably wouldn't have been able to see it without the cosmetic changes.

lib/gshadow.c
```diff
@@ -141,39 +141,15 @@ void endsgent (void)
...
-		}
-		buflen = BUFSIZ;
-	}
-
 	if (NULL == fp) {
 		return NULL;
 	}
 
-	if (fgets(buf, buflen, fp) == NULL)
+	if (getline(&buf, &buflen, fp) == -1)
```
zeha, line 148: I know the old code didn't, but should buf be free()ed?

alejandro-colomar: It uses a static pointer, so that it gets reused in the next call.  So, technically it's not a memory leak, since the pointer is not lost.

Still, I'm not a fan of that, which is a common thing in this project.  I agree that it's better to free it and forget the pointer.  It might be less efficient if you're reading millions of lines, but I don't think that's our case anyway.

Yeah, we could free this thing

zeha: Oh wow. Didn't see the `static`. Nevermind then.

alejandro-colomar: I'll free it in a separate PR.

alejandro-colomar: Meh, the change was short enough that I added it to this PR.  Thanks!

lib/gshadow.c
thesamesam: Did you consider my coccinelle suggestion?

lib/loginprompt.c
```diff
@@ -81,9 +84,8 @@ void login_prompt (char *name, int namesize)
 	 */
 
 	MEMZERO(buf);
-	if (fgets (buf, sizeof buf, stdin) != buf) {
+	if (fgets(buf, sizeof(buf), stdin) == NULL)
```
thesamesam, line 87: It would be helpful if formatting changes weren't mixed in.

lib/gshadow.c
```diff
@@ -129,57 +129,37 @@ void endsgent (void)
...
 /*
  * fgetsgent - convert next line in stream to (struct sgrp)
  *
  * fgetsgent() reads the next line from the provided stream and
  * converts it to a (struct sgrp).  NULL is returned on EOF.
  */
 
-/*@observer@*//*@null@*/struct sgrp *fgetsgent (/*@null@*/FILE * fp)
+/*@observer@*//*@null@*/struct sgrp *
+fgetsgent(/*@null@*/FILE *fp)
```
thesamesam, line 141: Is this null stuff really needed? Can't we use the attribute for it?

lib/fields.c
```diff
@@ -66,7 +68,8 @@ int valid_field (const char *field, const char *illegal)
  * prompt the user with the name of the field being changed and the
  * current value.
  */
-void change_field (char *buf, size_t maxsize, const char *prompt)
+void
```
thesamesam, line 71: Really, if you must do this, can't you just do a huge clang-format pass?

thesamesam:
I really feel this PR is still pretty noisy.

lib/gshadow.c
alejandro-colomar: I don't know how to use it.  If you point me to some docs and/or show me a small shell example session showing how it's useful, I'll be happy to try it.

lib/loginprompt.c
alejandro-colomar: I try to keep those to a minimum: only whitespace and (sometimes) parentheses.  And only on hunks where it (subjectively, admittedly) doesn't hurt too much readability (i.e., if I'm doing some huge transformations where I need your attention somewhere else, I'll try to not do it there.

I'll keep in mind your concern.

lib/gshadow.c
alejandro-colomar: I added some patches for reducing those comments some time ago, and replaced some of them by attributes and qualifiers

I should have another look at those at some point, and send another round of patches.

lib/fields.c
alejandro-colomar: Hmm, I'll have a look at using it.  Thanks for the suggestion.

lib/gshadow.c
alejandro-colomar: Hmmm.  <https://coccinelle.gitlabpages.inria.fr/website/sp.html>.  Sounds like it could be interesting.

thesamesam: That's why I mentioned it before ;)

I think it's both a good fit for the work you're interested in _and_ eases review of it.

alejandro-colomar: I'm trying to develop a coccinelle script that would do something similar to <https://github.com/shadow-maint/shadow/pull/1056/commits/941c951cba51d8b050ed82854276054f45485ddc>

So far, I've written 
```diff
$ cat fgets_cast.sp 
@@
expression a, b, c;
type t;

@@

- fgets(a, (t) (b), c)
+ fgets(a, b, c)

@@
expression a, b, c;
type t;

@@

- fgets(a, (t) b, c)
+ fgets(a, b, c)

@@
expression a, b, c;
type t;
@@

- fgetsx(a, (t) (b), c)
+ fgetsx(a, b, c)

@@
expression a, b, c;
type t;
@@

- fgetsx(a, (t) b, c)
+ fgetsx(a, b, c)

@@
expression a, b, c, p;
type t;
@@

- p->fgets(a, (t) (b), c)
+ p->fgets(a, b, c)

@@
expression a, b, c, p;
type t;
@@

- p->fgets(a, (t) b, c)
+ p->fgets(a, b, c)
```

but it results in false positives, such as 
```diff
diff --git i/lib/shadow.c w/lib/shadow.c
index 44436836..5759f259 100644
--- i/lib/shadow.c
+++ w/lib/shadow.c
@@ -202,7 +202,7 @@ struct spwd *fgetspent (FILE * fp)
                return (0);
        }
 
-       if (fgets (buf, sizeof buf, fp) != NULL)
+       if (fgets(buf, sizeof buf, fp) != NULL)
        {
                stpsep(buf, "\n");
                return my_sgetspent (buf);
```

How is that matching `(t)`, if there are no casts?

contrib/adduser.c
```diff
@@ -398,7 +398,7 @@ main (void)
 
   *environ = NULL;
 
-  bzero (cmd, sizeof (cmd));
+  bzero(cmd, sizeof(cmd));
```
alejandro-colomar, line 401: @thesamesam 

If you apply the semantic patch accompanied with the corresponding commit message, and then `git diff -w` against the actual commit, you should be able to ignore these unrelated whitespace changes.  I've written recommendations for how to better review each commit.

I hope that helps reviewing, while at the same time allowing me to apply some house style while doing the changes.

lib/gshadow.c
hallyn: Haven't had a chance to look closely here, but in general, switching from static memory to dynamically allocated and freed memory, if not useful, is a step downward, with added risks of a use-after-free etc.  (OTOH if it really does simplify reasoning about it, or is needed, then of course it's worthwhile)

We've had breakages before when a well meaning patch decided to add a "obviously correct" free that in fact was not always correct, right before exiting!

alejandro-colomar: It was already dynamic memory.  The full diff of the function body is:

```diff
 {
        static size_t buflen = 0;
        static char *buf = NULL;
 
-       char *cp;
-
-       if (0 == buflen) {
-               buf = MALLOC(BUFSIZ, char);
-               if (NULL == buf) {
-                       return NULL;
-               }
-               buflen = BUFSIZ;
-       }
-
        if (NULL == fp) {
                return NULL;
        }
 
-       if (fgets(buf, buflen, fp) == NULL)
+       if (getline(&buf, &buflen, fp) == -1)
+               return NULL;
+       if (stpsep(buf, "\n") == NULL)
                return NULL;
 
-       while (   (strrchr(buf, '\n') == NULL)
-              && (feof (fp) == 0)) {
-               size_t len;
-
-               cp = REALLOC(buf, buflen * 2, char);
-               if (NULL == cp) {
-                       return NULL;
-               }
-               buf = cp;
-               buflen *= 2;
-
-               len = strlen (buf);
-               if (fgets(&buf[len], buflen - len, fp) == NULL)
-                       return NULL;
-       }
-       stpsep(buf, "\n");
        return (sgetsgent (buf));
 }
```

alejandro-colomar: We could restrict the contract to read a limited line, and I would be happy to do it.  But that should be a separate change (and in any case, I'd do it after simplifying the current code to getline(3), which already does in an open-coded way.

alejandro-colomar:
@hallyn  Please review this when you have some time.

alejandro-colomar:
I've removed this PR from the list for this release.  Let's discuss it after the release.

PR #1058, opened 2024-07-31
lib/, src/: Use local time for human-readable dates

alejandro-colomar:
That is, use localtime_r(3) instead of gmtime_r(3).

Closes: <https://github.com/shadow-maint/shadow/issues/1057>
Reported-by: @kenion
Cc: @hallyn 
Cc: @eggert

PR #1059, opened 2024-07-31
DAY_TO_STR() localization

alejandro-colomar:
Cc: @eggert 
Cc: @timparenti
Cc: @kenion

---

Revisions:

<details>
<summary>v2</summary>

-  Print localized dates and times by default.
-  Do not use custom formats.
-  Print timezone information together with date strings, as an RFC 9557 suffix.
-  Fix STRFTIME() macro

```
$ git range-diff shadow/master gh/time time 
1:  2686a033 < -:  -------- src/chage.c: Always use ISO 8601 format (YYYY-MM-DD) for printing dates
-:  -------- > 1:  68a97728 lib/, src/: Use %F instead of %Y-%m-%d with strftime(3)
-:  -------- > 2:  56067591 src/: Use %c for printing localized date&times with strftime(3)
-:  -------- > 3:  928172c4 src/chage.c: print_day_as_date(): Use %x for printing localized dates with strftime(3)
2:  de6f93bf = 4:  e78f9782 src/chage.c: print_day_as_date(): Simplify error handling
-:  -------- > 5:  fc46d1f3 lib/time/, src/: day_to_str(): Add support for formatting localized dates
3:  bd7701fa ! 6:  b2c33c58 src/chage.c: print_day_as_date(): Call DAY_TO_STR() to de-duplicate code
    @@ src/chage.c: static int new_fields (void)
     -          return;
     -  }
     -
    --  STRFTIME(buf, "%Y-%m-%d", &tm);
    +-  STRFTIME(buf, iflg ? "%F" : "%x", &tm);
     -  (void) puts (buf);
    -+  DAY_TO_STR(buf, day);
    ++  DAY_TO_STR(buf, day, iflg);
     +  puts(buf);
      }
      
-:  -------- > 7:  9c0ee33c lib/time/day_to_str.h: day_to_str(): Print numeric timezone together with local dates
-:  -------- > 8:  41a12dfe lib/string/strftime.h: STRFTIME(): Tighten macro definition
```
</details>

<details>
<summary>v3</summary>

-  Use the time zone name/abbreviation rather than the numeric offset.  It's strongly preferred by RFC 9557.  [@eggert]

```
$ git range-diff shadow/master gh/time time 
1:  68a97728 = 1:  68a97728 lib/, src/: Use %F instead of %Y-%m-%d with strftime(3)
2:  56067591 = 2:  56067591 src/: Use %c for printing localized date&times with strftime(3)
[...]

PR #1060, opened 2024-08-03
Revert #1058

alejandro-colomar:
-  This reverts #1058.

```
    Revert "lib/, src/: Use local time for human-readable dates"
    
    This reverts commit 3f5b4b56268269fefed55aa106f382037297d663.
    
    Time-zone conversion of day-precision timestamps is lossy.  Since the
    timestamps are stored as UTC, let's print them as UTC.  That allows
    users to see the actual value that is stored, and thus the value that
    would be used for comparison, for example for deciding if a login is
    valid.
    
    In the following commit, we add time-zone information to the strings,
    which should allow users understand any discrepancy between the date
    input (by default in local time), and the printed one (always in UTC).
```

```
    lib/time/day_to_str.h: day_to_str(): Print time-zone information together with dates
    
    We print dates as UTC dates, but unqualified dates are understood to be
    in local time.  Thus, we need to expressely mark them as UTC dates.
    
    In ISO mode, append a 'Z' without space.  In the current locale, since
    we don't know the format, append the ' UTC'.
```

Link: <https://en.wikipedia.org/wiki/ISO_8601#Local_time_(unqualified)>
Link: <https://lists.gnu.org/archive/html/bug-gnulib/2024-08/msg00015.html>
Cc: @hallyn 
Cc: @ikerexxe 
Cc: @bhaible
Cc: @eggert 
Cc: @kenion

---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase

```
$ git range-diff gh/time..gh/time2 time..time2
1:  57b44b4c = 1:  7cdca6d8 Revert "lib/, src/: Use local time for human-readable dates"
2:  ded86651 = 2:  746e9c49 lib/time/day_to_str.h: day_to_str(): Print time-zone information together with dates
3:  c05f8ae1 < -:  -------- po/es.po: wsfix
```
</details>

### DIFF
--- a/lib/time/day_to_str.c
+++ b/lib/time/day_to_str.c
@@ -7,5 +7,8 @@
 
 #include "time/day_to_str.h"
 
+#include <stdbool.h>
+#include <stddef.h>
 
-extern inline void day_to_str(size_t size, char buf[size], long day);
+
+extern inline void day_to_str(size_t size, char buf[size], long day, bool iso);
--- a/lib/time/day_to_str.h
+++ b/lib/time/day_to_str.h
@@ -45,7 +45,7 @@ day_to_str(size_t size, char buf[size], long day, bool iso)
 		return;
 	}
 
-	if (strftime(buf, size, iso ? "%F" : "%x", &tm) == 0)
+	if (strftime(buf, size, iso ? "%FZ" : "%x UTC", &tm) == 0)
 		strtcpy(buf, "future", size);
 }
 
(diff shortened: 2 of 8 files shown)


alejandro-colomar:
This is queued after #1059 .

ikerexxe:
This PR and https://github.com/shadow-maint/shadow/pull/1059 are driving me a bit crazy, so I will refrain from commenting on them.

PR #1064, opened 2024-08-09
Deprecate non-standard date format

alejandro-colomar:
---

Revisions:

<details>
<summary>v2</summary>

-  Rebase

```
$ git range-diff master..gh/t3 shadow/master..t3 
1:  68a97728 < -:  -------- lib/, src/: Use %F instead of %Y-%m-%d with strftime(3)
2:  d2bfc356 ! 1:  66cbfd63 src/chage.c: print_day_as_date(): Deprecate the non-standard date format
    @@ src/chage.c: print_day_as_date(long day)
     +                stderr);
     +  }
     +
    -   STRFTIME(buf, iflg ? "%F" : "%b %d, %Y", &tm);
        (void) puts (buf);
      }
    + 
```
</details>

### DIFF
--- a/src/chage.c
+++ b/src/chage.c
@@ -252,6 +252,11 @@ print_day_as_date(long day)
 		return;
 	}
 
+	if (!iflg) {
+		fputs(_("This date format is deprecated; use the '-i' option.\n"),
+		      stderr);
+	}
+
 	(void) puts (buf);
 }
 


hallyn:
So the first patch is a noop, I'm ok with it.

I still feel the second one is not ideal.  I'd prefer we simply add a login.defs option to use the standardized formats, and put it in the default login.defs that we ship.  Distros can decide for themselves whether it's safe to change their default.

alejandro-colomar:
> So the first patch is a noop, I'm ok with it.
> 
> I still feel the second one is not ideal. I'd prefer we simply add a login.defs option to use the standardized formats, and put it in the default login.defs that we ship. Distros can decide for themselves whether it's safe to change their default.

If distros knew that they have scripts relying on these features, they could just change the scripts already, instead of login.defs.

I think it's random users who might have scripts that need the old format, and we don't want distros changing the default for them.

I think printing some message to stderr could let users know that the scripts they wrote could break in the future.

In any case, adding a message to stderr is harmless and useful.

What worries you of this change?

hallyn:
> > So the first patch is a noop, I'm ok with it.
> > I still feel the second one is not ideal. I'd prefer we simply add a login.defs option to use the standardized formats, and put it in the default login.defs that we ship. Distros can decide for themselves whether it's safe to change their default.
> 
> If distros knew that they have scripts relying on these features, they could just change the scripts already, instead of login.defs.

Yes, it's more that distros might know what types of users they have.

> I think it's random users who might have scripts that need the old format, and we don't want distros changing the default for them.

I think that's up to the distro.

> I think printing some message to stderr could let users know that the scripts they wrote could break in the future.

I think the combined number of people who either 1. ignore stderr or 2. wrongly parse stderr and will then be broken, will greatly out-number the ppl helped by the message.

> In any case, adding a message to stderr is harmless and useful.
> 
> What worries you of this change?

alejandro-colomar:
> > > So the first patch is a noop, I'm ok with it.
> > > I still feel the second one is not ideal. I'd prefer we simply add a login.defs option to use the standardized formats, and put it in the default login.defs that we ship. Distros can decide for themselves whether it's safe to change their default.
> > 
> > 
> > If distros knew that they have scripts relying on these features, they could just change the scripts already, instead of login.defs.
> 
> Yes, it's more that distros might know what types of users they have.
> 
> > I think it's random users who might have scripts that need the old format, and we don't want distros changing the default for them.
> 
> I think that's up to the distro.
> 
> > I think printing some message to stderr could let users know that the scripts they wrote could break in the future.
> 
> I think the combined number of people who either 1. ignore stderr or 2. wrongly parse stderr and will then be broken, will greatly out-number the ppl helped by the message.

Hmmm, could be.  Still, I don't like adding configurability.  I think I prefer not doing anything at all.

hallyn:
I mean, "it's just chage", I suppose it doesn't matter much either way.  Doing nothing should be fine.

alejandro-colomar:
I'll reopen, just to rebase and see what's remaining of this PR, and then close again.

Issue #1050, opened 2024-07-11
userdel -r -f ... can remove random files under '/'


wolfsage:
Howdy.

It appears that userdel has no protection against erasing a system if a user's home dir is set to '/'. (A lot of users like systemd-network, etc have their homedir set to '/').

If you remove one of them with userdel -r -f, one of two things might happen:

* userdel may open /proc first, attempt to delete something in there, fail, and abort, and you are safe
* userdel may open some other directory first (like /usr!), delete everything in there, and move on, trashing your system.

I can't imagine a good reason userdel should ever attempt to delete / and all files under it. Can this be prevented somehow?

I've seen this issue on Debian 6.1.94-1, and it seems highly repeatable (using a DigitalOcean Debian 12 Droplet).

For some reason, on Debian 6.1.38-2 and earlier, userdel always seems to find /proc first, and never deletes '/'.

This is Debian shadow-utils 4.13 in both cases

alejandro-colomar:
Hmmm, some question that may be dumb:

Does systemd-network (or any user) need to have its homedir set to `/`?

alejandro-colomar:
> Debian 6.1.94-1

What is that?  Debian 6 was dead many years ago.

wolfsage:
Woof, sorry, that's likely the kernel version
```
$ uname -a
Linux 6.1.0-22-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.94-1 (2024-06-21) x86_64 GNU/Linux
```

This is debian bookworm
```
$ cat /etc/debian_version 
12.6
```

alejandro-colomar:
Why should a user have its homedir set to `/`?

wolfsage:
It certainly seems weird to me, but adding this protection still seems like a good thing to do.

alejandro-colomar:
> It certainly seems weird to me,

Please report a bug to systemd (or maybe it's a distro issue?), and please add a link here to see what they say.

> but adding this protection still seems like a good thing to do.

I'm not sure.  Those are heuristics which increase the complexity of the software, possibly adding bugs, and I'm not sure we can write correctly those heuristics.  Should we also try to prevent removing /usr/bin? or /usr/sbin?  Where do we stop?

I'd rather keep it raw, and let one run `rm -rf /`.  If they broke their systems, so that userdel runs `rm -rf /`, it's not our business.  userdel is a symptom, not the cause.

wolfsage:
[...]

PR #1071, opened 2024-09-01
userdel: --force: Allow the flag to be specified twice, for more granularity

alejandro-colomar:
Previously, with a single --force, one could only chose between not forcing, and forcing (which could destroy one's system entirely).

Now one can pass --force, with only a small danger, to skip some checks, and a double --force --force with the meaning of the previous --force, which will possibly destroy the entire system.

Closes: <https://github.com/shadow-maint/shadow/issues/1050>
Reported-by: @wolfsage
Cc: @poettering
Cc: @hallyn 

---

I have NOT tested this.  Please test.

---

Revisions:

<details>
<summary>v2</summary>

-  Reword documentation

```
$ git range-diff shadow/master gh/ff ff 
1:  5d8ddfa5 ! 1:  11c0e8b5 userdel: --force: Allow the flag to be specified twice, for more granularity
    @@ man/userdel.8.xml
     +    </para>
     +    <para>
     +      If specified once,
    -+      it doesn't check if the user is still logged in.
    ++      users are removed
    ++      even if they're still logged in,
    ++      and groups are removed
    ++      even if they are the primary group of a user.
     +      If specified twice,
     +      it skips all safety checks.
          </para>
```
</details>

<details>
<summary>v2b</summary>

-  Reword paragraph.

```
$ git range-diff shadow/master gh/ff ff 
1:  11c0e8b5 ! 1:  cb580580 userdel: --force: Allow the flag to be specified twice, for more granularity
    @@ man/userdel.8.xml
     +    </para>
     +    <para>
     +      If specified once,
    -+      users are removed
    -+      even if they're still logged in,
    -+      and groups are removed
    -+      even if they are the primary group of a user.
    ++      a user is removed
    ++      even if it's still logged in,
    ++      and its primary group is removed
    ++      even if it's the primary group of another user.
     +      If specified twice,
     +      it skips all safety checks.
          </para>
```
</details>

<details>
<summary>v2c</summary>

-  Rebase

```
$ git range-diff gh/master..gh/ff master..ff 
1:  cb580580 = 1:  72af7d32 userdel: --force: Allow the flag to be specified twice, for more granularity
```
</details>

<details>
<summary>v2d</summary>

-  Rebase

```
$ git range-diff master..gh/ff shadow/master..ff 
1:  72af7d32 = 1:  372fdbe2 userdel: --force: Allow the flag to be specified twice, for more granularity
```
</details>

<details>
<summary>v2e</summary>

-  Rebase

```
$ git range-diff alx/master..gh/ff master..ff 
1:  372fdbe2 = 1:  d6d8095d userdel: --force: Allow the flag to be specified twice, for more granularity
```
</details>

<details>
<summary>v2f</summary>

-  Rebase

```
$ git range-diff alx/master..gh/ff shadow/master..ff 
1:  d6d8095d = 1:  cd0f9edb userdel: --force: Allow the flag to be specified twice, for more granularity
```
</details>

<details>
<summary>v2g</summary>

-  Rebase

```
$ git range-diff master..gh/ff shadow/master..ff 
1:  cd0f9edb = 1:  da377a98 userdel: --force: Allow the flag to be specified twice, for more granularity
```
</details>

<details>
<summary>v2h</summary>

-  Rebase

```
$ git range-diff master..gh/ff shadow/master..ff 
1:  da377a98 = 1:  49d5a39f userdel: --force: Allow the flag to be specified twice, for more granularity
```
</details>

<details>
<summary>v2i</summary>

-  Rebase

```
$ git range-diff master..gh/ff shadow/master..ff 
1:  49d5a39f = 1:  8ed1c0cb userdel: --force: Allow the flag to be specified twice, for more granularity
```
</details>

<details>
<summary>v2j</summary>

-  Rebase

```
$ git range-diff master..gh/ff shadow/master..ff 
1:  8ed1c0cb = 1:  2ba0f96a userdel: --force: Allow the flag to be specified twice, for more granularity
```
</details>

<details>
<summary>v2k</summary>

-  Rebase

```
$ git range-diff master..gh/ff shadow/master..ff 
1:  2ba0f96a ! 1:  6d8a7a22 userdel: --force: Allow the flag to be specified twice, for more granularity
    @@ src/userdel.c: int main (int argc, char **argv)
     @@ src/userdel.c: int main (int argc, char **argv)
         * a cron job may be started on her behalf, etc.
         */
    -   if ((prefix[0] == '\0') && !Rflg && user_busy (user_name, user_id) != 0) {
    +   if (streq(prefix, "") && !Rflg && user_busy(user_name, user_id) != 0) {
     -          if (!fflg) {
     +          if (fflg < 1) {
      #ifdef WITH_AUDIT
```
</details>

<details>
<summary>v2l</summary>

-  Rebase

```
$ git range-diff alx/master..gh/ff master..ff 
1:  6d8a7a22 = 1:  76a00097 userdel: --force: Allow the flag to be specified twice, for more granularity
```
</details>

### DIFF
--- a/man/userdel.8.xml
+++ b/man/userdel.8.xml
@@ -79,7 +79,16 @@
 	  <para>
 	    This option forces the removal of the user account
 	    and any other requested actions,
-	    skipping any safety checks.
+	    skipping safety checks.
+	  </para>
+	  <para>
+	    If specified once,
+	    a user is removed
+	    even if it's still logged in,
+	    and its primary group is removed
+	    even if it's the primary group of another user.
+	    If specified twice,
+	    it skips all safety checks.
 	  </para>
 	  <para>
 	    <emphasis>Note:</emphasis> This option is dangerous and may leave
--- a/src/userdel.c
+++ b/src/userdel.c
@@ -80,7 +80,7 @@ static uid_t user_id;
 static gid_t user_gid;
 static char *user_home;
 
-static bool fflg = false;
+static int fflg = 0;
 static bool rflg = false;
 #ifdef WITH_SELINUX
 static bool Zflg = false;
@@ -313,7 +313,7 @@ static void remove_usergroup (void)
 		return;
 	}
 
-	if (!fflg) {
+	if (fflg < 1) {
 		/*
 		 * Scan the passwd file to check if this group is still
 		 * used as a primary group.
@@ -833,7 +833,7 @@ static int remove_mailbox (void)
 		}
 	}
 
-	if (fflg) {
+	if (fflg >= 2) {
 		if (unlink (mailfile) != 0) {
 			fprintf (stderr,
 			         _("%s: warning: can't remove %s: %s\n"),
@@ -998,8 +998,8 @@ int main (int argc, char **argv)
 #endif				/* !WITH_SELINUX */
 		                         long_options, NULL)) != -1) {
 			switch (c) {
-			case 'f':	/* force remove even if not owned by user */
-				fflg = true;
+			case 'f':
+				fflg++;
 				break;
 			case 'h':
 				usage (E_SUCCESS);
@@ -1132,7 +1132,7 @@ int main (int argc, char **argv)
 	 * a cron job may be started on her behalf, etc.
 	 */
 	if (streq(prefix, "") && !Rflg && user_busy(user_name, user_id) != 0) {
-		if (!fflg) {
+		if (fflg < 1) {
 #ifdef WITH_AUDIT
 			audit_logger (AUDIT_DEL_USER, Prog,
 			              "deleting user logged in",
@@ -1161,7 +1161,7 @@ int main (int argc, char **argv)
 			         _("%s: %s home directory (%s) not found\n"),
 			         Prog, user_name, user_home);
 			rflg = 0;
-		} else if ((0 == home_owned) && !fflg) {
+		} else if ((0 == home_owned) && fflg < 2) {
 			fprintf (stderr,
 			         _("%s: %s not owned by %s, not removing\n"),
 			         Prog, user_home, user_name);
@@ -1173,7 +1173,7 @@ int main (int argc, char **argv)
 
 #ifdef EXTRA_CHECK_HOME_DIR
 	/* This may be slow, the above should be good enough. */
-	if (rflg && !fflg) {
+	if (rflg && fflg < 2) {
 		struct passwd *pwd;
 		/*
 		 * For safety, refuse to remove the home directory if it


src/userdel.c
```diff
@@ -312,7 +312,7 @@ static void remove_usergroup (void)
 		return;
 	}
 
-	if (!fflg) {
+	if (fflg < 1) {
 		/*
 		 * Scan the passwd file to check if this group is still
 		 * used as a primary group.
```
alejandro-colomar, line 318: Hmm, this is not covered by the wording I used in the manual page.  I'll try to come up with some better wording.

If anyone has better ideas for the wording, please let me know.

man/userdel.8.xml
```diff
@@ -79,7 +79,16 @@
...
+	    skipping safety checks.
+	  </para>
+	  <para>
+	    If specified once,
+	    a user is removed
+	    even if it's still logged in,
+	    and its primary group is removed
+	    even if it's the primary group of another user.
+	    If specified twice,
+	    it skips all safety checks.
```
alejandro-colomar, line 91: Does this distinction make sense?  Should the group thing go with `-ff`?

alejandro-colomar: I think it's ok, but I ask just in case.

alejandro-colomar:
@wolfsage, can you please check if this looks good to you?  Also, would you please test it?

ikerexxe:
I don't like this change and I think it'll be problematic to some users.

alejandro-colomar:
> I don't like this change and I think it'll be problematic to some users.

Yep, we have problems either way.  This is half broken, so users will complain about it being quite unsafe.  Yet, scripts may rely on that brokenness, so fixing it may break scripts.

I'm in favour of fixing it and potentially breaking some random scripts.  Any opinions?

ikerexxe:
From my perspective, the `force' option implies deleting the user regardless of the consequences. This option should be used with care, because as you mention, it can have fatal consequences for the system.
The double use of the `force' option is problematic and will cause more problems than it intends to solve. I would try to keep it as simple as possible and avoid such a complication.

alejandro-colomar:
> From my perspective, the `force' option implies deleting the user regardless of the consequences. This option should be used with care, because as you mention, it can have fatal consequences for the system. The double use of the `force' option is problematic and will cause more problems than it intends to solve. I would try to keep it as simple as possible and avoid such a complication.

There's prior art in other software for this schema of asking for double `-ff` for more dangerous stuff.

This is for example from git-clean(1):
```
     -f, --force
         If the Git configuration variable clean.requireForce is not
         set to false, git clean will refuse to delete files or
         directories unless given -f. Git will refuse to modify
         untracked nested git repositories (directories with a .git
         subdirectory) unless a second -f is given.
```

hallyn:
Alex, thank you for continuing to try to make this better.

Please don't make any changes yet based on what I say here, I'm just thinking aloud :)

First, there's a difference between "home is not owned by the user" and "home is / and we're all gonna die!".  Shadow is not that "generic" of a package.  I think it would be ok to add a list of "do not touch" paths, and always refuse to delete those no matter what: /, /usr, /etc, and /var at least.  I suppose we could even make the list configurable through /etc/login.defs, so distros can add their own entries.

Then, for the "home not owned by user" case, we just need to draw line between not breaking any current users and complicating code.  The main options are: 1. change nothing (which if we implement the never-delete list above becomes far less scary), 2. add a default-off DELETE_UNOWNED_HOME option to login.defs, 3. add -ff like you have, and 4. both 2 and 3.

alejandro-colomar:
> Shadow is not that "generic" of a package. I think it would be ok to add a list of "do not touch" paths, and always refuse to delete those no matter what: /, /usr, /etc, and /var at least.

How about refusing to add users with such homedirs in the first place?

(Just thinking out loud too.)

alejandro-colomar:
> Then, for the "home not owned by user" case, we just need to draw line between not breaking any current users and complicating code. The main options are: [...] 2. add a default-off DELETE_UNOWNED_HOME option to login.defs.  [...]

I tend to dislike configurability of programs.  It adds complexity to both sysadmins, and the program itself.  I think I would prefer a hardcoded list of directories that should not be removed.

hallyn:
> > Shadow is not that "generic" of a package. I think it would be ok to add a list of "do not touch" paths, and always refuse to delete those no matter what: /, /usr, /etc, and /var at least.
> 
> How about refusing to add users with such homedirs in the first place?
> 
> (Just thinking aloud too.)

The problem with that is that a mistaken or malicious /etc/passwd edit could easily change it afterwards.

alejandro-colomar:
> > > Shadow is not that "generic" of a package. I think it would be ok to add a list of "do not touch" paths, and always refuse to delete those no matter what: /, /usr, /etc, and /var at least.
> > 
> > 
> > How about refusing to add users with such homedirs in the first place?
> > (Just thinking aloud too.)
> 
> The problem with that is that a mistaken or malicious /etc/passwd edit could easily change it afterwards.

I think `-ff` is good in the long term (even if we may want to do other changes too).  What's your opinion about breaking existing scripts (or hardcoded brains)?  I don't know how much people would rely on `-f` for removing not-owned dirs, and other stuff we're hiding under `-ff`.

hallyn:
> What's your opinion about breaking existing scripts (or hardcoded brains)? I don't know how much people would rely on `-f` for removing non-owned dirs, and other stuff we're hiding under `-ff`.

Well actually, I would imagine there are a lot of places that have end-of-semester or whatever scripts that nuke the out-going students' accounts.  For those, having -f return error and -ff succeed would be a problem - if they ever have a situation like this.  It might be better if -f just skipped the deletion and returned success.  But that has its own problems.

alejandro-colomar:
> > What's your opinion about breaking existing scripts (or hardcoded brains)? I don't know how much people would rely on `-f` for removing non-owned dirs, and other stuff we're hiding under `-ff`.
> 
> Well actually, I would imagine there are a lot of places that have end-of-semester or whatever scripts that nuke the out-going students' accounts. For those, having -f return error and -ff succeed would be a problem - if they ever have a situation like this. It might be better if -f just skipped the deletion and returned success. But that has its own problems.

But I expect those user accounts to either have an owned dir, or have something like /nonexistent that shouldn't be removed.  In what case would one have a directory that isn't owned but they want to remove it anyway?

hallyn:
> > 
> 
> But I expect such user accounts to either have an owned dir, or have something like /nonexistent that shouldn't be removed. In what case would one have a directory that isn't owned but they want to remove it anyway?

That makes sense, yeah.  So system accounts pointing at some /var/lib/something owned by another system user would be more likely.

But, I was thinking based on the discussion that this patch would make it so that whereas
previously you would just not delete the homedir, how you'd return an error.  But that
doesn't appear to be the case.  So I'm good with this patch.

@ikerexxe do you still feel uncomfortable with this patch?

ikerexxe:
It seems counter-intuitive to me, but I won't object if you think it's the best solution

alejandro-colomar:
Turning into draft until after 4.17.0.

alejandro-colomar:
Since the people that were interested in this haven't responded, I'll close this.  Feel free to suggest re-opening if anybody is interested in this or has more time to test.

PR #1076, opened 2024-09-13
man: Fix information in manual pages

stoeckmann:
Trivial typo fix and list correct programs


ikerexxe (approved):
LGTM. Thanks for the patches.

PR #1077, opened 2024-09-13
lib/run_part: Minor cleanups

stoeckmann:
- Use correct data type
- Unify logging (at least the fprintf calls)
- Reduce visibility of locally used function

### DIFF
--- a/lib/run_part.c
+++ b/lib/run_part.c
@@ -15,7 +15,7 @@
 #include "shadowlog_internal.h"
 
 
-int run_part (char *script_path, const char *name, const char *action)
+static int run_part (char *script_path, const char *name, const char *action)
 {
 	pid_t pid;
 	int wait_status;
--- a/lib/run_part.c
+++ b/lib/run_part.c
@@ -17,9 +17,9 @@
 
 int run_part (char *script_path, const char *name, const char *action)
 {
-	int pid;
+	pid_t pid;
 	int wait_status;
-	int pid_status;
+	pid_t pid_status;
 	char *args[] = { script_path, NULL };
 
 	pid=fork();
--- a/lib/run_part.c
+++ b/lib/run_part.c
@@ -15,7 +15,7 @@
 #include "shadowlog_internal.h"
 
 
-static int run_part (char *script_path, const char *name, const char *action)
+static int run_part(char *script_path, const char *name, const char *action)
 {
 	pid_t pid;
 	int wait_status;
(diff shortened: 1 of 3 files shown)


lib/run_part.c
```diff
@@ -61,7 +61,8 @@ int run_parts (const char *directory, const char *name, const char *action)
 		struct stat  sb;
 
 		if (asprintf(&s, "%s/%s", directory, namelist[n]->d_name) == -1) {
-			fprintf(stderr, "could not allocate memory\n");
+			fprintf(shadow_logfd, "%s: could not allocate memory\n",
+			    namelist[n]->d_name);
```
alejandro-colomar, line 65: But there are also perror(3) calls.  The commit should also justify why those still use stderr.

lib/run_part.c
```diff
@@ -15,7 +15,7 @@
 #include "shadowlog_internal.h"
 
 
-int run_part (char *script_path, const char *name, const char *action)
+static int run_part (char *script_path, const char *name, const char *action)
```
alejandro-colomar, line 18: LGTM.

lib/run_part.c
```diff
@@ -17,9 +17,9 @@
 
 int run_part (char *script_path, const char *name, const char *action)
 {
-	int pid;
+	pid_t pid;
 	int wait_status;
-	int pid_status;
+	pid_t pid_status;
```
alejandro-colomar, line 22: LGTM

lib/run_part.c
stoeckmann: Or I just replace them as well. ;) Done.

alejandro-colomar: I'd like at least a justification of why not unify the other way around.  :)

stoeckmann: https://github.com/shadow-maint/shadow/blob/master/lib/shadowlog_internal.h#L5

stoeckmann: I'd accept arguing about keeping perror after the fork, but simply writing into stderr from a fork isn't clean either. Otherwise the rational is that the logging infrastructure is there, both cases were available in that file and the majority in lib/ uses the logging infrastructure (as expected).

alejandro-colomar: I don't say it's wrong, but the commit message should say something.  Something like that would work.

stoeckmann: Adjusted

lib/run_part.c
```diff
@@ -15,23 +15,23 @@
...
-		perror ("Could not fork");
+		fprintf (shadow_logfd, "Could not fork: %s\n", strerror(errno));
 		return 1;
 	}
 	if (pid==0) {
 		setenv ("ACTION",action,1);
 		setenv ("SUBJECT",name,1);
 		execv (script_path,args);
-		perror ("execv");
+		fprintf (shadow_logfd, "execv: %s\n", strerror(errno));
```
alejandro-colomar, line 34: While changing the perror after fork, it might be good to change the string to "fork: %s\n", just like the others, for consistency.

lib/run_part.c
```diff
@@ -61,7 +61,7 @@ int run_parts (const char *directory, const char *name, const char *action)
 		struct stat  sb;
 
 		if (asprintf(&s, "%s/%s", directory, namelist[n]->d_name) == -1) {
-			fprintf(stderr, "could not allocate memory\n");
+			fprintf (shadow_logfd, "could not allocate memory\n");
```
alejandro-colomar: I would rather do the opposite white-space change.  The tendency is to get rid of it.

alejandro-colomar (approved):
```
```

I'll let someone else merge.  Thanks!

lib/run_part.c
```diff
@@ -24,7 +24,7 @@ static int run_part (char *script_path, const char *name, const char *action)
 
 	pid=fork();
 	if (pid==-1) {
-		fprintf (shadow_logfd, "Could not fork: %s\n", strerror(errno));
+		fprintf (shadow_logfd, "fork: %s\n", strerror(errno));
```
ikerexxe: Can you add `failed` at the end to make it more understandable when reading the logs? Something like `fork failed: %s...`

lib/run_part.c
```diff
@@ -61,7 +61,7 @@ int run_parts (const char *directory, const char *name, const char *action)
 		struct stat  sb;
 
 		if (asprintf(&s, "%s/%s", directory, namelist[n]->d_name) == -1) {
-			fprintf (shadow_logfd, "could not allocate memory\n");
+			fprintf (shadow_logfd, "asprintf: %s\n", strerror(errno));
```
ikerexxe: Same here, if not I fear it'll be difficult to understand the problem. Example `asprintf failed...`

ikerexxe (commented):
Two minor comments. Once those are fixed I'll proceed to merge.

lib/run_part.c
alejandro-colomar: The format used here is common in many programs:

```sh
$ cat /nonexistent
cat: /nonexistent: No such file or directory
```

The errno description already implies an error or failure.

lib/run_part.c
alejandro-colomar:
```
asprintf: Cannot allocate memory
```

Should imply a failure.  :)

lib/run_part.c
stoeckmann: I agree with @alejandro-colomar here. It's pretty much the same output you would get with the `perror` calls as they are in this file right now as well.

lib/run_part.c
ikerexxe: You are right here

ikerexxe (approved):
LGTM! Thank you for the patches

PR #1078, opened 2024-09-19
lib: Eliminate dead code

stoeckmann:
The tz function is only called if ENV_TZ starts with a slash.

If the specified file cannot be read, the code implies that ENV_TZ would be returned if it does not start with a slash.

Since we know that it DOES start with a slash, the code can be simplified to state that "TZ=CST6CDT" is returned as a default if the specified file cannot be read.

Benefit of this change is that strcpy's use case here can be easier verified.

### DIFF
--- a/lib/tz.c
+++ b/lib/tz.c
@@ -32,27 +32,23 @@
 /*@observer@*/const char *tz (const char *fname)
 {
 	FILE *fp = NULL;
+	const char *result;
 	static char tzbuf[BUFSIZ];
-	const char *def_tz = "TZ=CST6CDT";
 
 	fp = fopen (fname, "r");
 	if (   (NULL == fp)
 	    || (fgets (tzbuf, sizeof (tzbuf), fp) == NULL)) {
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
Heh, indeed there's even another case of dead code in that function: The initialization of `def_tz` is unused:

```c
$ cat lib/tz.c | grep -e { -e } -e '\<if\>' -e def_tz
{
	const char *def_tz = "TZ=CST6CDT";
	if (   (NULL == fp)
	    || (fgets (tzbuf, sizeof (tzbuf), fp) == NULL)) {
		def_tz = getdef_str ("ENV_TZ");
		if ((NULL == def_tz) || ('/' == def_tz[0])) {
			def_tz = "TZ=CST6CDT";
		}
		strcpy (tzbuf, def_tz);
	} else {
	}
	if (NULL != fp) {
	}
}
```

since it's later overwritten inside the condition.

alejandro-colomar:
LGTM.  Thanks!

alejandro-colomar (approved):

PR #1084, opened 2024-10-02
src/check_subid_range.c: Remove dead code

alejandro-colomar:
I forgot to remove the setting of errno when I switched from strtoul_noneg() to str2ul().  strtoul(3) needs errno for determining success, but str2ul() does not.

Fixes: f3a1e1cf098d ("src/check_subid_range.c: Call str2ul() instead of strtoul_noneg()")


ikerexxe (approved):
LGTM. Thanks for the patch!

PR #1085, opened 2024-10-03
Simplify string handling (piecemeal)

alejandro-colomar:
This is a subset of the patches from #1048 for an easier review.

Cc: @hallyn 
Cc: @ikerexxe

### DIFF
--- a/lib/fields.c
+++ b/lib/fields.c
@@ -16,6 +16,7 @@
 #include <stdio.h>
 
 #include "prototypes.h"
+#include "string/strchr/stpspn.h"
 #include "string/strchr/strrspn.h"
 #include "string/strtok/stpsep.h"
 
@@ -91,12 +92,7 @@ void change_field (char *buf, size_t maxsize, const char *prompt)
 		 * entering a space.  --marekm
 		 */
 		stpcpy(strrspn(newf, " \t\n"), "");
-
-		cp = newf;
-		while (isspace (*cp)) {
-			cp++;
-		}
-
+		cp = stpspn(newf, " \t");
 		strcpy (buf, cp);
 	}
 }
--- a/lib/getdate.y
+++ b/lib/getdate.y
@@ -24,14 +24,15 @@
 # undef static
 #endif
 
-#include <stdio.h>
 #include <ctype.h>
+#include <stdio.h>
+#include <string.h>
 #include <time.h>
 
 #include "attr.h"
 #include "getdate.h"
+#include "string/strchr/stpspn.h"
 
-#include <string.h>
 
 /* Some old versions of bison generate parsers that use bcopy.
    That loses on systems that don't provide the function, so we have
@@ -746,8 +747,7 @@ yylex (void)
 
   for (;;)
     {
-      while (isspace (*yyInput))
-	yyInput++;
+      yyInput = stpspn(yyInput, " \t");
 
       if (isdigit (c = *yyInput) || c == '-' || c == '+')
 	{
--- a/lib/gshadow.c
+++ b/lib/gshadow.c
@@ -156,7 +156,7 @@ void endsgent (void)
 	}
 
 	if (fgetsx(buf, buflen, fp) == buf) {
-		while (   ((cp = strrchr (buf, '\n')) == NULL)
+		while (   (strrchr(buf, '\n') == NULL)
 		       && (feof (fp) == 0)) {
 			size_t len;
 
--- a/lib/limits.c
+++ b/lib/limits.c
@@ -36,6 +36,7 @@
 #include "atoi/str2i/str2s.h"
 #include "atoi/str2i/str2u.h"
 #include "string/memset/memzero.h"
+#include "string/strchr/stpspn.h"
 #include "typetraits.h"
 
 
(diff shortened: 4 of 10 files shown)


lib/fields.c
```diff
@@ -91,12 +92,7 @@ void change_field (char *buf, size_t maxsize, const char *prompt)
 		 * entering a space.  --marekm
 		 */
 		stpcpy(strrspn(newf, " \t\n"), "");
-
-		cp = newf;
-		while (isspace (*cp)) {
-			cp++;
-		}
-
+		cp = stpspn(newf, " \t");
```
ikerexxe, line 95: How was the tab calculated previously? Was it the equivalent of 1, 4, 8... characters? And now?

ikerexxe (commented):
Did you test the changes in the last commit? The change raises concerns about its potential to work correctly.

alejandro-colomar:
> Did you test the changes in the last commit? The change raises concerns about its potential to work correctly.

Do you mean this?

```diff
commit 5e0c907a5b72b1f8a506939182f283ed4c088dbb (HEAD -> str, gh/str)
Author: Alejandro Colomar <alx@kernel.org>
Date:   Wed Jul 3 23:49:04 2024 +0200

    src/suauth.c: check_su_auth(): Use pointers to simplify
    
    Signed-off-by: Alejandro Colomar <alx@kernel.org>

diff --git a/src/suauth.c b/src/suauth.c
index 895f1278..cbb33dc8 100644
--- a/src/suauth.c
+++ b/src/suauth.c
@@ -18,6 +18,7 @@
 
 #include "defines.h"
 #include "prototypes.h"
+#include "string/strchr/stpspn.h"
 #include "string/strchr/strrspn.h"
 #include "string/strtok/stpsep.h"
 
@@ -47,7 +48,6 @@ int check_su_auth (const char *actual_id,
                    const char *wanted_id,
                    bool su_to_root)
 {
-       int posn;
        const char field[] = ":";
        FILE *authfile_fd;
        char temp[1024];
@@ -73,6 +73,8 @@ int check_su_auth (const char *actual_id,
        }
 
        while (fgets (temp, sizeof (temp), authfile_fd) != NULL) {
+               char  *p;
+
                lines++;
 
                if (stpsep(temp, "\n") == NULL) {
@@ -84,12 +86,11 @@ int check_su_auth (const char *actual_id,
 
                stpcpy(strrspn(temp, " \t"), "");
 
-               posn = strspn(temp, " \t");
-
-               if (temp[posn] == '#' || temp[posn] == '\0') {
+               p = stpspn(temp, " \t");
+               if (*p == '#' || *p == '\0')
                        continue;
-               }
-               if (!(to_users = strtok (temp + posn, field))
+
+               if (!(to_users = strtok(p, field))
                    || !(from_users = strtok (NULL, field))
                    || !(action = strtok (NULL, field))
                    || strtok (NULL, field)) {
```

What do you think might be incorrect?

`stpcpn()` is defined as `s_ + strspn(s_, accept);`
so `p = stpspn(temp, ...)` would be equivalent to `p = temp + strspn(temp, ...)`
so `*p` would be equivalent to `*(temp + posn)`, so equivalent to `temp[posn]`.

But I haven't tested this.

lib/fields.c
alejandro-colomar:
```diff
$ git log -p alx/master..str -- lib/fields.c
commit 75c094da708e9da091e5b0e737b704f2e816b140
Author: Alejandro Colomar <alx@kernel.org>
Date:   Wed Jul 3 22:47:45 2024 +0200

    lib/, src/: Use stpspn() instead of its pattern
    
    Signed-off-by: Alejandro Colomar <alx@kernel.org>

diff --git a/lib/fields.c b/lib/fields.c
index 555f4bed..32399fe5 100644
--- a/lib/fields.c
+++ b/lib/fields.c
@@ -16,6 +16,7 @@
 #include <stdio.h>
 
 #include "prototypes.h"
+#include "string/strchr/stpspn.h"
 #include "string/strchr/strrspn.h"
 #include "string/strtok/stpsep.h"
 
@@ -91,12 +92,7 @@ void change_field (char *buf, size_t maxsize, const char *pr>
                 * entering a space.  --marekm
                 */
                stpcpy(strrspn(newf, " \t\n"), "");
-
-               cp = newf;
-               while (isspace (*cp)) {
-                       cp++;
-               }
-
+               cp = stpspn(newf, " \t");
                strcpy (buf, cp);
        }
 }
```

The tab is one byte.  It doesn't matter how it is interpreted; as it is stored in the string as a single `0x09` byte.  isspace('\t') returned true, an cp++ advanced one byte.

alejandro-colomar: See also ascii(7).  <https://www.man7.org/linux/man-pages/man7/ascii.7.html>

hallyn:
Let me take a closer look at this tonight.  I'm pretty sure I already reviewed it as part of 1048 but I'll make sure :)

hallyn (approved):
lgtm.  Only possible issue would still be replacing isspace() with only checking for ' ' and '\t', but the other chars like \f aren't really valid anyway, i believe.

ikerexxe:
> Do you mean this?
> 
> ```diff
> commit 5e0c907a5b72b1f8a506939182f283ed4c088dbb (HEAD -> str, gh/str)

Yes, I meant that one, and there's nothing I can point to that might fail. It was just a feeling more than anything else.

PR #1095, opened 2024-10-16
Avoid zero-size reallocation calls

alejandro-colomar:
Zero-length allocation calls are non-portable, and may result in nasty
bugs when porting software to other platforms.  Let's just make sure
that we never do that.  Some algorithms need malloc(0) or realloc(p, 0)
to just succeed and give a pointer that has zero elements (but is not
NULL); let's just over-allocate 1 byte, which guarantees success, and
doesn't hurt at all.

BTW, I wonder what's the history about this issue.  If anyone knows what
was the original behavior of malloc(0) and realloc(p, 0), and can
illuminate, it would be very interesting.  Was it ANSI C that goofed it
by allowing NULL to be returned?  Was it before?

---

Revisions:

<details>
<summary>v2</summary>

-  Do not avoid malloc(0) calls.  The only broken thing is realloc(p,0).  [@nabijaczleweli]

```
$ git range-diff master gh/alloc0 alloc0 
1:  39d2b51f = 1:  39d2b51f lib/alloc/realloc*.h: Rename macro parameter
2:  57b53dfe ! 2:  4b49c0c9 lib/alloc/: Always allocate at least 1 byte
    @@ Metadata
     
      ## Commit message ##
    -    lib/alloc/: Always allocate at least 1 byte
    +    lib/alloc/realloc*.h: Always reallocate at least 1 byte
     
    -    Zero-length allocation calls are non-portable, and may result in nasty
    -    bugs when porting software to other platforms.  Let's just make sure
    -    that we never do that.  Some algorithms need malloc(0) or realloc(p, 0)
    -    to just succeed and give a pointer that has zero elements (but is not
    -    NULL); let's just over-allocate 1 byte, which guarantees success, and
    -    doesn't hurt at all.
    -
    -    BTW, I wonder what's the history about this issue.  If anyone knows what
    -    was the original behavior of malloc(0) and realloc(p, 0), and can
    -    illuminate, it would be very interesting.  Was it ANSI C that goofed it
    -    by allowing NULL to be returned?  Was it before?
    +    glibc's realloc(3) is broken.  It was good until 2006/2007, when it was
    +    changed to conform to C89, which had a bogus specification that required
    +    that it returns NULL.  However, by the time glibc was changed to
    +    conform to the broken C89, the standard specification had already been
    +    fixed in C99.  Thus, the change in glibc made glibc non-conforming to
    +    C99 (it had been conforming previously).
     
    +    Link: <https://github.com/shadow-maint/shadow/pull/1095>
    +    Link: <https://nabijaczleweli.xyz/content/blogn_t/017-malloc0.html>
    +    Co-developed-by: наб <nabijaczleweli@nabijaczleweli.xyz>
     
    - ## lib/alloc/calloc.h ##
    -@@
    - 
    - #define CALLOC(n, type)                                                       \
    - (                                                                             \
    --  (type *) calloc(n, sizeof(type))                                      \
    -+  (type *) calloc((n) ?: 1, sizeof(type))                               \
    - )
    - 
    - 
    -
    - ## lib/alloc/malloc.h ##
    -@@
    - 
    - #define MALLOC(n, type)                                                       \
    - (                                                                             \
    --  (type *) mallocarray(n, sizeof(type))                                 \
    -+  (type *) mallocarray((n) ?: 1, sizeof(type))                          \
    - )
    - 
    - 
    -@@ lib/alloc/malloc.h: inline void *mallocarray(size_t nmemb, size_t size);
    - inline void *
    - mallocarray(size_t nmemb, size_t size)
    - {
    --  return reallocarray(NULL, nmemb, size);
    -+  return reallocarray(NULL, nmemb ?: 1, size);
    - }
    - 
    - 
    -
      ## lib/alloc/realloc.h ##
     @@
      
```
</details>

<details>
<summary>v2b</summary>

-  Signed-off-by @nabijaczleweli 

```
$ git range-diff master gh/alloc0 alloc0 
1:  39d2b51f = 1:  39d2b51f lib/alloc/realloc*.h: Rename macro parameter
2:  4b49c0c9 ! 2:  98af83c4 lib/alloc/realloc*.h: Always reallocate at least 1 byte
    @@ Commit message
         Link: <https://github.com/shadow-maint/shadow/pull/1095>
         Link: <https://nabijaczleweli.xyz/content/blogn_t/017-malloc0.html>
         Co-developed-by: наб <nabijaczleweli@nabijaczleweli.xyz>
    +    Signed-off-by: наб <nabijaczleweli@nabijaczleweli.xyz>
     
      ## lib/alloc/realloc.h ##
```
</details>

<details>
<summary>v2c</summary>

-  Reword commit message:
   -  C17 broke the standard definition of realloc(3) again.
   -  It seems glibc moved earlier to conform to C89.

```
$ git range-diff master gh/alloc0 alloc0 
1:  39d2b51f = 1:  39d2b51f lib/alloc/realloc*.h: Rename macro parameter
2:  98af83c4 ! 2:  12a0f304 lib/alloc/realloc*.h: Always reallocate at least 1 byte
    @@ Metadata
      ## Commit message ##
         lib/alloc/realloc*.h: Always reallocate at least 1 byte
     
    -    glibc's realloc(3) is broken.  It was good until 2006/2007, when it was
    -    changed to conform to C89, which had a bogus specification that required
    -    that it returns NULL.  However, by the time glibc was changed to
    -    conform to the broken C89, the standard specification had already been
    -    fixed in C99.  Thus, the change in glibc made glibc non-conforming to
    -    C99 (it had been conforming previously).
    +    glibc's realloc(3) is broken.  It was originally good (I believe) until
    +    at some point, when it was changed to conform to C89, which had a bogus
    +    specification that required that it returns NULL.  C99 fixed the mistake
    +    from C89, and so glibc's realloc(3) is non-conforming to
    +    C99/C11/POSIX.1-2008.  C17 broke again the definition of realloc(3).
     
         Link: <https://github.com/shadow-maint/shadow/pull/1095>
         Link: <https://nabijaczleweli.xyz/content/blogn_t/017-malloc0.html>
    +    Link: <https://inbox.sourceware.org/libc-alpha/5gclfbrxfd7446gtwd2x2gfuquy7ukjdbrndphyfmfszxlft76@wwjz7spd4vd7/T/#t>
         Co-developed-by: наб <nabijaczleweli@nabijaczleweli.xyz>
         Signed-off-by: наб <nabijaczleweli@nabijaczleweli.xyz>
```
</details>

<details>
<summary>v2d</summary>

-  Some review tags.  [@ikerexxe , @eggert ]

```
$ git range-diff master gh/alloc0 alloc0 
1:  39d2b51f = 1:  39d2b51f lib/alloc/realloc*.h: Rename macro parameter
2:  12a0f304 ! 2:  73a41048 lib/alloc/realloc*.h: Always reallocate at least 1 byte
    @@ Commit message
         Link: <https://inbox.sourceware.org/libc-alpha/5gclfbrxfd7446gtwd2x2gfuquy7ukjdbrndphyfmfszxlft76@wwjz7spd4vd7/T/#t>
         Co-developed-by: наб <nabijaczleweli@nabijaczleweli.xyz>
         Signed-off-by: наб <nabijaczleweli@nabijaczleweli.xyz>
    +    Reviewed-by: Iker Pedrosa <ipedrosa@redhat.com>
    +    Acked-by: Paul Eggert <eggert@cs.ucla.edu>
     
      ## lib/alloc/realloc.h ##
```
</details>

### DIFF
--- a/lib/alloc/realloc.h
+++ b/lib/alloc/realloc.h
@@ -13,7 +13,7 @@
 
 #define REALLOC(p, n, type)                                                   \
 (                                                                             \
-	_Generic(p, type *:  (type *) reallocarray(p, n, sizeof(type)))       \
+	_Generic(p, type *: (type *) reallocarray(p, (n) ?: 1, sizeof(type))) \
 )
 
 
--- a/lib/alloc/realloc.h
+++ b/lib/alloc/realloc.h
@@ -11,9 +11,9 @@
 #include <stdlib.h>
 
 
-#define REALLOC(ptr, n, type)                                                 \
+#define REALLOC(p, n, type)                                                   \
 (                                                                             \
-	_Generic(ptr, type *:  (type *) reallocarray(ptr, n, sizeof(type)))   \
+	_Generic(p, type *:  (type *) reallocarray(p, n, sizeof(type)))       \
 )
 
 
--- a/lib/alloc/reallocf.h
+++ b/lib/alloc/reallocf.h
@@ -14,9 +14,9 @@
 #include "attr.h"
 
 
-#define REALLOCF(ptr, n, type)                                                \
+#define REALLOCF(p, n, type)                                                  \
 (                                                                             \
-	_Generic(ptr, type *:  (type *) reallocarrayf(ptr, n, sizeof(type)))  \
+	_Generic(p, type *:  (type *) reallocarrayf(p, n, sizeof(type)))      \
 )
 
 
--- a/lib/alloc/reallocf.h
+++ b/lib/alloc/reallocf.h
@@ -16,7 +16,7 @@
 
 #define REALLOCF(p, n, type)                                                  \
 (                                                                             \
-	_Generic(p, type *:  (type *) reallocarrayf(p, n, sizeof(type)))      \
+	_Generic(p, type *: (type *) reallocarrayf(p, (n) ?: 1, sizeof(type)))\
 )
 
 
@@ -30,10 +30,9 @@ reallocarrayf(void *p, size_t nmemb, size_t size)
 {
 	void  *q;
 
-	q = reallocarray(p, nmemb, size);
+	q = reallocarray(p, nmemb ?: 1, size ?: 1);
 
-	/* realloc(p, 0) is equivalent to free(p);  avoid double free.  */
-	if (q == NULL && nmemb != 0 && size != 0)
+	if (q == NULL)
 		free(p);
 	return q;
 }


alejandro-colomar:
@nabijaczleweli

You play with old Unix systems.  Maybe you know what was the original behavior of v7, SysV, and BSD?  And when did that change?  It would be interesting to document that in malloc(3).

nabijaczleweli:
I don't play with them, I just look at the dumps. To that end:

SVr4 has malloc() => if(size < MINSIZE) => _smalloc(size) => 
```c
static VOID *   _smalloc(size)
size_t          size;
        {
        reg TREE        *tp;
        reg size_t      i;

        /**/ ASSERT(size%WORDSIZE == 0);
        /* want to return a unique pointer on malloc(0) */
        if(size == 0)
                size = WORDSIZE;
```
(it can also reuse the last freed block as the first step, but the minimum is always `WORDSIZE`, and the actual block is actually a little bigger).

lib/libc/port/gen/ malloc.h & mallint.h

---

Solaris 2.8, expectedly, is the same (with some mutex wrappers).

---

4.4BSD is malloc(nbytes) => `if (nbytes <= (n = pagesz - sizeof (*op) - RSLOP)) {` (RSLOP is 0/2 in release/debug) =>
```
                amt = 8;        /* size of first bucket */
                bucket = 0;
```
and `amt` is the block size.

usr/src/lib/libc/stdlib/malloc.c (get the CSRG CDs once the archive is back up)

---

V7 has malloc(nbytes) => `nw = (nbytes+WORD+WORD-1)/WORD;` (Number of Words)

usr/src/libc/gen/malloc.c (Hippe dump)

---

These are like 40 years old (sure, 30), you're much more interested in systems that are real *now*.

---

POSIX.1-1988 defers to C.

> 8. Language-Specific Services for the C Programming Language
> 8.1 Referenced C Language Routines.
> The functions listed below will be described in the indicated sections of the C Standard. IEEE Std 1003.1-1988 with the C Language Binding comprises these functions, the extensions to them described in this chapter, and the rest of the requirements stipulated in this standard. The functions appended with plus signs (+) have requirements beyond those set forth in the C Standard.
> 
> 4.10 General Utilities
Functions: abs, atof, atoi, atol, rand, srand, calloc, free, malloc, realloc, abort+, exit, getenv+, bsearch, qsort.

And appendix B says 

> Though malloc() §8.1 and free{) §8.1 are in the C Standard and are required by Referenced C Language Routines §8.1 of the present standard, neither
brk{) nor sbrk() occur in either standard (although they were in the 1984 /usr/group Standard), because this standard is designed to provide the basic
set of functions required to write a Conforming POSIX Application; the underlying implementation of malloc() or free{) is not an appropriate concern for the standard.

And later issues of the standard inline C malloc() explicitly.

But it sure looks like no main-line historical UNIX was `malloc(0) = NULL`. At least not by the end (I will not be diffing this, one has to assume that it's in every interstitial version as well).

idk if C has rationales or "History of Decisions Made"s like POSIX does. Not my barn, not my ass.

alejandro-colomar:
> I don't play with them, I just look at the dumps.

:)

> To that end:

[... Extensive proof that SVr4, Solaris 2.8, 4.4BSD, and Unix V7 all do the right thing ...]

> These are like 40 years old (sure, 30), you're much more interested in systems that are real _now_.

Actually those were very interesting to me.  I was wondering where the hell this brain-damaged trend of allowing NULL to be returned actually originated.

> POSIX.1-1988 defers to C.

[... The original POSIX is silent on the matter ...]

> And later issues of the standard inline C malloc() explicitly.
> 
> But it sure looks like no main-line historical UNIX was `malloc(0) = NULL`. At least not by the end (I will not be diffing this, one has to assume that it's in every interstitial version as well).

Hmmmm.  Very interesting.  It might have been a hallucination of ANSI C (X3J11).

> idk if C has rationales or "History of Decisions Made"s like POSIX does. Not my barn, not my ass.

It would be interesting to ask anyone that was present in the committee for standardizing ANSI C, X3J11.  Is anyone present in WG14?  Does anyone know who was part of X3J11?


> SVID2 (before this bit was cut for POSIX, obviously) says:
> 
> > The functions malloc, realloc, and calloc return a
> > NULL pointer if nbytes is 0 or if there is not enough available memory. When the function realloc returns NULL, the block pointed to by ptr is left intact.
> 
> As we know, this is not true (everything up to SVr3 carries something that is identical to the V7 allocator).

Humm.

> Absent any _actual system that does this_, this may be a hoax based entirely on the SVID (which is itself just the SysV manuals)!

It might be.  Maybe ANSI C just blindly standardized what the SysV r3 manual (SVID2) said without even checking.  That wouldn't be surprising.

alejandro-colomar:
> And to the "real system" end, I'm yet to encounter a system that _is_ `malloc(0) = 0` (and I just quickly checked glibc, musl, {Net,Free,Open}BSD, and the illumos gate).
> 
> I wouldn't go so far as to call it a hoax outright (because some systems that implemented _C_ probably _had_ done this at _some_ point once) but I'm yet to see anything UNIX-shaped where this were true.

While no Unix systems seem to return NULL from `malloc(0)`, some Unix(-like) systems (including glibc) do that for `realloc(p, 0)`, which is equally insane.  (At least, that's what n2464 <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2464.pdf> says; I'm blindly trusting @rcseacord.)  The BSDs are the only ones that have a sane `realloc(p, 0)` these days.

I wonder why glibc decided to have this bad behavior inconsistent with `malloc(0)`.  @fweimer-rh , any insider insights?

alejandro-colomar:
> I asked PhD, he either knows (and his word as the editor is law) or has access to documents the general public doesn't.
> 
> I'll start writing a real post so this isn't scattered grey literature and you can reference something in your manual.

BTW, @nabijaczleweli , could you please check the arcane V7, SVr4, and 4.4BSD for realloc(p, 0)?  It would be interesting to see if those were sane (I assume they are).

nabijaczleweli:
Within the area I surveyed (AT&T UNIX, the BSD, SunOS and 'til-Y2K Solaris), every `realloc(…, 0)` behaved congruently with its `malloc`

rcseacord:
I wrote the paper for WG14 which made this UB:  
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2464.pdf 
The paper contains some of the history.  The original problem I would say is that the standard wording was somewhat ambiguous and implementations in not clearly stating that these functions return 0 on error and implementations interpreted the standard to mean they had to implement these functions in a weird, non-intuitive manner.

nabijaczleweli:
https://nabijaczleweli.xyz/content/blogn_t/017-malloc0.html

alejandro-colomar:
> I wrote the paper for WG14 which made this UB: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2464.pdf The paper contains some of the history. The original problem I would say is that the standard wording was somewhat ambiguous and implementations in not clearly stating that these functions return 0 on error and implementations interpreted the standard to mean they had to implement these functions in a weird, non-intuitive manner.

Which itself seems to derive from the bogus SVID, which declared standard behavior what had been a bug in a weird optimized alternative allocator from SysV r3.

alejandro-colomar:
> I wrote the paper for WG14 which made this UB: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2464.pdf The paper contains some of the history. The original problem I would say is that the standard wording was somewhat ambiguous and implementations in not clearly stating that these functions return 0 on error and implementations interpreted the standard to mean they had to implement these functions in a weird, non-intuitive manner.

@rcseacord 

BTW, you could have made it implementation-defined behavior instead of UB.  It would have allowed POSIX to define the behavior anyway.  And now a lot of people are complaining about making it UB (so might wipe your hard drive).

alejandro-colomar:
@nabijaczleweli 

Thanks a lot for the research!

I'm having trouble finding the origin of the sentence

```
If size is 0 and ptr is not a null pointer, the object pointed to is freed.
```

in POSIX.1-2001.  <https://pubs.opengroup.org/onlinepubs/009695399/functions/realloc.html>.

POSIX says:

```
CHANGE HISTORY

    First released in Issue 1. Derived from Issue 1 of the SVID.

Issue 6

    Extensions beyond the ISO C standard are marked.

    The following new requirements on POSIX implementations derive from alignment with the Single UNIX Specification:

        In the RETURN VALUE section, if there is not enough available memory, the setting of errno to [ENOMEM] is added.

        The [ENOMEM] error condition is added.

```

But I've read SVID Issue 2 <https://bitsavers.org/pdf/att/unix/SVID/System_V_Interface_Definition_Issue_2_Volume_1_1986.pdf>, and it doesn't say that.  So, at some point, this sentence was invented without a proper rationale (I can guess they wanted to fix the leak, and they did it in the worst possible way), and without documenting it.

BTW, the link for SVID Issue 3 that you provide is not working for me (at least at the moment).  Could you please check, and maybe provide a different one?  Or is it a temporary issue with the wayback machine?

It seems the same exact sentence was in ANSI C (C89):

```
If size is zero and ptr is not a null pointer, the object it points to is freed.
```

So POSIX probably got it from ANSI C.  But why is that undocumented in the POSIX page?!  Also where the hell did ANSI C hallucinate that sentence from?

Incidentally, that bogus POSIX/ANSI sentence seems to be what caused the bogus glibc behavior:

```
commit 11bf311edc76f5ddc469a8c396e313e82d76be15
Author: Ulrich Drepper <drepper@redhat.com>
Date:   Thu Jan 11 21:51:07 2007 +0000

    [BZ #2510, BZ #2830, BZ #3137, BZ #3313, BZ #3426, BZ #3465, BZ #3480, BZ #3483, BZ #
3493, BZ #3514, BZ #3515, BZ #3664, BZ #3673, BZ #3674]

[...]

    2006-12-08  Ulrich Drepper  <drepper@redhat.com>
            * malloc/memusage.c: Handle realloc with new size of zero and
            non-NULL pointer correctly.

[...]
```

where "correctly" means follow the brain-damaged POSIX/ANSI specification.

alejandro-colomar:
Oh, and it seems that while POSIX.1-2008 retained that bogus sentence, C99 had removed it.  So this is one of those rare cases where ISO C had a better definition than POSIX.1.

So, the change in C23 that allowed the POSIX bug to continue living was actually a bad thing.  Instead, POSIX should have been fixed like C99 did.

And the change authored by Ulrich Drepper was a bug, since it made glibc not conform to ISO C99.  It is the only non-conforming implementation, AFAICS, so I'll report a bug.

nabijaczleweli:
It's in [XPG4](https://pubs.opengroup.org/onlinepubs/009656499/toc.pdf#page=495) and XPG3-XPG4 Base Migration Guide, Version 2.pdf says
> realloc( )
Issue 4:
The description is updated to align with the ISO C standard. These changes are
largely superficial, although they should be noted as they may have a marginal
effect on some applications. Specifically:
• The order and contiguity of storage allocated by successive calls to this
function is unspecified.
• Each allocation yields a pointer to an object disjoint from any other object.
• The returned pointer points to the lowest byte address of the allocation.
• If size is zero, either a null pointer or a unique pointer that can be successfully
passed to free( ) is returned.

expectedly, this is not in any SVIDs, and is verbatim in C89, with rationale

> 4.10.3.4 The realloc function
A null first argument is permissible. If the first argument is not null, and the second
argument is 0, then the call frees the memory pointed to by the first argument, and
a null argument may be returned; this specification is consistent with the policy of
not allowing zero-size objects.

This is probably the best we're gonna get.

The internet archive is down, yes.

alejandro-colomar:
@rcseacord , @ThePhD

This seems conclusive proof that C89/POSIX.1-2001 were actually hallucinating, and [C99, C17] and POSIX.1-2008 were fixed.  Could anyone issue a NB Comment against C23's change that makes realloc(,0) UB?  C17 was good, and C23 is harmful/regressive.

The only bogus implementations that are non-conforming to C17 are MSVC and glibc (at least from those listed in @rcseacord 's paper), and at least in the case of glibc, this was a recent (2006~7) change (ironically, a so-called bugfix).  Let's just force them to conform.

alejandro-colomar:
@nabijaczleweli I've changed the patch to only avoid realloc(p,0) but not malloc(0), and have added a Co-developed-by you, since you did the investigation that led to the change.  Would you mind confirming the tag by issuing a Signed-off-by?

rcseacord:
C23 is done and there is no changing it.  Since it is UB implementations
are allowed to do the right thing.

Probably someone should submit a paper for C2Y. If adopted, this would
provide further evidence to implementers about the direction of the library.

Thanks,
rCs

alejandro-colomar:
> C23 is done and there is no changing it. Since it is UB implementations are allowed to do the right thing. Probably someone should submit a paper for C2Y. If adopted, this would provide further evidence to implementers about the direction of the library. Thanks, rCs

Makes sense.  I've sent a bug report to glibc, and will prepare a paper to reinstate the C17 wording in C2y.

I'll also prepare a paper to disallow malloc(0) returning NULL, and thus also realloc(,0), as it should be congruent, but this will be a second paper.

Thanks!

rcseacord:
responses below

On Thu, Oct 17, 2024 at 10:27 AM Alejandro Colomar ***@***.***>
wrote:

> C23 is done and there is no changing it. Since it is UB implementations
> are allowed to do the right thing. Probably someone should submit a paper
> for C2Y. If adopted, this would provide further evidence to implementers
> about the direction of the library. Thanks, rCs
>
> Makes sense. I've sent a bug report to glibc, and will prepare a paper to
> reinstate the C17 wording in C2y.
>
It would also be helpful if the maintainer from GCC could contribute to
this paper and or attend the meeting where we discuss it because I'm sure
the committee would like to hear what GCC intends to do here.


> I'll also prepare a paper to disallow malloc(0) returning NULL, and thus
> also realloc(,0), as it should be congruent, but this will be a second
> paper.
>

Sorry if I haven't been following this very carefully, but why would you
want to disallow malloc(0) from returning NULL?
malloc() should return a null pointer on failure only.  malloc(0)  can fail
because even a zero allocation requires some memory for control structures
such as boundary tags.  If this memory is not available, malloc(0) should
fail and return 0.

Thanks,
rCs

alejandro-colomar:
On Thu, Oct 17, 2024 at 08:06:47AM GMT, Robert C. Seacord wrote:
> responses below
> 
> On Thu, Oct 17, 2024 at 10:27 AM Alejandro Colomar ***@***.***>
> wrote:
> 
> > C23 is done and there is no changing it. Since it is UB implementations
> > are allowed to do the right thing. Probably someone should submit a paper
> > for C2Y. If adopted, this would provide further evidence to implementers
> > about the direction of the library. Thanks, rCs
> >
> > Makes sense. I've sent a bug report to glibc, and will prepare a paper to
> > reinstate the C17 wording in C2y.
> >
> It would also be helpful if the maintainer from GCC could contribute to
> this paper and or attend the meeting where we discuss it because I'm sure
> the committee would like to hear what GCC intends to do here.

Sure.  I'll CC gcc@ and you on the mail thread I've opened in
***@***.***  That way we'll all be able to discuss this issue at once.

> > I'll also prepare a paper to disallow malloc(0) returning NULL, and thus
> > also realloc(,0), as it should be congruent, but this will be a second
> > paper.
> >
> 
> Sorry if I haven't been following this very carefully, but why would you
> want to disallow malloc(0) from returning NULL?
> malloc() should return a null pointer on failure only.  malloc(0)  can fail
> because even a zero allocation requires some memory for control structures
> such as boundary tags.  If this memory is not available, malloc(0) should
> fail and return 0.

Yeah, my bad.  I meant NULL should not be allowed on success.  That is,
a successfull malloc(0) should return a unique pointer.

Cheers,
Alex

> 
> Thanks,
> rCs
> 
> > Message ID: ***@***.***>
> >
> 
> 
> -- 
> https://github.com/shadow-maint/shadow/pull/1095#issuecomment-2419810910
> 
> Message ID: ***@***.***>

-- 

ThePhD:
To be clear about an ask about changing this behavior for C2y: absolutely not.

The problem isn't that any one given implementation isn't self-consistent (or trying to be). The problem is porting from one implementation (e.g., from Bionic to BSD) has such varying implementation behavior that coding to one platform's expectations violates another platforms so badly you end up with a vulnerability. Some implementations like glibc don't even provide a full range of implementation-defined errno values to tell you what things have potentially gone wrong.

I vehemently do not support the change from UB back to IB. I will oppose any such paper unless it is accompanied by a guarantee from vendors large and small (BSD, Bionic, musl-libc, glibc, MSVC, and others) to harmonize in such a way that the impl-def behavior on one platform and its corner cases do not turn into vulnerabilities on other platforms: https://awakened1712.github.io/hacking/hacking-whatsapp-gif-rce/

alejandro-colomar:
> To be clear about an ask about changing this behavior for C2y: absolutely not.
> 
> The problem isn't that any one given implementation isn't self-consistent (or trying to be). The problem is porting from one implementation (e.g., from Bionic to BSD) has such varying implementation behavior that coding to one platform's expectations violates another platforms so badly you end up with a vulnerability.

You're conflating two bugs into one.  The two bugs are:

-  Some implementations return NULL from malloc(0), and others return a unique pointer.
-  Some implementations are consistent between malloc(0) and realloc(,0) and others are not.

The whole POSIX never had the first issue in libc.
And the **only** POSIX system that has the second issue is glibc.

Anything else is an absurd artificially created by the nonsensical decisions of X3J11 (ANSI C), and only affects non-POSIX systems.  Ironically, C23 has changed realloc(p,0) in order to supposedly help POSIX, but POSIX never had this problem in the first place.  We only need to convince glibc to fix that, and the problem is solved; portability will be great across POSIX systems.  Don't worry, I won't get out of POSIX, anyway.

This looks very much like the nullptr problem which nobody ever attempted to solve, or so it seems.

> Some implementations like glibc don't even provide a full range of implementation-defined errno values to tell you what things have potentially gone wrong.

We can ignore `errno`.  As long as it is set after an allocation call returns `NULL`, we have an error.  The precise error is unimportant, since most programs should just clean up and exit(1).  (Airplanes do want to differ, but let's ignore those for now.)

> 
> I vehemently do not support the change from UB back to IB. I will oppose any such paper unless it is accompanied by a guarantee from vendors large and small (BSD, Bionic, musl-libc, glibc, MSVC, and others)

I am working on that.  That's why I asked @nabijaczleweli to help on the research of how this crap ever came into being.  Please let me fix this; just like I fixed NULL recently.  I only need to fix glibc (and maybe bionic, I don't know what they do; but if I convince glibc, convincing @enh will be easy).  The rest of POSIX **already** does the right thing.  And I don't give a **** about what Win32 does.

> to harmonize in such a way that the impl-def behavior on one platform and its corner cases do not turn into vulnerabilities on other platforms: https://awakened1712.github.io/hacking/hacking-whatsapp-gif-rce/

alejandro-colomar:
> to harmonize in such a way that the impl-def behavior on one platform and its corner cases do not turn into vulnerabilities on other platforms: https://awakened1712.github.io/hacking/hacking-whatsapp-gif-rce/

That vulnerability is based on code that does something like:

```c
q = realloc(p, 100);
realloc(q, 0);
realloc(q, 0);
```

That is a bogus program, which followed the bogus specification of C89, assuming that that was the truth.

Please demand the members of X3J11 for being responsible for any losses produced by that vulnerability.

Don't trust ANSI/ISO C to take good decisions is the moral of the story.

paulfloyd:
I added a warning for realloc size zero to Valgrind in March 2023. No plans to remove it since C23 made it UB. There is also an option --realloc-zero-bytes-frees in case you need to match the implementation behaviour of your libc/malloc library.

ikerexxe:
The conversation has been long, so just to make sure I understood the problem correctly. Currently, realloc() can fail and not return NULL if `size` is equal to zero, making us fail miserably. The patch takes this into account and when size is 0 realloc's a size of 1 to avoid this problem.

alejandro-colomar:
> The conversation has been long, so just to make sure I understood the problem correctly.


> Currently, realloc() can fail and not return NULL if `size` is equal to zero,

The opposite, actually.  realloc(p, 0) will succeed but yet return NULL, which is used as a sentinel value that is understood as an error, so it will miserably confuse the program.

> making us fail miserably.

Yes.

> The patch takes this into account and when size is 0 realloc's a size of 1 to avoid this problem.

Yep, when we want to allocate 0 bytes, we can just as well work with 1 byte, since over-allocating is not a problem (other than wasting 1 byte, and maybe some static analysis degradation), which will succeed without returning NULL.

ikerexxe (approved):
> The opposite, actually. realloc(p, 0) will succeed but yet return NULL, which is used as a sentinel value that is understood as an error, so it will miserably confuse the program.

Lucky me that I confirmed.

LGTM! Still I'd like somebody else to give the thumbs up before merging.

alejandro-colomar:
> > The opposite, actually. realloc(p, 0) will succeed but yet return NULL, which is used as a sentinel value that is understood as an error, so it will miserably confuse the program.
> 
> Lucky me that I confirmed.
> 
> LGTM! Still I'd like somebody else to give the thumbs up before merging.

I'm sure @eggert or @fweimer-rh can confirm.  :-)

alejandro-colomar:
And just for more fun, you can see the alternating pattern:

```c
alx@debian:~/tmp/gcc/realloc$ cat brain_dmg.c 
#include <errno.h>
#include <stdio.h>
#include <stdlib.h>

int
main(void)
{
	void *p, *q;

	errno = 0;

	p = malloc(0);
	printf("malloc(0): %p\n", p);
	printf("      errno: %d\n", errno);
	free(p);

	q = NULL;
	q = realloc(q, 0);
	printf("realloc(q, 0): %p\n", q);
	printf("      errno: %d\n", errno);
	q = realloc(q, 0);
	printf("realloc(q, 0): %p\n", q);
	printf("      errno: %d\n", errno);
	q = realloc(q, 0);
	printf("realloc(q, 0): %p\n", q);
	printf("      errno: %d\n", errno);
	q = realloc(q, 0);
	printf("realloc(q, 0): %p\n", q);
	printf("      errno: %d\n", errno);
	free(q);
}
alx@debian:~/tmp/gcc/realloc$ cc -Wall -Wextra brain_dmg.c 
alx@debian:~/tmp/gcc/realloc$ ./a.out 
malloc(0): 0x5563eaa742a0
      errno: 0
realloc(q, 0): 0x5563eaa742a0
      errno: 0
realloc(q, 0): (nil)
      errno: 0
realloc(q, 0): 0x5563eaa742a0
      errno: 0
realloc(q, 0): (nil)
      errno: 0
```

eggert:
> > LGTM! Still I'd like somebody else to give the thumbs up before merging.
> 
> I'm sure @eggert or @fweimer-rh can review and confirm. :-)

Not sure why our opinions would be needed. The proposed patch is obviously a win, given the disaster of existing realloc (p, 0) implementations.

enh-google:
> I am working on that. That's why I asked @nabijaczleweli to help on the research of how this crap ever came into being. Please let me fix this; just like I fixed NULL recently. I only need to fix glibc (and maybe bionic, I don't know what they do; but if I convince glibc, convincing @enh will be easy).

fwiw, bionic already does the right thing (`malloc(0)` doesn't return null on success, `realloc(p, 0)` is equivalent to `free(p)` for non-null p, and `realloc(nullptr, 0)` is equivalent to `malloc(0)`).

this is true of both the current scudo allocator and the jemalloc allocator. the tests did not exist back when we still used dlmalloc, so i have no idea what its behavior was. but there are CTS tests for all of these cases, and all Android implementations are required to pass CTS, so even though you could theoretically ship an Android device with a modified allocator (or even a completely different allocator), it would have to behave like this in these cases.

alejandro-colomar:
@enh-google 

> > I am working on that. That's why I asked @nabijaczleweli to help on the research of how this crap ever came into being. Please let me fix this; just like I fixed NULL recently. I only need to fix glibc (and maybe bionic, I don't know what they do; but if I convince glibc, convincing @enh will be easy).
> 
> fwiw, bionic already does the right thing (`malloc(0)` doesn't return null on success, `realloc(p, 0)` is equivalent to `free(p)` for non-null p, and `realloc(nullptr, 0)` is equivalent to `malloc(0)`).

Well, that is not the right thing.  :-)

The right thing is that `realloc(p, 0)` does `free(p)` and `malloc(0)`, regardless of `p`.  That is the original behavior, and the one that is self-consistent.  Only the BSDs do that today.

That's not what the standards say is the right behavior, but the standards are very bad in this regard.  They were responsible for this mess in the first place, in their attempt to remove 0-size objects.

enh-google:
> @enh-google
> 
> > > I am working on that. That's why I asked @nabijaczleweli to help on the research of how this crap ever came into being. Please let me fix this; just like I fixed NULL recently. I only need to fix glibc (and maybe bionic, I don't know what they do; but if I convince glibc, convincing @enh will be easy).
> > 
> > 
> > fwiw, bionic already does the right thing (`malloc(0)` doesn't return null on success, `realloc(p, 0)` is equivalent to `free(p)` for non-null p, and `realloc(nullptr, 0)` is equivalent to `malloc(0)`).
> 
> Well, that is not the right thing to do. :-)
> 
> The right thing is that `realloc(p, 0)` does `free(p)` and `malloc(0)`, regardless of `p`. That is the original behavior, and the one that is self-consistent. Only the BSDs do that today.
> 
> That's not what the standards say is the right behavior, but the standards are very bad in this regard. They were responsible for this mess in the first place, in their attempt to remove 0-size objects.

ah, interesting. macOS (and thus presumably iOS) agrees with you. @cferris1000

alejandro-colomar:
@enh-google 

> ah, interesting. macOS (and thus presumably iOS) agrees with you. @cferris1000

Do you have interest in fixing Bionic to align with the BSDs?  You'll have to check some samples of code to see if anything would break, but @eggert and I suspect that code that is well-written for glibc (or Bionic, since it behaves just like glibc) should also behave well for the BSDs.  It is possible that some code might have more memory leaks than before, but I doubt we would introduce leaks in bug-free code.

See <https://inbox.sourceware.org/libc-alpha/rtfbobe6nvu2oxndc7qeltk2jvcuc4wodrsukfcmukgxyfaxh6@5wm6pmsz75sc/T/#mde1d787f2b80488879a1a2a33ab82ba48ac01125>.

alejandro-colomar:
> > > LGTM! Still I'd like somebody else to give the thumbs up before merging.
> > 
> > 
> > I'm sure @eggert or @fweimer-rh can review and confirm. :-)
> 
> Not sure why our opinions would be needed. The proposed patch is obviously a win, given the disaster of existing realloc (p, 0) implementations.

@ikerexxe That's the confirmation you were looking for, I think.  Feel free to merge.  Thank you both!  ;)

alejandro-colomar:
> My definition of "perfectly valid code" is "compiles, links, and runs and is shipped to one of the top 10 used communication applications in the world".

Okay.  I can agree to that definition.

Then it will be easy to transform that code into a minimal reproducible example program that I can reproduce to "run" on a system of your choice with behavior that does not run into a double free, and then can reproduce in a BSD/Unix V7-like system in which it would result in a double free.

Can you please provide a double_free_mre.c file?

alejandro-colomar:
@ThePhD 

> My definition of "perfectly valid code" is "compiles, links, and runs and is shipped to one of the top 10 used communication applications in the world".

Okay.  I can agree to that definition.

Then it will be easy to transform that code into a minimal reproducible example program that I can reproduce to "run" on a system of your choice with behavior that does not run into a double free, and then can reproduce in a BSD/Unix V7-like system in which it would result in a double free.

Can you please provide a double_free_mre.c file?

You will need to tell me in which system(s) it should work well.

paulfloyd:
> My definition of "perfectly valid code" is "compiles, links, and runs and is shipped to one of the top 10 used communication applications in the world".

"compiles, links and runs", famous last words.

paulfloyd:
> I vehemently do not support the change from UB back to IB. I will oppose any such paper unless it is accompanied by a guarantee from vendors large and small (BSD, Bionic, musl-libc, glibc, MSVC, and others) to harmonize in such a way that the impl-def behavior on one platform and its corner cases do not turn into vulnerabilities on other platforms: https://awakened1712.github.io/hacking/hacking-whatsapp-gif-rce/

Agree anything other than UB would be bad.

It's not just vulns. There is also the issue of leaks. I reckon that a lot of Linux glibc-using code just assumes that realloc works like free and doesn't check the return code. So if someone switches to using, say, snmalloc (since these are replaceable functions) from Linux glibc then they go from realloc-zero-bytes-frees to something else (maybe free then alloc minimum size, maybe do nothing). A long running exe that does a lot of zero-size reallocs risks memory exhaustion.

alejandro-colomar:
> > I vehemently do not support the change from UB back to IB. I will oppose any such paper unless it is accompanied by a guarantee from vendors large and small (BSD, Bionic, musl-libc, glibc, MSVC, and others) to harmonize in such a way that the impl-def behavior on one platform and its corner cases do not turn into vulnerabilities on other platforms: https://awakened1712.github.io/hacking/hacking-whatsapp-gif-rce/
> 
> Agree anything other than UB would be bad.
> 
> It's not just vulns. There is also the issue of leaks. I reckon that a lot of Linux glibc-using code just assumes that realloc works like free and doesn't check the return code.

That assumption is not quite right.  I've seen some code (very rare, but exists) that implements a free(3)-like wrapper by actually calling `realloc(p,0)`.  However, free(3) accepts NULL.  If you pass NULL to this realloc(3) call, you'll actually get a leak in glibc.  So, yeah, the leaks would likely increase on a Unix V7-like system, but it's not like they weren't there before.

> So if someone switches to using, say, snmalloc (since these are replaceable functions) from Linux glibc then they go from realloc-zero-bytes-frees to something else (maybe free then alloc minimum size, maybe do nothing). A long running exe that does a lot of zero-size reallocs risks memory exhaustion.

PR #1104, opened 2024-10-29
Added option -a to lastlog for listing active users only

pranlawate:


ikerexxe:
Hello, thank you for your contribution.

Unfortunately and although it is not very well documented `lastlog` is deprecated since [shadow v4.14](https://github.com/shadow-maint/shadow/releases/tag/4.14.0). If you want to include this functionality for the lastlog replacemente, lastlog2, I would recommend you to go to https://github.com/util-linux/util-linux/ since that is where it is hosted.

Depending on the distribution you are using, it is possible that lastlog2 is already available in new versions, if not, I would recommend you to open a ticket in the distribution system to have it replaced.

pranlawate:
@ikerexxe , Thank you for your quick response. I will check with the new Repo and also raise internal JIRA for Red Hat for this RFE.

hallyn (changes requested):
Actually I'd be ok taking this patch, but note that instead of the two strstrs you have, you could simply do the following patch:

```
diff --git a/src/lastlog.c b/src/lastlog.c
index 3bd5c62f..91f988f3 100644
--- a/src/lastlog.c
+++ b/src/lastlog.c
@@ -165,6 +165,8 @@ static void print_one (/*@null@*/const struct passwd *pw)
                cp = ptime;
        }
        if (ll.ll_time == (time_t) 0) {
+               if (!aflag)
+                       return;
                cp = _("**Never logged in**\0");
        }

```

(plus the flag parsing that you have).

pranlawate:
@hallyn Thank you so much for accepting this PR.

You are so right about that suggestion !
I found a similar way while working with my lastlog2 local file. ( I gave up raising any PR since there is no good way to test it without a proper sqlitedb on RHEL or Fedora) . I will not waste your time with that story here.
Let me get this short patch committed to my-branch and here.

I just want to ensure I am following correctly.
As much as I understand this code flow, if we 'return' for when time is '0' (zero converted to time_t) that will break the print_one function's that run. So the lines of users who never log in(zero time) are never printed.


Also Just to be sure, here is what I did to restart working with lastlog again after deleting previously cloned directory.
-- Resynced my fork
-- cloned my fork
-- changed to my-branch

Plan to do :
-- making changes to the code (reverting my 2 big bulky conditions with strstr and adding condition to return in print_one.
-- commit , push to my fork
-- merge commit in my fork
-- hoping to see the change to this PR.

pranlawate:
I have doubt here,
The condition will trigger with when aflag is false, i.e. default behavior, when -a is not used.  I don't mind it to be a default behavior to omit 'Never logged in'  entries . But wouldn't that annoy the users who need that information. 

As of now , I want to omit 'Never logged in' lines when aflag is used (`lastlog -a` should show lines with valid times ) 

So thinking of using  `if(aflag)` 
>  if (ll.ll_time == (time_t) 0) {
> +               if (aflag)
> +                       return;
>                 cp = _("**Never logged in**\0");
>         }

do let me know if I am missing something or if this has any other reason to change.

pranlawate:
I have tested this patch with if (aflg) return , it works as I intended it to.

hallyn:
Looks good, but please squash the two commits into one :)

(I can have github do it, but then it squashes the commit messages too and that's ugly)

pranlawate:
After considerable amount of googling   I figured what's squash and rebase.
I rebased my local ,  35 commits  , 1st -> my og change, 35th my current change.  34 in between were of fork resync.

I moved my 1st commit to 34th place, reword it and squashed the 35th into it .
 Now that should make it single commit.

Currently my repo is 17 commits behind the main. I hope that won't cause conflict issues unless lastlog was modified.
Let me know if I have to resync my repo or anythings else is needed.

hallyn:
> After considerable amount of googling I figured what's squash and rebase. I rebased my local , 35 commits , 1st -> my og change, 35th my current change. 34 in between were of fork resync.
> 
> I moved my 1st commit to 34th place, reword it and squashed the 35th into it . Now that should make it single commit.
> 
> Currently my repo is 17 commits behind the main. I hope that won't cause conflict issues unless lastlog was modified. Let me know if I have to resync my repo or anythings else is needed.

Oh dear, sorry about that.  I probably could've just done the rebase on my laptop and pushed it back to your branch.  But the branch looks good.  I did

```
wget https://github.com/shadow-maint/shadow/pull/1104.patch
```

to see the actual patch contents.  The only oddness is it has two signed-off-bys.  But let's just merge that as is.

Thanks!

hallyn:
Well, it's up to you - you can do a

git commit --amend

to pull up an editor and remove one of the signed-off-bys if you like.  Or I can merge as is.  Please let me know.

pranlawate:
@hallyn ,

It's better I figure the processes out myself as I would need them sooner or later.
Thanks for the command to ammend that commit info it saved time.

I had to 'git push origin +my-changes' , to get it up here.

I have corrected my email address as well along with removal of extra sign off line.
I prefer clean code so thanks for giving me chance to fix this up before merging.

hallyn (changes requested):
Oh!  This looks good, but you should also update the manual page at

man/lastlog.8.xml

If you like I'm ok merging this and accepting a separate PR later to do that update, just let me know.

pranlawate:
Hey @hallyn ,
Sure please merge this one.
I will raise a new PR today to get the man page updated as well.

hallyn (approved):

PR #1113, opened 2024-11-08
src/gpasswd.c: is_valid_user_list(): Separate declarations from code

alejandro-colomar:


hallyn:
Thanks, but I want to say no to this.  I don't think it's any clearer or easier to read, and it adds a line of code.  Open to being pursuaded, if you want to re-open...

alejandro-colomar:
> Thanks, but I want to say no to this. I don't think it's any clearer or easier to read, and it adds a line of code. Open to being pursuaded, if you want to re-open...

Yeah, I don't think a PR is worth it.  I would have passed it with the patch that fixed the bug, as a second patch, but a separate PR for it doesn't feel as good.  Might do it some day if there's a reason.  :)

hallyn:
Spent some time thinking about this.  I think the guidance I would want to give is:

In simple cases, I prefer to see variable initializations to happen at declaration.  It's actually easier to read and easier to find.  The case where I would want the initialization to happen separately is when it is based on another variable declared in the same function.

alejandro-colomar:
Regarding readability:

I think having the declaration together with the initialization is more readable in the case where the function is big and it's hard to go up and down tracking everything.  I think those functions are already bad, and should be split into smaller functions (or their bodies should not open-code so much stuff).  Once functions are small, separate declarations are quite readable, and I'd say more than joint ones.  IMO, it's like a book; first, the function presents the protagonists, and then it tells you the story.  In some blocks, you're introduced some extra characters that will be important in that block, and then you forget them after the block.  :)

For examples of this, I'd take functions I've written myself from scratch:

<https://github.com/shadow-maint/shadow/blob/master/lib/alloc/reallocf.h>
<https://github.com/shadow-maint/shadow/blob/master/lib/string/sprintf/stpeprintf.h>
<https://github.com/shadow-maint/shadow/blob/master/lib/adds.h>

---

Regarding initialization vs assignment:

Initializing at declaration is bad for static analysis.
Consider the case where you forget some path.  If you didn't assign any value yet to a variable, a static analyzer will notice, and let you know about the use of a uninitialized value.  If you initialize it to a value you won't use, you're creating dead code, and also, the static analyzer cannot know your intention, so it must think you did it on purpose and want to use that value in that path, so there won't be any diagnostics.

---

I disagree with those guidances.  But we'll see/discuss when we find some code where they actually matter.  Theoretical talks are usually not convincing.  :-)

PR #1117, opened 2024-11-11
lib/idmapping.c: Fix get_map_ranges regression

stoeckmann:
If get_map_ranges is called on a 32 bit system, where UINT_MAX equals ULONG_MAX, the count check allows illegal values if uid and loweruid are both UINT_MAX due to unsigned integer underflow.

Always use the fix size UINT_MAX as count limitation due to guarantee that sizeof(unsigned int) <= sizeof(unsigned long) in C language.

Fixes: 7c43eb2c4ea6 ("lib/idmapping.c: get_map_ranges(): Move range check to a2ul() call")

Proof of Concept (32 bit system):
```
newuidmap $$ 4294967295 4294967295 10
```

Output:
```
newuidmap: uid range [4294967295-9) -> [4294967295-9) not allowed
```

Expected:
```
newuidmap: subuid overflow detected.
usage: newuidmap [<pid>|fd:<pidfd>] <uid> <loweruid> <count> [ <uid> <loweruid> <count> ] ...
```

### DIFF
--- a/lib/idmapping.c
+++ b/lib/idmapping.c
@@ -52,23 +52,19 @@ get_map_ranges(int ranges, int argc, char **argv)
 	/* Gather up the ranges from the command line */
 	m = mappings;
 	for (int i = 0; i < ranges * 3; i+=3, m++) {
-		if (a2ul(&m->upper, argv[i + 0], NULL, 0, 0, UINT_MAX) == -1) {
+		if (a2ul(&m->upper, argv[i + 0], NULL, 0, 0, UINT_MAX - 1) == -1) {
 			if (errno == ERANGE)
 				fprintf(log_get_logfd(), _( "%s: subuid overflow detected.\n"), log_get_progname());
 			free(mappings);
 			return NULL;
 		}
-		if (a2ul(&m->lower, argv[i + 1], NULL, 0, 0, UINT_MAX) == -1) {
+		if (a2ul(&m->lower, argv[i + 1], NULL, 0, 0, UINT_MAX - 1) == -1) {
 			if (errno == ERANGE)
 				fprintf(log_get_logfd(), _( "%s: subuid overflow detected.\n"), log_get_progname());
 			free(mappings);
 			return NULL;
 		}
-		if (a2ul(&m->count, argv[i + 2], NULL, 0, 0,
-		         MIN(MIN(UINT_MAX, ULONG_MAX - 1) - m->lower,
-		             MIN(UINT_MAX, ULONG_MAX - 1) - m->upper))
-		    == -1)
-		{
+		if (a2ul(&m->count, argv[i + 2], NULL, 0, 1, UINT_MAX - MAX(m->lower, m->upper)) == -1) {
 			if (errno == ERANGE)
 				fprintf(log_get_logfd(), _( "%s: subuid overflow detected.\n"), log_get_progname());
 			free(mappings);


lib/idmapping.c
```diff
@@ -64,11 +64,7 @@ get_map_ranges(int ranges, int argc, char **argv)
 			free(mappings);
 			return NULL;
 		}
-		if (a2ul(&m->count, argv[i + 2], NULL, 0, 0,
-		         MIN(MIN(UINT_MAX, ULONG_MAX - 1) - m->lower,
-		             MIN(UINT_MAX, ULONG_MAX - 1) - m->upper))
-		    == -1)
-		{
+		if (a2ul(&m->count, argv[i + 2], NULL, 0, 0, UINT_MAX - MIN(m->lower, m->upper)) == -1) {
```
alejandro-colomar, line 67: Hmm, you're right; I wrote a bug there.

However, I'm having a hard time checking that your transformation is valid.

In a 32ILP platform, before I introduced the bug, assuming `m->lower = 42` and `m->upper == ULONG_MAX`, there were no valid values for `m->count`.

After my bug, any value in `[0, ULONG_MAX - 1 - 42]` is accepted.

With your patch, the accepted range becomes `[0, UINT_MAX - 42]`.

There's no way to tell a2sl(3) that no input is valid at all, so we need to think more.  How about restricting m->lower and m->upper to be limited by `MIN(UINT_MAX, ULONG_MAX - 1)`?  Or maybe we can make it simpler and limit to `UINT_MAX - 1`.  That would reduce by one the maximum values for 64-bit platforms, but does anybody care?

Thanks for the report!

alejandro-colomar: The following diff would revert back to something close to the original behavior, I think:

```diff
diff --git i/lib/idmapping.c w/lib/idmapping.c
index 248269c8..a4869346 100644
--- i/lib/idmapping.c
+++ w/lib/idmapping.c
@@ -52,13 +52,13 @@ get_map_ranges(int ranges, int argc, char **argv)
        /* Gather up the ranges from the command line */
        m = mappings;
        for (int i = 0; i < ranges * 3; i+=3, m++) {
-               if (a2ul(&m->upper, argv[i + 0], NULL, 0, 0, UINT_MAX) == -1) {
+               if (a2ul(&m->upper, argv[i + 0], NULL, 0, 0, MIN(UINT_MAX, ULONG_MAX - 1)) == -1) {
                        if (errno == ERANGE)
                                fprintf(log_get_logfd(), _( "%s: subuid overflow detected.\n"), log_get_progname());
                        free(mappings);
                        return NULL;
                }
-               if (a2ul(&m->lower, argv[i + 1], NULL, 0, 0, UINT_MAX) == -1) {
+               if (a2ul(&m->lower, argv[i + 1], NULL, 0, 0, MIN(UINT_MAX, ULONG_MAX - 1)) == -1) {
                        if (errno == ERANGE)
                                fprintf(log_get_logfd(), _( "%s: subuid overflow detected.\n"), log_get_progname());
                        free(mappings);
```

alejandro-colomar: And then, from your suggestion, I take the following simplification, which is a no-op, but makes the code more readable (I had a hard time trying to simplify it back then, and couldn't reach this; thanks!).

```diff
diff --git i/lib/idmapping.c w/lib/idmapping.c
index 248269c8..fb5f28af 100644
--- i/lib/idmapping.c
+++ w/lib/idmapping.c
@@ -65,8 +65,7 @@ get_map_ranges(int ranges, int argc, char **argv)
                        return NULL;
                }
                if (a2ul(&m->count, argv[i + 2], NULL, 0, 0,
-                        MIN(MIN(UINT_MAX, ULONG_MAX - 1) - m->lower,
-                            MIN(UINT_MAX, ULONG_MAX - 1) - m->upper))
+                        MIN(UINT_MAX, ULONG_MAX - 1) - MAX(m->lower, m->upper))
                    == -1)
                {
                        if (errno == ERANGE)
```

alejandro-colomar: Feel free to add those changes, and if so, add

```
Co-authored-by: Alejandro Colomar <alx@kernel.org>
```

stoeckmann: > assuming you meant to use `MAX()` instead of `MIN()`

Oops. Yeah. That's what I get for just testing my one edge case ... :)

> we should clarify what ranges we intend to allow in the future.

According to `user_namespaces(7)` (and testing all this with no checks at all in code to verify what I read):

- The three fields must be valid numbers, and the last field must be greater than 0.

So, range 0 is invalid in all cases. This can be adjusted by setting minimum accepted value to 1.

> Or maybe we can make it simpler and limit to `UINT_MAX - 1`.

That's what we have to do. If we try to use UINT_MAX with a range of 1, we get:

1. Start a new user namespace
```
# unshare -U /bin/bash
nobody$ echo $$
12345
```
2. Modify the user map
```
# newuidmap 12345 0 4294967295 1
newuidmap: write to uid_map failed: Invalid Argument
```

Since range cannot be 0, UINT_MAX is never valid for a uid here. Also, we can skip the ULONG_MAX value because the maximum UID is UINT_MAX, also on 64 bit systems.

stoeckmann: Adjusted to have 1 as a minimum. I don't like the error message a lot, but ... is it worth to add extra checks here?

```
# newuidmap 12345 0 100 0
newuidmap: subuid overflow detected.
usage: newuidmap [<pid>|fd:<pidfd>] <uid> <loweruid> <count> [ <uid> <loweruid> <count> ] ...
```

Haven't added you as `Signed-off-by` yet, since you didn't sign off the latest version of this PR yet. ;)

alejandro-colomar: Hmmmm, sounds good to me.

If you do all in one commit (which probably makes sense, since it has never been good), maybe it would be interesting to also add a Fixes: tag for the commits that added the checks before me, since they were already bad (not as bad as mine, though).

```
Fixes: 7c43eb2c4ea6 (2024-07-11, "lib/idmapping.c: get_map_ranges(): Move range check to a2ul() call")
Fixes: ff2baed5dbf8 (2016-08-14, "idmapping: add more checks for overflow")
Fixes: 94da3dc5c853 (2016-08-14, "also check upper for wrap")
Fixes: 7f5a14817d30 (2016-07-31, "get_map_ranges: check for overflow")
```

The commit looks good to me, so

```
```

Thanks!

alejandro-colomar: Although, the commit message probably needs to updated to detail all of what we discussed.  :-)

alejandro-colomar (approved):

alejandro-colomar:
(I've approved it, but it's really waiting for the commit message update.)

lib/idmapping.c
stoeckmann: Adjusted now.

stoeckmann:
> (I've approved it, but it's really waiting for the commit message update.)

Almost just in time ... ;)

alejandro-colomar:
> > (I've approved it, but it's really waiting for the commit message update.)
> 
> Almost just in time ... ;)

:)  Thanks!  I'll merge.

alejandro-colomar:
I'm happy that refactoring/simplifying the old code, even though it introduced a bug, it made it more obvious that the previous code was bad, and we've been able to improve the checks beyond what was before my regression.  :)

Issue #1120, opened 2024-11-13
Deprecation of 0 as a null pointer constant


alejandro-colomar:
GCC 15 will add -Wzero-as-null-pointer-constant working towards the deprecation of 0 as a null pointer constant in C.  Let's remove all of our uses of 0 as a null pointer constant already.

I'm also working on a paper for deprecating it in ISO C2y.

Link: <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117059>
Link: <https://software.codidact.com/posts/292718/292759#answer-292759>

PR #1121, opened 2024-11-13
lib/, src/: Use NULL instead of 0 as a null pointer constant

alejandro-colomar:
GCC 15 will add -Wzero-as-null-pointer-constant for deprecating it, and I'm working on a paper for deprecating it from ISO C too. Let's remove any uses in our code base.

I've done this change by building GCC from master, adding -Werror=zero-as-null-pointer-constant to ./autogen.sh, and fixing every error that showed up.

Closes: <https://github.com/shadow-maint/shadow/issues/1120>
Link: <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117059>
Link: <https://software.codidact.com/posts/292718/292759#answer-292759>

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -1240,7 +1240,7 @@ int commonio_rewind (struct commonio_db *db)
 
 	if (!db->isopen) {
 		errno = EINVAL;
-		return 0;
+		return NULL;
 	}
 	if (NULL == db->cursor) {
 		db->cursor = db->head;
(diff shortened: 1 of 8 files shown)


alejandro-colomar:
> Oh, interesting. I've seen other cases (in golang prs) where actually the lines of context were completely wrong. I wonder if git was to blame there too then.

Most likely.  :)

PR #1093, opened 2024-10-14
Replace most strtok(3) calls by strsep(3)

alejandro-colomar:
This is a subset of #1048 for easier review.

---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase

```
$ git range-diff gh/strtcpy..gh/strsep strtcpy..strsep 
1:  1ada0038 = 1:  b806041d lib/, src/: Use strsep(3) instead of strtok(3)
```
</details>

<details>
<summary>v1c</summary>

-  Rebase

```
$ git range-diff gh/strtcpy..gh/strsep master..strsep 
1:  b806041d = 1:  b87f79c6 lib/, src/: Use strsep(3) instead of strtok(3)
```
</details>

<details>
<summary>v1d</summary>

-  Expand commit message

```
$ git range-diff master gh/strsep strsep 
1:  b87f79c6 ! 1:  75903c36 lib/, src/: Use strsep(3) instead of strtok(3)
    @@ Metadata
      ## Commit message ##
         lib/, src/: Use strsep(3) instead of strtok(3)
     
    +    strsep(3) is stateless, and so is easier to reason about.
    +
    +    It also has a slight difference: strtok(3) jumps over empty fields,
    +    while strsep(3) respects them as empty fields.  In most of the cases
    +    where we were using strtok(3), it makes more sense to respect empty
    +    fields, and this commit probably silently fixes a few bugs.
    +
    +    In other cases (most notably filesystem paths), contiguous delimiters
    +    ("//") should be collapsed, so strtok(3) still makes more sense there.
    +    This commit doesn't replace such strtok(3) calls.
    +
         Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
      ## lib/addgrps.c ##
```
</details>

<details>
<summary>v1e</summary>

-  Rebase

```
$ git range-diff master..gh/strsep shadow/master..strsep 
1:  75903c36 = 1:  fa952752 lib/, src/: Use strsep(3) instead of strtok(3)
```
</details>

<details>
<summary>v2</summary>

-  Several rebases

```
$ git range-diff fa952752^..fa952752 shadow/master..gh/strsep 
1:  fa952752 = 1:  eb9e6257 lib/, src/: Use strsep(3) instead of strtok(3)
```
</details>

<details>
<summary>v2b</summary>

-  Rebase

```
$ git range-diff alx/master..gh/strsep shadow/master..strsep 
[...]

PR #1123, opened 2024-11-14
lib/addgrps.c: add_groups(): Simplify

alejandro-colomar:
Queued after #1093


lib/search/lfind/lfind.h
```diff
@@ -0,0 +1,34 @@
...
+    int (*cmp)(const void *k, const void *elt));
+
+
+inline void *
+lfind_(const void *k, const void *a, size_t n, size_t ksize,
+    int (*cmp)(const void *k, const void *elt))
+{
+	// lfind(3) wants a pointer to n for historic reasons.
+	return lfind(k, a, &n, ksize, cmp);
+}
```
alejandro-colomar, line 31: @uecker

Would you mind reviewing this macro?

alejandro-colomar:
> Does it make sense as a replacement for a simple for loop?

I don't like empty loops too much, and I think this improves the readability slightly.  But yeah, I also had doubts.

> I find the utility of such macros questionable.

I think it's more questionable if to call lfind(3) in the first place.  Once we agree to call lfind(3), this macro is probably better, even if just for lfind_(), which already hides the `*n` historical accident.

> Looking at the surrounding code, i would also generally make indices / dims "int" for improved overflow detection.

Hmm, yep, since getgroups(3) returns an int, it doesn't make much sense to use size_t, which forces us to use casts often.  Thanks!

> But to the macro: It does multiple expansion of k, n, a which should be avoided.

Hmmm.

-  `n` is only used once.
-  `a` is used twice, but one of them is within typeof(), within a _Generic selector (which is never evaluated, right?).  Nevertheless, getting an __auto_type local variable and using a static_assert(3) would help readability, so I'll do it.
-  `k` is used 3 times.  One of them is the controlling expression, which is never evaluated.  I should fix the other two, right?

> I also do not see why you need the inline function. You could create a compound literal locally &(size_t){ (n) } if you need a pointer here.

I hadn't thought of it.  :)

> (BTW: the manpage for lfind should use *.nmemb).

I don't understand this.  Would you mind suggesting a patch?  (Or at least some clarification.)

> Also, if you insist on a generic macro version, why not also make the return type correct?

Hmmm, yeah.


How about this?:

```c
$ grepc LFIND .
./lib/search/lfind/lfind.h:#define LFIND(k, a, n, cmp)                                           \
({                                                                    \
	__auto_type  k_ = k;                                          \
	__auto_type  a_ = a;                                          \
                                                                      \
	static_assert(is_same_typeof(k_, a_), "");                    \
                                                                      \
	/* lfind(3) wants a 'size_t *n' for historic reasons.  */     \
	(typeof(k_)) lfind_(k_, a_, &(size_t){(n)}, sizeof(*k_), cmp);\
})
```
```c
$ grepc is_same_typeof .
./lib/must_be.h:#define is_same_typeof(a, b)                                                  \
	is_same_type(typeof(a), typeof(b))                                    \
)
$ grepc is_same_type .
./lib/must_be.h:#define is_same_type(a, b)                                                    \
	__builtin_types_compatible_p(a, b)                                    \
)
```

uecker:
> > But to the macro: It does multiple expansion of k, n, a which should be avoided.
> 
> Hmmm.
> 
>     * `n` is only used once.
> 
>     * `a` is used twice, but one of them is within typeof(), within a _Generic selector (which is never evaluated, right?).  Nevertheless, getting an __auto_type local variable and using a static_assert(3) would help readability, so I'll do it.
> 
>     * `k` is used 3 times.  One of them is the controlling expression, which is never evaluated.  I should fix the other two, I think.
>

There can be evaluation in typeof / sizeof when used with variably modified types. But I was thinking mainly about macro expansion itself. Maybe this is not a problem here, but if you have several macros nested this can explode quickly.

> > (BTW: the manpage for lfind should use *.nmemb).
> 
> I don't understand this. Would you mind suggesting a patch? (Or at least some clarification.)

My local version has [.size * .nmemb] but this is not correct if .nmemb is a pointer, then it should be [.size * (* .nmemb)]
 
> > Also, if you insist on a generic macro version, why not also make the return type correct?
> 
> Yep, thanks!
> 
> How about this?:
> 
> ```c
> $ grepc LFIND .
> ./lib/search/lfind/lfind.h:#define LFIND(k, a, n, cmp)                                           \
> ({                                                                    \
> 	__auto_type  k_ = k;                                          \
> 	__auto_type  a_ = a;                                          \
> 	size_t       n_ = n;                                          \
> 	size_t       *np_;                                            \
>                                                                       \
> 	static_assert(is_same_typeof(k_, a_), "");                    \
>                                                                       \
> 	/* lfind(3) wants a 'size_t *n' for historic reasons.  */     \
> 	np_ = &(size_t){n_};                                          \
>                                                                       \
> 	(typeof(k_)) lfind(k_, a_, np_, sizeof(*k_), cmp);            \
> })

Seems ok to me. Do we need () around the arguments? (not sure)

alejandro-colomar:
> > > But to the macro: It does multiple expansion of k, n, a which should be avoided.
> > 
> > 
> > Hmmm.
> > ```
> > * `n` is only used once.
> > 
> > * `a` is used twice, but one of them is within typeof(), within a _Generic selector (which is never evaluated, right?).  Nevertheless, getting an __auto_type local variable and using a static_assert(3) would help readability, so I'll do it.
> > 
> > * `k` is used 3 times.  One of them is the controlling expression, which is never evaluated.  I should fix the other two, I think.
> > ```
> 
> There can be evaluation in typeof / sizeof when used with variably modified types.

Does `typeof(vmt)` evaluate within a _Generic selector?  That's one of those corner cases I don't know.

> But I was thinking mainly about macro expansion itself. Maybe this is not a problem here, but if you have several macros nested this can explode quickly.

Ahhh, I see.  That shouldn't be a problem here.  Nevertheless, for evaluation reasons it makes sense to avoid it, by using `auto`.

> 
> > > (BTW: the manpage for lfind should use *.nmemb).
> > 
> > 
> > I don't understand this. Would you mind suggesting a patch? (Or at least some clarification.)
> 
> My local version has [.size * .nmemb] but this is not correct if .nmemb is a pointer, then it should be [.size * (* .nmemb)]

Ahh, you're right.  I'll do `[*.nmemb * .size]` to avoid the parentheses.

> 
> > > Also, if you insist on a generic macro version, why not also make the return type correct?
> > 
> > 
> > Yep, thanks!
> > How about this?:
> > ```c
> > $ grepc LFIND .
> > ./lib/search/lfind/lfind.h:#define LFIND(k, a, n, cmp)                                           \
> > ({                                                                    \
> > 	__auto_type  k_ = k;                                          \
> > 	__auto_type  a_ = a;                                          \
> > 	size_t       n_ = n;                                          \
> > 	size_t       *np_;                                            \
> >                                                                       \
> > 	static_assert(is_same_typeof(k_, a_), "");                    \
> >                                                                       \
> > 	/* lfind(3) wants a 'size_t *n' for historic reasons.  */     \
> > 	np_ = &(size_t){n_};                                          \
> >                                                                       \
> > 	(typeof(k_)) lfind(k_, a_, np_, sizeof(*k_), cmp);            \
> > })
> > ```
> 
> Seems ok to me. Do we need () around the arguments? (not sure)

AFAIK, they're redundant.  See <https://software.codidact.com/posts/290205>.

BTW, I've reverted to using the inline function lfind_(), to make the macro slimmer.  I don't like macros growing too much.


```c
$ grepc LFIND .
./lib/search/lfind/lfind.h:#define LFIND(k, a, n, cmp)                                           \
({                                                                    \
	__auto_type  k_ = k;                                          \
	__auto_type  a_ = a;                                          \
                                                                      \
	static_assert(is_same_typeof(k_, a_), "");                    \
                                                                      \
	(typeof(k_)) lfind_(k_, a_, n, sizeof(*k_), cmp);             \
})
```
```c
$ grepc -tfd lfind_ .
./lib/search/lfind/lfind.h:inline void *
lfind_(const void *k, const void *a, size_t n, size_t ksize,
    int (*cmp)(const void *k, const void *elt))
{
	// lfind(3) wants a pointer to n for historic reasons.
	return lfind(k, a, &n, ksize, cmp);
}
```

I'll push in a moment v2 of this PR.

uecker:
> > 
> > There can be evaluation in typeof / sizeof when used with variably modified types.
> 
> Does `typeof(vmt)` evaluate within a _Generic selector? That's one of those corner cases I don't know.

No, nothing in the _Generic selector is ever evaluated.

alejandro-colomar:
> > > There can be evaluation in typeof / sizeof when used with variably modified types.
> > 
> > 
> > Does `typeof(vmt)` evaluate within a _Generic selector? That's one of those corner cases I don't know.
> 
> No, nothing in the _Generic selector is ever evaluated.

Thanks!  I've pushed v2.

alejandro-colomar:
@uecker 

The use of lfind(3) showed that we could also do some more work to use lsearch(3), which is more compelling as a simplification.

Also, I've come up with auto-detection of the right comparison function through a _Generic(3) selector.  :-)

```c
$ grepc LSEARCH .
./lib/search/l/lsearch.h:#define LSEARCH(k, a, n)                                              \
({                                                                    \
	__auto_type  k_ = k;                                          \
	__auto_type  a_ = a;                                          \
                                                                      \
	static_assert(is_same_typeof(k_, a_), "");                    \
                                                                      \
	(typeof(k_)) lsearch(k_, a_, n, sizeof(*k_), CMP(typeof(k_)));\
})
```
```c
$ grepc CMP .
./lib/search/cmp/cmp.h:#define CMP(TYPE)                                                     \
(                                                                     \
	_Generic((TYPE) 0,                                            \
		int *:            cmp_int,                            \
		long *:           cmp_long,                           \
		unsigned int *:   cmp_uint,                           \
		unsigned long *:  cmp_ulong                           \
	)                                                             \
)
```

That CMP() could grow as needed.  Once there's something like void-that-binds, this might get simpler (no need for CMP()), but for now, it's quite nice.

alejandro-colomar:
Fun story:

1)  I do a Debian code search to see what's being done in other projects that call getgroups(2), out of curiosity.
   <https://codesearch.debian.net/search?q=getgroups+*%5C%28&literal=0>.
   I was initially interested in util-linux, but out of curiosity I decided to search in Debian to see more projects.

2)  I find in the first page a few interesting projects, including util-linux, gnupg, and systemd.
3)  util-linux is correct, and does exactly what I do here.
4)  gnupg has an off-by-one bug, but it's overallocating, so I'm a bit lazy to report the bug.  Maybe in a few days I get the energy to send them a patch.
5)  systemd has some bizarre wrapper, called getgroups_alloc(), which is so brain-damaged that I start simplifying it just to be able to understand it.  After simplification, a bug (double-free) is made shallow.  Here's the bug fix for systemd: <https://github.com/systemd/systemd/pull/35275>.

6)   Fun fact, I wrote the systemd fix in a non-systemd system (Devuan); and the PR shows that.  It was totally casual (my Debian machine has a hardware failure, and I won't be able to use it in a few weeks); I hope they don't take offense.  :-)

alejandro-colomar:
I've opened this for review.

This is a bit large.  If you feel like merging it before the cut, ok.  If it's too much, also ok.  I've put it last in the priority list.

alejandro-colomar:
Queued after the release of 4.17.0.

alejandro-colomar:
WG14 is in tha house, it seems  :)

lib/shadow/grp/agetgroups.h
```diff
@@ -0,0 +1,52 @@
...
+	if (n == -1)
+		return NULL;
+
+	gids = MALLOC(n, gid_t);
+	if (gids == NULL)
+		return NULL;
+
+	n = getgroups(n, gids);
+	if (n == -1) {
+		free(gids);
```
hallyn, line 43: free() will overwrite errno, and you'll perror() if this failed, so that is worth preserving.

Other than that, this set looks good, thanks.

alejandro-colomar: The free(3) function is the only libc function (AFAIR) that doesn't overwrite errno.

```
$ man free.3 | grep preserves.errno
     The free() function returns no value, and preserves errno.
```

I'll take that as a reviewed-by.  :-)

alejandro-colomar:
<details>
<summary>v15d</summary>

-  Add 'Reviewed-by' @hallyn .  (taken from the informal <https://github.com/shadow-maint/shadow/pull/1123#discussion_r1928025570>)

```
$ git range-diff master gh/add_groups add_groups 
 1:  32aa0e56 !  1:  8065468f lib/, src/: Reduce scope of variables
    @@ Metadata
      ## Commit message ##
         lib/, src/: Reduce scope of variables
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
 2:  b2ff8388 !  2:  cd1f5841 lib/, src/: Un-spageticize code
    @@ Metadata
      ## Commit message ##
         lib/, src/: Un-spageticize code
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
 3:  8725e904 !  3:  bc711ed3 lib/, src/: Simplify allocation of buffer
    @@ Commit message
         according to gnulib, that's only NeXTstep 3.2, which we don't support.
     
         Link: <https://www.gnu.org/software/gnulib/manual/html_node/getgroups.html>
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
 4:  a215cf29 !  4:  db2ffb6c lib/search/cmp/, lib/, tests/: CMP(), cmp_*(): Add macro and functions
    @@ Commit message
     
         These macros are for use with bsearch(3),lfind(3),qsort(3).
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/Makefile.am ##
 5:  43a64b36 !  5:  0a39fdf1 lib/search/l/: LFIND(): Add macro
    @@ Commit message
     
         Co-authored-by: Martin Uecker <uecker@tugraz.at>
         Cc: Jorenar <dev@jorenar.com>
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/Makefile.am ##
 6:  e2b99c88 !  6:  d16535d7 lib/, src/: Use LFIND() instead of open-coded search loops
    @@ Metadata
      ## Commit message ##
         lib/, src/: Use LFIND() instead of open-coded search loops
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
 7:  48efbcec !  7:  fa9b36cd lib/addgrps.c: add_groups(): Remove useless cast
    @@ Metadata
      ## Commit message ##
         lib/addgrps.c: add_groups(): Remove useless cast
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
 8:  d3f427d9 !  8:  de465949 lib/addgrps.c: add_groups(): Allocate earlier
    @@ Commit message
     
         This will allow using lsearch(3).
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
 9:  1babdce7 !  9:  1ced5d45 lib/addgrps.c: add_groups(): Simplify redundant code with a goto
    @@ Metadata
      ## Commit message ##
         lib/addgrps.c: add_groups(): Simplify redundant code with a goto
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
10:  2eedee6d ! 10:  ee9d7538 lib/, src/: Unconditionally call setgroups(2)
    @@ Commit message
         simplifies the surrounding code, by removing the 'added' variable, and
         allowing to call lsearch(3) instead of lfind(3).
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
11:  de33dcce ! 11:  8d33eaf0 lib/, src/: Replace redundant checks by actual error handling
    @@ Commit message
     
         Instead of our check, let's report errors from setgroups(2).
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
12:  ea0e93ab ! 12:  52bbccba lib/search/l/: LSEARCH(): Add macro
    @@ Metadata
      ## Commit message ##
         lib/search/l/: LSEARCH(): Add macro
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/Makefile.am ##
13:  e34b5319 ! 13:  3fbd7e88 lib/, src/: Use LSEARCH() instead of its pattern
    @@ Metadata
      ## Commit message ##
         lib/, src/: Use LSEARCH() instead of its pattern
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
14:  f1ff7da6 ! 14:  95a6f2a0 lib/addgrps.c: add_groups(): Split variable to avoid sign-mismatch diagnostics
    @@ Metadata
      ## Commit message ##
         lib/addgrps.c: add_groups(): Split variable to avoid sign-mismatch diagnostics
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
15:  11714da7 ! 15:  88d40c2f lib/string/strchr/: strchrscnt(): Add function
    @@ Metadata
      ## Commit message ##
         lib/string/strchr/: strchrscnt(): Add function
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/Makefile.am ##
16:  9eee7b05 ! 16:  e90c6f24 lib/addgrps.c: add_groups(): Reallocate at once
    @@ Commit message
         element after the last delimiter).  This avoids reallocating +1 in a
         loop.
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
17:  a6f9c732 ! 17:  b101363f lib/, src/: Rename variables
    @@ Commit message
         Use the more common convention of just using plural.  Also, use 'gids'
         to distinguish it from other group representations.
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
18:  b7caa4a0 ! 18:  439c003a lib/addgrps.c: add_groups(): Remove arbitrary limit
    @@ Metadata
      ## Commit message ##
         lib/addgrps.c: add_groups(): Remove arbitrary limit
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
19:  cd18f202 ! 19:  a1f480eb lib/search/sort/: QSORT(): Add macro
    @@ Metadata
      ## Commit message ##
         lib/search/sort/: QSORT(): Add macro
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/Makefile.am ##
20:  587d8967 ! 20:  17216aa4 lib/adds.h: addslN(): Use QSORT() instead of its pattern
    @@ Metadata
      ## Commit message ##
         lib/adds.h: addslN(): Use QSORT() instead of its pattern
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/adds.h ##
21:  ea53e9d1 ! 21:  9b104bfb configure.ac, lib/, src/: Use gid_t instead of GETGROUPS_T
    @@ Commit message
           provide an accurate prototype for getgroups in unistd.h, and to
           implement it as specified in POSIX.
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## configure.ac ##
22:  72657f46 ! 22:  1fc3ecbd lib/shadow/grp/: agetgroups(): Add function
    @@ Commit message
         This encapsulates the logic for calling getgroups(3), which requires two
         calls plus a malloc(3) call to do it correctly.
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/Makefile.am ##
23:  8b7f0815 ! 23:  654ca3eb lib/, src/: Use agetgroups() instead of its pattern
    @@ Metadata
      ## Commit message ##
         lib/, src/: Use agetgroups() instead of its pattern
     
    +    Reviewed-by: Serge Hallyn <serge@hallyn.com>
     
      ## lib/addgrps.c ##
```
</details>

lib/shadow/grp/agetgroups.h
alejandro-colomar: Thanks a lot for reviewing!  This one wasn't small.

hallyn: Oh, my manpage doesn't have that text.

hallyn (approved):

lib/shadow/grp/agetgroups.h
alejandro-colomar:
```sh
$ cat man/man3/free.3 
.so man3/malloc.3
$ cat man/man3/free.3 | cut -f2 -d' ' | xargs realpath | xargs git blame -- | grep preserves
cfc381be29 man3/malloc.3     (Paul Eggert         2021-08-10 12:37:07 -0700 193) function returns no value, and preserves
$ git ref2 cfc381be29
cfc381be29fb (2021-08-10, 2021-08-31; "malloc.3: Modernize for glibc 2.34")
$ git log -1 cfc381be29
commit cfc381be29fb7b4f642b8b333882f22628112abe
Author: Paul Eggert <eggert@cs.ucla.edu>
Date:   Tue Aug 10 12:37:07 2021 -0700

    malloc.3: Modernize for glibc 2.34
    
    glibc has tightened up its rules for replacing the memory
    allocator.  I went through the malloc man page and looked for how
    it documented malloc() and related functions, and fixed
    discrepancies with glibc malloc() documentation and/or
    implementation.  I also reorganized the portability discussion so
    that portability issues can be seen more clearly.
    
    Signed-off-by: Michael Kerrisk <mtk.manpages@gmail.com>
```

POSIX.1-2024 also added this requirement to free(3):
<https://pubs.opengroup.org/onlinepubs/9799919799/functions/free.html>

Quoting it:

```
DESCRIPTION
    ...

    [CX] The free() function shall not modify errno if ptr is a null pointer or a pointer previously returned as if by malloc() and not yet deallocated.
```

```
Issue 8

    Austin Group Defect 385 is applied, adding a requirement that free() does not modify errno when passed a pointer to an object than can be freed.

    Austin Group Defect 1218 is applied, adding [reallocarray()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/reallocarray.html).

    Austin Group Defect 1302 is applied, aligning this function with the ISO/IEC 9899:2018 standard.
```

PR #1130, opened 2024-11-22
CI: avoid cancelling all jobs when one fails

ikerexxe:
If a job in a matrix fails we don't want to cancel all jobs, thus we need to set `fail-fast: false` as a strategy property.

### DIFF
--- a/.github/workflows/runner.yml
+++ b/.github/workflows/runner.yml
@@ -69,6 +69,7 @@ jobs:
   container-build:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         os: [alpine, debian, fedora]
 


ikerexxe:
In https://github.com/shadow-maint/shadow/pull/1131 Alpine build was supposed to fail due to https://gitlab.alpinelinux.org/alpine/aports/-/issues/16633, but apparently it was fixed and deployed before I was able to finish my work

alejandro-colomar:
That would increase the load, and probably slow down CI.

Since such bugs are rare, I suggest manually triggering the disabled jobs, which can be done easily.

On the other hand, if you think the slow down is not problematic, I don't have a strong opinion, so I'm okay if you want it.

alejandro-colomar (approved):
I'll approve it.  Feel free to merge it or drop it.  I'm okay with either of them.  ;)

alejandro-colomar:
Self correction:

Github has a bug, and while it supposedly allows enabling the job manually, it gets immediately canceled again if there's another job that has failed.

So, we need your patch.

ikerexxe:
> That would increase the load, and probably slow down CI.
> 
> Since such bugs are rare, I suggest manually triggering the canceled jobs, which can be done easily.
> 
> On the other hand, if you think the increased load is not problematic, I don't have a strong opinion, so I'm okay if you want it.

If I remember correctly Github offers 20 runners per organization. In our case we need 9 runners every time the actions are triggered and they take less than 10 minutes to run, so yes, I think it's fine.

PR #1131, opened 2024-11-22
System test framework

ikerexxe:
As discussed at length, this is the implementation of the new system
tests framework for shadow. This is a proof of concept that contains the
key elements to be able to run basic user (i.e. useradd, usermod) and
group (i.e. usermod) tests. If you like the framework the rest of the
functionality will be added in the future.

Some useful facts:
* It is implemented in python
* It is based on pytest and pytest-mh
* It works on all the distributions that are part of our CI
* It can be run in the cloud (VM or container) as well as on-premises
* After the execution of each test the environment is cleaned up
* Logs and other artifacts for failed tests are collected
* It has a rich API that can be extended and extended to cover new
  functionalities

In addition, the PR contains a set of tests that make use of the framework
to test the functionality. These tests are located under `tests/system/tests`.
They offer a good example of the high-level API and how to model the test
cases.

Finally, the system test are run in the CI. This way we make sure that the
code can run successfully and avoid any regressions.

Closes: https://github.com/shadow-maint/shadow/issues/835

### DIFF
--- a/etc/login.defs
+++ b/etc/login.defs
@@ -440,7 +440,7 @@ USERGROUPS_ENAB yes
 # This option is overridden with the -M or -m flags on the useradd(8)
 # command-line.
 #
-#CREATE_HOME     yes
+CREATE_HOME     yes
 
 #
 # Force use shadow, even if shadow passwd & shadow group files are
--- a/share/ansible/roles/ci_run/tasks/alpine.yml
+++ b/share/ansible/roles/ci_run/tasks/alpine.yml
@@ -12,6 +12,7 @@
       - build-base
       - byacc
       - cmocka-dev
+      - coreutils
       - expect
       - gettext-dev
       - git
@@ -64,3 +65,9 @@
     - "/usr/local/src/shadow/config.h"
     - "/usr/local/src/shadow/build.log"
     - "/usr/local/src/shadow/tests/unit/test-suite.log"
+
+- name: Copy configuration file for testing
+  ansible.builtin.copy:
+    src: /usr/local/src/shadow/etc/login.defs
+    dest: /etc/login.defs
+    remote_src: yes
--- a/tests/system/framework/hosts/shadow.py
+++ b/tests/system/framework/hosts/shadow.py
@@ -1,0 +1,175 @@
+"""shadow multihost host."""
+
+from __future__ import annotations
+
+from pathlib import PurePosixPath
+from typing import Any
+
+from pytest_mh.conn import ProcessLogLevel
+
+from .base import BaseHost, BaseLinuxHost
+
+__all__ = [
+    "ShadowHost",
+]
+
+
+class ShadowHost(BaseHost, BaseLinuxHost):
+    """
+    shadow host object.
+
+    This is the host where the tests are run.
+
+    .. note::
+
+        Full backup and restore of shadow state is supported.
+    """
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+        self._features: dict[str, bool] | None = None
+        """Features dictionary."""
+
+        self._backup_path: PurePosixPath | None = None
+        """Path to backup files."""
+
+        self._verify_files: [dict[str, str]] = [
+            {"origin": "/etc/passwd", "backup": "passwd"},
+            {"origin": "/etc/shadow", "backup": "shadow"},
+            {"origin": "/etc/group", "backup": "group"},
+            {"origin": "/etc/gshadow", "backup": "gshadow"},
+        ]
+        """Files to verify for mismatch."""
+
+    def pytest_setup(self) -> None:
+        super().pytest_setup()
+
+    def start(self) -> None:
+        """
+        Not supported.
+
+        :raises NotImplementedError: _description_
+        """
+        raise NotImplementedError("Starting shadow service is not implemented.")
+
+    def stop(self) -> None:
+        """
+        Not supported.
+
+        :raises NotImplementedError: _description_
+        """
+        raise NotImplementedError("Stopping shadow service is not implemented.")
+
+    def backup(self) -> Any:
+        """
+        Backup all shadow data.
+
+        :return: Backup data.
+        :rtype: Any
+        """
+        self.logger.info("Creating backup of shadow host")
+
+        result = self.conn.run(
+            """
+            set -ex
+
+            function backup {
+                if [ -d "$1" ] || [ -f "$1" ]; then
+                    cp --force --archive "$1" "$2"
+                fi
+            }
+
+            path=`mktemp -d`
+            backup /etc/login.defs "$path/login.defs"
+            backup /etc/default/useradd "$path/useradd"
+            backup /etc/passwd "$path/passwd"
+            backup /etc/shadow "$path/shadow"
+            backup /etc/group "$path/group"
+            backup /etc/gshadow "$path/gshadow"
+            backup /etc/subuid "$path/subuid"
+            backup /etc/subgid "$path/subgid"
+            backup /home "$path/home"
+            backup /var/log/secure "$path/secure"
+
+            echo $path
+            """,
+            log_level=ProcessLogLevel.Error,
+        )
+
+        self._backup_path = PurePosixPath(result.stdout_lines[-1].strip())
+
+        return PurePosixPath(result.stdout_lines[-1].strip())
+
+    def restore(self, backup_data: Any | None) -> None:
+        """
+        Restore all shadow data.
+
+        :return: Backup data.
+        :rtype: Any
+        """
+        if backup_data is None:
+            return
+
+        if not isinstance(backup_data, PurePosixPath):
+            raise TypeError(f"Expected PurePosixPath, got {type(backup_data)}")
+
+        backup_path = str(backup_data)
+
+        self.logger.info(f"Restoring shadow data from {backup_path}")
+        self.conn.run(
+            f"""
+            set -ex
+
+            function restore {{
+                rm --force --recursive "$2"
+                if [ -d "$1" ] || [ -f "$1" ]; then
+                    cp --force --archive "$1" "$2"
+                fi
+            }}
+
+            rm --force --recursive /var/log/secure
+            restore "{backup_path}/login.defs" /etc/login.defs
+            restore "{backup_path}/useradd" /etc/default/useradd
+            restore "{backup_path}/passwd" /etc/passwd
+            restore "{backup_path}/shadow" /etc/shadow
+            restore "{backup_path}/group" /etc/group
+            restore "{backup_path}/gshadow" /etc/gshadow
+            restore "{backup_path}/subuid" /etc/subuid
+            restore "{backup_path}/subgid" /etc/subgid
+            restore "{backup_path}/home" /home
+            restore "{backup_path}/secure" /var/log/secure
+            """,
+            log_level=ProcessLogLevel.Error,
+        )
+
+    def detect_file_mismatches(self) -> None:
+        """
+        Shadow binaries modify a number of files, but usually do not modify all of them. This is why we add an
+        additional check at the end of the test to verify that the files that should not have been modified are still
+        intact.
+        """
+        self.logger.info(f"Detecting mismatches in shadow files {self._backup_path}")
+
+        for x in self._verify_files:
+            result = self.conn.run(
+                f"""
+                set -ex
+
+                cmp {x['origin']} {self._backup_path}/{x['backup']}
+                """,
+                log_level=ProcessLogLevel.Error,
+                raise_on_error=False,
+            )
+            if result.rc != 0:
+                self.logger.error(f"File mismatch in '{x['origin']}' and '{self._backup_path}/{x['backup']}'")
+                result.throw()
+
+    def discard_file(self, origin: str) -> None:
+        """
+        Discard modified files from the files that should be verified.
+        """
+        for x in self._verify_files:
+            if x["origin"] == origin:
+                self._verify_files.remove(x)
+                break
--- a/tests/system/framework/roles/shadow.py
+++ b/tests/system/framework/roles/shadow.py
@@ -1,0 +1,129 @@
+"""shadow multihost role."""
+
+from __future__ import annotations
+
+import shlex
+from typing import Dict
+
+from pytest_mh.conn import ProcessLogLevel, ProcessResult
+
+from ..hosts.shadow import ShadowHost
+from .base import BaseLinuxRole
+
+__all__ = [
+    "Shadow",
+]
+
+
+class Shadow(BaseLinuxRole[ShadowHost]):
+    """
+    shadow role.
+
+    Provides unified Python API for managing and testing shadow.
+    """
+
+    def __init__(self, *args, **kwargs) -> None:
+        """
+        Set up the environment.
+        """
+        super().__init__(*args, **kwargs)
+
+    def teardown(self) -> None:
+        """
+        Detect file mismatches before cleaning up the environment.
+        """
+        self.host.detect_file_mismatches()
+        """
+        Clean up the environment.
+        """
+        super().teardown()
+
+    def _parse_args(self, *args) -> Dict[str, str]:
+        args_list = shlex.split(*args[0])
+        name = args_list[-1]
+
+        return {"name": name}
+
+    def useradd(self, *args) -> ProcessResult:
+        """
+        Create user.
+        """
+        args_dict = self._parse_args(args)
+        self.logger.info(f'Creating user "{args_dict["name"]}" on {self.host.hostname}')
+        cmd = self.host.conn.run("useradd " + args[0], log_level=ProcessLogLevel.Error)
+
+        self.host.discard_file("/etc/passwd")
+        self.host.discard_file("/etc/shadow")
+        self.host.discard_file("/etc/group")
+        self.host.discard_file("/etc/gshadow")
+
+        return cmd
+
+    def usermod(self, *args) -> ProcessResult:
+        """
+        Modify user.
+        """
+        args_dict = self._parse_args(args)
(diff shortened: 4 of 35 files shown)


ikerexxe:
@hallyn  I'd appreciate your feedback on this PR, particularly the tests in the `tests/system/tests` folder. This PR is quite extensive, so I'd like to focus our initial review on these tests. They showcase the new testing framework and how it interacts with the shadow utilities and other tools.  Your input on this core part will be valuable as we move forward with the review.

hallyn:
Thanks for the ping, I'm putting this on my next_actions list.

ikerexxe:
For the moment, let's just ignore the CI errors. I'll fix them when I figure out how to do it

etc/login.defs
```diff
@@ -440,7 +440,7 @@ USERGROUPS_ENAB yes
 # This option is overridden with the -M or -m flags on the useradd(8)
 # command-line.
 #
-#CREATE_HOME     yes
+CREATE_HOME     yes
```
alejandro-colomar, line 443: I think this change is unrelated to the test framework and should be addressed in a separate PR.  Or are you using it for testing purposes?

alejandro-colomar:
How should I run the tests in my computer?

Also, are those tests (potentially) destructive?  That is, do I need to run them inside a container?

Or does the test suite spin up containers as necessary?

tests/system/framework/hosts/shadow.py
```diff
@@ -0,0 +1,128 @@
...
+            backup /etc/subuid "$path/subuid"
+            backup /etc/subgid "$path/subgid"
+            backup /var/log/secure "$path/secure"
+
+            echo $path
+            """,
+            log_level=ProcessLogLevel.Error,
+        )
+
+        return PurePosixPath(result.stdout_lines[-1].strip())
```
hallyn, line 87: Hm, but the script does set -e, so result.stdout_lines could in theory be empty?

I guess it'll just give us an indexerror so no big deal?  However, catching a return value != 0 early might be less confusing

tests/system/framework/misc/ssh.py
```diff
@@ -0,0 +1,175 @@
...
+        :type input: str | None, optional
+        :param log_level: Log level, defaults to ProcessLogLevel.Full
+        :type log_level: ProcessLogLevel, optional
+        """
+        if env is None:
+            env = {}
+
+        argv = to_list_of_strings(argv)
+        command = shlex.join(argv)
+        pidfile = "/tmp/.mh.sshkillableprocess.pid"
```
hallyn, line 47: This feels racy.  Not sure how we can do better though.  (Racy as in - if two tests happen to try an ssh killable process at the same time)

hallyn:
I'm only partly through this (at patch 5), but it looks good so far, thanks.

etc/login.defs
ikerexxe: I'm using it for testing purposes.

One of the objectives of the framework is to be able to test the same functionality in all distributions. In order to do this, a number of things have to be in place, including configuration sharing. For this particular case, both Debian and Fedora at distribution level have this configuration enabled, but Alpine does not. I could have created a new `login.defs file` just for testing but I decided to reuse the existing one. This will give more value to this sample `login.defs file` file, as it will see how it works in conjunction with the code.

tests/system/framework/hosts/shadow.py
ikerexxe: How about the following?
```
if not result.stdout_lines:
    return PurePosixPath(result.stdout_lines[-1].strip())
else:
    return -1
```

tests/system/framework/misc/ssh.py
ikerexxe: Tests run sequentially and this is a hard requirement for the framework, so two tests can run at the same time

ikerexxe:
> How should I run the tests in my computer?
> 
> Also, are those tests (potentially) destructive? That is, do I need to run them inside a container?
> 
> Or does the test suite spin up containers as necessary?

The instructions from `doc/contributions/build_install.md` still apply. Running them will create the container, build and install shadow and run the tests inside the container.

You can of course run the test on you own system, but I'd recommend against it since you may break your system with a shadow change. I plan to extend the documentation to explain how the framework works, the API, how to integrate it, etc. But that will come in another PR, once we all agree that the PoC is in a good shape and it can be merged.

> I'm only partly through this (at patch 5), but it looks good so far, thanks.

So you've already gone through the first test cases. What do you think about them? Is the API clear to manage shadow and other utilities clear enough?

tests/system/framework/misc/ssh.py
alejandro-colomar, line 47: The standard location for PID files is $runstatedir, that is, <$prefix/var/run/program-name/>.

(The .../program-name/ directory is only necessary when the program stores more than one file.  If there's only the PID file, you can put it directly under the main `$prefix/var/run/` dir.)

See also:

-  <https://www.gnu.org/prep/standards/html_node/Directory-Variables.html#Directory-Variables>
-  <https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s13.html>
-  <https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html>

alejandro-colomar: Although for a test suite there might be reasons to not follow the standard...

share/ansible/roles/ci_run/tasks/alpine.yml
```diff
@@ -71,3 +71,22 @@
...
+- name: Ensure test dependencies are installed
+  community.general.apk:
+    name:
+      - bash
+    state: present
+
+#The following is a workaround
+- name: Change shell for tests
+  ansible.builtin.command:
+    chsh -s /bin/bash
```
alejandro-colomar, line 85: Do we need to report a bug to pytest-mh?  I'd like to know that the workaround is temporary if there's a bug in a software we're using.

alejandro-colomar:
> > How should I run the tests in my computer?
> > Also, are those tests (potentially) destructive? That is, do I need to run them inside a container?
> > Or does the test suite spin up containers as necessary?
> 
> The instructions from `doc/contributions/build_install.md` still apply. Running them will create the container, build and install shadow and run the tests inside the container.
> 
> You can of course run the test on you own system, but I'd recommend against it since you may break your system with a shadow change. I plan to extend the documentation to explain how the framework works, the API, how to integrate it, etc. But that will come in another PR, once we all agree that the PoC is in a good shape and it can be merged.

Just to check if I've understood that:

So, if I run `share/container-build.sh`, this new test suite will be run?

tests/system/framework/misc/ssh.py
ikerexxe: @pbrezina your input here is much welcome

share/ansible/roles/ci_run/tasks/alpine.yml
ikerexxe: I used the word workaround for lack of a better one. I was talking about this problem with @pbrezina (the pytest-mh maintainer) and this is the first time he encounters the problem of having to use another shell than bash, so the use of this shell is hardcoded. This by itself would not be a compelling enough reason to implement this change on our part, of course if other reasons are found or someone wants to implement this change I don't think there would be any problem.

ikerexxe:
> So, if I run `share/container-build.sh`, this new test suite will be run?

Yes, the test suite will run within `share/container-build.sh`

ikerexxe:
Ok, now the CI is in the state where I wanted it to be last week. Alpine build is failing due to https://gitlab.alpinelinux.org/alpine/aports/-/issues/16633. This is already fixed but it takes some time for those changes to arrive to users.

tests/system/framework/misc/ssh.py
alejandro-colomar: This NGINX Unit discussion also contains useful info: <https://github.com/nginx/unit/pull/1235>

share/ansible/roles/ci_run/tasks/alpine.yml
alejandro-colomar: But code that needs bash mush use a bash shebang (`#!/usr/bin/env bash`) or something equivalent.  Why is the workaround needed?

alejandro-colomar: As for installing dependencies, how is pytest-mh installed?  Doesn't it pull dependencies as usual via the system package manager?  Why do we need to manually specify deps?

ikerexxe: > Ahh, yeah, that sounds more reasonable. So, the program fails if bash is not installed, and when bash is installed it works fine without needing to do anything special, right?

Yes, you are right, but this is because I force bash as a shell later on.

> As for installing dependencies, how is pytest-mh installed? Doesn't it pull dependencies as usual via the system package manager? Why do we need to manually specify deps?

Python dependencies are installed in a virtual environment using pip. They are listed in the [requirements](https://github.com/shadow-maint/shadow/pull/1131/files#diff-f3c0fe44e24c0b0d0ed6ce1b65a761c0ea590440d10644c7b1704516365fed60) file.

share/ansible/roles/ci_run/tasks/alpine.yml
```diff
@@ -71,3 +71,10 @@
     src: /usr/local/src/shadow/etc/login.defs
     dest: /etc/login.defs
     remote_src: yes
+
+#The following is a workaround
+- name: Create a symbolic link
+  ansible.builtin.file:
+    src: /bin/bash
+    dest: /usr/bin/bash
```
alejandro-colomar, line 79: Ahh, so the problem is that they use bash in a path that doesn't exist in Alpine?  Good.  That's a smaller problem to deal with.  Adding a symlink sounds like a reasonable workaround.

LGTM.

alejandro-colomar: pytest-mh (CC: @pbrezina ) may want to use `#!/usr/bin/env bash` as their shebang, since that's significantly more portable.  It's technically not guaranteed by POSIX, but all systems I know have env(1) there.

alejandro-colomar: @ikerexxe , could you point at the line of code in pytest-mh that requires this workaround?  That would be very useful to note in the commit message.

alejandro-colomar: I suspect it's this one:
<https://github.com/next-actions/pytest-mh/blob/101d4d45a3f482d7de22d67edbfeb451cbedd2a4/pytest_mh/conn/__init__.py#L1040>

```py
        super().__init__("bash", "/usr/bin/bash -c")
```

and a few other similar ones.

@pbrezina , how about calling `/usr/bin/env bash -c` instead?

alejandro-colomar: As an easy example, my laptop is one of those systems where your command will break:

```sh
alx@devuan:~$ cat /etc/os-release 
PRETTY_NAME="Devuan GNU/Linux 5 (daedalus)"
NAME="Devuan GNU/Linux"
VERSION_ID="5"
VERSION="5 (daedalus)"
VERSION_CODENAME="daedalus"
ID=devuan
ID_LIKE=debian
HOME_URL="https://www.devuan.org/"
SUPPORT_URL="https://devuan.org/os/community"
BUG_REPORT_URL="https://bugs.devuan.org/"
alx@devuan:~$ which bash
/bin/bash
alx@devuan:~$ /usr/bin/bash -c "echo foo"
bash: /usr/bin/bash: No such file or directory
alx@devuan:~$ /usr/bin/env bash -c "echo foo"
foo
```

pbrezina: Yes, please, open a ticket. I am also thinking about switching just to /bin since calls to other programs might break as well. All distros that done usrmerge have /bin as a symlink anyway.

alejandro-colomar: I think `/bin/bash` might be problematic in MacOS, where the version in /bin/bash is usually too old and problematic.  People usually install a more modern bash in /usr/local/bin/bash, or elsewhere.

And in the BSDs this might also be problematic, where they also install bash as a port in /usr/local/bin/bash, AFAIK.

I'll open a ticket.  :-)

pbrezina: https://github.com/next-actions/pytest-mh/pull/96

ikerexxe: I removed the workaround as it was tackled in pytest-mh so I'm closing this conversation

hallyn:
> > 
> 
> So you've already gone through the first test cases. What do you think about them? Is the API clear to manage shadow and other utilities clear enough?

I think so, yes.

tests/system/tests/test_groupdel.py
```diff
@@ -0,0 +1,30 @@
...
+    :steps:
+        1. Group doesn't exist
+    :expectedresults:
+        1. Group is not found
+    :customerscenario: False
+    """
+    shadow.groupadd("tgroup")
+    shadow.groupdel("tgroup")
+
+    result = shadow.tools.getent.group("tgroup")
```
hallyn, line 29: Now, arguably the things being tested for here are not the same.  The original test checks for no changes at all to passwd, shadow etc.  Theoretically if groupadd or groupdel wrongly made a change to another entry in the db, the old test would find it and the new would not.

At first that felt like an improvement (more precision in the verification step), but I worry that a dumb string parsing snafu in some complicated patch could in fact cause something like this.

What do you think?

tests/system/framework/hosts/shadow.py
pbrezina: It shouldn't be empty, if the script fails it raises `ProcessError` from `run`  since `raise_on_error=True` by default.

tests/system/framework/roles/shadow.py
```diff
@@ -0,0 +1,161 @@
...
+        Clean up the environment.
+        """
+        cmd = ""
+
+        if self._users:
+            cmd += "\n".join([f"userdel '{x}' --force --remove" for x in self._users])
+            cmd += "\n"
+        if self._groups:
+            cmd += "\n".join([f"groupdel '{x}' --force" for x in self._groups])
+            cmd += "\n"
```
pbrezina, line 42: It may not be the best idea to call tool that you are actually testing in teardown. I think you should backup/restore the files that shadow-utils is touching (/etc/passwd,groups and /etc/home?), perhaps backup/restore in the host class is sufficient?

pbrezina (commented):
Hi, I went over the code briefly.

I think this needs clean up. It looks like you copy&paste lost of stuff from sssd-test-framework that you actually don't use. I suggest you to go over it and remove what is unused (like SSHKillableProcess, BaseTopologyController, ShadowTopologyController) to simplify the code.

Some things, maybe, you do not need to copy&paste but import it instead (like stuff in tools.py). But that is of course up to your preference.

tests/system/tests/test_groupdel.py
ikerexxe: I see your point here, and that would apply to all tests, right? I mean, `useradd` could also be doing something wrong to another entry of the db when adding a new user.

In general terms, do we want to check it? This would add some complexity to all the tests, but I think it is manageable. Database management APIs are stable (and have been stable for a long time), so the key question would be, how likely is this to happen?

tests/system/framework/roles/shadow.py
ikerexxe: Do you mean `ShadowHost`? I'm already doing the backup, except for /home, would that be enough?

tests/system/framework/misc/ssh.py
ikerexxe: As Pavel mentioned this class is not needed, so I removed it

tests/system/framework/roles/shadow.py
ikerexxe: Done. `ShadowHost` takes care of backup and restore for each test case

tests/system/tests/test_groupdel.py
hallyn: What I like about your version is it makes clear what we are testing for.

Most of the tests should result in most of the shadow related files not being modified, right?  Perhaps we should actually tell the testsuite which files should change, and the testsuite always checks for changes in any others?

hallyn: I'm not *quite* sure how different that is, so I might be misunderstanding, but it sounds good to me :)

ikerexxe:
@alejandro-colomar @hallyn it'd be nice to have a final review from your side

alejandro-colomar:
> @alejandro-colomar @hallyn it'd be nice to have a final review from your side

Mine will be more an opinion than an actual review:

I don't understand Python.  I can maybe read 5 or 10 lines for a small script, but this is huge, and inscrutable to me.

I wish this was done in bash(1), which is a language I can read much better (and I guess many Linux/Unix programmers too, but I don't know).

However, the old framework was written as shell scripts, and I only used it maybe once or twice, so it wasn't much better.

So if you and Serge understand this framework and make it easy for others to use, it's a net improvement.  I'll use it, but won't be able to contribute to it.

Thank you for your work!  I appreciate it, even if I don't understand it.  :-)

Anyway, if I ever have the time to write a testing framework in bash(1), I'll let you know.  For the time being, this is better than no tests (nobody was adding tests with the old test suite).  I have other food in my plate for the time being.  :-)

ikerexxe:
At last Alpine's change has landed in its stable version. So it would only remain your review @hallyn  to be able to merge these changes.

hallyn (approved):

hallyn:
Thanks.

etc/login.defs
alejandro-colomar: This was accidentally merged in the end.

Is that intended?

hallyn: Hm, indeed.  I'll revert this patch in a bit if you haven't already spun up a pr.

jubalh:
Very cool!

PR #1132, opened 2024-11-25
share/container-build.sh: Fix path

alejandro-colomar:
The instructions are written so that this script should be run from the root of the repository.  Specify the path from the root of the repo. Before this fix, the command needed to be run from within `<share/>`.


ikerexxe (approved):
LGTM!

PR #1133, opened 2024-11-26
Allow setting an empty homedir

votdev:
With the PR 352 it is not possible anymore to run `usermod --home "" <USERNAME>`. This PR will fix that regression.

Related to: https://github.com/shadow-maint/shadow/pull/352

### DIFF
--- a/src/usermod.c
+++ b/src/usermod.c
@@ -64,6 +64,7 @@
 #include "shadowlog.h"
 #include "string/memset/memzero.h"
 #include "string/sprintf/xasprintf.h"
+#include "string/strcmp/streq.h"
 #include "string/strdup/xstrdup.h"
 #include "time/day_to_str.h"
 
@@ -1062,7 +1063,7 @@ process_flags(int argc, char **argv)
 				}
 				dflg = true;
 				user_newhome = optarg;
-				if (user_newhome[0] != '/') {
+				if ((user_newhome[0] != '/') && !streq(user_newhome, "")) {
 					fprintf (stderr,
 					         _("%s: homedir must be an absolute path\n"),
 					         Prog);


src/usermod.c
```diff
@@ -1062,7 +1062,8 @@ process_flags(int argc, char **argv)
 				}
 				dflg = true;
 				user_newhome = optarg;
-				if (user_newhome[0] != '/') {
+				/* If no empty string, check whether the path is absolute */
+				if ((user_newhome[0] != '\0') && (user_newhome[0] != '/')) {
```
votdev, line 1066:
```suggestion
				if ((strlen(user_newhome) > 0) && (user_newhome[0] != '/')) {
```
I think this is better to read, but it is more costly than checking `user_newhome[0] != '\0'`.

alejandro-colomar: Please use `!streq(user_home, "")`.  We have it available with `#include "string/strcmp/streq.h"`.

alejandro-colomar: Also, since `/` is a more common case, it should probably go first.  Both for optimization reasons (although I care 0 about optimizing this), and for readability reasons (the brain --at least in my case-- cares more about the first condition, and only cares about the second secondarily; just like the compiler).

src/usermod.c
```diff
@@ -1062,7 +1063,8 @@ process_flags(int argc, char **argv)
 				}
 				dflg = true;
 				user_newhome = optarg;
-				if (user_newhome[0] != '/') {
+				/* If no empty string, check whether the path is absolute */
```
alejandro-colomar, line 1066: I think I would remove that comment.  It doesn't say anything that the code doesn't say.

If anything needs to be clarified further, I'd do it in the commit message.

votdev: This is of course true, but written text can be understood more quickly than code if it is a little more complicated. In the end, of course, it's a matter of opinion and taste. Personally, I hate source code without comments.

alejandro-colomar: > This is of course true, but written text can be understood more quickly than code if it is a little more complicated.

Then the goal is to never write code that is more complicated.  That's a very difficult thing, actually, but doable, IMO.  :-)

alejandro-colomar:
I've added @hallyn as reviewer, as he requested himself.

alejandro-colomar:
> Can you add 'Closes #352' to the commit message? That should auto-close the issue.

That discussion was within a closed PR.  There's nothing to close, actually.  :)

> 
> But +1.
> 
> Thanks!

hallyn:
Oh well then let'r rip.

PR #1146, opened 2024-12-07
lib/sgetgrent.c: sgetgrent(): Simplify by calling strdup(3)

alejandro-colomar:
While at it, rename the function parameter.


alejandro-colomar:
> Looks good,

Thanks!

> but is there any reason why you can't just drop the 'static' modifier and stop doing the first 'free'?

I remember having looked into it, but I don't remember my findings.  :)

I'd like to attempt that in a separate PR in the future.

Issue #1149, opened 2024-12-08
Should we remove --badname?


alejandro-colomar:
<https://lwn.net/Articles/1001215/#Comments>

That should be a reminder of the dangers.

My opinion is now stronger that names shall be limited to ASCII.  I'm not happy allowing distros to lessen the requirements.

If one needs to manually fix an old system with such users, they can still manually edit the shadow files via coreutils tools.  The files are text files, and so they can be fixed easily.

For continued use, I would like to ban any bad names.

How about removing --badname in 4.18?

Cc: @hallyn , @ikerexxe, @thesamesam , @jubalh , @zeha , @Zugschlus, @rbalint

See also:  <https://dwheeler.com/essays/fixing-unix-linux-filenames.html>

ikerexxe:
I would also feel more comfortable if we eliminated that option.

From what I see in the history, https://github.com/shadow-maint/shadow/issues/121 was opened to re-add a way to bypass the username checks. That ticket mentions https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=42874, where it is discussed that Debian lost that ability because shadow removed it previously. So, shadow has had a way of bypassing username checks for some time already.

Before proceeding to remove the option I would like to understand the reasons why Debian still insists on having that option, and whether they would welcome its removal.

alejandro-colomar:
> I would also feel more comfortable if we eliminated that option.
> 
> From what I see in the history, #121 was opened to re-add a way to bypass the username checks. That ticket mentions https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=42874, where it is discussed that Debian lost that ability because shadow removed it previously. So, shadow has had a way of bypassing username checks for some time already.
> 
> Before proceeding to remove the option I would like to understand the reasons why Debian still insists on having that option, and whether they would welcome its removal.

They're discussing about internationalization of usernames.

The thread started in November here:
<https://lists.debian.org/debian-devel/2024/11/msg00250.html>.
And continued in December here:
<https://lists.debian.org/debian-devel/2024/12/msg00012.html>
(for some reason, their archive software for the mailing list can't follow from one month to the other, and you need a link to both).

Some people are in favor, some are against.

alejandro-colomar:
LWN has also covered it on the news (behind a paywall): <https://lwn.net/Articles/1000485/>

ikerexxe:
[...]

PR #1151, opened 2024-12-09
lib/, man/, src/: Do not allow bad names

alejandro-colomar:
Closes: <https://github.com/shadow-maint/shadow/issues/1149>
Link: <https://lwn.net/Articles/1001215/>
Link: <https://dwheeler.com/essays/fixing-unix-linux-filenames.html>
Link: <https://github.com/shadow-maint/shadow/issues/121>
Link: <https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=42874>
Link: <https://lists.debian.org/debian-devel/2024/11/msg00250.html>
Link: <https://lists.debian.org/debian-devel/2024/12/msg00012.html>
Link: <https://lwn.net/Articles/1000485/>
Cc: @ikerexxe 
Cc: @hallyn 
Cc: @thesamesam 
Cc: @jubalh 
Cc: @zeha 
Cc: @rbalint 
Cc: @zugschlus

While at it:

src/: Report errors in user or groups names more consistently

-  Don't print the user name; if it's bad, it might be dangerous.
-  Print the string "user" or "group" before the error message.
-  Print the errno string to be consistent.


hallyn:
I'd prefer aliasing --allow-badnames to --allow-unsafe-names.  (But keeping the old name so anyone using it doesn't get their scripts broken).

Calling it allow-unsafe-names gives the clear message that we don't advise doing it.  But I don't think it's our place to try to force people to stop.

If shadow itself is breaking on such names, then we should fix those places in shadow.  The conern is that other userspace will break on the unsafe names.  If someone makes a very tight full system where they're convinced unsafe names are fine for them, that's their prerogative.

zeha:
> I'd prefer aliasing --allow-badnames to --allow-unsafe-names. (But keeping the old name so anyone using it doesn't get their scripts broken).
> 
> Calling it allow-unsafe-names gives the clear message that we don't advise doing it. But I don't think it's our place to try to force people to stop.

> If shadow itself is breaking on such names, then we should fix those places in shadow. 

I agree whatever breaks in shadow should be fixed in shadow. It might be debatable what will be called broken though.

* `useradd -m --badname ../foo`?  (this makes a user called `../foo` with a homedir of `/home/../foo`, resulting in `/foo` existing)
* `useradd -m --badname "$(printf '\n')"`? (this makes a user with a zero-length name)

> The conern is that other userspace will break on the unsafe names. If someone makes a very tight full system where they're convinced unsafe names are fine for them, that's their prerogative.

I agree, but I see it as not pushing the rest of userspace forward to a world where more names will be allowed in general.

alejandro-colomar:
> If shadow itself is breaking on such names, then we should fix those places in shadow.

That's easier said than done.  If one insists on using characters such as a newline, can we really "work"?

```sh
root@359f6c0bddb0:/# cp /etc/passwd /tmp/passwd
root@359f6c0bddb0:/# useradd --badname '\n'
```
```diff
root@359f6c0bddb0:/# diff -u /tmp/passwd /etc/passwd
--- /tmp/passwd	2024-12-10 22:56:06.914745615 +0000
+++ /etc/passwd	2024-12-10 22:56:57.662149407 +0000
@@ -16,3 +16,4 @@
 irc:x:39:39:ircd:/run/ircd:/usr/sbin/nologin
 _apt:x:42:65534::/nonexistent:/usr/sbin/nologin
 nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin
+\n:x:1000:1000::/home/\n:/bin/sh
```
```sh
root@359f6c0bddb0:/# useradd --badname '^M'
```
```diff
root@359f6c0bddb0:/# diff -u /tmp/passwd /etc/passwd
--- /tmp/passwd	2024-12-10 22:56:06.914745615 +0000
+++ /etc/passwd	2024-12-10 22:57:43.377648075 +0000
@@ -16,3 +16,5 @@
 irc:x:39:39:ircd:/run/ircd:/usr/sbin/nologin
 _apt:x:42:65534::/nonexistent:/usr/sbin/nologin
 nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin
+\n:x:1000:1000::/home/\n:/bin/sh
:/bin/sh1001::/home/
```
```sh
root@359f6c0bddb0:/# useradd --badname ':,'
useradd: failure while writing changes to /etc/passwd
root@359f6c0bddb0:/# useradd --badname ':'
useradd: failure while writing changes to /etc/passwd
root@359f6c0bddb0:/# useradd --badname ','
```
```diff
root@359f6c0bddb0:/# diff -u /tmp/passwd /etc/passwd
--- /tmp/passwd	2024-12-10 22:56:06.914745615 +0000
+++ /etc/passwd	2024-12-10 22:58:45.041004746 +0000
@@ -16,3 +16,6 @@
 irc:x:39:39:ircd:/run/ircd:/usr/sbin/nologin
 _apt:x:42:65534::/nonexistent:/usr/sbin/nologin
 nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin
+\n:x:1000:1000::/home/\n:/bin/sh
:/bin/sh1001::/home/
+,:x:1002:1002::/home/,:/bin/sh
```

PR #1156, opened 2024-12-16
useradd: Do not automatically add supplements groups for system users

SgAkErRu:
Do not automatically add supplements groups for system users that is created via the `-r` flag. 

Before this change, groups from the `GROUPS` parameter from the `/etc/default/useradd` file were automatically added for the system user, which is most often undesirable (more precisely, I do not even know the case when it would be necessary).

Related to: https://github.com/shadow-maint/shadow/pull/586/


SgAkErRu:
> Could you please explain why we want this?
> 
> Isn't this another way of having an empty GROUPS list? Why not have an empty GROUPS list instead? (I'm not sure I liked the GROUPS list in the first place.)

`GROUPS` allows maintainers to set a list of additional (supplementary) default groups, for example `users`, `dialout`, `audio`, `serial` and so on.

Or, for example, only `users`, which is then expanded through the `libnss-role`, i.e. these groups (`dialout`, `audio`, `serial`) are added through `libnss-role` to those users who are in the `users` group.

But at this moment, it works so that when creating a **system** user via `useradd --system` or `useradd -r`, these groups are also added to it, which is most often undesirable (more precisely, I do not even know the case when it would be necessary).

Therefore, this new option was created, which allows maintainers to disable auto-addition of additional groups for **system** users.

src/useradd.c
```diff
@@ -849,6 +846,18 @@ static int get_groups (char *list)
...
+ */
+static void free_groups (void)
+{
+	int i = 0;
+	while (NULL != user_groups[i]) {
+		free(user_groups[i]);
+		user_groups[i++] = NULL;
+	}
+}
+
```
alejandro-colomar, line 860: This looks like a generic function for freeing a (NULL-terminated) list.  We have list interfaces in <lib/list.c>.

Would you mind calling this API free_list() and moving it to lib/list.c?

Also, the loop looks more like a for loop IMO.  How about this?

```c
void
free_list(char **ls)
{
	for (size_t i = 0; ls != NULL; i++) {
		free(ls[i]);
		ls[i] = NULL;
	}
}
```

src/useradd.c
```diff
@@ -764,11 +765,7 @@ static int get_groups (char *list)
 	/*
 	 * Free previous group list before creating a new one.
 	 */
-	int i = 0;
-	while (NULL != user_groups[i]) {
-		free(user_groups[i]);
-		user_groups[i++] = NULL;
-	}
+	free_groups ();
```
alejandro-colomar, line 768: I would prefer passing the list, to reduce the use of global variables.

alejandro-colomar, line 768: Please do this refactor in a preparation patch, so that the commit that adds the feature is minimal.

ikerexxe:
> But at this moment, it works so that when creating a **system** user via `useradd --system` or `useradd -r`, these groups are also added to it, which is most often undesirable (more precisely, I do not even know the case when it would be necessary).

Since you are the original author of this functionality and you don't know the case, why don't you filter the group membership addition for non-system users instead of creating a new option? I think that would be the most sensible approach.

SgAkErRu:
> > But at this moment, it works so that when creating a **system** user via `useradd --system` or `useradd -r`, these groups are also added to it, which is most often undesirable (more precisely, I do not even know the case when it would be necessary).
> 
> Since you are the original author of this functionality and you don't know the case, why don't you filter the group membership addition for non-system users instead of creating a new option? I think that would be the most sensible approach.

Because I am not original author of this functionality. 

So, that's why I suggest this option, which will keep the possibility of the original behavior (when these groups automatically adds to system users).

src/useradd.c
SgAkErRu: I don't mind. 

Should adding the `free_list()` function and using it here in `useradd.c` be in the same PR or in two different ones?

ikerexxe:
> Because I am not original author of this functionality.

Ups sorry, I have mixed up the names.

@AZaugg any opinion on the following topic?

> > But at this moment, it works so that when creating a **system** user via `useradd --system` or `useradd -r`, these groups are also added to it, which is most often undesirable (more precisely, I do not even know the case when it would be necessary).
> 
> Since you are the original author of this functionality and you don't know the case, why don't you filter the group membership addition for non-system users instead of creating a new option? I think that would be the most sensible approach.

src/useradd.c
alejandro-colomar: You can do it in the same PR, in a separate commit.

alejandro-colomar: Oh, and it can be simplified into this:

```c
void
free_list(char **ls)
{
	for (size_t i = 0; ls != NULL; i++)
		free(ls[i]);
	ls[0] = NULL;
}

File: src/useradd.c [hunk 3]
```diff
@@ -764,11 +764,7 @@ static int get_groups (char *list)
 	/*
 	 * Free previous group list before creating a new one.
 	 */
-	int i = 0;
-	while (NULL != user_groups[i]) {
-		free(user_groups[i]);
-		user_groups[i++] = NULL;
-	}
+	free_list (user_groups);
```
alejandro-colomar, line 767: Please don't add that whitespace.

src/useradd.c
```diff
@@ -1591,6 +1591,13 @@ static void process_flags (int argc, char **argv)
 		if (getdef_bool ("CREATE_HOME")) {
 			mflg = true;
 		}
+	} else {
+		/* If SYS_USER_AUTO_GROUPS_ENAB is disabled,
+		 * then do not automatically add supplements groups for system users. */
+		if (!getdef_bool ("SYS_USER_AUTO_GROUPS_ENAB") && !Gflg && do_grp_update) {
```
alejandro-colomar, line 1597: Please don't add whitespace here in `getdef_bool (` either.  I know the old code has it, but I'm working on slowly getting rid of that.

Thanks!

alejandro-colomar: Thanks!

SgAkErRu:
@ikerexxe Hi! I can redo this PR as you suggested (filter the group membership addition for non-system users instead of creating a new option) - since the author @AZaugg of the original change has not expressed any objections for quite a long time.

ikerexxe:
@SgAkErRu yes, go ahead

SgAkErRu:
@ikerexxe 

> @SgAkErRu yes, go ahead

Done

src/useradd.c
```diff
@@ -1552,6 +1552,12 @@ static void process_flags (int argc, char **argv, struct option_flags *flags)
 		if (getdef_bool ("CREATE_HOME")) {
 			mflg = true;
 		}
+	} else {
+		/* Do not automatically add supplements groups for system users. */
+		if (!Gflg && do_grp_update) {
+			free_list(user_groups);
+			do_grp_update = false;
+		}
```
ikerexxe: Having this in an else condition is confusing because it mixes two different things (i.e. home directory and secondary groups). Can you make it more obvious?
```
if (rflg && !Gflg && do_grp_update) {
...
```

ikerexxe (changes requested):
The current implementation contains a flaw. The `grp_add()` function also sets `do_grp_update = true` in [line 1950](https://github.com/shadow-maint/shadow/blob/master/src/useradd.c#L1950). That function is called when `useradd` is called with `-U` flag. That means that if a user is created with `-U` and without `-G` flags, then the logic you implemented will be executed. That's an undesired execution and it will leave some users without their default supplementary groups.

In addition, I added an inline comment. Probably the end code won't look like that once you fix the aforementioned flaw but it should be something similar.

Finally, one of the commits mentions `useradd: fix test 69_useradd_default_GROUPS_name`, what is it exactly you are fixing? I'm assuming you are adapting to the new GROUPS behaviour you are trying to implement, but I'd make it more obvious in the commit message.

SgAkErRu:
> The current implementation contains a flaw. The `grp_add()` function also sets `do_grp_update = true` in [line 1950](https://github.com/shadow-maint/shadow/blob/master/src/useradd.c#L1950). That function is called when `useradd` is called with `-U` flag. That means that if a user is created with `-U` and without `-G` flags, then the logic you implemented will be executed. That's an undesired execution and it will leave some users without their default supplementary groups.

But the new code check `do_grp_update` and sets `do_grp_update = false` inside the `process_flags()` function, which is called in `main()` BEFORE calling the `grp_add()` when the `-U` flag is set.

So, the new code check `do_grp_update`, which was changed to true inside the `get_defaults` function for the `GROUPS` variable case, and based on the order in which commands are executed, it should not conflict with this logic with `-U` flag.

Or am I wrong?

SgAkErRu:
> Finally, one of the commits mentions useradd: fix test 69_useradd_default_GROUPS_name, what is it exactly you are fixing? I'm assuming you are adapting to the new GROUPS behaviour you are trying to implement, but I'd make it more obvious in the commit message.

No, the behavior of the test itself does not change in any way and does not adapt to the current PR, because the current PR does not change the behavior for the regular user, only for the system user. And a new test number 70 has been created for the system user.

The corrections are visible in the commit itself:

1. Missing files have been added for the test to work: `gshadow`, `passwd`, `shadow`. Without them, the foo user was created with a different UID and thus the test failed.

2. Added an empty line at the end of the file in `config/group`

3. The extra space at the beginning of the file in `data/group` was removed and an empty line was added at the end of the file. That extra space also led to a test failure due to a mismatch between the initial line of the `group` file in the data and config subfolders.

4. Fixed a typo in the log message in `useradd.test`

If necessary, I can add this information to the description of the commit.

src/useradd.c
SgAkErRu: No problem, I can redo it, if necessary.

I just took it as such a general check of the -r flag, and not as a check for creating a home dir. Because it was written with a nested condition, and not with &&.

I mean now it is like:

```

if (!rflg) {
    /* for system accounts defaults are ignored and we
     * do not create a home dir */
    if (getdef_bool ("CREATE_HOME")) {
        mflg = true;
    }
}

```

Instead of:

```

/* for system accounts defaults are ignored and we
 * do not create a home dir */
if (!rflg && getdef_bool ("CREATE_HOME")) {
    mflg = true;
}

```

So I just followed the existing style.

ikerexxe: It's not so much about the style but about grouping different things together. The original condition does some processing for the home directory, while the new code is about secondary groups. I'd like to keep those things separate

ikerexxe:
> Or am I wrong?

No, you are right. I must have misunderstood something but checking it again it behaves as you specified.

>     1. Missing files have been added for the test to work: `gshadow`, `passwd`, `shadow`. Without them, the foo user was created with a different UID and thus the test failed.
> 
>     2. Added an empty line at the end of the file in `config/group`
> 
>     3. The extra space at the beginning of the file in `data/group` was removed and an empty line was added at the end of the file. That extra space also led to a test failure due to a mismatch between the initial line of the `group` file in the data and config subfolders.
> 
>     4. Fixed a typo in the log message in `useradd.test`
> 
> 
> If necessary, I can add this information to the description of the commit.

Please mention 1 and you can merge 2, 3 and 4 with `minor improvements` or something like similar.

src/useradd.c
SgAkErRu: Ok, done

ikerexxe (approved):
LGTM. Thank you for the patches

Issue #1160, opened 2024-12-28
4.17.0: su - does not invoke login shell


hdejin:
After updating to 4.17.0, running su - no longer invokes a login shell.
I've built the source with --with-libpam and using the following pam config for su:

auth        required        pam_env.so
auth        required        pam_nologin.so successok file=/etc/nologin
auth        sufficient      pam_rootok.so
auth        sufficient      pam_unix.so likeauth nullok
auth        required        pam_deny.so

account     sufficient      pam_unix.so likeauth nullok
account     required        pam_deny.so

password    required        pam_unix.so nullok use_authtok md5 shadow
password    required        pam_deny.so

session     required        pam_unix.so

dchmelik:
I and several other Slackware-current users have same problem.

stoeckmann:
The PR https://github.com/shadow-maint/shadow/pull/1163 should fix this issue.

alejandro-colomar:
Thanks for the report and the quick fix to all of you!  And sorry for the bug.

PR #1163, opened 2024-12-30
su: Fix su - regression

stoeckmann:
Launch a login shell again if requested through "su -" or "su -l".

Fixes: d9923431eb38 ("src/: Use xasprintf() instead of its pattern")
Closes: <https://github.com/shadow-maint/shadow/issues/1160>

### DIFF
--- a/src/su.c
+++ b/src/su.c
@@ -1212,6 +1212,7 @@ int main (int argc, char **argv)
 		}
 
 		xasprintf(&arg0, "-%s", cp);
+		cp = arg0;
 	} else {
 		cp = Basename (shellstr);
 	}


glangshaw:
> Launch a login shell again if requested through "su -" or "su -l".
> 
> Fixes: [d992343](https://github.com/shadow-maint/shadow/commit/d9923431eb3878243a38c7cc4e11897daebc779d) ("src/: Use xasprintf() instead of its pattern") Closes: #1160

Instead, why not just : 
xasprintf(&cp, "-%s", cp);

and remove the arg0 .

stoeckmann:
Should work. Is there a guarantee somewhere that `vasprintf` always has to properly handle situations where first argument is also part of the variable argument list?

Since I didn't find this guarantee now, I'll let maintainers decide and keep it as it is now to highlight the regression without doing more refactoring.

glangshaw:
Fair enough.  I just noticed that cp was being reused throughout that section of code, so it seemed in keeping with the general flow.   In this case, the first parameter is the address of the cp pointer, while the 3rd parameter is the value of the cp pointer so I don't believe it will be a problem here. That said, the stdarg variadic stuff  in C has always struck me as a bit weird and kludgy.  I usually try and avoid using it in my own code. wherever possible.

Anyway, thanks for PR and the response,. As you say,, the shadow guys can decide for themselves.  In all honesty, I preferred the old code before they switched to xasprintf().

stoeckmann:
> In this case, the first parameter is the address of the cp pointer, while the 3rd parameter is the value of the cp pointer

Oh right, it's call by value for the var args. In that case your suggestion works as well. Thanks for clarification.

alejandro-colomar:
It seems glibc is missing a `access(write_only, 1)` attribute.

```c
$ grepc vasprintf /usr/include/
/usr/include/stdio.h:extern int vasprintf (char **__restrict __ptr, const char *__restrict __f,
		      __gnuc_va_list __arg)
     __THROWNL __attribute__ ((__format__ (__printf__, 2, 0))) __wur;
```

It should have `__attribute__((access(write_only, 1)))`, I think.

I'll CC a few maintainers (we should also email their mailing list, but I'll CC them here for completeness).

Cc: @fweimer-rh , @eggert

alejandro-colomar (approved):
@hallyn should we release from master?  Since we haven't yet applied any commits to master, I think this would be simpler.

alejandro-colomar:
```
```

alejandro-colomar:
> Since I didn't find this guarantee now, I'll let maintainers decide and keep it as it is now to highlight the regression without doing more refactoring.

Agree.  Your patch better shows the regression.  I'll merge now.

hallyn:
Sorry what does "release from master" mean @alejandro-colomar

alejandro-colomar:
> Sorry what does "release from master" mean @alejandro-colomar

Should you release 4.17.1 just like you do the main releases directly in the `master` branch?  Or should I create a `4.17.x` branch and release 4.17.1 from there?

hallyn:
Oh.  I can't think of a particular advantage to either one.

hallyn:
Sure I can do from master branch.

hallyn:
(seems we should have a testcase for this, using expect...)

alejandro-colomar:
> > Launch a login shell again if requested through "su -" or "su -l".
> > Fixes: [d992343](https://github.com/shadow-maint/shadow/commit/d9923431eb3878243a38c7cc4e11897daebc779d) ("src/: Use xasprintf() instead of its pattern") Closes: #1160
> 
> Instead, why not just : xasprintf(&cp, "-%s", cp);
> 
> and remove the arg0 .

On the API design space, I don't like this API.  I can't blame the authors, since it was designed a long time ago, way before the malloc() GNU attribute was invented (I think), and before static analysis was so powerful.

These days, it would have been better to return a pointer to the newly allocated string, just like strdup(3).  That would allow a much better static analysis.  It also makes it obvious that the old value of the pointer is unimportant and ignored.  Since most callers don't care about the length, but some may, I would have a `size_t *len` argument which can be NULL to ignore, or can be a pointer to a size_t for getting the length of the new string.

That would result in `cp = xasprintf2(NULL, "-%s", cp);`, which I think is cleaner.

alejandro-colomar:
> > > Launch a login shell again if requested through "su -" or "su -l".
> > > Fixes: [d992343](https://github.com/shadow-maint/shadow/commit/d9923431eb3878243a38c7cc4e11897daebc779d) ("src/: Use xasprintf() instead of its pattern") Closes: #1160
> > 
> > 
> > Instead, why not just : xasprintf(&cp, "-%s", cp);
> > and remove the arg0 .
> 
> On the API design space, I don't like this API. I can't blame the authors, since it was designed a long time ago, way before the malloc() GNU attribute was invented (I think), and before static analysis was so powerful.
> 
> These days, it would have been better to return a pointer to the newly allocated string, just like strdup(3). That would allow a much better static analysis. It also makes it obvious that the old value of the pointer is unimportant and ignored. Since most callers don't care about the length, but some may, I would have a `size_t *len` argument which can be NULL to ignore, or can be a pointer to a size_t for getting the length of the new string.
> 
> That would result in `cp = xasprintf2(NULL, "-%s", cp);`, which I think is cleaner.

Hmmm, I would go further and say that those who care about the lenght, would avoid asprintf(3) altogether, and use some faster APIs.  And if one wants the length ocasionally, they can just call strlen(3); the performance loss by calling strlen(3) is negligible after a call to asprintf(3).  So a good API would just omit the length at all, and just return the newly allocated string.

New API in the oven, while listening to the New Year's Concert by the Vienna Philharmonic.  :)

PR #1164, opened 2024-12-31
Use HTTPS in link to Wikipedia article on password strength

scottdotweb:
The link to the Wikipedia article was added here in 2008, and Wikipedia went HTTPS-only in 2015. (https://diff.wikimedia.org/2015/06/12/securing-wikimedia-sites-with-https/)

### DIFF
--- a/man/passwd.1.xml
+++ b/man/passwd.1.xml
@@ -137,7 +137,7 @@
 
       <para>
 	You can find advice on how to choose a strong password on
-	http://en.wikipedia.org/wiki/Password_strength
+	https://en.wikipedia.org/wiki/Password_strength
       </para>
     </refsect2>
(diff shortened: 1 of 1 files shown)


ikerexxe:
The changes themselves look good, but the commit message does not follow the project practices.

We usually limit the commit title to 50 characters and the message lines to 72. In addition, we mention the folder in which we are committing the changes in the title and add the links at the end. Here is an example to show you what is expected: https://github.com/shadow-maint/shadow/commit/0b8c0c893c004b204ca55d303cc3b83a25f47019

scottdotweb:
Hi Iker and thanks for the guidance 🙂 This project is quite outside my usual area of contributions. I've modified the commit per those rules.

PR #1168, opened 2025-01-01
[x][v]asprintf_():  Add APIs, and use them

alejandro-colomar:


alejandro-colomar:
As a curiosity, this asprint() API seems to be exactly the same that Plan9 calls smprint(3).

<https://9fans.github.io/plan9port/man/man3/print.html>

alejandro-colomar:
This API has been discussed in the C Committee for inclusion in C2y, and there seems to be interest in the API.  Here's a draft of a proposal, which will be discussed in the next committee meeting: <https://www.alejandro-colomar.es/src/alx/alx/wg14/alx-0007.git/tree/alx-0007.txt>.

lib/semanage.c
```diff
@@ -23,27 +23,23 @@
...
                                      const char *fmt, ...)
 {
-	int ret;
 	char * message = NULL;
 	va_list ap;
 
-
 	va_start (ap, fmt);
-	ret = vasprintf (&message, fmt, ap);
+	message = vaprintf(fmt, ap);
```
hallyn, line 39: Why not use aprintf here and drop the va_list?

alejandro-colomar: semanage_error_callback() is a variadic function.  You need to use va_list to reference its argument list.  It's impossible to do use aprintf() in this code, unless I'm not understanding what you're suggesting.

hallyn: Oh.. right

PR #1170, opened 2025-01-06
src/login_nopam.c: Fix compiler warnings

stoeckmann:
The function `user_match` actually modifies the string passed as its first argument, so use `char *` instead of `const char *`.

### DIFF
--- a/src/login_nopam.c
+++ b/src/login_nopam.c
@@ -68,9 +68,9 @@
 #define TABLE	"/etc/login.access"
 #endif
 
-static bool list_match (char *list, const char *item, bool (*match_fn) (const char *, const char *));
-static bool user_match (const char *tok, const char *string);
-static bool from_match (const char *tok, const char *string);
+static bool list_match (char *list, const char *item, bool (*match_fn) (char *, const char *));
+static bool user_match (char *tok, const char *string);
+static bool from_match (char *tok, const char *string);
 static bool string_match (const char *tok, const char *string);
 static const char *resolve_hostname (const char *string);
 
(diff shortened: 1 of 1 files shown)


src/login_nopam.c
```diff
@@ -289,7 +289,7 @@ static const char *resolve_hostname (const char *string)
 
 /* from_match - match a host or tty against a list of tokens */
 
-static bool from_match (const char *tok, const char *string)
+static bool from_match (char *tok, const char *string)
```
alejandro-colomar, line 292: This one doesn't write, no?  Am I missing something?

alejandro-colomar: Ahhh, it's for compatibility with the function pointer.  Sorry for the noise.

stoeckmann: You are correct, it doesn't write. But if it's const, then it does not match the signature required by list_match. So ... unfortunately we are forced to do that here.

alejandro-colomar (approved):
LGTM.  Thanks!

```
```

ikerexxe (approved):
LGTM!

src/login_nopam.c
alejandro-colomar: WG14 has a paper proposing relaxing these rules for C2y.  It basically proposes allowing certain implicit conversions between function pointers that are known to be safe.  It was uploaded the day after we discussed this.
<https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3449.pdf>

Coincidence or not?  :-)

PR #1171, opened 2025-01-06
Add cheap defense mechanisms

stoeckmann:
Before you read any further: No reachable issues exist in current code base (as far as I can tell)

But these defense mechanisms would have made reviewing easier because I could have skipped a few searches:

1. Use `_exit` instead of `exit` after execution errors

After a fork and a failing exec, calling `exit` is not the best choice. Use `_exit` instead because tools like `groupmod` etc. use `add_cleanup` to clean up when `exit` is called. If the child, which inherits these instructions, calls `exit` by itself, these cleanup routines could release passwd/group locks too early. Again: Does not happen, because the code launches sub processes only after these cleanup routines have been already released again. Yet, better be safe than sorry.

2. Set close on exec on critical file descriptors

Even though `fopen` is used in lib/gshadow.c and lib/shadow.c, we can add the "e" mode to trigger O_CLOEXEC for systems which support this glibc extension. The file descriptor which keeps /etc/gshadow or /etc/shadow open is then guaranteed to not be passed down to a sub process. Again: Does not happen. The code carefully closes these file descriptors through dedicated function calls before sub processes are launched. Yet, better be safe than sorry again.

### DIFF
--- a/lib/gshadow.c
+++ b/lib/gshadow.c
@@ -55,7 +55,7 @@ void setsgent (void)
 	if (NULL != shadow) {
 		rewind (shadow);
 	} else {
-		shadow = fopen (SGROUP_FILE, "r");
+		shadow = fopen (SGROUP_FILE, "re");
 	}
 }
 
--- a/lib/run_part.c
+++ b/lib/run_part.c
@@ -32,7 +32,7 @@ static int run_part(char *script_path, const char *name, const char *action)
 		setenv("SUBJECT",name,1);
 		execv(script_path,args);
 		fprintf(shadow_logfd, "execv: %s\n", strerror(errno));
-		exit(1);
+		_exit(1);
 	}
 
 	pid_status = wait(&wait_status);
--- a/lib/shadow.c
+++ b/lib/shadow.c
@@ -41,7 +41,7 @@ void setspent (void)
 	if (NULL != shadow) {
 		rewind (shadow);
 	}else {
-		shadow = fopen (SHADOW_FILE, "r");
+		shadow = fopen (SHADOW_FILE, "re");
 	}
 }
 
--- a/lib/spawn.c
+++ b/lib/spawn.c
@@ -35,11 +35,11 @@ run_command(const char *cmd, const char *argv[],
 		(void) execve (cmd, (char * const *) argv,
 		               (char * const *) envp);
 		if (ENOENT == errno) {
-			exit (E_CMD_NOTFOUND);
+			_exit (E_CMD_NOTFOUND);
 		}
 		fprintf (shadow_logfd, "%s: cannot execute %s: %s\n",
 		         shadow_progname, cmd, strerror (errno));
-		exit (E_CMD_NOEXEC);
+		_exit (E_CMD_NOEXEC);
 	} else if ((pid_t)-1 == pid) {
 		fprintf (shadow_logfd, "%s: cannot execute %s: %s\n",
 		         shadow_progname, cmd, strerror (errno));


alejandro-colomar:
It would be interesting to patch the fork(2) manual page to do the same thing in the EXAMPLES section.  Would you mind sending a patch to linux-man@ too?

alejandro-colomar (approved):
LGTM.  Thanks!

```
```

stoeckmann:
> It would be interesting to patch the fork(2) manual page to do the same thing in the EXAMPLES section. Would you mind sending a patch to linux-man@ too?

I've looked at the EXAMPLES section now, but I'm not sure if it really helps there. No exec call is performed and the whole main-function is printed without any atexit calls. Leaving the forked process with exit in this case is no problem and introducing _exit might make the code less readable or more confusing imho.

alejandro-colomar:
> > It would be interesting to patch the fork(2) manual page to do the same thing in the EXAMPLES section. Would you mind sending a patch to linux-man@ too?
> 
> I've looked at the EXAMPLES section now, but I'm not sure if it really helps there. No exec call is performed and the whole main-function is printed without any atexit calls. Leaving the forked process with exit in this case is no problem and introducing _exit might make the code less readable or more confusing imho.

I guess _exit(2) is never wrong there?  Or is it sometimes wrong?  If _exit(2) is always better (or at least equivalent), maybe it would be good to show it in the example, to promote it.  And in the commit message we could add a detailed explanation.

What do you think?

stoeckmann:
> I guess _exit(2) is never wrong there? Or is it sometimes wrong? If _exit(2) is always better (or at least equivalent), maybe it would be good to show it in the example, to promote it. And in the commit message we could add a detailed explanation.
> 
> What do you think?

I'll prepare a patch. Promoting that _exit exists might help people out there to at least read its manual page. We can discuss this further on man-pages mailing list if needed.

alejandro-colomar:
> > I guess _exit(2) is never wrong there? Or is it sometimes wrong? If _exit(2) is always better (or at least equivalent), maybe it would be good to show it in the example, to promote it. And in the commit message we could add a detailed explanation.
> > What do you think?
> 
> I'll prepare a patch. Promoting that _exit exists might help people out there to at least read its manual page. We can discuss this further on man-pages mailing list if needed.

Thanks!

Indeed, I knew it existed, but had only seen it sporadically, and never knew what it was really good for, until you sent this PR.  Using it in the manual page will hopefully help people know about it, and give it more visibility.

Karlson2k:
Doesn't use of glibc extensions break functioning with non-glibc, like musl?

Isn't it safe to use constructs like 
``` C
shadow = fopen (SGROUP_FILE, "re");
if (NULL == shadow )
  shadow = fopen (SGROUP_FILE, "r");
```
?

Or, alternatively, detect extension in `configure`?

alejandro-colomar:
Hi,

On Sun, Feb 16, 2025 at 06:15:18PM -0800, Karlson2k wrote:
> Karlson2k left a comment (shadow-maint/shadow#1171)
> 
> Doesn't use of glibc extensions break functioning with non-glibc, like musl?

Hmmm, I didn't know musl doesn't support this.  It would be interesting
to get them to support it.  I've CCd several interested parties in this
email.

> 
> Isn't it safe to use constructs like 
> ``` C
> shadow = fopen (SGROUP_FILE, "re");
> if (NULL == shadow )
>   shadow = fopen (SGROUP_FILE, "r");
> ```
> ?

Is 'e' only available in glibc?  Do other libraries consciously not
support O_CLOEXEC in fopen(3)?

I see that POSIX.1-2024 added the 'e' mode string character, so we're
using standard features (yeah, very modern ones, but still standard).
Is there any reason to not implement them, or is it just a matter of
time and contributors?

<https://pubs.opengroup.org/onlinepubs/9799919799/functions/fopen.html>

> Or, alternatively, detect extension in `configure`?

If we have to...


Have a lovely day!
Alex

-- 

alejandro-colomar:
Hi,

On Mon, Feb 17, 2025 at 10:42:11AM +0100, Alejandro Colomar wrote:
> On Sun, Feb 16, 2025 at 06:15:18PM -0800, Karlson2k wrote:
> > Karlson2k left a comment (shadow-maint/shadow#1171)
> > 
> > Doesn't use of glibc extensions break functioning with non-glibc, like musl?

After reading musl's source code, it does support 'e'.  And it does so
since 2012.  Which libc are you using?  And which version?

	***@***.***:~/src/musl/libc/master$ grepc fopen .
	./include/stdio.h:FILE *fopen(const char *__restrict, const char *__restrict);
	./src/stdio/fopen.c:FILE *fopen(const char *restrict filename, const char *restrict mode)
	{
		FILE *f;
		int fd;
		int flags;

		/* Check for valid initial mode character */
		if (!strchr("rwa", *mode)) {
			errno = EINVAL;
			return 0;
		}

		/* Compute the flags to pass to open() */
		flags = __fmodeflags(mode);

		fd = sys_open(filename, flags, 0666);
		if (fd < 0) return 0;
		if (flags & O_CLOEXEC)
			__syscall(SYS_fcntl, fd, F_SETFD, FD_CLOEXEC);

		f = __fdopen(fd, mode);
		if (f) return f;

		__syscall(SYS_close, fd);
		return 0;
	}
	***@***.***:~/src/musl/libc/master$ grepc __fmodeflags .
	./src/stdio/__fmodeflags.c:int __fmodeflags(const char *mode)
	{
		int flags;
		if (strchr(mode, '+')) flags = O_RDWR;
		else if (*mode == 'r') flags = O_RDONLY;
		else flags = O_WRONLY;
		if (strchr(mode, 'x')) flags |= O_EXCL;
		if (strchr(mode, 'e')) flags |= O_CLOEXEC;
		if (*mode != 'r') flags |= O_CREAT;
		if (*mode == 'w') flags |= O_TRUNC;
		if (*mode == 'a') flags |= O_APPEND;
		return flags;
	}
	./src/internal/stdio_impl.h:hidden int __fmodeflags(const char *);
	***@***.***:~/src/musl/libc/master$ git blame -- src/stdio/__fmodeflags.c | grep CLOEXEC
	892cafff6 (Rich Felker 2012-10-24 21:16:06 -0400 11) 	if (strchr(mode, 'e')) flags |= O_CLOEXEC;
	***@***.***:~/src/musl/libc/master$ git show 892cafff6 | grep -e CLOEXEC -e ^diff | head -n5
	diff --git a/src/internal/stdio_impl.h b/src/internal/stdio_impl.h
	diff --git a/src/stdio/__fmodeflags.c b/src/stdio/__fmodeflags.c
	+	if (strchr(mode, 'e')) flags |= O_CLOEXEC;
	diff --git a/src/stdio/fopen.c b/src/stdio/fopen.c
	-	if (strchr(mode, 'e')) flags |= O_CLOEXEC;
	***@***.***:~/src/musl/libc/master$ git blame 892cafff6^ -- src/stdio/fopen.c | grep CLOEXEC
	8582a6e9f (Rich Felker 2012-09-29 18:09:34 -0400 20) 	if (strchr(mode, 'e')) flags |= O_CLOEXEC;
	***@***.***:~/src/musl/libc/master$ git log -1 8582a6e9f
	commit 8582a6e9f25dd7b87d72961f58008052a4cac473
	Author: Rich Felker ***@***.***>
	Date:   Sat Sep 29 18:09:34 2012 -0400

	    add 'e' modifier (close-on-exec) to fopen and fdopen
	    
	    this feature will be in the next version of POSIX, and can be used
	    internally immediately. there are many internal uses of fopen where
	    close-on-exec is needed to fix bugs.


Cheers,
Alex

-- 

alejandro-colomar:
Hi,

On Mon, Feb 17, 2025 at 10:42:11AM +0100, Alejandro Colomar wrote:
> On Sun, Feb 16, 2025 at 06:15:18PM -0800, Karlson2k wrote:
> > Karlson2k left a comment (shadow-maint/shadow#1171)
> >
> > Doesn't use of glibc extensions break functioning with non-glibc, like musl?

After reading musl's source code, it does support 'e'.  And it does so
since 2012.  Which libc are you using?  And which version?

```
        alx@devuan:~/src/musl/libc/master$ grepc fopen .
        ./include/stdio.h:FILE *fopen(const char *__restrict, const char *__restrict);
        ./src/stdio/fopen.c:FILE *fopen(const char *restrict filename, const char
+*restrict mode)
        {
                FILE *f;
                int fd;
                int flags;

                /* Check for valid initial mode character */
                if (!strchr("rwa", *mode)) {
                        errno = EINVAL;
                        return 0;
                }

                /* Compute the flags to pass to open() */
                flags = __fmodeflags(mode);

                fd = sys_open(filename, flags, 0666);
                if (fd < 0) return 0;
                if (flags & O_CLOEXEC)
                        __syscall(SYS_fcntl, fd, F_SETFD, FD_CLOEXEC);

                f = __fdopen(fd, mode);
                if (f) return f;

                __syscall(SYS_close, fd);
                return 0;
        }
```
```
        alx@devuan:~/src/musl/libc/master$ grepc __fmodeflags .
        ./src/stdio/__fmodeflags.c:int __fmodeflags(const char *mode)
        {
                int flags;
                if (strchr(mode, '+')) flags = O_RDWR;
                else if (*mode == 'r') flags = O_RDONLY;
                else flags = O_WRONLY;
                if (strchr(mode, 'x')) flags |= O_EXCL;
                if (strchr(mode, 'e')) flags |= O_CLOEXEC;
                if (*mode != 'r') flags |= O_CREAT;
                if (*mode == 'w') flags |= O_TRUNC;
                if (*mode == 'a') flags |= O_APPEND;
                return flags;
        }
        ./src/internal/stdio_impl.h:hidden int __fmodeflags(const char *);
```
```
        alx@devuan:~/src/musl/libc/master$ git blame -- src/stdio/__fmodeflags.c | grep
+CLOEXEC
        892cafff6 (Rich Felker 2012-10-24 21:16:06 -0400 11)    if (strchr(mode, 'e'))
+flags |= O_CLOEXEC;
```
```
        alx@devuan:~/src/musl/libc/master$ git show 892cafff6 | grep -e CLOEXEC -e
+^diff | head -n5
        diff --git a/src/internal/stdio_impl.h b/src/internal/stdio_impl.h
        diff --git a/src/stdio/__fmodeflags.c b/src/stdio/__fmodeflags.c
        +       if (strchr(mode, 'e')) flags |= O_CLOEXEC;
        diff --git a/src/stdio/fopen.c b/src/stdio/fopen.c
        -       if (strchr(mode, 'e')) flags |= O_CLOEXEC;
```
```
        alx@devuan:~/src/musl/libc/master$ git blame 892cafff6^ -- src/stdio/fopen.c |
+grep CLOEXEC
        8582a6e9f (Rich Felker 2012-09-29 18:09:34 -0400 20)    if (strchr(mode, 'e'))
+flags |= O_CLOEXEC;
```
```
        alx@devuan:~/src/musl/libc/master$ git log -1 8582a6e9f
        commit 8582a6e9f25dd7b87d72961f58008052a4cac473
        Author: Rich Felker <dalias@aerifal.cx>
        Date:   Sat Sep 29 18:09:34 2012 -0400

            add 'e' modifier (close-on-exec) to fopen and fdopen

            this feature will be in the next version of POSIX, and can be used
            internally immediately. there are many internal uses of fopen where
            close-on-exec is needed to fix bugs.
```

Cheers,
Alex

Karlson2k:
On GNU/Linux I'm using glibc mainly, but my comment is not related to my specific use.

The generic approach to writing reliable portable software: 
* do not use private extension if possible (they are unportable by their nature and could be even removed in new releases)
* if private extension is important, test for extension availability before use
* do not blindly use standard features if the standard is new, check for feature availability instead
* use fallback paths if possible

Talking about this flag. It is supported by glibc, musl and uclibc-ng. Not supported by dietlibc.

alejandro-colomar:
> On GNU/Linux I'm using glibc mainly, but my comment is not related to my specific use.
> 
> The generic approach to writing reliable portable software:
> 
>     * do not use private extension if possible (they are unportable by their nature and could be even removed in new releases)
> 
>     * if private extension is important, test for extension availability before use
> 
>     * do not blindly use standard features if the standard is new, check for feature availability instead
> 
>     * use fallback paths if possible
> 



Since this extension is old (available in glibc since 2007, and in musl since 2012), and it provides important security features, and it has been standardized, I think it's okay to use it.

Even back in 2012, it had already been accepted into the next POSIX standard (which would eventually become POSIX.1-2024), as the commit message from musl documents:

```
        commit 8582a6e9f25dd7b87d72961f58008052a4cac473
        Author: Rich Felker <dalias@aerifal.cx>
        Date:   Sat Sep 29 18:09:34 2012 -0400

            add 'e' modifier (close-on-exec) to fopen and fdopen

            this feature will be in the next version of POSIX, and can be used
            internally immediately. there are many internal uses of fopen where
            close-on-exec is needed to fix bugs.
```

> Talking about this flag. It is supported by glibc, musl and uclibc-ng. Not supported by dietlibc.

Good; it has wide support.  dietlibc should implement it.

Karlson2k:
Sure, it is up to maintainers whether to improve code portability or wait for others to adapt. :wink:
On the other hand, the code already has a number of uses of unportable (or not fully portable) features, so this particular part is not a problem as the whole code may work only in specific environments, where these extensions are automatically available.

mirabilos:
(Not that there’s any need for the shadow package on MirBSD, but…)

On Mon, 17 Feb 2025, Alejandro Colomar wrote:

>Thorsten, is it an undocumented feature?  Or does MirBSD not implement
>O_CLOEXEC in fopen(3)?

MirBSD didn’t even have O_CLOEXEC in open until recently (2016),
and I don’t expect it to provide latest POSIX features any time
soon other than on a need basis for now. I’m stuck fixing some
locale-related issues at the moment…

fopen(3) and friends currently support /^[rwa]b?\+?/ with [rwa]
and [+] actually doing something (src/lib/libc/stdio/flags.c).

In general, MirBSD doesn’t really target POSIX, that’s pretty
much insurmountable (especially the threads part) and in parts
(leap seconds) even against a law in my country, but I try to
bring mksh as close to POSIX as possible/sensible (all current
shells ignore the part where you’re supposed to assign a path
for builtins). Expect porting to MirBSD to be similar to porting
to a classic BSD or other Unix, not to a modern one. And, given
its slow progress at the moment it’s sensible to not consider it
much when porting software (not calling it abandoned, but I’ve
yet to get quite some working on it done before I can even use
it as daily box myself again).

Still thanks for thinking about it,
//mirabilos

PR #1172, opened 2025-01-07
lib/attr.h: use C23 attributes only with gcc >= 10

kanavin:
These are not available on earlier versions and builds break there.

### DIFF
--- a/lib/attr.h
+++ b/lib/attr.h
@@ -5,7 +5,7 @@
 #include "config.h"
 
 
-#if defined(__GNUC__)
+#if (__GNUC__ >= 10)
 # define MAYBE_UNUSED                [[gnu::unused]]
 # define NORETURN                    [[gnu::__noreturn__]]
 # define format_attr(type, fmt, va)  [[gnu::format(type, fmt, va)]]


alejandro-colomar:
> These are not available on earlier versions and builds break there.

Hmmm, I was wondering why no-one had complained about them.  I hoped that all users would have a recent enough GCC.  Where are you building shadow that triggers this?

I'm fine with the change, I'm just curious.  :)

Thanks for the patch!

alejandro-colomar (approved):
```
```

lib/attr.h
```diff
@@ -5,7 +5,7 @@
 #include "config.h"
 
 
-#if defined(__GNUC__)
+#if defined(__GNUC__) && (__GNUC__ >= 10)
```
alejandro-colomar, line 8: Hmm, we can simplify by removing the defined() clause.  Undefined macros have the value 0 in `#if` directives:

```c
alx@devuan:~/tmp/gcc$ cat cpp.c
#if (ASD > 1)
#warning TRUE
#else
#warning FALSE
#endif

#if (ASD < 1)
#warning TRUE
#else
#warning FALSE
#endif

#if (ASD == 0)
#warning TRUE
#else
#warning FALSE
#endif
alx@devuan:~/tmp/gcc$ cc -Wall -Wextra -S cpp.c 
cpp.c:4:2: warning: #warning FALSE [-Wcpp]
    4 | #warning FALSE
      |  ^~~~~~~
cpp.c:8:2: warning: #warning TRUE [-Wcpp]
    8 | #warning TRUE
      |  ^~~~~~~
cpp.c:14:2: warning: #warning TRUE [-Wcpp]
   14 | #warning TRUE
      |  ^~~~~~~
```

kanavin: Done

alejandro-colomar:
BTW, is this a regression compared to 4.16?  Or did 4.16 also not build there?

I guess we should release 4.17.2.  Since the master branch is still in a good state, I'll let @hallyn release that one.

I'll merge this patch now.  Thanks!

kanavin:
> BTW, is this a regression compared to 4.16? Or did 4.16 also not build there?

4.16 was fine. I think the problem appeared here:
https://github.com/shadow-maint/shadow/commit/060b0849a6433a17685cf73b1816603b00657f95

alejandro-colomar:
> > BTW, is this a regression compared to 4.16? Or did 4.16 also not build there?
> 
> 4.16 was fine. I think the problem appeared here: [060b084](https://github.com/shadow-maint/shadow/commit/060b0849a6433a17685cf73b1816603b00657f95)

Ahh, yep, the others were covered in version checks already.  Thanks!

PR #1175, opened 2025-01-08
Man patches

zeha:


alejandro-colomar:
I picked a patch from here into <https://github.com/shadow-maint/shadow/pull/1186>.

man/chfn.1.xml
```diff
@@ -80,7 +81,10 @@
       numbers. The <emphasis remap='I'>other</emphasis> field is used to
       store accounting information used by other applications.
     </para>
-
+    <para>
+      If <replaceable>LOGIN</replaceable> is not specified, <command>chfn</command>
+      prompts for the current user account.
```
ikerexxe, line 86: I can't use chfn from shadow as my distribution uses util-linux, but I have a doubt. Does it prompt for the user account? Or does it assume the current user account?

alejandro-colomar: It seems to me that it assumes the current user:

```sh
alx@devuan:~$ which chfn | xargs dpkg -S | cut -f1 -d:
passwd
```
```sh
alx@devuan:~$ chfn
Password: 
chfn: PAM: Authentication failure
```

hallyn: @zeha do you mind updating this to reflect that?

alejandro-colomar: I can do it.  @zeha is just forwarding the diffs that other people proposed to Debian, and I promised to transform them into proper patches.

hallyn (approved):

PR #1177, opened 2025-01-08
Update French translation

zeha:
Bug-Debian: https://bugs.debian.org/1076260

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -1,7 +1,7 @@
 # translation of shadow-man.po to French
 # French translation of the shadow's man pages
 # Traduction des pages de manuel livrées avec shadow
-# Copyright (C) 2011-2022  Debian French l10n team <debian-l10n-french@lists.debian.org>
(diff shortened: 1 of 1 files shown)


alejandro-colomar (approved):

po/fr.po
```diff
@@ -325,7 +320,7 @@ msgstr "trop court"
 
 #, c-format
 msgid "Bad password: %s.  "
-msgstr "Mot de passe erroné : %s."
+msgstr "Mot de passe erroné : %s. "
```
alejandro-colomar, line 323: A minor comment here:

The correct inter-sentence space ammount is two.

<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/patches/description#n44>
<https://web.archive.org/web/20171217060354/http://www.heracliteanriver.com/?p=324>

Anyway, since you're not the author, let's take the patch as is.

PR #1181, opened 2025-01-11
Revert "etc/login.defs: enable CREATE_HOME"

hallyn:
This reverts commit dcb6b0f4152c1507a137640fea42401a91f22e2c.

### DIFF
--- a/etc/login.defs
+++ b/etc/login.defs
@@ -440,7 +440,7 @@ USERGROUPS_ENAB yes
 # This option is overridden with the -M or -m flags on the useradd(8)
 # command-line.
 #
-CREATE_HOME     yes
+#CREATE_HOME     yes
 
 #
 # Force use shadow, even if shadow passwd & shadow group files are
--- a/share/ansible/roles/ci_run/tasks/alpine.yml
+++ b/share/ansible/roles/ci_run/tasks/alpine.yml
@@ -68,6 +68,6 @@
 
 - name: Copy configuration file for testing
   ansible.builtin.copy:
-    src: /usr/local/src/shadow/etc/login.defs
+    src: /usr/local/src/shadow/tests/system/etc/login.defs
     dest: /etc/login.defs
     remote_src: yes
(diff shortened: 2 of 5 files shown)


hallyn:
I *suspect* this is not a big deal, and most distributions just ship their own version verbatim like debian/login.defs.  But if there *is* a distro - or even a person - using this as is from upstream, then we dont' want to break them.  So let's undo this and use an etc/login.defs.test for the testing if needed.

etc/login.defs.test
```diff
@@ -0,0 +1,473 @@
+#
+# /etc/login.defs - Configuration control definitions for the shadow package.
+#
+#	$Id$
+#
```
alejandro-colomar, line 5: Maybe we should have a tests/etc/ subdirectory?  I'm not very happy mixing actual stuff and tests.

alejandro-colomar:
For commit 1:

```
```

Thanks!

etc/login.defs.test
ikerexxe: Yes, I'd also prefer to ship it under the tests folder. What about `tests/system/config/login.defs`? It's highly unlikely that this file will have any use in bash system tests and unit tests. In addition, we can use the folder to store any configuration file needed for python system tests.

alejandro-colomar: > Yes, I'd also prefer to ship it under the tests folder. What about `tests/system/config/login.defs`?

The usual spelling of .../system/config/ is `.../etc/`.  I prefer etc for consistency with the FHS.
<https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html>

In fact, GNU coding standards say that build systems should use the `$sysconfdir` variable, and define it to `$(prefix)/etc`.
<https://www.gnu.org/prep/standards/html_node/Directory-Variables.html#Directory-Variables>

ikerexxe: In this case system refers to system tests not system config. I'm fine with calling it `tests/system/etc/login.defs`

ikerexxe (approved):
Looks good to me now

alejandro-colomar (approved):

alejandro-colomar:
Thanks!  LGTM too.

PR #1182, opened 2025-01-12
man/: Install suauth.5 only if feature exists

stoeckmann:
Support for `/etc/suauth` only exists if su is installed without PAM support. If su is not installed (--without-su) or if PAM support is enabled (default), do not install suauth.5 manual page.

### DIFF
--- a/man/Makefile.am
+++ b/man/Makefile.am
@@ -39,7 +39,6 @@ man_MANS = \
 	man1/sg.1 \
 	man3/shadow.3 \
 	man5/shadow.5 \
-	man5/suauth.5 \
 	man8/useradd.8 \
 	man8/userdel.8 \
 	man8/usermod.8 \
@@ -57,6 +56,7 @@ man_nopam = \
 
 if WITH_SU
 man_MANS += man1/su.1
+man_nopam += man5/suauth.5
 endif
 
 if !USE_PAM


alejandro-colomar:
> Support for `/etc/suauth` only exists if su is installed without PAM support.

Could you please show where in the source code that happens?

alejandro-colomar:
> > Could you please show where in the source code that happens?
> 
> Here: https://github.com/shadow-maint/shadow/blob/master/src/su.c#L558-L581 Because of: https://github.com/shadow-maint/shadow/blob/master/configure.ac#L628-L654

Thanks!  Would you mind adding those links to the commit message?

In any case:

```
```

man/Makefile.am
```diff
@@ -57,6 +56,7 @@ man_nopam = \
 
 if WITH_SU
 man_MANS += man1/su.1
+man_nopam += man5/suauth.5
 endif
```
alejandro-colomar, line 60: This doesn't seem to remove the page if WITH_SU but !USE_PAM, right?

stoeckmann: That's the point. If you have su without PAM, /etc/suauth is parsed.

alejandro-colomar: Oops, I probably meant WITH_SU and USE_PAM.

alejandro-colomar:
Since there have been no comments, I'll merge.  Thanks!

PR #1183, opened 2025-01-12
src/: Make line number overflows less likely

stoeckmann:
First pull request based on current work of refactoring login_nopam.c:

Huge files could trigger signed integer overflows if enough lines are within the file. Use intmax_t which is at least 64 bit to move this event far into the future. I don't see a need to handle THAT case in any more effort.

Also turn a variable from unsigned int to size_t just to make sure that even systems with huge amount of memory cannot trigger an out of boundary access.

### DIFF
--- a/src/chgpasswd.c
+++ b/src/chgpasswd.c
@@ -14,6 +14,7 @@
 #include <fcntl.h>
 #include <getopt.h>
 #include <pwd.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 
@@ -426,7 +427,7 @@ int main (int argc, char **argv)
 	const struct group *gr;
 	struct group newgr;
 	int errors = 0;
-	int line = 0;
+	intmax_t line = 0;
 
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
@@ -463,7 +464,7 @@ int main (int argc, char **argv)
 	while (fgets (buf, (int) sizeof buf, stdin) != NULL) {
 		line++;
 		if (stpsep(buf, "\n") == NULL) {
-			fprintf (stderr, _("%s: line %d: line too long\n"),
+			fprintf (stderr, _("%s: line %jd: line too long\n"),
 			         Prog, line);
 			errors++;
 			continue;
@@ -482,7 +483,7 @@ int main (int argc, char **argv)
 		cp = stpsep(name, ":");
 		if (cp == NULL) {
 			fprintf (stderr,
-			         _("%s: line %d: missing new password\n"),
+			         _("%s: line %jd: missing new password\n"),
 			         Prog, line);
 			errors++;
 			continue;
@@ -533,7 +534,7 @@ int main (int argc, char **argv)
 		gr = gr_locate (name);
 		if (NULL == gr) {
 			fprintf (stderr,
-			         _("%s: line %d: group '%s' does not exist\n"), Prog,
+			         _("%s: line %jd: group '%s' does not exist\n"), Prog,
 			         line, name);
 			errors++;
 			continue;
@@ -593,7 +594,7 @@ int main (int argc, char **argv)
 		if (NULL != sg) {
 			if (sgr_update (&newsg) == 0) {
 				fprintf (stderr,
-				         _("%s: line %d: failed to prepare the new %s entry '%s'\n"),
+				         _("%s: line %jd: failed to prepare the new %s entry '%s'\n"),
 				         Prog, line, sgr_dbname (), newsg.sg_name);
 				errors++;
 				continue;
@@ -605,7 +606,7 @@ int main (int argc, char **argv)
 		{
 			if (gr_update (&newgr) == 0) {
 				fprintf (stderr,
-				         _("%s: line %d: failed to prepare the new %s entry '%s'\n"),
+				         _("%s: line %jd: failed to prepare the new %s entry '%s'\n"),
 				         Prog, line, gr_dbname (), newgr.gr_name);
 				errors++;
 				continue;
--- a/src/newusers.c
+++ b/src/newusers.c
@@ -28,6 +28,7 @@
 #include <getopt.h>
 #include <ctype.h>
 #include <errno.h>
+#include <stddef.h>
 #include <stdint.h>
 #include <string.h>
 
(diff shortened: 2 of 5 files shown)


src/chgpasswd.c
```diff
@@ -426,7 +426,7 @@ int main (int argc, char **argv)
 	const struct group *gr;
 	struct group newgr;
 	int errors = 0;
-	int line = 0;
+	unsigned int line = 0;
```
alejandro-colomar, line 429: I prefer signed integers.

The bad thing about unsigned is that wrapping is legal, which means that if we accidentally wrap, a static analyzer is unable to warn us.  On the other hand, if we overflow integers, a static analyzer might let us know.

Instead, I suggest using a `long`.

src/newusers.c
```diff
@@ -1333,7 +1333,7 @@ int main (int argc, char **argv)
 	sssd_flush_cache (SSSD_DB_PASSWD | SSSD_DB_GROUP);
 
 #ifdef USE_PAM
-	unsigned int i;
+	size_t i;
 	/* Now update the passwords using PAM */
 	for (i = 0; i < nusers; i++) {
```
alejandro-colomar, line 1338: While you're at it, it might be interesting to reduce the scope of `i`, by using a C99 loop variable declaration

alejandro-colomar: I like the second patch.

For "src/newusers.c: Turn nusers into size_t":

```
```

Thanks!

src/chgpasswd.c
stoeckmann: For me, it happens rarely that I count up to negative numbers when I start with 0. ;)

Using long has 0 effect on a 32 bit system, since it's 32 bit. Staying with a signed data type fixes nothing here and your hypothetical static analyzer is broken if it doesn't keep complaining about a possible overflow here.

Alternative would be to use the `addsl` function to prevent the overflow, but then the debug output is still incorrect, since it stops counting upwards instead of wrapping around. And just using an unsigned data type has less overhead effect compared to adding overflow checks.

stoeckmann: You could use pipes instead of files, avoiding any such boundaries. `chgpasswd` for example reads from stdin, so it could be `yes | chgpasswd`.

Since these "counters are too small" scenarios are almost just theoretical (but proofable), we are quite academic here. And that's why I prefer the easiest/fastest solution.

alejandro-colomar: Hmmmm, then let's use `long long`.

lslebodn: I agree with @stoeckmann we are quite academic here.

but in case of `long long` it would be better to use `unsigned long long`.  It does not make a sense to reserve half of range for negative numbers to to check for -1.

Or even better one could use `uintmax_t` which is part of c99 standard.
If one would like to be safe we can still add a check for `line < UINTMAX_MAX` before incrementing the counter :-)

alejandro-colomar: > I agree with @stoeckmann we are quite academic here.
> 
> but in case of `long long` it would be better to use `unsigned long long`. It does not make a sense to reserve half of range for negative numbers to to check for -1.

I follow Google's C++ style guide regarding unsigned integers:

<https://google.github.io/styleguide/cppguide.html#Integer_Types>

```
    On Unsigned Integers

Unsigned integers are good for representing bitfields and modular arithmetic.
Because of historical accident, the C++ standard also uses unsigned integers
to represent the size of containers - many members of the standards body
believe this to be a mistake, but it is effectively impossible to fix at this point.
The fact that unsigned arithmetic doesn't model the behavior of a simple
integer, but is instead defined by the standard to model modular arithmetic
(wrapping around on overflow/underflow), means that a significant class of
bugs cannot be diagnosed by the compiler. In other cases, the defined
behavior impedes optimization.

That said, mixing signedness of integer types is responsible for an equally
large class of problems. The best advice we can provide: try to use iterators
and containers rather than pointers and sizes, try not to mix signedness, and
try to avoid unsigned types (except for representing bitfields or modular
arithmetic). Do not use an unsigned type merely to assert that a variable is
non-negative.
```

I don't have in mind an example of bugs that become silent with unsigned types.
If I remember eventually, I'll try to document it here.  I remember having discussed
some of those in the past (not here).

> Or even better one could use `uintmax_t` which is part of c99 standard.

I'm indiferent to intmax_t vs long long.  intmax_t is shorter to type, so yeah, we could you it.

> If one would like to be safe we can still add a safe check for `line < UINTMAX_MAX` before incrementing the counter :-)

Nah, the good thing about `long long` is that you can safely assume you'll never overflow it.  At one line per nanosecond (and we're quite slower), it would take in the order of 1k years to overflow.

alejandro-colomar: Cc: @eggert 

Maybe Paul has an example of this.

stoeckmann: Too much chatter about this nothing-burger:
- long
- addsl

If there's more, feel free to edit this pull request.

eggert: > Cc: @eggert
> 
> Maybe Paul has an example of this.

Not sure how I got dragged into this, but the Google advice is good advice, and we independently derived it for the GNU core utilities and similar projects: don't use unsigned, size_t, etc. unless forced to, and here you're not forced to.

I also agree there's no practical need to check for intmax_t overflow if you're just adding 1 at a time.

stoeckmann: Adjusted. intmax_t, no check. I am rather confident that @alejandro-colomar was actually refering to a dynamic and not a static analyzer in first comment. In that case we can simply adjust the data type, since dynamically it won't happen. But that's what they pretty much said about 640 KB as well.

src/newusers.c
stoeckmann: Adjusted the scope.

alejandro-colomar (approved):
Thanks!  LGTM.

```
```

src/newusers.c
```diff
@@ -1070,7 +1070,7 @@ int main (int argc, char **argv)
 	intmax_t *lines = NULL;
 	char **usernames = NULL;
 	char **passwords = NULL;
-	unsigned int nusers = 0;
+	size_t nusers = 0;
```
alejandro-colomar, line 1073: Please `#include <stddef.h>` for using `size_t` (IWYU guidelines).

alejandro-colomar:
No objections have been raised, and it's quite simple.  I'll merge now.  Thanks!

PR #1184, opened 2025-01-14
chage: Drop PAM support

stoeckmann:
The PAM support was only enabled with configure option `--enable-account-tools-setuid`. The other account tools would use PAM then to verify that the user is granted elevated permissions for actions which normally only root can do.

In chage, however, any non-root user who does not specify the `-l` command line option is denied access in `check_perms`. The check for being root or not is done with `getuid`, so non-root users cannot change user account's aging information in any possible way since more than 18 years by now.

It's safe to say that nobody misses this non-existing feature. Biggest benefit is to get chage out of the ACCT_TOOLS_SETUID group of tools.


alejandro-colomar (approved):
I'm all for removing stuff.  Thanks!

```
```

ikerexxe (commented):
Thank you for taking care of this. I wanted to do it myself a few months ago but I didn't have the time.

PR #1185, opened 2025-01-14
man/passwd.1.xml: -P disables PAM support

stoeckmann:
If passwd is called with -P, then PAM handling is disabled ([src/passwd.c line 749](https://github.com/shadow-maint/shadow/blob/master/src/passwd.c#L746-L751)). The manual page claims that host files would be used, which is not true.

### DIFF
--- a/man/passwd.1.xml
+++ b/man/passwd.1.xml
@@ -292,8 +292,7 @@
 	    found under the directory <replaceable>PREFIX_DIR</replaceable>.
 	    This option does not chroot and is intended for preparing a cross-compilation
 	    target.  Some limitations: NIS and LDAP users/groups are
-	    not verified.  PAM authentication is using the host files.
-	    No SELINUX support.
+	    not verified.  No PAM support.  No SELINUX support.
 	  </para>
 	</listitem>
       </varlistentry>


alejandro-colomar (approved):
LGTM.  Thanks!

```
```

man/passwd.1.xml
```diff
@@ -292,8 +292,7 @@
 	    found under the directory <replaceable>PREFIX_DIR</replaceable>.
 	    This option does not chroot and is intended for preparing a cross-compilation
 	    target.  Some limitations: NIS and LDAP users/groups are
-	    not verified.  PAM authentication is using the host files.
-	    No SELINUX support.
+	    not verified.  No PAM support.  No SELINUX support.
```
alejandro-colomar, line 295: The text seems to have been introduced in

```
01f6258df744 (2023-10-20, 2023-10-26; "man: document --prefix option in chage, chpasswd and passwd")
```
01f6258df744

Cc: @jubalh

alejandro-colomar: Please add:

```
Fixes: 01f6258df744 (2023-10-26; "man: document --prefix option in chage, chpasswd and passwd")
```

BTW, @jubalh , do you have any comments?  Was it a mistake in the manual?  Or something else?

alejandro-colomar: Since that patch documented another patch from a different author, let's CC that one too:

Cc: @Dzejrou

alejandro-colomar: And people that participated in the PRs:

@ferivoz , @ancorgs, @hallyn , @ikerexxe

hallyn: I would guess it was looking at a recent release, given the manpage was patched in October and the pam behavior was changed in May.

ferivoz: Buenas @alejandro-colomar! I think that PAM and SELINUX together with root and prefix flags could need some rework. I have stopped in the past because I'm not an SELINUX expert, but it seems that it's never disabled with prefix. The manual page entries are probably copy & pasted without adjustments.

What is the purpose of using PAM with host files when handling prefixes anyway? Should it be root-only just like root prefix requires it? From what I understand, it was introduced for installation media or preparing another root filesystem or image.

alejandro-colomar: Buenas @ferivoz!

> Buenas @alejandro-colomar! I think that PAM and SELINUX together with root and prefix flags could need some rework. I have stopped in the past because I'm not an SELINUX expert, but it seems that it's never disabled with prefix. The manual page entries are probably copy & pasted without adjustments.

Most likely, it seems; just wanted some confirmation.  Thanks!

> What is the purpose of using PAM with host files when handling prefixes anyway? Should it be root-only just like root prefix requires it? From what I understand, it was introduced for installation media or preparing another root filesystem or image.

I'm even less of an expert.  :-)

hallyn: @ferivoz in fact, at least passwd, chage and chpasswd have been modified to explicitly not support PAM with prefix, which, we all appear to agree, is nonsense, since the local PAM presumably knows nothing about the prefixed root environment.

PR #1193, opened 2025-01-20
lib/, src/: Turn error counters into flags

stoeckmann:
If we are not interested in the amount of errors but only if errors exist, use a flag instead of a counter. This eliminates the chance of signed integer overflows and better reflects the meaning of variable.

Keeping variable name and basically copied from src/faillog.c.

### DIFF
--- a/src/userdel.c
+++ b/src/userdel.c
@@ -119,7 +119,7 @@ static void user_cancel (const char *);
 static bool path_prefix (const char *, const char *);
 #endif				/* EXTRA_CHECK_HOME_DIR */
 static int is_owner (uid_t, const char *);
-static int remove_mailbox (void);
+static bool remove_mailbox (void);
 #ifdef WITH_TCB
 static int remove_tcbdir (const char *user_name, uid_t user_id);
 #endif				/* WITH_TCB */
@@ -789,9 +789,10 @@ static int is_owner (uid_t uid, const char *path)
 	return (st.st_uid == uid) ? 1 : 0;
 }
 
-static int remove_mailbox (void)
+static bool remove_mailbox (void)
 {
-	int         i, errors = 0;
+	int         i;
+	bool        errors = false;
 	char        *mailfile;
 	const char  *maildir;
 
@@ -844,7 +845,7 @@ static int remove_mailbox (void)
 			              "deleting mail file",
 			              user_name, user_id, SHADOW_AUDIT_FAILURE);
 #endif				/* WITH_AUDIT */
-			errors = 1;
+			errors = true;
 			/* continue */
 		}
 #ifdef WITH_AUDIT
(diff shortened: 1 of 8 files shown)


alejandro-colomar (approved):
Thanks!

```
```

src/userdel.c
```diff
@@ -1152,7 +1153,7 @@ int main (int argc, char **argv)
 	update_groups ();
 
 	if (rflg) {
-		errors += remove_mailbox ();
+		errors = remove_mailbox () && errors;
```
alejandro-colomar, line 1156: This should be `||`, right?

src/userdel.c
```diff
@@ -1273,13 +1274,13 @@ int main (int argc, char **argv)
 	}
 
 #ifdef WITH_TCB
-	errors += remove_tcbdir (user_name, user_id);
+	errors = remove_tcbdir (user_name, user_id) && errors;
```
alejandro-colomar, line 1277: Same here.  (`||`)

src/userdel.c
stoeckmann: Oh, right. Thanks for the review! It's supposed to preserve true, not false.

src/su.c
```diff
@@ -313,6 +313,7 @@ static void prepare_pam_close_session (void)
 	}
 
 	/* parent only */
+	closelog();
```
alejandro-colomar, line 316: ?

src/userdel.c
```diff
@@ -1152,7 +1153,7 @@ int main (int argc, char **argv)
 	update_groups ();
 
 	if (rflg) {
-		errors += remove_mailbox ();
+		errors = remove_mailbox () || errors;
```
alejandro-colomar, line 1156: Should we do `|=`?  It should work as good as a hypothetical `||=`, I guess.

stoeckmann: I'd prefer to stay with boolean operations here just to clarify what's going on. Heh, it helped with finding the logic bug I introduced, right?

src/su.c
stoeckmann: Removed again. Sorry, had work in progress changes in tree.

alejandro-colomar: Yup, I guessed that.  :)

src/userdel.c
alejandro-colomar: On the one hand yes, it helped see the bug.  I honestly don't know if I would have catched it as `&=`.

On the other hand, if you had written `errors = errors || remove_mailbox();` I wouldn't probably have noticed that we're short-cutting the call, and so `|=` might have been safer in that regard.

I don't know what to do.  What do you think?  I guess I'll take whatever you prefer, since you're the author.

alejandro-colomar: Cc: @hallyn

alejandro-colomar (approved):
```
```

(but please reply to the still-open dialog about `|=` vs `||`.)

src/userdel.c
alejandro-colomar: Hmmm, sounds good.  Thanks!

hallyn (approved):
Thanks guys, indeed, let's always aim for readability over cleverness or optimizations :)

PR #1199, opened 2025-01-27
Audit update

ikerexxe:
This is a set of changes that Fedora introduced in a downstream patch. The idea is to converge upstream and downstream versions to reduce the maintenance burden. The patch was created a long time ago and we have been adapting it, but it may not be up to date with current upstream code standards, if so, please inform me so I can update the patch.

### DIFF
--- a/lib/audit_help.c
+++ b/lib/audit_help.c
@@ -25,6 +25,8 @@
 #include "attr.h"
 #include "prototypes.h"
 #include "shadowlog.h"
+#include "string/sprintf/snprintf.h"
+
 int audit_fd;
 
 void audit_help_open (void)
@@ -46,10 +48,14 @@ void audit_help_open (void)
 
 /*
  * This function will log a message to the audit system using a predefined
- * message format. Parameter usage is as follows:
+ * message format. For additional information on the user account lifecycle
+ * events check
+ * <https://github.com/linux-audit/audit-documentation/wiki/SPEC-User-Account-Lifecycle-Events>
  *
- * type - type of message: AUDIT_USER_CHAUTHTOK for changing any account
- *	  attributes.
+ * Parameter usage is as follows:
+ *
+ * type - type of message. A list of possible values is available in
+ *        "audit-records.h" file.
  * pgname - program's name
  * op  -  operation. "adding user", "changing finger info", "deleting group"
  * name - user's account or group name. If not available use NULL.
@@ -66,6 +72,47 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
 		audit_log_acct_message (audit_fd, type, NULL, op, name, id,
 		                        NULL, NULL, NULL, result);
 	}
+}
+
+/*
+ * This function will log a message to the audit system using a predefined
+ * message format. For additional information on the group account lifecycle
+ * events check
+ * <https://github.com/linux-audit/audit-documentation/wiki/SPEC-User-Account-Lifecycle-Events>
+ *
+ * Parameter usage is as follows:
+ *
+ * type - type of message. A list of possible values is available in
+ *        "audit-records.h" file.
+ * op  -  operation. "adding-user", "modify-group", "deleting-user-from-group"
+ * name - user's account or group name. If not available use NULL.
+ * id  -  uid or gid that the operation is being performed on. This is used
+ *	  only when user is NULL.
+ * grp_type - type of group: "grp" or "new_group"
+ * grp - group name associated with event
+ */
+void
+audit_logger_with_group(int type, const char *op, const char *name,
+    id_t id, const char *grp_type, const char *grp,
+    shadow_audit_result result)
+{
+	int len;
+	char enc_group[GROUP_NAME_MAX_LENGTH * 2 + 1];
+	char buf[NITEMS(enc_group) + 100];
+
+	if (audit_fd < 0)
+		return;
+
+	len = strnlen(grp, sizeof(enc_group)/2);
+	if (audit_value_needs_encoding(grp, len)) {
+		SNPRINTF(buf, "%s %s=%s", op, grp_type,
+			audit_encode_value(enc_group, grp, len));
+	} else {
+		SNPRINTF(buf, "%s %s=\"%s\"", op, grp_type, grp);
+	}
+
+	audit_log_acct_message(audit_fd, type, NULL, buf, name, id,
+		               NULL, NULL, NULL, result);
 }
 
 void audit_logger_message (const char *message, shadow_audit_result result)
--- a/lib/cleanup_group.c
+++ b/lib/cleanup_group.c
@@ -62,7 +62,7 @@ void cleanup_report_mod_group (void *cleanup_info)
 	         gr_dbname (),
 	         info->action));
 #ifdef WITH_AUDIT
-	audit_logger (AUDIT_USER_ACCT, log_get_progname(),
+	audit_logger (AUDIT_GRP_MGMT, log_get_progname(),
 	              info->audit_msg,
 	              info->name, AUDIT_NO_ID,
 	              SHADOW_AUDIT_FAILURE);
@@ -80,7 +80,7 @@ void cleanup_report_mod_gshadow (void *cleanup_info)
 	         sgr_dbname (),
 	         info->action));
 #ifdef WITH_AUDIT
-	audit_logger (AUDIT_USER_ACCT, log_get_progname(),
+	audit_logger (AUDIT_GRP_MGMT, log_get_progname(),
 	              info->audit_msg,
 	              info->name, AUDIT_NO_ID,
 	              SHADOW_AUDIT_FAILURE);
@@ -101,7 +101,7 @@ void cleanup_report_add_group_group (void *group_name)
 	SYSLOG ((LOG_ERR, "failed to add group %s to %s", name, gr_dbname ()));
 #ifdef WITH_AUDIT
 	audit_logger (AUDIT_ADD_GROUP, log_get_progname(),
-	              "adding group to /etc/group",
+	              "adding-group",
 	              name, AUDIT_NO_ID,
 	              SHADOW_AUDIT_FAILURE);
 #endif
@@ -120,8 +120,8 @@ void cleanup_report_add_group_gshadow (void *group_name)
 
 	SYSLOG ((LOG_ERR, "failed to add group %s to %s", name, sgr_dbname ()));
 #ifdef WITH_AUDIT
-	audit_logger (AUDIT_ADD_GROUP, log_get_progname(),
-	              "adding group to /etc/gshadow",
+	audit_logger (AUDIT_GRP_MGMT, log_get_progname(),
+	              "adding-shadow-group",
 	              name, AUDIT_NO_ID,
 	              SHADOW_AUDIT_FAILURE);
 #endif
@@ -143,8 +143,8 @@ void cleanup_report_del_group_group (void *group_name)
 	         "failed to remove group %s from %s",
 	         name, gr_dbname ()));
 #ifdef WITH_AUDIT
-	audit_logger (AUDIT_ADD_GROUP, log_get_progname(),
-	              "removing group from /etc/group",
+	audit_logger (AUDIT_DEL_GROUP, log_get_progname(),
+	              "removing-group",
 	              name, AUDIT_NO_ID,
 	              SHADOW_AUDIT_FAILURE);
 #endif
@@ -166,8 +166,8 @@ void cleanup_report_del_group_gshadow (void *group_name)
 	         "failed to remove group %s from %s",
 	         name, sgr_dbname ()));
 #ifdef WITH_AUDIT
-	audit_logger (AUDIT_ADD_GROUP, log_get_progname(),
-	              "removing group from /etc/gshadow",
+	audit_logger (AUDIT_GRP_MGMT, log_get_progname(),
+	              "removing-shadow-group",
 	              name, AUDIT_NO_ID,
 	              SHADOW_AUDIT_FAILURE);
 #endif
@@ -187,7 +187,7 @@ void cleanup_unlock_group (MAYBE_UNUSED void *arg)
 		         log_get_progname(), gr_dbname ());
 		SYSLOG ((LOG_ERR, "failed to unlock %s", gr_dbname ()));
 #ifdef WITH_AUDIT
-		audit_logger_message ("unlocking group file",
+		audit_logger_message ("unlocking-group",
 		                      SHADOW_AUDIT_FAILURE);
 #endif
 	}
@@ -207,7 +207,7 @@ void cleanup_unlock_gshadow (MAYBE_UNUSED void *arg)
 		         log_get_progname(), sgr_dbname ());
 		SYSLOG ((LOG_ERR, "failed to unlock %s", sgr_dbname ()));
 #ifdef WITH_AUDIT
-		audit_logger_message ("unlocking gshadow file",
+		audit_logger_message ("unlocking-gshadow",
 		                      SHADOW_AUDIT_FAILURE);
 #endif
 	}
--- a/lib/prototypes.h
+++ b/lib/prototypes.h
@@ -189,6 +189,9 @@ extern void audit_logger (int type, const char *pgname, const char *op,
                           const char *name, unsigned int id,
                           shadow_audit_result result);
 void audit_logger_message (const char *message, shadow_audit_result result);
+void audit_logger_with_group(int type, const char *op, const char *name,
+    id_t id, const char *grp_type, const char *grp,
+    shadow_audit_result result);
 #endif
 
 /* limits.c */
--- a/src/gpasswd.c
+++ b/src/gpasswd.c
@@ -382,20 +382,14 @@ static void open_files (void)
 
 static void log_gpasswd_failure (const char *suffix)
 {
-#ifdef WITH_AUDIT
-	char  buf[1024];
-#endif
-
 	if (aflg) {
 		SYSLOG ((LOG_ERR,
 		         "%s failed to add user %s to group %s%s",
 		         myname, user, group, suffix));
 #ifdef WITH_AUDIT
-		SNPRINTF(buf, "%s failed to add user %s to group %s%s",
-		         myname, user, group, suffix);
-		audit_logger (AUDIT_USER_ACCT, Prog,
-		              buf,
-		              group, AUDIT_NO_ID,
+		audit_logger_with_group (AUDIT_USER_MGMT,
+		              "add-user-to-group",
+		              user, AUDIT_NO_ID, "grp", group,
 		              SHADOW_AUDIT_FAILURE);
 #endif
 	} else if (dflg) {
@@ -403,11 +397,9 @@ static void log_gpasswd_failure (const char *suffix)
 		         "%s failed to remove user %s from group %s%s",
 		         myname, user, group, suffix));
 #ifdef WITH_AUDIT
-		SNPRINTF(buf, "%s failed to remove user %s from group %s%s",
-		         myname, user, group, suffix);
-		audit_logger (AUDIT_USER_ACCT, Prog,
-		              buf,
-		              group, AUDIT_NO_ID,
+		audit_logger_with_group (AUDIT_USER_MGMT,
+		              "delete-user-from-group",
+		              user, AUDIT_NO_ID, "grp", group,
 		              SHADOW_AUDIT_FAILURE);
 #endif
 	} else if (rflg) {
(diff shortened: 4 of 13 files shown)


ikerexxe:
@stevegrubb before I take this out of draft status, are there any other distributions besides Fedora/CentOS/RHEL that might be affected by these changes? I'd like to ping the maintainers of those distributions to review these code changes.

lib/audit_help.c
```diff
@@ -68,6 +68,39 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
...
+	char enc_group[(GROUP_NAME_MAX_LENGTH*2)+1], buf[1024];
+	if (audit_fd < 0) {
+		return;
+	}
+	len = strnlen(grp, sizeof(enc_group)/2);
+	if (audit_value_needs_encoding(grp, len)) {
+		snprintf(buf, sizeof(buf), "%s grp=%s", op,
+			audit_encode_value(enc_group, grp, len));
+	} else {
+		snprintf(buf, sizeof(buf), "%s grp=\"%s\"", op, grp);
```
alejandro-colomar, line 98: These should call SNPRINTF().

```c
$ grepc SNPRINTF .
./lib/string/sprintf/snprintf.h:#define SNPRINTF(s, fmt, ...)                                                 \
	snprintf_(s, NITEMS(s), fmt __VA_OPT__(,) __VA_ARGS__)                \
)
```

lib/audit_help.c
```diff
@@ -68,6 +68,39 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
...
+ * id  -  uid or gid that the operation is being performed on. This is used
+ *	  only when user is NULL.
+ * grp - group name associated with event
+ */
+void audit_logger_with_group (int type, MAYBE_UNUSED const char *pgname,
+		const char *op, const char *name, unsigned int id,
+		const char *grp, shadow_audit_result result)
+{
+	int len;
+	char enc_group[(GROUP_NAME_MAX_LENGTH*2)+1], buf[1024];
```
alejandro-colomar, line 89: That 1024 seems a bit arbitrary.  Could it maybe be less than sizeof(enc_group), which would be problematic?


Also,May we should do this?:

```c
char  enc_group[GROUP_NAME_MAX_LENGTH * 2 + 1];
char  buf[NITEMS(enc_group) + some value];  // How much would be enough?  Maybe 100?

```

(Also, the parentheses are redundant.)

lib/audit_help.c
```diff
@@ -68,6 +68,39 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
...
+	if (audit_fd < 0) {
+		return;
+	}
+	len = strnlen(grp, sizeof(enc_group)/2);
+	if (audit_value_needs_encoding(grp, len)) {
```
alejandro-colomar, line 94: By default, we don't allow names that would need encoding.  This reminds me that we have <https://github.com/shadow-maint/shadow/pull/1158>.

I will add `"` to the list of strictly disallowed characters, and then we don't need to encode anything here.

lib/prototypes.h
```diff
@@ -182,12 +182,21 @@ extern int audit_fd;
 extern void audit_help_open (void);
 /* Use AUDIT_NO_ID when a name is provided to audit_logger instead of an ID */
 #define AUDIT_NO_ID	((unsigned int) -1)
+#ifndef AUDIT_GRP_MGMT
+#define AUDIT_GRP_MGMT          1132    /* Group account was modified */
+#endif
+#ifndef AUDIT_GRP_CHAUTHTOK
+#define AUDIT_GRP_CHAUTHTOK     1133    /* Group account password was changed */
+#endif
```
alejandro-colomar, line 190: Any reason for those specific values?

src/gpasswd.c
```diff
@@ -512,23 +494,19 @@ static void log_gpasswd_success (const char *suffix)
 		         "user %s added by %s to group %s%s",
 		         user, myname, group, suffix));
 #ifdef WITH_AUDIT
-		SNPRINTF(buf, "user %s added by %s to group %s%s",
-		         user, myname, group, suffix);
-		audit_logger (AUDIT_USER_ACCT, Prog,
-		              buf,
-		              group, AUDIT_NO_ID,
+		audit_logger_with_group (AUDIT_USER_MGMT, Prog,
+		              "add-user-to-group",
```
alejandro-colomar, line 498: Maybe we should use enums?  That would centralize an array with all the strings.

```c
enum audit_op {
    AUDIT_ADD_USER_TO_GROUP,
    ...
};

static const char audit_ops[][64] = {
    [.AUDIT_ADD_USER_TO_GROUP] = "add-user-to-group",
    ...
};
```

src/groupadd.c
```diff
@@ -120,6 +120,15 @@ usage (int status)
...
 
+static void fail_exit(int status)
+{
+#ifdef WITH_AUDIT
+	audit_logger(AUDIT_ADD_GROUP, Prog, "add-group", group_name,
+				 AUDIT_NO_ID, SHADOW_AUDIT_FAILURE);
+#endif
+	exit (status);
+}
+
```
alejandro-colomar, line 131: Would it maybe be interesting to register the audit_logger() call with atexit(3)?  That would avoid changing so many lines from exit() to fail_exit().

alejandro-colomar:
```
atexit(3)               Library Functions Manual              atexit(3)

NAME
     atexit - register a function to be called at normal process termi‐
     nation

LIBRARY

SYNOPSIS
     #include <stdlib.h>

     int atexit(typeof(void (void)) *function);

DESCRIPTION
     The atexit() function registers the given function to be called at
     normal  process termination, either via exit(3) or via return from
     the program’s main().  Functions so registered are called  in  the
     reverse order of their registration; no arguments are passed.
```

Then, on successful exit, we could call `fflush(NULL); _exit(EXIT_SUCCESS);`, to not call the handler.

alejandro-colomar:
```c
void
audit_atexit(void)
{
    audit_logger(AUDIT_ADD_GROUP, Prog, "add-group", group_name,
				 AUDIT_NO_ID, SHADOW_AUDIT_FAILURE);
}

atexit(&audit_atexit);
```

lib/prototypes.h
ikerexxe: They come from `audit-records.h` and probably they weren't available in all distributions when this patch was originally created (that happened years ago). I'm fine with removing this part of the patch as probably they will be available in all distributions.

src/groupadd.c
ikerexxe: The thing is that we don't always want to record an audit message when we exit this program, there are times when you just exit the program.

lib/audit_help.c
ikerexxe: Updated. Thanks! I suspected I must be missing something like this.

lib/audit_help.c
ikerexxe: Sounds good to me. I've updated the code accordingly.

src/gpasswd.c
alejandro-colomar: Okay, sounds reasonable.  Anyway, once can easily grep for all such strings, so it's ok.

lib/audit_help.c
alejandro-colomar: Yes, that's fine.  Thanks!

lib/cleanup_group.c
```diff
@@ -62,7 +62,7 @@ void cleanup_report_mod_group (void *cleanup_info)
 	         gr_dbname (),
 	         info->action));
 #ifdef WITH_AUDIT
-	audit_logger (AUDIT_USER_ACCT, log_get_progname(),
+	audit_logger (AUDIT_GRP_MGMT, log_get_progname(),
```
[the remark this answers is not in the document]
alejandro-colomar: > Regarding
> 
> [b2c8eff](https://github.com/shadow-maint/shadow/commit/b2c8effb0b51e1835ef105e9532474a8259cb0e4)
> 
> You could move the filename in the subject to the prefix:
> 
> ```
> lib/cleanup_group.c: Update audit messages
> ```

Or will you squash the commits?  (I think it would make sense to squash them.)

stevegrubb:
@ikerexxe Most major distributions enable auditing. So, they are all affected. That said, they are all broken except Fedora. This fixes everyone and Fedora can drop it's patch.

lib/cleanup_group.c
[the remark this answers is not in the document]
ikerexxe: mmm I'm fine either way. Which one do you prefer?

ikerexxe:
> @ikerexxe Most major distributions enable auditing. So, they are all affected. That said, they are all broken except Fedora. This fixes everyone and Fedora can drop it's patch.

Thank you for the clarification. I'm moving the PR out of draft state.

lib/cleanup_group.c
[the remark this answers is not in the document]
alejandro-colomar: I prefer squashing them.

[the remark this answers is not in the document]
ikerexxe: Done

alejandro-colomar, line 65: Could you please explain the change `AUDIT_USER_ACCT` => `AUDIT_GRP_MGMT` in the commit message?

lib/audit_help.c
```diff
@@ -48,7 +50,7 @@ void audit_help_open (void)
  * This function will log a message to the audit system using a predefined
  * message format. Parameter usage is as follows:
  *
- * type - type of message: AUDIT_USER_CHAUTHTOK for changing any account
+ * type - type of message: AUDIT_USER_MGMT for changing any account
  *	  attributes.
```
alejandro-colomar, line 54: Why do we only document this one type? Maybe we should actually point to some documentation that documents all of them, instead of documenting one and leaving the others undocumented?

lib/audit_help.c
```diff
@@ -68,6 +70,43 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
...
+		const char *op, const char *name, unsigned int id,
+		const char *grp, shadow_audit_result result)
+{
+	int len;
+	char enc_group[GROUP_NAME_MAX_LENGTH * 2 + 1];
+	char buf[NITEMS(enc_group) + 100];
+
+	if (audit_fd < 0) {
+		return;
+	}
```
alejandro-colomar, line 96: Please don't use braces for single-statement blocks.

lib/audit_help.c
```diff
@@ -68,6 +70,43 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
...
+	len = strnlen(grp, sizeof(enc_group)/2);
+	if (audit_value_needs_encoding(grp, len)) {
+		SNPRINTF(buf, "%s grp=%s", op,
+			audit_encode_value(enc_group, grp, len));
+	} else {
+		SNPRINTF(buf, "%s grp=\"%s\"", op, grp);
+	}
+
+	audit_log_acct_message (audit_fd, type, NULL, buf, name, id,
+		                        NULL, NULL, NULL, (int) result);
```
alejandro-colomar, line 107: Is this cast necessary?

lib/audit_help.c
stevegrubb: The documentation around the user and group account lifecycle wrt auditing is here: https://github.com/linux-audit/audit-documentation/wiki/SPEC-User-Account-Lifecycle-Events  This might help clarify some changes.

alejandro-colomar: I think we should point to that link, instead of arbitrarily documenting one of the types.

lib/cleanup_group.c
ikerexxe: Done

lib/audit_help.c
ikerexxe: I've added the link to the interfaces and to the commit message, so that it's easier to find

lib/audit_help.c
```diff
@@ -68,6 +70,43 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
...
+ * type - type of message: AUDIT_USER_MGMT for changing any account
+ *	  attributes.
+ * pgname - program's name
+ * op  -  operation. "adding user", "changing finger info", "deleting group"
+ * name - user's account or group name. If not available use NULL.
+ * id  -  uid or gid that the operation is being performed on. This is used
+ *	  only when user is NULL.
+ * grp - group name associated with event
+ */
+void audit_logger_with_group (int type, MAYBE_UNUSED const char *pgname,
```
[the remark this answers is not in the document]
ikerexxe: No, but I think that was a copy from `audit_logger()`. I just removed it for `audit_logger_with_group()`.

lib/audit_help.c
ikerexxe: Done

lib/audit_help.c
ikerexxe: No, not really, so I'm removing it

lib/audit_help.c
```diff
@@ -68,6 +74,44 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
...
 
+/*
+ * This function will log a message to the audit system using a predefined
+ * message format. For additional information on the group account lifecycle
+ * events check
+ * <https://github.com/linux-audit/audit-documentation/wiki/SPEC-User-Account-Lifecycle-Events>
+ *
+ * Parameter usage is as follows:
+ *
+ * type - type of message.A list of possible values is available in
```
alejandro-colomar, line 85: Missing space here.

lib/audit_help.c
```diff
@@ -68,6 +74,44 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
...
+	len = strnlen(grp, sizeof(enc_group)/2);
+	if (audit_value_needs_encoding(grp, len)) {
+		SNPRINTF(buf, "%s grp=%s", op,
+			audit_encode_value(enc_group, grp, len));
+	} else {
+		SNPRINTF(buf, "%s grp=\"%s\"", op, grp);
+	}
+
+	audit_log_acct_message (audit_fd, type, NULL, buf, name, id,
+		                        NULL, NULL, NULL, result);
```
alejandro-colomar, line 112: Please remove the space before the `(`.  Also, please align the continuation line correctly.

lib/audit_help.c
```diff
@@ -68,6 +74,44 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
...
+ * This function will log a message to the audit system using a predefined
+ * message format. For additional information on the group account lifecycle
+ * events check
+ * <https://github.com/linux-audit/audit-documentation/wiki/SPEC-User-Account-Lifecycle-Events>
+ *
+ * Parameter usage is as follows:
+ *
+ * type - type of message.A list of possible values is available in
+ *        "audit-records.h" file.
+ * op  -  operation. "adding user", "changing finger info", "deleting group"
```
alejandro-colomar, line 87: The strings you're actually passing don't have spaces.  Please revise this documentation.

src/newgrp.c
```diff
@@ -821,12 +826,20 @@ int main (int argc, char **argv)
...
+			SNPRINTF(audit_buf, "changing new_group=%s", group);
+		}
 		audit_logger (AUDIT_CHGRP_ID, Prog,
-		              audit_buf, NULL, getuid (), 0);
+		              audit_buf, NULL, getuid (), SHADOW_AUDIT_FAILURE);
 	} else {
 		audit_logger (AUDIT_CHGRP_ID, Prog,
-		              "changing", NULL, getuid (), 0);
+		              "changing", NULL, getuid (), SHADOW_AUDIT_FAILURE);
 	}
```
alejandro-colomar, line 843: This looks like open-coding the function you're adding in the second commit.  Can we encapsulate it in a function?  Why is this open-coded here?

lib/audit_help.c
ikerexxe: Updated

lib/audit_help.c
ikerexxe: Updated

lib/audit_help.c
ikerexxe: Yeah, and `changing finger info` doesn't even exist. Updated to provide real-world examples.

src/newgrp.c
ikerexxe: The original patch predates my introduction to the professional software world. I tried updating the code to the upstream standards, but I was unable to do it completely.

I have updated the three instances where this type of code was present to use `audit_logger_with_group()` instead. I had to add an additional argument to the function so that everything matched correctly. Take a look at the new code.

alejandro-colomar: Makes sense.  Thanks!

lib/audit_help.c
```diff
@@ -68,6 +74,47 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
...
+ * op  -  operation. "adding-user", "modify-group", "deleting-user-from-group"
+ * name - user's account or group name. If not available use NULL.
+ * id  -  uid or gid that the operation is being performed on. This is used
+ *	  only when user is NULL.
+ * grp_type - type of group: "grp" or "new_group"
+ * grp - group name associated with event
+ */
+void
+audit_logger_with_group(int type, const char *op, const char *name,
+    unsigned int id, const char *grp_type, const char *grp,
```
alejandro-colomar, line 96: I prefer using `id_t`, which is the right type for user and group id values.

src/gpasswd.c
```diff
@@ -127,7 +127,7 @@ static void usage (int status)
 	(void) fputs (_("  -d, --delete USER             remove USER from GROUP\n"), usageout);
 	(void) fputs (_("  -h, --help                    display this help message and exit\n"), usageout);
 	(void) fputs (_("  -Q, --root CHROOT_DIR         directory to chroot into\n"), usageout);
-	(void) fputs (_("  -r, --remove-password         remove the GROUP's password\n"), usageout);
+	(void) fputs (_("  -r, --delete-password         remove the GROUP's password\n"), usageout);
```
alejandro-colomar, line 130: I don't see justification for this name change in the commit message.  Can you please justify?

src/useradd.c
```diff
@@ -252,6 +252,8 @@ static FILE *fmkomstemp(char *template, unsigned int flags, mode_t m);
  */
 static void fail_exit (int code)
 {
+	int type;
```
alejandro-colomar, line 255: This declaration should be wrapped in ifdef, since it's used in ifdef code.

lib/audit_help.c
ikerexxe: Updated to use `id_t`

src/useradd.c
ikerexxe: Done

src/useradd.c
```diff
@@ -263,66 +267,46 @@ static void fail_exit (int code)
...
 
 #ifdef WITH_AUDIT
-	audit_logger(AUDIT_ADD_USER, Prog, "adding user",
+	if (code == E_PW_UPDATE || code >= E_GRP_UPDATE)
+		type = AUDIT_USER_MGMT;
+	else
+		type = AUDIT_ADD_USER;
+
+	audit_logger (type, Prog,
+	              "add-user",
```
alejandro-colomar, line 309: Maybe we could also use a ternary here to remove the variable.  Up to you; I don't feel strongly either way.

alejandro-colomar (approved):
I've made a couple of comments, but feel free to do what you like with them.  I'm happy enough with the changes as they are now, so:

```
```

src/useradd.c
alejandro-colomar:
```c
audit_logger ((code == E_PW_UPDATE || code >= E_GRP_UPDATE) ? AUDIT_USER_MGMT : AUDIT_ADD_USER,
              ...);
```

src/groupmod.c
```diff
@@ -511,7 +511,14 @@ static void close_files (void)
...
+		if (pflg)
+			audit_logger (AUDIT_GRP_CHAUTHTOK, Prog,
+		              info_gshadow.audit_msg,
+		              group_name, AUDIT_NO_ID,
+		              SHADOW_AUDIT_SUCCESS);
+		else
+			audit_logger (AUDIT_GRP_MGMT, Prog,
 		              info_gshadow.audit_msg,
 		              group_name, AUDIT_NO_ID,
 		              SHADOW_AUDIT_SUCCESS);
```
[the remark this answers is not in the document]
ikerexxe: I would prefer to keep it as it is so that the code looks as much like the existing downstream code as possible.

src/useradd.c
ikerexxe: I would prefer to keep it as it is so that the code looks as much like the existing downstream code as possible.

ikerexxe:
> I've made a couple of comments, but feel free to do what you like with them.

I'd prefer to keep it as it is for now.

> I'm happy enough with the changes as they are now, so:
> 
> ```
> ```

I added the tag, feel free to merge if you are fine with the current proposal.

lib/audit_help.c
```diff
@@ -68,6 +74,47 @@ void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
...
+ * grp_type - type of group: "grp" or "new_group"
+ * grp - group name associated with event
+ */
+void
+audit_logger_with_group(int type, const char *op, const char *name,
+    id_t id, const char *grp_type, const char *grp,
+    shadow_audit_result result)
+{
+	int len;
+	char enc_group[GROUP_NAME_MAX_LENGTH * 2 + 1];
```
alejandro-colomar, line 100: @ikerexxe  Where is GROUP_NAME_MAX_LENGTH defined?

PR #1201, opened 2025-02-03
src/: Prefer return over exit in main function

stoeckmann:
Argued with myself about creating this PR or not, so here are my pro and contra arguments:

### return over exit
- Allows the stack protector of main function to work

### exit over return
- Prevents copy&paste errors if code is moved from main into another function
- Counter argument to stack protector: Writing far enough behind the end of stack could modify the stack protector behavior itself
- If no stack protector is compiled in, this change actually enables stack attacks in main function

### Either/or
- Unifies the code base a bit more

### Keep as is
- Every change has potential of failure
- Counter argument to stack protector: Pretty much every error handling with exit in a function circumvents the protector

Since I created the PR, I see benefits in allowing the stack protector to work and unification process is always nice.

But let's not rush this. Small benefit anyway.

### DIFF
--- a/src/check_subid_range.c
+++ b/src/check_subid_range.c
@@ -36,20 +36,20 @@ main(int argc, char **argv)
 	log_set_logfd(stderr);
 
 	if (argc != 5)
-		exit(1);
+		return 1;
 
 	owner = argv[1];
 	check_uids = argv[2][0] == 'u';
 	if (get_uid(argv[3], &start) == -1)
-		exit(1);
+		return 1;
 	if (str2ul(&count, argv[4]) == -1)
-		exit(1);
+		return 1;
 	if (check_uids) {
 		if (have_sub_uids(owner, start, count))
 			exit(0);
-		exit(1);
+		return 1;
 	}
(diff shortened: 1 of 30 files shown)


src/check_subid_range.c
```diff
@@ -36,20 +36,20 @@ main(int argc, char **argv)
...
-		exit(1);
+		return 1;
 	if (str2ul(&count, argv[4]) == -1)
-		exit(1);
+		return 1;
 	if (check_uids) {
 		if (have_sub_uids(owner, start, count))
 			exit(0);
-		exit(1);
+		return 1;
```
alejandro-colomar, line 50: For error handling, I very much prefer exit(3).

For successful exiting, I still have some preference for exit(3), but could be convinced of using return.
I think exit(3) has more readability, as it tells that the program is over, while with return you have to remember (and check) that you're in main().

BTW, right here there's a successful exit(3) call unchanged.  :)

stoeckmann: I think we can close this PR. I'm not fully convinced but wanted to offer in case someone really wants this. So ... No. :)

PR #1213, opened 2025-02-17
strsep(3) transition in steps

alejandro-colomar:
Hi @zeha ,

After investigating the code transformation, I can't see the mistake.  In this PR I've reintroduced the change, but this time in steps, to be able to bisect the bug.

Would you mind bisecting this to see which transformation reintroduces the bug you're experiencing?

Thanks!
Alex

### DIFF
--- a/src/chfn.c
+++ b/src/chfn.c
@@ -216,27 +216,32 @@ static void new_fields (void)
  */
 static char *copy_field (char *in, char *out, char *extra)
 {
+	char *cp = NULL;
+
 	while (NULL != in) {
-		char  *f;
+		cp = strchr (in, ',');
+		if (NULL != cp) {
+			*cp++ = '\0';
+		}
 
-		f = strsep(&in, ",");
-
-		if (strchr(f, '=') == NULL)
+		if (strchr (in, '=') == NULL) {
 			break;
+		}
 
 		if (NULL != extra) {
 			if (!streq(extra, "")) {
 				strcat (extra, ",");
 			}
 
-			strcat(extra, f);
+			strcat (extra, in);
 		}
+		in = cp;
 	}
 	if ((NULL != in) && (NULL != out)) {
 		strcpy (out, in);
 	}
 
-	return in;
+	return cp;
 }
 
 /*
--- a/src/chfn.c
+++ b/src/chfn.c
@@ -220,18 +220,20 @@ static char *copy_field (char *in, char *out, char *extra)
 	char *cp = NULL;
 
 	while (NULL != in) {
+		char  *f;
+
+		f = in;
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
Hmmmm, bisecting says this transformation is bogus:

```diff
commit 53b40c5d6ff01a528c9a85311d50755f725077f7 (HEAD -> tmp)
Author: Alejandro Colomar <alx@kernel.org>
Date:   Mon Feb 17 13:51:40 2025 +0100

    src/chfn.c: Use strsep(3) instead of its pattern
    
    Signed-off-by: Alejandro Colomar <alx@kernel.org>

diff --git a/src/chfn.c b/src/chfn.c
index 60fbd841..23945325 100644
--- a/src/chfn.c
+++ b/src/chfn.c
@@ -36,7 +36,6 @@
 #include "string/strcmp/streq.h"
 #include "string/strcpy/strtcpy.h"
 #include "string/strdup/xstrdup.h"
-#include "string/strtok/stpsep.h"
 
 
 /*
@@ -222,8 +221,8 @@ static char *copy_field (char *in, char *out, char *extra)
        while (NULL != in) {
                char  *f;
 
-               f = in;
-               cp = stpsep(in, ",");
+               f = strsep(&in, ",");
+               cp = in;
 
                if (strchr(f, '=') == NULL)
                        break;
@@ -235,7 +234,6 @@ static char *copy_field (char *in, char *out, char *extra)
 
                        strcat(extra, f);
                }
-               in = cp;
        }
        if ((NULL != in) && (NULL != out)) {
                strcpy (out, in);
```

zeha:
53b40c5d6ff01a528c9a85311d50755f725077f7 apparently

alejandro-colomar:
Let's analyze this:

stpsep()-based code:

`f` points to the left-hand-side (LHS) of the separated string.
`cp` points to the RHS, or is NULL if there was no delimiter.
`in` points to the LHS, and at the end of the iteration it's updated to point to the RHS (or to be NULL, if there was no separation).

strsep(3)=based code:

`f` points to the LHS.
`in` is immediately updated to be the RHS (or NULL).  But since it's not used in the loop, it doesn't matter.
`cp` is the RHS or NULL.

Anyone can see the bug?

It doesn't seem to be a bug in stpsep() itself, because the code using stpsep() works correctly.  And strsep(3) is not mine, so we can also asume strsep(3) works correctly.  It is necessarily my fault, but I don't see it.

alejandro-colomar:
Ahhhhh!  I've seen it now.  The `break`!

alejandro-colomar:
The good news is that I've checked the original commit that introduced the bug for similar issues with `break` or other keywords that jump out, and this is the only one.

PR #1215, opened 2025-02-17
chsh: do not warn about blank shell

zeha:
chsh(1) documents the -s option with this extra sentence:
> Setting this field to blank causes the system to select the default login shell.

As such warning about a blank string seems wrong.

Closes: <https://bugs.debian.org/876542>
Reported-by: 積丹尼 Dan Jacobson <jidanni@jidanni.org>

### DIFF
--- a/src/chsh.c
+++ b/src/chsh.c
@@ -555,9 +555,11 @@ int main (int argc, char **argv)
 		fprintf (stderr, _("%s: Invalid entry: %s\n"), Prog, loginsh);
 		fail_exit (1);
 	}
-	if (loginsh[0] != '/'
-			|| is_restricted_shell (loginsh)
-			|| (access (loginsh, X_OK) != 0)) {
+	if (!streq(loginsh, "")
+	    && (loginsh[0] != '/'
+	        || is_restricted_shell (loginsh)
+	        || (access (loginsh, X_OK) != 0)))
+	{
 		if (amroot) {
 			fprintf (stderr, _("%s: Warning: %s is an invalid shell\n"), Prog, loginsh);
 		} else {
@@ -567,10 +569,13 @@ int main (int argc, char **argv)
 	}
 
 	/* Even for root, warn if an invalid shell is specified. */
-	if (access (loginsh, F_OK) != 0) {
-		fprintf (stderr, _("%s: Warning: %s does not exist\n"), Prog, loginsh);
-	} else if (access (loginsh, X_OK) != 0) {
-		fprintf (stderr, _("%s: Warning: %s is not executable\n"), Prog, loginsh);
+	if (!streq(loginsh, "")) {
+		/* But not if an empty string is given, documented as meaning the default shell */
+		if (access (loginsh, F_OK) != 0) {
+			fprintf (stderr, _("%s: Warning: %s does not exist\n"), Prog, loginsh);
+		} else if (access (loginsh, X_OK) != 0) {
+			fprintf (stderr, _("%s: Warning: %s is not executable\n"), Prog, loginsh);
+		}
 	}
 
 	update_shell (user, loginsh);


zeha:
I guess the if() formatting wont be right.

src/chsh.c
```diff
@@ -555,9 +555,9 @@ int main (int argc, char **argv)
 		fprintf (stderr, _("%s: Invalid entry: %s\n"), Prog, loginsh);
 		fail_exit (1);
 	}
-	if (loginsh[0] != '/'
+	if (loginsh[0] != '\0' && (loginsh[0] != '/'
```
alejandro-colomar, line 558: Please use `!streq(loginsh, "")`.

src/chsh.c
```diff
@@ -567,10 +567,13 @@ int main (int argc, char **argv)
 	}
 
 	/* Even for root, warn if an invalid shell is specified. */
-	if (access (loginsh, F_OK) != 0) {
-		fprintf (stderr, _("%s: Warning: %s does not exist\n"), Prog, loginsh);
-	} else if (access (loginsh, X_OK) != 0) {
-		fprintf (stderr, _("%s: Warning: %s is not executable\n"), Prog, loginsh);
+	if (loginsh[0] != '\0') {
```
alejandro-colomar, line 570: Same here.  Please use `streq()`.

alejandro-colomar:
I see that you're only disabling the diagnostics.  I assume that it behaved correctly (except for the diagnostics), that is, the default shell is used.  Is that correct?

alejandro-colomar:
I think we could change in the commit message:

```diff
-Bug-Debian: https://bugs.debian.org/876542
+Closes: <https://bugs.debian.org/876542>
```

By the URI, we already see it's a Debian bug.  This uses less specialized tags.  Plus, URIs are better enclosed within <>, I think (although these tags being like HTTP header fields, which don't use it --AFAIR--, I don't have a strong opinion).

src/chsh.c
zeha: done

src/chsh.c
zeha: done

zeha:
> I see that you're only disabling the diagnostics. I assume that it behaved correctly (except for the diagnostics), that is, the default shell is used. Is that correct?

It writes an empty string into /etc/passwd for the shell. "The system" (I assume PAM) then uses the default shell (/bin/sh).

src/chsh.c
```diff
@@ -555,9 +555,9 @@ int main (int argc, char **argv)
 		fprintf (stderr, _("%s: Invalid entry: %s\n"), Prog, loginsh);
 		fail_exit (1);
 	}
-	if (loginsh[0] != '/'
+	if (!streq(loginsh, "") && (loginsh[0] != '/'
 			|| is_restricted_shell (loginsh)
-			|| (access (loginsh, X_OK) != 0)) {
+			|| (access (loginsh, X_OK) != 0))) {
```
alejandro-colomar, line 560: I think the most common formatting for such a conditional in this project is:

```c
if (A
    && (B
        || C
        || D))
}
```

alejandro-colomar (changes requested):

src/chsh.c
```diff
@@ -555,9 +555,10 @@ int main (int argc, char **argv)
 		fprintf (stderr, _("%s: Invalid entry: %s\n"), Prog, loginsh);
 		fail_exit (1);
 	}
-	if (loginsh[0] != '/'
+	if (!streq(loginsh, "")
+		&& (loginsh[0] != '/'
 			|| is_restricted_shell (loginsh)
-			|| (access (loginsh, X_OK) != 0)) {
+			|| (access (loginsh, X_OK) != 0))) {
```
alejandro-colomar, line 561: Please move the opening brace to the next line.

zeha:
Other code in the same file uses spaces to align, so I assume this is the right thing to do.

alejandro-colomar (approved):
```
```

alejandro-colomar:
> Other code in the same file uses spaces to align, so I assume this is the right thing to do.

Yes it is.  I could live with the tabs, but I do prefer spaces.  Thanks!

alejandro-colomar:
I'll cherry-pick this in 4.17.3.

zeha:
> I'll cherry-pick this in 4.17.3. Thanks!

Thanks!

alejandro-colomar:
There have been no comments in a few days, so I'll merge now.

PR #1222, opened 2025-02-22
lib/fs/readlink/areadlink.h: areadlink(): Avoid inconditionally using PATH_MAX

sthibaul:
86451e374b90 ("lib/fs/readlink/areadlink.h: areadlink(): Use PATH_MAX instead of a magic value") introduced using PATH_MAX, but e.g. GNU/Hurd does not define such a limitation.

### DIFF
--- a/lib/fs/readlink/areadlink.h
+++ b/lib/fs/readlink/areadlink.h
@@ -9,11 +9,11 @@
 #include <config.h>
 
 #include <errno.h>
-#include <limits.h>
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdlib.h>
 
+#include "defines.h"
 #include "alloc/malloc.h"
 #include "attr.h"
 #include "fs/readlink/readlinknul.h"
--- a/lib/defines.h
+++ b/lib/defines.h
@@ -65,6 +65,14 @@
 #endif
 #endif
 
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
+
+#ifndef MAXHOSTNAMELEN
+#define MAXHOSTNAMELEN 64
+#endif
+
 #include <syslog.h>
 
 #ifndef LOG_WARN
--- a/lib/get_pid.c
+++ b/lib/get_pid.c
@@ -10,6 +10,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 
+#include "defines.h"
 #include "atoi/getnum.h"
 #include "defines.h"
 #include "prototypes.h"
--- a/src/login_nopam.c
+++ b/src/login_nopam.c
@@ -57,6 +57,7 @@
 #include <syslog.h>
 #include <unistd.h>
 
+#include "defines.h"
 #include "prototypes.h"
 #include "sizeof.h"
 #include "string/strcmp/strcaseeq.h"


lib/fs/readlink/areadlink.h
```diff
@@ -27,7 +27,11 @@ inline char *areadlink(const char *link);
 inline char *
 areadlink(const char *link)
 {
+#ifdef PATH_MAX
 	size_t  size = PATH_MAX;
+#else
+	size_t  size = 1024;
+#endif
```
alejandro-colomar, line 34: I guess we should rather check if PATH_MAX doesn't exist and then define it ourselves, like we did in 8b36662205ed (2025-01-07; "lib/chkname.c: login_name_max_size(): Put limits for LOGIN_NAME_MAX and sysconf(_SC_LOGIN_NAME_MAX)").

sthibaul: That would look odd to me. Here we are just talking about an initial guess for the buffer size, it's not actually a maximum limit (which doesn't even exist).

alejandro-colomar: I have patches for adding that guess in other places, so I think having a name for that guess is a good thing.  The magic number 1024 doesn't tell much.

alejandro-colomar: Plus it avoids ifdefs (one vs many).

alejandro-colomar: BTW, there are other places using PATH_MAX already:

```
lib/fs/readlink/areadlink.h:30:	size_t  size = PATH_MAX;
lib/commonio.c:79:	char resolved_path[PATH_MAX];
lib/get_pid.c:55:	char   proc_dir_name[PATH_MAX];
```

You'll want to fix all of them at once, I think.

alejandro-colomar: > `commonio.c` already has a codepath that avoids using `PATH_MAX` when on glibc, which provides the posix 2008 behavior of dynamically allocating the string. I don't see `PATH_MAX` being used in `get_pid.c`?

Ahhh, I grepped in my working tree, where I have a patch adding one of those PATH_MAX.

```diff
commit 1eb18292941123548612c145f9c2f2c12ea9b3a8
Author: Alejandro Colomar <alx@kernel.org>
Date:   Sat Feb 22 14:47:22 2025 +0100

    lib/get_pid.c: pid_t is a signed integer
    
    Print it with "%d".  Also, we need to adapt the size of the allocation.
    To avoid making mistakes, just make it PATH_MAX, and let the compiler
    optimize that if it's smart enough.
    
    Signed-off-by: Alejandro Colomar <alx@kernel.org>

diff --git a/lib/get_pid.c b/lib/get_pid.c
index f4d31cfd..27e0899d 100644
--- a/lib/get_pid.c
+++ b/lib/get_pid.c
@@ -52,22 +52,21 @@ int get_pidfd_from_fd(const char *pidfdstr)
 int open_pidfd(const char *pidstr)
 {
        int    proc_dir_fd;
-       char   proc_dir_name[32];
+       char   proc_dir_name[PATH_MAX];
        pid_t  target;
 
        if (get_pid(pidstr, &target) == -1)
                return -ENOENT;
 
-       /* max string length is 6 + 10 + 1 + 1 = 18, allocate 32 bytes */
-       if (SNPRINTF(proc_dir_name, "/proc/%u/", target) == -1) {
-               fprintf(stderr, "snprintf of proc path failed for %u: %s\n",
+       if (SNPRINTF(proc_dir_name, "/proc/%d/", target) == -1) {
+               fprintf(stderr, "snprintf of proc path failed for %d: %s\n",
                        target, strerror(errno));
                return -EINVAL;
        }
 
        proc_dir_fd = open(proc_dir_name, O_DIRECTORY);
        if (proc_dir_fd < 0) {
-               fprintf(stderr, _("Could not open proc directory for target %u: %s\n"),
+               fprintf(stderr, _("Could not open proc directory for target %d: %s\n"),
                        target, strerror(errno));
                return -EINVAL;
        }
```

> 
> > The magic number 1024 doesn't tell much
> 
> There is nothing to tell here, it's just a pick-up number for an initial guess at the path length...

Since we have a different value in each file, I had plans to use PATH_MAX everywhere we deal with patches, and similar macros for other purposes too.  I think it's easier to understand when reading a declaration that has PATH_MAX elements what it's going to hold (in this case, the path name also helps).

> 
> It's _very_ different from calling `realpath` without the posix 2008 behavior, which _imposes_ the buffer to be at least `PATH_MAX`-sized, otherwise it may overflow. We do not want to guess what `PATH_MAX` could be there.

Hmmmm, this has opened another can of worms.  :)

sthibaul: > Hmmmm, this has opened another can of worms. :)

That's why `PATH_MAX` should just be avoided.

alejandro-colomar: > > Hmmmm, this has opened another can of worms. :)
> 
> That's why `PATH_MAX` should just be avoided.

What do implementations do for `realpath(, NULL);` (including some oldish ones)?

Is it safe to assume the POSIX.1-2008 behavior?

Cc: @richfelker

sthibaul: [I don't know any system that doesn't provide the posix 2008 behavior.](https://stackoverflow.com/questions/60802732/are-realpath-portability-concerns-obsolete) says it works fine on various systems

alejandro-colomar: > [I don't know any system that doesn't provide the posix 2008 behavior.](https://stackoverflow.com/questions/60802732/are-realpath-portability-concerns-obsolete) says it works fine on various systems

Thanks!  Then, we should remove the path that doesn't use the new behavior.

alejandro-colomar: > > `commonio.c` already has a codepath that avoids using `PATH_MAX` when on glibc, which provides the posix 2008 behavior of dynamically allocating the string. I don't see `PATH_MAX` being used in `get_pid.c`?
> 
> Ahhh, I grepped in my working tree, where I have a patch adding one of those PATH_MAX.
> 
> ```diff
> commit 1eb18292941123548612c145f9c2f2c12ea9b3a8
> Author: Alejandro Colomar <alx@kernel.org>
> Date:   Sat Feb 22 14:47:22 2025 +0100
> 
>     lib/get_pid.c: pid_t is a signed integer
>     
>     Print it with "%d".  Also, we need to adapt the size of the allocation.
>     To avoid making mistakes, just make it PATH_MAX, and let the compiler
>     optimize that if it's smart enough.
>     
>     Signed-off-by: Alejandro Colomar <alx@kernel.org>
> 
> diff --git a/lib/get_pid.c b/lib/get_pid.c
> index f4d31cfd..27e0899d 100644
> --- a/lib/get_pid.c
> +++ b/lib/get_pid.c
> @@ -52,22 +52,21 @@ int get_pidfd_from_fd(const char *pidfdstr)
>  int open_pidfd(const char *pidstr)
>  {
>         int    proc_dir_fd;
> -       char   proc_dir_name[32];
> +       char   proc_dir_name[PATH_MAX];
>         pid_t  target;
>  
>         if (get_pid(pidstr, &target) == -1)
>                 return -ENOENT;
>  
> -       /* max string length is 6 + 10 + 1 + 1 = 18, allocate 32 bytes */
> -       if (SNPRINTF(proc_dir_name, "/proc/%u/", target) == -1) {
> -               fprintf(stderr, "snprintf of proc path failed for %u: %s\n",
> +       if (SNPRINTF(proc_dir_name, "/proc/%d/", target) == -1) {
> +               fprintf(stderr, "snprintf of proc path failed for %d: %s\n",
>                         target, strerror(errno));
>                 return -EINVAL;
>         }
>  
>         proc_dir_fd = open(proc_dir_name, O_DIRECTORY);
>         if (proc_dir_fd < 0) {
> -               fprintf(stderr, _("Could not open proc directory for target %u: %s\n"),
> +               fprintf(stderr, _("Could not open proc directory for target %d: %s\n"),
>                         target, strerror(errno));
>                 return -EINVAL;
>         }
> ```


I still prefer if you define PATH_MAX, as I'm going to add more uses of it, such as in that code I showed there.

alejandro-colomar: @sthibaul We have already merged the code above.  Please revise this PR.  I really recommend defining PATH_MAX in this project, since it's used in several places.

sthibaul:
Here it is

lib/fs/readlink/areadlink.h
sthibaul: Here it is

sthibaul:
I have also done the same for MAXHOSTNAMELEN

lib/alloc/malloc.h
```diff
@@ -9,10 +9,19 @@
...
 
 
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
+
+#ifndef MAXHOSTNAMELEN
+#define MAXHOSTNAMELEN 64
+#endif
+
```
alejandro-colomar: Thanks!  This is nicer.

But please move it to "lib/defines.h", since "lib/alloc/malloc.h" isn't very appropriate for these values.

sthibaul: moved so

PR #1223, opened 2025-02-22
lib/commonio.c: Rely on the POSIX.1-2008 behavior of realpath(3)

alejandro-colomar:
Link: <https://github.com/shadow-maint/shadow/pull/1222#discussion_r1966612238>
Link: <https://stackoverflow.com/questions/60802732/are-realpath-portability-concerns-obsolete>
Reported-by: @sthibaul

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -74,17 +74,10 @@ int lrename (const char *old, const char *new)
 {
 	int res;
 	char *r = NULL;
-
-#ifndef __GLIBC__
-	char resolved_path[PATH_MAX];
-#endif				/* !__GLIBC__ */
 	struct stat sb;
+
 	if (lstat (new, &sb) == 0 && S_ISLNK (sb.st_mode)) {
-#ifdef __GLIBC__ /* now a POSIX.1-2008 feature */
 		r = realpath (new, NULL);
-#else				/* !__GLIBC__ */
-		r = realpath (new, resolved_path);
-#endif				/* !__GLIBC__ */
 		if (NULL == r) {
 			perror ("realpath in lrename()");
 		} else {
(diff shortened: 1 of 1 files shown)


ikerexxe:
I only have a question here. Is it POSIX.1-2008 or POSIX.1-2018? I would say it is the second but I prefer to have it confirmed.

alejandro-colomar:
> I only have a question here. Is it POSIX.1-2008 or POSIX.1-2018? I would say it is the second but I prefer to have it confirmed.

```
$ man standards 2>/dev/null | grep '^       POSIX.1'
       POSIX.1‐1988
       POSIX.1‐1990
       POSIX.1b (formerly known as POSIX.4)
       POSIX.1c (formerly known as POSIX.4a)
       POSIX.1d
       POSIX.1g
       POSIX.1j
       POSIX.1‐1996
       POSIX.1‐2001
       POSIX.1‐2008
       POSIX.1‐2017
```

You're probably asking about POSIX.1-2017.  POSIX.1-2008 is the major revision of the POSIX standard.  In 2017, a Technical Corrigenda was issued, but which didn't meaningfully change the standard, so we keep referring to the 2008 one.  It's similar to C11 and C17.  C11 was a major revision of ISO C, and C17 is just the same with a Technical Corrigenda applied.

If you go to the online publication of POSIX.1-2017, it says that at the top (in other words):
```
The Open Group Base Specifications Issue 7, 2018 edition
IEEE Std 1003.1™-2017 (Revision of IEEE Std 1003.1-2008)
```
<https://pubs.opengroup.org/onlinepubs/9699919799/>

(BTW, it is the 2018 edition because it was finally published in 2018, but the name is 2017 because it was finalized in 2017; just like C23 is called C23 even if it was published by ISO in 2024.)

ikerexxe (approved):
Thanks for the explanation, now I see why I got confused. LGTM!

PR #1225, opened 2025-02-24
add tags file to .gitignore

hallyn:


hallyn:
> What are these tags files?

It's the ctags(1) file.

> Do you mind adding an additional blank line between them and the "test files"?

Sure I'll change that later today.

I wasn't quite sure whether to do /tags or tags.  Normally one
only keeps it at the top level.  But if you "cd lib; vi log.c"
you'll then not automatically have tags, and so you might
sometimes want more targetted tags files for certain subdirs.
Like I say, just not quite sure.

alejandro-colomar:
> > What are these tags files?
> 
> It's the ctags(1) file.
> 
> > Do you mind adding an additional blank line between them and the "test files"?
> 
> Sure I'll change that later today.
> 
> I wasn't quite sure whether to do /tags or tags. Normally one only keeps it at the top level. But if you "cd lib; vi log.c" you'll then not automatically have tags, and so you might sometimes want more targetted tags files for certain subdirs. Like I say, just not quite sure.

Just in case you didn't know, you can have a personal gitignore file, which will work for all projects, not needing to add rules for every project you work in.

I have the following for myself:

```sh
alx@debian:~$ cat /etc/gitconfig | sed -n '/^.core/,/^[^\t]/p'
[core]
	editor = vim
	attributesFile = /etc/git/attributes
	excludesFile = /etc/git/ignore
[diff]
alx@debian:~$ cat /etc/git/ignore 
# vim(1) tmp files
*.swp
*.swo

# Store patches here:
/patches/
```

And then for example in the Linux man-pages repo I only ignore stuff which is part of the project itself:

```sh
alx@debian:~/src/linux/man-pages/man-pages/master$ cat .gitignore 
# Default $builddir:
/.tmp/

# checkpatch spurious output.  I hope we fix that some day.
/.checkpatch-camelcase.*
```

hallyn:
> Just in case you didn't know, you can have a personal gitignore file, which will work for all projects, not needing to add rules for every project you work in.

All right all right.  I'll close this.

I didn't know about excludesFile= option.  Neat.

> (you can have the same under `~/`, BTW. I have it under `/etc/` because I find it nice to have the same config when I do something as root.)

Haha, I prefer everything to be wonky when I'm root, so that I know that
hey, I shouldn't be doing anything, I'm root :)

Issue #1224, opened 2025-02-24
configure: derives PASSWD_PROGRAM from installed(!) /usr/bin/passwd


zeha:
configure today checks if it finds /usr/bin/passwd, and if so it sets PASSWD_PROGRAM=/usr/bin/passwd and PASSWD_PROGRAM=/bin/passwd otherwise.
This seems wrong: this package will install the passwd program, and it already knows where it will go (into exec_prefix/bin).

The existing check makes building of shadow vary between an environment which already has shadow installed and one that does not, cf. https://reproduce.debian.net/amd64/api/v0/builds/216764/diffoscope

I would propose applying something like this, but it can probably be simpler:

```patch
--- a/configure.ac
+++ b/configure.ac
@@ -150,13 +150,7 @@ AC_DEFINE_UNQUOTED(LASTLOG_FILE, "$shadow_cv_logdir/lastlog",
 AC_DEFINE_UNQUOTED(FAILLOG_FILE, "$shadow_cv_logdir/faillog",
 	[Path for faillog file.])
 
-AC_CACHE_CHECK([location of the passwd program], shadow_cv_passwd_dir,
-[if test -f /usr/bin/passwd; then
-	shadow_cv_passwd_dir=/usr/bin
-else
-	shadow_cv_passwd_dir=/bin
-fi])
-AC_DEFINE_UNQUOTED(PASSWD_PROGRAM, "$shadow_cv_passwd_dir/passwd",
+AC_DEFINE_UNQUOTED(PASSWD_PROGRAM, "$prefix/bin/passwd",
 	[Path to passwd program.])
 
 AC_ARG_ENABLE(shadowgrp,
```

I've avoided using $exec_prefix here, as configure.ac has some hacks to clear exec_prefix if prefix=/usr. That might also want to be cleaned up.

alejandro-colomar:
Cc: @thesamesam

alejandro-colomar:
> configure today

That's confusing, especially since today we had a release.  Do you mean a regression in the release from today?  I guess not.

You probably mean "configure up until today", right?

> checks if it finds /usr/bin/passwd, and if so it sets PASSWD_PROGRAM=/usr/bin/passwd and PASSWD_PROGRAM=/bin/passwd otherwise. This seems wrong: this package will install the passwd program, and it already knows where it will go (into exec_prefix/bin).

Heh, pretty wrong, I'd say.

> 
[...]

PR #1227, opened 2025-02-24
configure.ac: be deterministic about passwd location

zeha:
Statically set PASSWD_PROGRAM depending on exec_prefix, and not by where the passwd program was at configure time.

Depending on the specific build situation before, this may or may not change the embedded passwd program path. Also configure.ac sets exec_prefix=/ for prefix=/usr, so this might be a bit confusing, but at least deterministic.

Closes: #1224

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -115,13 +115,7 @@ AC_DEFINE_UNQUOTED(LASTLOG_FILE, "$shadow_cv_logdir/lastlog",
 AC_DEFINE_UNQUOTED(FAILLOG_FILE, "$shadow_cv_logdir/faillog",
 	[Path for faillog file.])
 
-AC_CACHE_CHECK([location of the passwd program], shadow_cv_passwd_dir,
-[if test -f /usr/bin/passwd; then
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
Please add a Closes tag to the commit itself.

zeha:
done!

alejandro-colomar:
I think you should mention as a caveat the issue with exec_prefix.  If this might eventually be considered a regression, we should have the information about why.

Also, please open a new issue for that, because maybe it would be better to first address that issue, and then fix this one.

alejandro-colomar:
Other distros, please confirm if this is fine for you.

Cc: @thesamesam , @jubalh , @ikerexxe

thesamesam (approved):
Thanks. It looks right but I've not tested it.

ikerexxe (approved):
LGTM! Let's wait some days for other maintainers to review before merging.

alejandro-colomar (approved):

alejandro-colomar:
No comments so far, so I'll merge now.

PR #1230, opened 2025-02-24
configure.ac, lib/shadow.c: Remove dead file

alejandro-colomar:
We didn't even have prototypes for these APIs since long ago, when the prototypes were removed, but misteriously the implementations remained.

Both glibc and musl provide getspnam(3), so this file was effectively being ignored by the compiler.  Just remove it.

Also remove the check for getspnam, which isn't used elsewhere.

Fixes: 0ee095abd8db (2007-10-07; "[svn-upgrade] Integrating new upstream version, shadow (4.0.7)")


alejandro-colomar:
> [4bb6df2](https://github.com/shadow-maint/shadow/commit/4bb6df211e178c3bdf3f8160c3ad22f89770cf9e) commit message has a stray "f" at the end

Thanks!  That was a glitch of squashing a commit with a dummy message, and forgetting to remove the name.

alejandro-colomar:
@ikerexxe This one is really trivial.  I'll push it higher in the priority list.  Please review.

alejandro-colomar:
> > While reviewing the `getspnam` related code, I found that `lib/xgetspnam.c` still contains a reference to it. I guess we should remove that file too.
> 
> Hmmm, thanks! I'll push a commit removing that file. Let's see if it passes CI cleanly.

Ahh, that file defines `xgetspnam`, not `getspnam` (see `XFUNCTION_NAME` in `xgetXXbyYY.c`.  So I'll revert that.

ikerexxe:
> Ahh, that file defines `xgetspnam`, not `getspnam` (see `XFUNCTION_NAME` in `xgetXXbyYY.c`. So I'll revert that.

I see. Sorry for the noise.

ikerexxe (approved):
LGTM! Let's merge the changes.

alejandro-colomar:
Thank you both!  

> > Ahh, that file defines `xgetspnam`, not `getspnam` (see `XFUNCTION_NAME` in `xgetXXbyYY.c`. So I'll revert that.
> 
> I see. Sorry for the noise.

No problem.  That code is rather convoluted and difficult to read correctly.  That's why I do what I do, indeed.  :-)

PR #1232, opened 2025-02-24
configure.ac: stop checking for utmp location

zeha:
Nothing seems to use it.


alejandro-colomar:
Nobody complained, so let's merge now.  Thanks!

PR #1217, opened 2025-02-18
Put get_date() on a diet

alejandro-colomar:
Cc: @zeha, @timparenti

This applies further removal of code in get_date().  I'll queue this patch for after the bugfix from #1214 .

This won't be cherry-picked for a stable release, as it's not supposed to fix a reported bug.  It just makes get_date() less brain damaged.

PR #1233, opened 2025-03-03
Test expiration date

ikerexxe:
Testing for https://github.com/shadow-maint/shadow/pull/1217

### DIFF
--- a/tests/system/tests/test_chage.py
+++ b/tests/system/tests/test_chage.py
@@ -1,0 +1,127 @@
+"""
+Test chage
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from pytest_mh.conn import ProcessError
+from framework.roles.shadow import Shadow
+from framework.topology import KnownTopology
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+@pytest.mark.parametrize(
+    "expiration_date, expected_date",
+    [
+        ("0", 0),  # 1970-01-01
+        ("1", 1),  # 1970-01-02
+        ("20089", 20089),  # 2025-01-01
+        ("1000000", 1000000),  # This will happen in a very long time
+        ("1970-01-01", 0),
+        ("1970-01-02", 1),
+        ("2025-01-01", 20089),
+    ],
+)
+def test_chage__set_expire_date_with_valid_date(shadow: Shadow, expiration_date: str, expected_date: int | None):
+    """
+    :title: Set valid account expiration date
+    :setup:
+        1. Create user
+    :steps:
+        1. Set valid account expiration date
+        2. Check user exists and expiration date
+    :expectedresults:
+        1. The expiration date is correctly set
+        2. User is found and expiration date is valid
+    :customerscenario: False
+    """
+    shadow.useradd("tuser1")
+
+    shadow.chage(f"-E {expiration_date} tuser1")
+
+    result = shadow.tools.getent.shadow("tuser1")
+    assert result is not None, "User should be found"
+    assert result.name == "tuser1", "Incorrect username"
+    assert result.expiration_date == expected_date, "Incorrect expiration date"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+@pytest.mark.parametrize(
+    "expiration_date",
+    [
+        "-2",  # Dates can't be in negative numbers
+        "-1000",  # Dates can't be in negative numbers
+        "2025-18-18",  # That month and day don't exist
+        "1969-01-01",  # This is before 1970-01-01
+        "2025-13-01",  # That month doesn't exist
+        "2025-01-32",  # That day doesn't exist
+        "today",
+        "tomorrow",
+    ],
+)
+def test_chage__set_expire_date_with_invalid_date(shadow: Shadow, expiration_date: str):
+    """
+    :title: Set invalid account expiration date
+    :setup:
+        1. Create user
+    :steps:
+        1. Set invalid account expiration date
+        2. Check user exists and expiration date
+    :expectedresults:
+        1. The process fails and the expiration date isn't changed
+        2. User is found and expiration date is empty
+    :customerscenario: False
+    """
+    shadow.useradd("tuser1")
+
+    with pytest.raises(ProcessError):
+        shadow.chage(f"-E {expiration_date} tuser1")
+
+    result = shadow.tools.getent.shadow("tuser1")
+    assert result is not None, "User should be found"
+    assert result.name == "tuser1", "Incorrect username"
+    assert result.expiration_date is None, "Expiration date should be empty"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+@pytest.mark.parametrize(
+    "expiration_date",
+    [
+        "-1",
+        "''",
+    ],
+)
+def test_chage__set_expire_date_with_empty_date(shadow: Shadow, expiration_date: str):
+    """
+    :title: Set empty account expiration date
+    :setup:
+        1. Create user
+    :steps:
+        1. Set account expiration date
+        2. Check user exists and expiration date
+        3. Empty account expiration date
+        4. Check user exists and expiration date
+    :expectedresults:
+        1. The expiration date is correctly set
+        2. User is found and expiration date is valid
+        3. The expiration date is correctly emptied
+        4. User is found and expiration date is empty
+    :customerscenario: False
+    """
+    shadow.useradd("tuser1")
+
+    shadow.chage("-E 10 tuser1")
+
+    result = shadow.tools.getent.shadow("tuser1")
+    assert result is not None, "User should be found"
+    assert result.name == "tuser1", "Incorrect username"
+    assert result.expiration_date == 10, "Incorrect expiration date"
+
+    shadow.chage(f"-E {expiration_date} tuser1")
+
+    result = shadow.tools.getent.shadow("tuser1")
+    assert result is not None, "User should be found"
+    assert result.name == "tuser1", "Incorrect username"
+    assert result.expiration_date is None, "Expiration date should be empty"
--- a/tests/system/tests/test_chage.py
+++ b/tests/system/tests/test_chage.py
@@ -7,8 +7,125 @@ from __future__ import annotations
 import pytest
 
 from pytest_mh.conn import ProcessError
+from framework.misc import days_since_epoch
 from framework.roles.shadow import Shadow
 from framework.topology import KnownTopology
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+@pytest.mark.parametrize(
+    "last_changed_date, expected_date",
+    [
+        ("0", 0),  # 1970-01-01
+        ("1", 1),  # 1970-01-02
+        ("20089", 20089),  # 2025-01-01
+        ("1000000", 1000000),  # This will happen in a very long time
+        ("1970-01-01", 0),
+        ("1970-01-02", 1),
+        ("2025-01-01", 20089),
+    ],
+)
+def test_chage__set_last_changed_with_valid_date(shadow: Shadow, last_changed_date: str, expected_date: int | None):
+    """
+    :title: Set valid last changed date
+    :setup:
+        1. Create user
+    :steps:
+        1. Set valid account last changed date
+        2. Check user exists and last changed date
+    :expectedresults:
+        1. The last changed date is correctly set
+        2. User is found and last changed date is valid
+    :customerscenario: False
+    """
+    shadow.useradd("tuser1")
+
+    shadow.chage(f"-d {last_changed_date} tuser1")
+
+    result = shadow.tools.getent.shadow("tuser1")
+    assert result is not None, "User should be found"
+    assert result.name == "tuser1", "Incorrect username"
+    assert result.last_changed == expected_date, "Incorrect expiration date"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+@pytest.mark.parametrize(
+    "last_changed_date",
+    [
+        "-2",  # Dates can't be in negative numbers
+        "-1000",  # Dates can't be in negative numbers
+        "2025-18-18",  # That month and day don't exist
+        "1969-01-01",  # This is before 1970-01-01
+        "2025-13-01",  # That month doesn't exist
+        "2025-01-32",  # That day doesn't exist
+        "today",
+        "tomorrow",
+    ],
+)
+def test_chage__set_last_changed_with_invalid_date(shadow: Shadow, last_changed_date: str):
+    """
+    :title: Set invalid last changed date
+    :setup:
+        1. Create user
+    :steps:
+        1. Set invalid last changed date
+        2. Check user exists and last changed date
+    :expectedresults:
+        1. The process fails and the last changed date isn't changed
+        2. User is found and last changed is not changed
+    :customerscenario: False
+    """
+    shadow.useradd("tuser1")
+
+    with pytest.raises(ProcessError):
+        shadow.chage(f"-d {last_changed_date} tuser1")
+
+    result = shadow.tools.getent.shadow("tuser1")
+    assert result is not None, "User should be found"
+    assert result.name == "tuser1", "Incorrect username"
+    assert result.last_changed == days_since_epoch(), "Expiration date should not be changed"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+@pytest.mark.parametrize(
+    "last_changed",
+    [
+        "-1",
+        "''",
+    ],
+)
+def test_chage__set_last_changed_with_empty_date(shadow: Shadow, last_changed: str):
+    """
+    :title: Set empty last changed date
+    :setup:
+        1. Create user
+    :steps:
+        1. Set account last changed date
+        2. Check user exists and last changed date
+        3. Empty account last changed date
+        4. Check user exists and last changed date
+    :expectedresults:
+        1. The last changed date is correctly set
+        2. User is found and last changed date is valid
+        3. The last changed date is correctly emptied
+        4. User is found and last changed date is empty
+    :customerscenario: False
+    """
+    shadow.useradd("tuser1")
+
+    shadow.chage("-d 10 tuser1")
+
+    result = shadow.tools.getent.shadow("tuser1")
+    assert result is not None, "User should be found"
+    assert result.name == "tuser1", "Incorrect username"
+    assert result.last_changed == 10, "Incorrect expiration date"
+
+    shadow.chage(f"-d {last_changed} tuser1")
+
+    result = shadow.tools.getent.shadow("tuser1")
+    assert result is not None, "User should be found"
+    assert result.name == "tuser1", "Incorrect username"
+    assert result.last_changed is None, "Incorrect expiration date"
 
 
 @pytest.mark.topology(KnownTopology.Shadow)
--- a/tests/system/tests/test_usermod.py
+++ b/tests/system/tests/test_usermod.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import pytest
 
+from pytest_mh.conn import ProcessError
 from framework.roles.shadow import Shadow
 from framework.topology import KnownTopology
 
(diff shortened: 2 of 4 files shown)


ikerexxe:
@alejandro-colomar this is the testing for https://github.com/shadow-maint/shadow/pull/1217. It contains all your commits there + the implementation to run `getent shadow $name` from the test framework + the test themselves. Do you mind taking a look to the last commit and confirm that I have tested all possible cases? Nothing else came to my mind, but you might have additional ideas.

In addition, one of the test cases (`usermod -e -1 $name`) returns success (0) even though the expiration date wasn't changed. Do you mind taking a look at it? I think it's a failure in the code.

Finally, does the aforementioned PR also change the date behaviour for useradd and chage? If so, I'll add additional tests for them.

tests/system/tests/test_usermod.py
```diff
@@ -41,3 +42,76 @@ def test_usermod__rename_user(shadow: Shadow):
...
+    assert result is not None, "User should be found"
+    assert result.name == "tuser1", "Incorrect username"
+    assert result.expiration_date == expected_date, "Incorrect expiration date"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+@pytest.mark.parametrize(
+    "expiration_date",
+    [
+        "-1",  # Dates can't be in negative numbers
```
alejandro-colomar, line 87: Dates can really be negative numbers.  It just means days before the Epoch.

alejandro-colomar:
> @alejandro-colomar this is the testing for #1217. It contains all your commits there + the implementation to run `getent shadow $name` from the test framework + the test themselves. Do you mind taking a look to the last commit and confirm that I have tested all possible cases? Nothing else came to my mind, but you might have additional ideas.
> 
> In addition, one of the test cases (`usermod -e -1 $name`) returns success (0) even though the expiration date wasn't changed. Do you mind taking a look at it? I think it's a failure in the code.
> 
> Finally, does the aforementioned PR also change the date behaviour for useradd and chage? If so, I'll add additional tests for them.

It affects all of these (since I'm changing strtoday()):

```
$ grep -rli strtoday 
ChangeLog
src/usermod.c
src/useradd.c
src/chage.c
lib/prototypes.h
lib/strtoday.c
lib/Makefile.am
po/POTFILES.in
NEWS
```

tests/system/tests/test_usermod.py
ikerexxe: mmm let's rephrase it then. usermod doesn't accept days before the Epoch. Do you see it more appropriate?

ikerexxe: You are right, it does. `An input of -1 or an empty string will blank the account expiration field in the shadow password file.`. So, there I have two additional test cases that should work :wink:

alejandro-colomar: And apart from that special case of -1, I think a negative number (e.g., -42) will just work, meaning 42 days before Epoch.

ikerexxe: I didn't check with your changed applied, but `usermod -e -2 tuser` reports the error `usermod: invalid date '-2'`. Which makes sense, as it doesn't make much sense to set a date before Epoch.

alejandro-colomar: I would expect it to succeed, because of this code:

```c
$ grepc strtoday .
./lib/prototypes.h:extern long strtoday (const char *);
./lib/strtoday.c:long strtoday (const char *str)
{
	time_t t;
	const char *s = str;

	/*
	 * get_date() interprets an empty string as the current date,
	 * which is not what we expect, unless you're a BOFH :-).
	 * (useradd sets sp_expire = current date for new lusers)
	 */
	if ((NULL == str) || streq(str, "")) {
		return -1;
	}

	/* If a numerical value is provided, this is already a number of
	 * days since EPOCH.
	 */
	if ('-' == *s) {
		s++;
	}
	s = stpspn(s, " ");
	if (strisdigit(s)) {
		long retdate;

		if (str2sl(&retdate, str) == -1)
			return -2;
		return retdate;
	}

	t = get_date(str, NULL);
	if ((time_t) - 1 == t) {
		return -2;
	}
	return t / DAY;
}
```

which is currently in master.

ikerexxe: `-2`, `-1000` and `1969-01-01` are incorrect dates, and as such, when running the corresponding binary you can't change the expiration date.

alejandro-colomar: Ahhh, you're correct.  Even though strtoday() supports negative dates, the callers reject them.

```c
$ grep -rn 'strtoday *(' -A1 src/ lib*
src/usermod.c:1085:				user_newexpire = strtoday (optarg);
src/usermod.c-1086-				if (user_newexpire < -1) {
--
src/useradd.c:1235:					user_expire = strtoday (optarg);
src/useradd.c-1236-					if (user_expire < -1) {
--
src/useradd.c:1534:		user_expire = strtoday (def_expire);
src/useradd.c-1535-	}
--
src/chage.c:188:		lstchgdate = strtoday (buf);
src/chage.c-189-		if (lstchgdate <= -1) {
--
src/chage.c:215:		expdate = strtoday (buf);
src/chage.c-216-		if (expdate <= -1) {
--
src/chage.c:366:			lstchgdate = strtoday (optarg);
src/chage.c-367-			if (lstchgdate < -1) {
--
src/chage.c:376:			expdate = strtoday (optarg);
src/chage.c-377-			if (expdate < -1) {
--
lib/prototypes.h:437:extern long strtoday (const char *);
lib/prototypes.h-438-
--
lib/strtoday.c:23: * strtoday() now uses get_date() (borrowed from GNU shellutils)
lib/strtoday.c-24- * which can handle many date formats, for example:
--
lib/strtoday.c:36:long strtoday (const char *str)
lib/strtoday.c-37-{
```

alejandro-colomar:
> @alejandro-colomar do you know why alpine fails when useradd/usermod/chage set the expiration date as YYYY-MM-DD? They all fail even though the date is valid (i.e. 2025-01-01).

Is it a regression or is it a failure in master?  Could you please open a branch without my changes, to test the tests against master?

ikerexxe:
> I don't know. It might be related to the signedness of time_t, or one of those little details. I'll check this weekend when I'm back home and have better internet to create the docker container and install all the deps.

@alejandro-colomar did you have time to look at this?

ikerexxe:
Did you setup everything manually? I'm building the container using the automation:
```
$ ansible-playbook share/ansible/playbook.yml -i share/ansible/inventory.ini -e 'distribution=alpine'
...
$ podman exec -ti builder bash
# useradd tt
# getent shadow tt
tt:!:20174:0:99999:7:::
# usermod -e 10 tt
# getent shadow tt
tt:!:20174:0:99999:7::10:
# usermod -e 1970-10-20 tt
usermod: invalid date '1970-10-20'
# usermod -e '1970-10-20' tt
usermod: invalid date '1970-10-20'
# usermod -e "1970-10-20" tt
usermod: invalid date '1970-10-20'
```

ikerexxe:
It fails even with the default shell:
```
$ podman exec -ti builder sh
# usermod -e 1970-10-20 tt
# usermod: invalid date '1970-10-20'
```

alejandro-colomar:
> Did you setup everything manually? I'm building the container using the automation:
> 
> ```
> $ ansible-playbook share/ansible/playbook.yml -i share/ansible/inventory.ini -e 'distribution=alpine'
> ...
> $ podman exec -ti builder bash
> # useradd tt
> # getent shadow tt
> tt:!:20174:0:99999:7:::
> # usermod -e 10 tt
> # getent shadow tt
> tt:!:20174:0:99999:7::10:
> # usermod -e 1970-10-20 tt
> usermod: invalid date '1970-10-20'
> # usermod -e '1970-10-20' tt
> usermod: invalid date '1970-10-20'
> # usermod -e "1970-10-20" tt
> usermod: invalid date '1970-10-20'
> ```

Yep, everything manual.  I'll try again with ansible.

alejandro-colomar:
> > Did you setup everything manually? I'm building the container using the automation:
> > ```
> > $ ansible-playbook share/ansible/playbook.yml -i share/ansible/inventory.ini -e 'distribution=alpine'
> > ...
> > $ podman exec -ti builder bash
> > # useradd tt
> > # getent shadow tt
> > tt:!:20174:0:99999:7:::
> > # usermod -e 10 tt
> > # getent shadow tt
> > tt:!:20174:0:99999:7::10:
> > # usermod -e 1970-10-20 tt
> > usermod: invalid date '1970-10-20'
> > # usermod -e '1970-10-20' tt
> > usermod: invalid date '1970-10-20'
> > # usermod -e "1970-10-20" tt
> > usermod: invalid date '1970-10-20'
> > ```
> 
> Yep, everything manual. I'll try again with ansible.

With ansible, it also works for me:

```sh
$ podman exec -ti builder bash
WARN[0000] Using cgroups-v1 which is deprecated in favor of cgroups-v2 with Podman v5 and will be removed in a future version. Set environment variable `PODMAN_IGNORE_CGROUPSV1_WARNING` to hide this warning. 
44da80a259d4:/# cat /etc/os-release 
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.21.3
PRETTY_NAME="Alpine Linux v3.21"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
44da80a259d4:/# ls
bin  etc   lib	  mnt  proc  run   srv	tmp  var
dev  home  media  opt  root  sbin  sys	usr
44da80a259d4:/# useradd tt
44da80a259d4:/# getent shadow tt
tt:!:20175:0:99999:7:::
44da80a259d4:/# usermod -e 10 tt
44da80a259d4:/# getent shadow tt
tt:!:20175:0:99999:7::10:
44da80a259d4:/# usermod -e 1970-10-20 tt
44da80a259d4:/# getent shadow tt
tt:!:20175:0:99999:7::292:
44da80a259d4:/# 
```

The bad thing about using ansible is that debugging becomes more difficult.  I really preferred the old Dockerfile.

ikerexxe:
Ok, I've been able to reproduce the problem in a basic Alpine Linux container (no need to use ansible or build shadow) with the following reproducer:
```
#define _XOPEN_SOURCE
#include <stdio.h>
#include <string.h>
#include <time.h>

int
main(void)
{
    struct tm tm;
    const char  *p;

    memset(&tm, 0, sizeof(tm));
    
    p = strptime("UTC", "%Z", &tm);
    if (p == NULL) {
    	printf("Error!\n");
        return -1;
    }
    
    printf("Success\n");
}
```

This code mimics in a very basic way [get_date()](https://github.com/shadow-maint/shadow/pull/1217/files#diff-e30b3208541dfcc01de8c9b99891f546ac46000a8edc5d5442aff297d65d417dR55) function and it should return `Error!` but it should succeed. I guess musl doesn't implement strptime() completely :disappointed:

alejandro-colomar:
> Ok, I've been able to reproduce the problem in a basic Alpine Linux container (no need to use ansible or build shadow) with the following reproducer:
> 
> ```
> #define _XOPEN_SOURCE
> #include <stdio.h>
> #include <string.h>
> #include <time.h>
> 
> int
> main(void)
> {
>     struct tm tm;
>     const char  *p;
> 
>     memset(&tm, 0, sizeof(tm));
>     
>     p = strptime("UTC", "%Z", &tm);
>     if (p == NULL) {
>     	printf("Error!\n");
>         return -1;
>     }
>     
>     printf("Success\n");
> }
> ```
> 
> This code mimics in a very basic way [get_date()](https://github.com/shadow-maint/shadow/pull/1217/files#diff-e30b3208541dfcc01de8c9b99891f546ac46000a8edc5d5442aff297d65d417dR55) function and it should return `Error!` but it should succeed. I guess musl doesn't implement strptime() completely 😞

musl does have a proper strptime(3), from what I can see in the source code.  It must be an issue about dependencies in musl.

alejandro-colomar:
> 		case 'Z':
> 			if (!strncmp(s, tzname[0], len = strlen(tzname[0]))) {
> 				tm->tm_isdst = 0;
> 				s += len;
> 			} else if (!strncmp(s, tzname[1], len=strlen(tzname[1]))) {
> 				tm->tm_isdst = 1;
> 				s += len;
> 			} else {
> 				/* FIXME: is this supposed to be an error? */
> 				while ((*s|32)-'a' <= 'z'-'a') s++;
> 			}
> 			break;


Sorry, my bad.  Actually, musl has a bug.  It doesn't support `%Z` correctly.

Cc: @richfelker

ikerexxe:
@alejandro-colomar your latest updates have partially fixed the problem with dates in ISO 8601 format. Right now in Alpine (with musl) dates that are not correct, for example 2025-18-18 or tomorrow, do not return an error when they should.

alejandro-colomar:
> @alejandro-colomar your latest updates have partially fixed the problem with dates in ISO 8601 format. Right now in Alpine (with musl) dates that are not correct, for example 2025-18-18 or tomorrow, do not return an error when they should.

How can I reproduce that in Alpine?  I'm getting an error even for valid dates.

```c
/ # cat strptime.c 
#define _XOPEN_SOURCE
#include <stdio.h>
#include <time.h>

int
main(void)
{
	struct tm   tm;
	const char  *p;

	p = strptime("2025-01-18", "%F", &tm);
	if (p == NULL)
		perror("A");
	else if (*p != 0)
		perror("B");

	return 0;
}

/ # gcc -Wall -Wextra strptime.c 
/ # ./a.out 
A: No error information
```
```sh
/ # cat /etc/os-release 
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.21.3
PRETTY_NAME="Alpine Linux v3.21"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```

alejandro-colomar:
> > @alejandro-colomar your latest updates have partially fixed the problem with dates in ISO 8601 format. Right now in Alpine (with musl) dates that are not correct, for example 2025-18-18 or tomorrow, do not return an error when they should.
> 
> How can I reproduce that in Alpine? I'm getting an error even for valid dates.
> 
> ```c
> / # cat strptime.c 
> #define _XOPEN_SOURCE
> #include <stdio.h>
> #include <time.h>
> 
> int
> main(void)
> {
> 	struct tm   tm;
> 	const char  *p;
> 
> 	p = strptime("2025-01-18", "%F", &tm);
> 	if (p == NULL)
> 		perror("A");
> 	else if (*p != 0)
> 		perror("B");
> 
> 	return 0;
> }
> 
> / # gcc -Wall -Wextra strptime.c 
> / # ./a.out 
> A: No error information
> ```
> 
> ```shell
> / # cat /etc/os-release 
> NAME="Alpine Linux"
> ID=alpine
> VERSION_ID=3.21.3
> PRETTY_NAME="Alpine Linux v3.21"
> HOME_URL="https://alpinelinux.org/"
> BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
> ```

Hmmm, I've tried replacing `"%F"` by `"%Y-%m-%d"` and it works.  Okay.  :|

alejandro-colomar:
So, now that I can work with strptime(3) in Alpine, I've tried this:

```c
/ # cat strptime.c 
#define _XOPEN_SOURCE
#include <stdio.h>
#include <time.h>

int
main(void)
{
	struct tm   tm;
	const char  *p;

	p = strptime("2025-01-18", "%Y-%m-%d", &tm);
	if (p == NULL)
		perror("1A");
	else if (*p != 0)
		perror("1B");

	p = strptime("2025-11-18", "%Y-%m-%d", &tm);
	if (p == NULL)
		perror("2A");
	else if (*p != 0)
		perror("2B");

	p = strptime("2025-18-18", "%Y-%m-%d", &tm);
	if (p == NULL)
		perror("3A");
	else if (*p != 0)
		perror("3B");

	return 0;
}

/ # gcc -Wall -Wextra strptime.c 
/ # ./a.out 
3A: No error information
```

It seems invalid dates are correctly reported, aren't they?  It seems consistent with glibc.  @ikerexxe can you reproduce the invalid date bug you're reporting with a simple program that can be run in a docker container?

ikerexxe:
> Hmmm, I've tried replacing `"%F"` by `"%Y-%m-%d"` and it works. Okay. :|

I have also replaced `"%F"` by `"%Y-%m-%d"` and now it works correctly.

alejandro-colomar:
> > Hmmm, I've tried replacing `"%F"` by `"%Y-%m-%d"` and it works. Okay. :|
> 
> I have also replaced `"%F"` by `"%Y-%m-%d"` and now it works correctly.

Thanks!  I've fixed that in my PR.  Please check now.

ikerexxe:
Good! That solved all issues and now CI is green

ikerexxe:
Rebased on top of latest changes to https://github.com/shadow-maint/shadow/pull/1217

alejandro-colomar:
I'm unable to review this; let's see if @hallyn can.

BTW, how do you feel about these tests?  Are they comprehensive about my changes?  Do you feel my changes are safe with this?

alejandro-colomar:
@ikerexxe 

Once @danlavu reviews your tests, I propose the following:

-  Drop my patches from this PR, so that only your tests remain.
-  Once your tests are alone, they should still pass CI, which means that they verify that the old behavior is the same as the new one.
-  Then, we should merge this PR with only your tests.
-  Then, we can merge my code changes.

Does that sound good?

tests/system/tests/test_chage.py
```diff
@@ -0,0 +1,244 @@
...
+        ("0", 0),  # 1970-01-01
+        ("1", 1),  # 1970-01-02
+        ("20089", 20089),  # 2025-01-01
+        ("1000000", 1000000),  # This will happen in a very long time
+        ("1970-01-01", 0),
+        ("1970-01-02", 1),
+        ("2025-01-01", 20089),
+    ],
+)
+def test_chage__last_changed_correct_date(shadow: Shadow, last_changed_date: str, expected_date: int | None):
```
danlavu, line 28: Consider making the test names more descriptive, **test_chage_set_last_change_with_correct_date**

I've found it particularly helpful with the SSSD test transformation and viewing only the test case name. This comment also stems from my lack of knowledge about shadow utilities. Adding the verbs to the name will make it more straightforward to me, but it's up to you; this is entirely my preference.

I think the word **valid** instead of correct is better; it is entirely a grammatical nitpick.

tests/system/tests/test_chage.py
```diff
@@ -0,0 +1,244 @@
+"""
+Test chage
```
danlavu, line 2: I didn't know about the chage command. I thought it was a typo for the word change. **Test chage (Change Age)** would be nice for anybody who may not be familiar with the shadow utilities.

danlavu (approved):
I only have optional nitpicks. This is because I'm unfamiliar with shadow utilities. These suggestions would have helped me understand the tests faster. Anytime I paused looking at the code, I made a comment. I'm digging deep to find any comments. Everything looks great.

ikerexxe:
By the way, @danlavu  what do you think about the test themselves? Do you think we are checking all possibilities regarding the input/output values?

tests/system/tests/test_chage.py
alejandro-colomar: I disagree with this.  However, we could say `chage(1)`, to clarify that we refer to the program.

tests/system/tests/test_chage.py
alejandro-colomar: > What do you think of the following test names?
> 
>     * test_chage__set_last_changed_with_valid_date
> 
>     * test_chage__set_last_changed_with_invalid_date
> 
>     * test_chage__set_last_changed_with_empty_date
> 
>     * test_chage__set_expire_date_valid_date
> 
>     * test_chage__set_expire_date_invalid_date
> 
>     * test_chage__set_expire_date_empty_date
> 
>     * test_useradd__set_expire_date_valid_date
> 
>     * test_useradd__set_expire_date_invalid_date
> 
>     * test_useradd__set_expire_date_empty_date
> 
>     * test_usermod__set_expire_date_valid_date
> 
>     * test_usermod__set_expire_date_invalid_date
> 
>     * test_usermod__set_expire_date_empty_date

I like them.

> I think the word **valid** instead of correct is better; it is entirely a grammatical nitpick.

Agree.

alejandro-colomar: I prefer `date`.  It's more explicit about what we test.

alejandro-colomar: I like @ikerexxe 's suggested names

ikerexxe: I'll add the `with` to all test cases names (for some reason I removed it unintentionally) and I'll keep date as these are dates and no other thing.

alejandro-colomar (approved):
I don't know if @hallyn wants to review or comment anything.  I'm fine with merging the tests already.

Disclaimer: I can't read python, so I haven't reviewed much, other than having a quick look at it.

I approve the nomenclature of the functions, which is the only thing I feel qualified to review.  :)

From @danlavu 's comments, it seems the tests are okay, so I'll approve so that you can merge if you want.

Please remember to remove my commits before merging.

alejandro-colomar:
> Ok, some of the invalid dates aren't recognized as such without Alejandro's changes. I don't want to merge and leave CI broken for days, so I'll wait until Monday before proceeding.
> 
> @alejandro-colomar will you be available on Monday to work on the pending PR? This way CI won't be broken for a long time.

Could you please mention which tests don't pass, and why?  It would be good to be explicit about what is changing with my code.

ikerexxe:
Failed test cases are related to invalid dates that test dates outside the limits:
* 2025-18-18
* 2025-13-01
* 2025-01-32

IMO, the fortifications added at https://github.com/shadow-maint/shadow/pull/1217 improve the code.

tests/system/tests/test_chage.py
ikerexxe: Done

ikerexxe:
Since everyone agrees and the CI is green I will proceed to merge this PR

alejandro-colomar:
> Failed test cases are related to invalid dates that test dates outside the limits:
> 
>     * 2025-18-18
> 
>     * 2025-13-01
> 
>     * 2025-01-32
> 
> 
> IMO, the fortifications added at #1217 improve the code.

Ahhh, that makes sense.  It's because the yacc(1) code was passing the dates to mktime(3), which accepts that kind of thing, unlike strptime(3).

Issue #1216, opened 2025-02-18
4.17.4 release plan


alejandro-colomar:
Stuff to be cherry-picked for 4.17.3: 
 
-  <https://github.com/shadow-maint/shadow/pull/1215>  [Cc: @zeha]
-  <https://github.com/shadow-maint/shadow/pull/1212>  [Cc: @zeha]
-  <https://github.com/shadow-maint/shadow/pull/1214>  [Cc: @zeha]
-  <https://github.com/shadow-maint/shadow/pull/1207>  [Cc: @zeha]
-  <https://github.com/shadow-maint/shadow/pull/1196>  [Cc: @yonecle]
-  <https://github.com/shadow-maint/shadow/pull/1227>  [Cc: @zeha]

I expect to release it in a week or so.  It depends on all of the above being merged first, so I don't have a fixed date yet.

hallyn:
I'm proceeding with the 4.17.4 build, however,

Issue #1210, opened 2025-02-17
chfn -w "" fails


zeha:
This was found by systemd tests, involving adduser.

However it seems like it ought to work?

```
# useradd test1
# /bin/chfn "-f" "foo" "-r" "" "test1"
# /bin/chfn -w "" "test1"
chfn: invalid name: ',,'

# getent passwd test1
test1:x:1001:1001:foo,,,:/home/test1:/bin/sh
```

Would be a regression from 4.16.

https://bugs.debian.org/1096187

zeha:
git bisect points to commit 16cb664865541162c504a6f5ef5ca4b38b5e0c9a

alejandro-colomar:
Ughhh.  Thanks for the report!  I'll try to have priority for this.

alejandro-colomar:
> > git bisect points to commit [16cb664](https://github.com/shadow-maint/shadow/commit/16cb664865541162c504a6f5ef5ca4b38b5e0c9a)
> 
> Reverting just the chfn.c change from that commit seems to unbreak it.

Reverting it brings a lot of conflicts.  Did you revert the whole commit, or just the part that affects src/chfn.c?

PR #1234, opened 2025-03-03
Release 4.17.4

alejandro-colomar:
Closes: <https://github.com/shadow-maint/shadow/issues/1216>
Cc: Chris Hofstaedtler <zeha@debian.org>
Cc: <https://github.com/yonecle>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4,7 +4,7 @@ m4_define([libsubid_abi_major], 5)
 m4_define([libsubid_abi_minor], 0)
 m4_define([libsubid_abi_micro], 0)
 m4_define([libsubid_abi], [libsubid_abi_major.libsubid_abi_minor.libsubid_abi_micro])
-AC_INIT([shadow], [4.17.3], [pkg-shadow-devel@lists.alioth.debian.org], [],
+AC_INIT([shadow], [4.17.4], [pkg-shadow-devel@lists.alioth.debian.org], [],
 	[https://github.com/shadow-maint/shadow])
 AM_INIT_AUTOMAKE([1.11 foreign dist-xz subdir-objects tar-pax])
 AC_CONFIG_MACRO_DIRS([m4])


ikerexxe (approved):
LGTM!

jubalh:
We should come up with an idea how we want to write changelogs. We have both a [NEWS](https://github.com/shadow-maint/shadow/blob/master/NEWS) and a [ChangeLog](https://github.com/shadow-maint/shadow/blob/master/ChangeLog) file. Neither of which we update.

The git commit history is a mess to just drop in as a changefile due to having this ugly `directory/:` or `file:` infront of each commit. I don't get why we do this. We see the files changed in the rest of the commit. We don't need this in the subject and it just clutters it. Makes it harder to grasp what have been done and harder to generate changelogs.
We could have such prefixes for subsystems like `ci:`, `doc:` etc. But the directory? 

Compare this to a git commit with the subject [Improve the login.defs unknown item error message](https://github.com/shadow-maint/shadow/commit/fa68441bc4be8edeaad8d104344e987b31ed44e6). This can easily be dropped in a changelog.

Question1: *Why* do we prefix the dir/file on the git subject line?

Additionally we have a billion of commits that are more or less the same just regard different files. Like `src/useradd.c: Use !strcaseeq() instead of its pattern` and `lib/, src/: Use strcaseeq() instead of its pattern`.

To have this in separate commits is good practise. In a changelog we would want only one entry for something like this though.

Earlier we had nice merge commits. Which could be used exactly for this reason.
Take for example: https://github.com/shadow-maint/shadow/commit/39eea79d8d26472b756c76470ccc67f18504153e it sais `Resolve several compiler warnings` which is perfect for a changelog and the invididual commit messages that are mostly interesting for developers or if you want to check them are like `Declare file local functions static` and `Use strict prototypes` etc.
So to have merge commits makes total sense. Then we could have changelog entries like `Resolve several compiler warning #487`. And everybody would have an easy time to understand what changed in general, then could check the PR for the discussion and details.

Question2: *Why* did we get rid of merge commits?

One option could be to write changelogs from hand. Or we could use GitHub or any other tool to let it write them for us. But with our current git history those will only have suboptimal results.

The best way would be to have *sane* defaults so that we can use the proper tools that automate the changelog process.
For stable distributions with a long lifetime our current approach won't be the best.

Additionally I think we should create system tests upon each report of a regression like for example https://github.com/shadow-maint/shadow/issues/1210 to make sure it doesn't happen again.

So my suggestion would be to go back to have merge commits, drop the prefix in git subject lines and generate the changelog automatically upon a new release based on the merge commits. I would also volunteer to edit them by hand to have them in a better state in case automation will not lead to desired results.
I will also try to keep an eye open for issues and try to contribute tests to catch regressions earlier.

I think we should unify the NEWS and ChangeLog file into one and just link to the other. In my opinion having a file with just git history isn't necessary and one of those files should just be information for the end user to understand what changed and what to look out for.

alejandro-colomar:
Please don't merge.  While rebasing my -Werror= branch I have detected issues in master.

alejandro-colomar:
> > I personally don't find the changelogs useful for myself.
> 
> Because you are not a packager/distribution maintainer ;) People who only sporadically check this project and don't have deep insights what's being worked on and got changed need changelogs to understand that.
> 
> They need to know:
> 
>     * which new features are there
> 
>     * which bugs got fixed
> 
>     * which incompatibilities are there? did a feature vanish? was a binary dropped? support for PAM?
> 

These contents seem reasonable, and are what I include in the Linux man-pages changelog.  We could do that.

> 
> They usually don't need to know details about implementation. Until they suddenly sometimes do. Which is why the issue number or PR number from a merge commit would be helpful to read up on those details. Both the details of the discussion and seeing the individual commits.
> 
> > I refuse to have an automated changelog, but a manual one
> 
> The discussion about having automated or manually written changelogs is as old as having projects. There are various pros and cons. I listed above how to have them automated. The downside of the manual ones is obviously that someone needs to do it. And usually this is the point why projects (after some time) fall back to automated ones. Which is why I proposed a mix approach, with mostly automation.

When releasing versions of the Linux man-pages I usually take half an hour (or a couple of hours if the release is really huge) for writing it.  I wouldn't mind doing it myself.  I can take a similar approach.  Whenever we want to release, I can do the work on the Changelog.  Let's start with 4.18.0 for that?

ikerexxe:
> We should come up with an idea how we want to write changelogs. We have both a [NEWS](https://github.com/shadow-maint/shadow/blob/master/NEWS) and a [ChangeLog](https://github.com/shadow-maint/shadow/blob/master/ChangeLog) file. Neither of which we update.

We stopped using and updating those files some time ago, and I'd prefer to keep it that way. Those files are there for history purposes. Having said that, I agree that we should have some kind of release notes that are helpful for the audience (see below).

> The git commit history is a mess to just drop in as a changefile due to having this ugly `directory/:` or `file:` infront of each commit. I don't get why we do this. We see the files changed in the rest of the commit. We don't need this in the subject and it just clutters it. Makes it harder to grasp what have been done and harder to generate changelogs. We could have such prefixes for subsystems like `ci:`, `doc:` etc. But the directory?

The important topic here is the audience. git history is mainly used by developers, and I agree with Alejandro that having those type of messages is helpful for this type of audience. And now if I switch my hat to a distribution maintainer, I agree that those messages aren't very useful.  Maybe the developers should try to improve the commit body message so that other people apart from themselves understand what is happening.

> To have this in separate commits is good practise. In a changelog we would want only one entry for something like this though.

The release notes should be informative for a more general audience, like distribution maintainers and sysadmins, so I agree that they should contain more information about the actual issues fixed in a given release.

I would also like to have good release notes and have them created **automatically**. If you have any ideas on how to achieve this feel free to propose it.

> Additionally I think we should create system tests upon each report of a regression like for example #1210 to make sure it doesn't happen again.

As a final note. I completely disagree with having tests for each regression. I've worked in several projects where this happened and test maintenance was a nightmare. I'd prefer to have comprehensive tests where the most common use cases are tested. If some corner case is missing, that's fine, as long as the general use cases are working and users can use the software in their day to day.

ikerexxe:
> Release notes or News file - basically I don't care. But since the file already exist and had such a purpose.. Why not use it?

I'd say most people read Github release notes page. If anything.

> Which I just did in the comment above :)

Right, I don't know I wrote that.

> That's interesting, what exactly was a nightmare? I can't see any downsides and thought the purpose of tests is to make sure functionality is as expected.

The sheer number of tests is the problem and the little value they provide. Since we are working on shadow let me show you test example that was written a **LONG** time ago and that we currently run to test our distributions:

```
#!/bin/bash                                                                      
                                                                                 
# Make sure faillog doesn't take ages on 64-bit systems                          
# BZ #177017 for more details.                                                   
                                                                                 
# to prevent other error messages                                                
if [ $EUID -ne 0 ]; then                                                         
        echo "This test requires root privileges"                                
        exit 1                                                                   
fi                                                                               
                                                                                 
touch /var/log/faillog                                                           
faillog -m 32767 -r
```

This test was created to test https://bugzilla.redhat.com/show_bug.cgi?id=177017. In my opinion, the value of this test today is almost nil.

From a more theoretical point of view, a feature like useradd would need thousands of tests to be fully tested. Given that it is unlikely that we will ever have that many tests, and given that we have users using this feature every day it is quite likely that someone will come up with the extraordinary idea that the ID can be a set of alphanumeric characters. Example: `useradd -u ten` or `useradd -u 0xae89`. The user could argue that nowhere `ten` or `0xae89` are numbers, but expressed in another way, or that it used to work like that twenty years ago. These would be valid regression tests, but their value would be very low.

I hope you understand my position on creating regression tests for all cases. I am not saying that regression tests should not be created, just that we should use our heads before creating them because this part is usually easy but then maintaining them during the years that the software is used is another matter.

hallyn:
For the kind of thing @jubalh is asking for, I don't think spending a few hours at release time is the best way.  It would work better to try and get PRs to update the changelog when they do something a package maintainer would want to know.  The trick of course is how to know when that's the case.

That, and then I start feeling like people who's changes aren't called out will feel slighted :)

So I end up just doing 'git shortlog', because actually I think that does pretty well.

alejandro-colomar:
> For the kind of thing @jubalh is asking for, I don't think spending a few hours at release time is the best way. It would work better to try and get PRs to update the changelog when they do something a package maintainer would want to know. The trick of course is how to know when that's the case.

That would add more friction to the PRs.  For example, GCC does this in the commit messages, and I find it very cumbersome to update the changelogs correctly.  In their case, it's fair, since their releases are huge, and updating the changelog at release time wouldn't be doable.

> That, and then I start feeling like people who's changes aren't called out will feel slighted :)

Yep, that's a thing.  I try to make it objective:

-  If behavior changes, it should be noted.
-  If distros need to update anything, it should be noted.

Otherwise, I tend to not mention it.  Unless for some reason I make an exception.  Exceptions break that objectivity, but hopefully not too much.  Also, I add a few lines saying that there are changes not mentioned there, but that are present in the release.

Example from this morning, from the release of man-pages-6.13:
<https://lore.kernel.org/linux-man/ft2zbwghk6hbnhlawjigckegrk3yekyeiudp43gafoububm5ma@mbkqcvh4idyf/T/#u>
 
> So I end up just doing 'git shortlog', because actually I think that does pretty well.

I'm okay if you prefer that.  But if we go that way, my only ask is that the changelog file not be updated.  I'd keep that only in the release notes.

hallyn:
Yeah, that's what I'm doing, putting it in the release notes.

If you're already clickity-clicking in the github webui, I think that's a lot easier than alternatives.

alejandro-colomar:
@hallyn , master is clean IMO, so you can release from there if you want.  How about you release on St. Joseph's say?

hallyn:
> @hallyn , master is clean IMO, so you can release from there if you want. How about you release .4 on St. Joseph's say?

This coming Wednesday?

alejandro-colomar:
> > @hallyn , master is clean IMO, so you can release from there if you want. How about you release .4 on St. Joseph's say?
> 
> This coming Wednesday?

Yep.  :)

jubalh:
> That, and then I start feeling like people who's changes aren't called out will feel slighted :)

I think a software like shadow should primarily focus on what are the big changes. So not each change will be mentioned and probably contributors will feel fine since they are not the only one left out. Another approach would be to have some kind of credits at the end of the release notes where just every contributor for this release is listed under "thanks" or something.

alejandro-colomar:
> > That, and then I start feeling like people who's changes aren't called out will feel slighted :)
> 
> I think a software like shadow should primarily focus on what are the big changes. So not each change will be mentioned and probably contributors will feel fine since they are not the only one left out. Another approach would be to have some kind of credits at the end of the release notes where just every contributor for this release is listed under "thanks" or something.

In the Linux man-pages I consider contributor everyone who reported bugs, contributed code, reviewed, or commented on a discussion (I keep track of those via Cc: and other tags in commits) for the given release.  If only people who wrote code are thanked, other kinds of contributors might feel slighted.

hallyn:
Oops, sorry, I forgot you did this as a pr not an issue, so i merged my own.

While I went with git shortlog, I'm still open to further discussions about other ways to generate release notes.  I agree we should provide to the distro packagers what is useful to them.

alejandro-colomar:
> Oops, sorry, I forgot you did this as a pr not an issue, so i merged my own.

No problem.  Thanks!

PR #1236, opened 2025-03-10
Use templates to avoid repetition, and to add type safety

alejandro-colomar:
Cc: @uecker 
Cc: @chrisbazley

This is an experiment about mimicking templates in C with macros.  If we get rid of the trailing ` \`, this looks quite nice.


lib/atoi/strtoi/strton.h
```diff
@@ -0,0 +1,69 @@
...
+#include "attr.h"
+
+
+#define template_STRTON(T)                                            \
+ATTR_STRING(1) ATTR_ACCESS(write_only, 2) ATTR_ACCESS(write_only, 6)  \
+inline T                                                       \
+STRTON(T,                                                             \
+    const char *s, char **restrict endp, int base,                    \
+    T min, T max, int *restrict status)                               \
+{                                                                     \
```
alejandro-colomar, line 27: Except for the attribute parameter references being off-by-one, this looks very cool.

uecker:
Am Montag, dem 10.03.2025 um 09:39 -0700 schrieb Alejandro Colomar:
> @alejandro-colomar commented on this pull request.
> In lib/atoi/strtoi/strton.h:
> > +                                                                      \
> +	if (*endp == s)                                               \
> +		*status = ECANCELED;                                  \
> +	else if (errno == ERANGE || n < min || n > max)               \
> +		*status = ERANGE;                                     \
> +	else if (**endp != '\0')                                      \
> +		*status = ENOTSUP;                                    \
> +	else                                                          \
> +		*status = 0;                                          \
> +                                                                      \
> +	errno = e;                                                    \
> +                                                                      \
> +	return MAX(min, MIN(max, n));                                 \
> +}
> +//#enddef
> +#define STRTON(T, ...)  STRTON__ ## T(__VA_ARGS__)
> I could actually fix the off-by-one attributes by moving the T to the end of the parameter list, and using the explicit parameter list here instead of ....

I would just write it as STRTON__ ## in the definition
instead of using the STRTON macro, which would get rid
fo the off-by-one.  I also hides less of mechanism which
I think is good.

I also wonder whether you can replace the code in
the c file with something such as

extern inline typeof(STRTON__ ## uintmax_t) STRTON__ ## uintmax_t;

instead of repeating all the arguments.

alejandro-colomar:
> Am Montag, dem 10.03.2025 um 09:39 -0700 schrieb Alejandro Colomar:
> I would just write it as STRTON__ ## in the definition instead of using the STRTON macro, which would get rid fo the off-by-one. I also hides less of mechanism which I think is good.

The good thing about using STRTON() is that it looks like a function definition.  grepc(1) (my regex-based program for grepping C code) finds it just fine.  However, if I use ##, then the regex doesn't find it.

> I also wonder whether you can replace the code in the c file with something such as extern inline typeof(STRTON__ ## uintmax_t) STRTON__ ## uintmax_t; instead of repeating all the arguments.

I'm not too worried, since that repetition is checked by the compiler.  Also, the same argument applied: regex-based programs will find the prototypes using STRTON().

alejandro-colomar:
> > Am Montag, dem 10.03.2025 um 09:39 -0700 schrieb Alejandro Colomar:
> > I would just write it as STRTON__ ## in the definition instead of using the STRTON macro, which would get rid fo the off-by-one. I also hides less of mechanism which I think is good.
> 
> The good thing about using STRTON() is that it looks like a function definition. grepc(1) (my regex-based program for grepping C code) finds it just fine. However, if I use ##, then the regex doesn't find it.
> 
> > I also wonder whether you can replace the code in the c file with something such as extern inline typeof(STRTON__ ## uintmax_t) STRTON__ ## uintmax_t; instead of repeating all the arguments.
> 
> I'm not too worried, since that repetition is checked by the compiler. Also, the same argument applied: regex-based programs will find the prototypes using STRTON().
> 
> https://www.alejandro-colomar.es/src/alx/alx/grepc.git/tree/bin/grepc

@uecker 

Actually, I correct myself.  grepc(1) is able to grep the template definition with `##` in its name as long as you include that in the search term.

```c
$ grepc CMP .
./cmp.h:#define CMP(T)  cmp__ ## T  // For use as a callback
```
```c
$ grepc -C1 'cmp__ ## T' .
./cmp.h-#def template_cmp(T)
./cmp.h:inline int
cmp__ ## T(const void *key, const void *elt)
{
	const T  *k = key;
	const T  *e = elt;

	if (*k < *e)
		return -1;
	if (*k > *e)
		return +1;
	return 0;
}
./cmp.h-#enddef
```

So, I'll follow your suggestion and make it explicit that I'm using `##` in the template definition.  It will make the machinery more readable.

alejandro-colomar:
> I also wonder whether you can replace the code in the c file with something such as extern inline typeof(STRTON__ ## uintmax_t) STRTON__ ## uintmax_t; instead of repeating all the arguments.
> […](#)

No, you can't.

```c
search/cmp/cmp.c:12:28: error: stray '##' in program
   12 | extern inline typeof(CMP__ ## int)     CMP__ ## int;
      |                            ^~
search/cmp/cmp.c:12:22: error: 'CMP__' undeclared here (not in a function); did you mean 'CMP'?
   12 | extern inline typeof(CMP__ ## int)     CMP__ ## int;
      |                      ^~~~~
      |                      CMP
search/cmp/cmp.c:12:27: error: expected ')' before 'int'
   12 | extern inline typeof(CMP__ ## int)     CMP__ ## int;
      |                     ~     ^   ~~~
      |                           )
```

Concatenation doesn't work outside of macros.  I guess I'll either concatenate myself, or write a small macro for that.

PR #1237, opened 2025-03-12
Extend basic test cases to check shadow and gshadow entries

ikerexxe:
The test framework PoC only provided basic checks. I've added additional
functionality to the framework by checking shadow and gshadow entries
and I've extended the basic test cases to check those too.

This should give us more coverage over the changes we make to the basic
functionality and ensure that we don't change anything inadvertently.

### DIFF
--- a/tests/system/framework/misc/__init__.py
+++ b/tests/system/framework/misc/__init__.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import datetime
 from typing import Any
 
 
@@ -40,3 +41,14 @@ def to_list_of_strings(value: Any | list[Any] | None) -> list[str]:
     :rtype: list[str]
     """
     return [str(x) for x in to_list(value)]
+
+
+def days_since_epoch():
+    """
+    Gets the current date and returns the number of days since 1970-01-01 UTC, this time is also referred to as the
+    Unix epoch.
+    """
+    epoch = datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone.utc)
+    now_utc = datetime.datetime.now(datetime.timezone.utc)
+    delta = now_utc - epoch
+    return delta.days
--- a/tests/system/etc/login.defs
+++ b/tests/system/etc/login.defs
@@ -446,7 +446,7 @@ CREATE_HOME     yes
 # Force use shadow, even if shadow passwd & shadow group files are
 # missing.
 #
-#FORCE_SHADOW    yes
+FORCE_SHADOW    yes
 
 #
 # Allow newuidmap and newgidmap when running under an alternative
--- a/tests/system/framework/hosts/base.py
+++ b/tests/system/framework/hosts/base.py
@@ -54,7 +54,8 @@ class BaseLinuxHost(MultihostHost[ShadowMultihostDomain]):
         """
         self.logger.info(f"Detecting distro information on {self.hostname}")
         os_release = self.fs.read("/etc/os-release")
-        self._os_release = dict(csv.reader([x for x in os_release.splitlines() if x], delimiter="="))
+        valid_lines = [line for line in os_release.splitlines() if line and not line.startswith("#")]
+        self._os_release = dict(csv.reader(valid_lines, delimiter="="))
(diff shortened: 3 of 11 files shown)


ikerexxe:
This PR is blocked by https://gitlab.alpinelinux.org/alpine/aports/-/issues/16979. Apparently Alpine's getent doesn't provide `gshadow` entry reading, so I requested it to be included.

ikerexxe:
@jubalh do you know why opensuse builder is failing? It seems a dependency installation problem:
```
Problem: 1: the to be installed libsemanage-devel-3.8-1.2.x86_64 requires 'libsemanage2 = 3.8', but this requirement cannot be provided
not installable providers: libsemanage2-3.8-1.2.x86_64[repo-oss]
Problem: 2: the to be installed libselinux-devel-3.8-2.2.x86_64 requires 'libselinux1 = 3.8', but this requirement cannot be provided
not installable providers: libselinux1-3.8-2.2.x86_64[repo-oss]
```

danlavu (commented):
I'd like it a lot more if it followed the structure of the other roles.

shadow.user.add()
shadow.user.get()
shadow.user.modify()
shadow.group.add()
shadow.group.get()
shadow.group.modify()
tools.getent_shadow()

So we can abstract the class?

ikerexxe:
> shadow.user.add() shadow.user.get() shadow.user.modify() shadow.group.add() shadow.group.get() shadow.group.modify()

We have decided to follow an approach as similar as possible to the CLI in shadow. So, instead of useradd or groupadd the framework should provide shadow.useradd() or shadow.groupadd(). If you want a little more context I recommend you read https://github.com/shadow-maint/shadow/issues/835#issuecomment-1886825156.

> tools.getent_shadow()

The behaviour with getent is the same one as in SSSD. What you mention there would be different.

danlavu:
> >  If you want a little more context I recommend you read [#835 (comment)](https://github.com/shadow-maint/shadow/issues/835#issuecomment-1886825156).
> 

Ack, I will read and continue the review then. Thank you for the clarification.

tests/system/framework/misc/__init__.py
```diff
@@ -40,3 +41,13 @@ def to_list_of_strings(value: Any | list[Any] | None) -> list[str]:
     :rtype: list[str]
     """
     return [str(x) for x in to_list(value)]
+
+
+def days_since_epoch():
+    """
+    Gets the current date and returns the number of days since the Unix epoch (1970-01-01 UTC).
```
danlavu, line 48: Interesting, I did not know what the Unix epoch was, rephrasing it a little would clarify it. 

``` Gets the current date and returns the number of days since 01-01-1970 UTC, this time is also referred to as the Unix epoch. ```

danlavu (approved):
Everything looks great, there is only one thing I do not like, and it's mostly a personal preference. Just the number of test files, I find it easier to work on in one file than having to open several files to view the tests. My only suggestion is to create one test file ``` test_shadow.py ``` and have test case names like ``` test_shadow__groupadd ```, etc. 

Going to approve it, kindly take the single test file into consideration. Having a file for one test seems strange to me.

tests/system/framework/misc/__init__.py
ikerexxe: Updated

ikerexxe: Updated

danlavu:
That makes perfect sense, disregard my comment then.

ikerexxe:
I'd like to proceed merging this PR but I need at least one approval (Dan only has read-only permissions and apparently doesn't count). @hallyn @alejandro-colomar can any of you approve the PR?

By the way, Serge, would you mind increasing Dan's permissions so that his reviews count? He's a colleague of mine and has a great track record in testing.

alejandro-colomar (approved):
Thanks!

Disclaimer:  I have not reviewed this, but trust you if you think it's okay.  If you want more review, maybe wait for @hallyn .

ikerexxe:
No, I think Dan's review is enough so let's merge it. Thank you for "approving" :wink:

PR #1244, opened 2025-03-22
passwd: document exit code when PAM has errored

### LINKED ISSUE #1219: undocumented return values from passwd
Hi!

passwd returns exit code 10 which is not defined in the manual page.

```
[2/5164]mh@swivel:~ $ passwd
Changing password for mh.
Current password:
passwd: Authentication token manipulation error
passwd: password unchanged
10 [3/5165]mh@swivel:~ $ passwd
Changing password for mh.
Current password:
New password:
Retype new password:
Sorry, passwords do not match.
passwd: Authentication token manipulation error
passwd: password unchanged
10 [4/5165]mh@swivel:~ $
```

passwd(1) only documens exit codes 1 through 6.

Please also consider giving distinct exit codes for "current password entered wrong" and "password repeat doesnt match".

Greetings
Marc

hallyn:
closes #1219

When pam returns an error, we were exiting with exit code 10, which was hardcoded and not documented.  Create a name for it, and document it in the manpage.

Reported-by: Marc Haber <githubvisible@zugschlus.de>

### DIFF
--- a/man/passwd.1.xml
+++ b/man/passwd.1.xml
@@ -484,6 +484,12 @@
 	    <para>invalid argument to option</para>
 	  </listitem>
 	</varlistentry>
+	<varlistentry>
+	  <term><replaceable>10</replaceable></term>
+	  <listitem>
+	    <para>an error was returned by pam(3).</para>
+	  </listitem>
+	</varlistentry>
       </variablelist>
     </para>
   </refsect1>
--- a/lib/pam_pass.c
+++ b/lib/pam_pass.c
@@ -40,7 +40,7 @@ void do_pam_passwd (const char *user, bool silent, bool change_expired)
 	if (ret != PAM_SUCCESS) {
 		fprintf (shadow_logfd,
 			 _("passwd: pam_start() failed, error %d\n"), ret);
-		exit (10);	/* XXX */
+		exit (E_PAM_ERR);
 	}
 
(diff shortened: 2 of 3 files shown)


man/passwd.1.xml
```diff
@@ -484,6 +484,12 @@
 	    <para>invalid argument to option</para>
 	  </listitem>
 	</varlistentry>
+	<varlistentry>
+	  <term><replaceable>10</replaceable></term>
+	  <listitem>
+	    <para>an error was returned by pam(3).</para>
```
alejandro-colomar, line 490: The other bullets don't have a trailing '.'.

Also, the man(3) reference should use formatting:

```
<refentrytitle>pam</refentrytitle><manvolnum>3</manvolnum>
```

alejandro-colomar (approved):
Except for the minor issues noted in the comment,

```
```

@zugschlus is this enough for you?  (I think the separation of error codes would be a separate thing.)

man/passwd.1.xml
hallyn: Oh, thanks for that.

It does make the raw xml a lot harder to read :)

alejandro-colomar: I can rewrite these pages to man(7), if you want.  ;-)

alejandro-colomar:
Thanks!  I'll merge.

PR #1246, opened 2025-03-28
Accept /usr/sbin/nologin as an alternate to /sbin/nologin

zeha:
Relevant on fully-usr-merged distributions.

Reported-by: Marc Haber @zugschlus

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -1377,6 +1377,7 @@ static void process_flags (int argc, char **argv)
 				if (!streq(optarg, "")
 				     && '*'  != optarg[0]
 				     && !streq(optarg, "/sbin/nologin")
+				     && !streq(optarg, "/usr/sbin/nologin")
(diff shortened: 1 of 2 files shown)


src/useradd.c
```diff
@@ -1377,6 +1377,7 @@ static void process_flags (int argc, char **argv)
 				if (!streq(optarg, "")
 				     && '*'  != optarg[0]
 				     && !streq(optarg, "/sbin/nologin")
+				     && !streq(optarg, "/usr/sbin/nologin")
```
alejandro-colomar, line 1380: Don't such systems have a symlink /sbin => /usr/sbin?

I'm okay with the change, but is it really necessary?

alejandro-colomar: Although yes, one may specify it as /usr/... in the command line, in which case we need to support it.

alejandro-colomar (approved):
```
```

Thanks!

PR #1248, opened 2025-04-09
Add LOGIN_ENV_SAFELIST to FOREIGNDEFS

stanislav-brabec:
util-linux-2.41 introduced new variable: LOGIN_ENV_SAFELIST. Add it to known login.defs variables.

### DIFF
--- a/lib/getdef.c
+++ b/lib/getdef.c
@@ -76,6 +76,7 @@ struct itemdef {
 #define FOREIGNDEFS				\
 	{"ALWAYS_SET_PATH", NULL},		\
 	{"ENV_ROOTPATH", NULL},			\
+	{"LOGIN_ENV_SAFELIST", NULL},		\
 	{"LOGIN_KEEP_USERNAME", NULL},		\
 	{"LOGIN_PLAIN_PROMPT", NULL},		\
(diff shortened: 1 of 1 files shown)


ikerexxe (approved):
LGTM! Thanks for making the effort to port this here.

stanislav-brabec:
We have a tool for that in openSUSE. Would the upstream be interested? It is an ugly hack that parses precompiler output of shadow and sources of pam and util-linux:
https://build.opensuse.org/projects/Base:System/packages/shadow/files/shadow-login_defs-check.sh?expand=1

And additional patches:

Things that needs to be changed in login.defs to switch from shadow login to util-linux login:
https://build.opensuse.org/projects/Base:System/packages/shadow/files/shadow-util-linux.patch?expand=1

Things that cannot be used if pam tools are used in favor of shadow tools:
https://build.opensuse.org/projects/Base:System/packages/shadow/files/shadow-login_defs-unused-by-pam.patch?expand=1

Comments improvements (again, may depend on the binary selection between shadow, util-linux and pam):
https://build.opensuse.org/projects/Base:System/packages/shadow/files/shadow-login_defs-comments.patch?expand=1

PR #1252, opened 2025-04-25
ci: add `gawk` as a fedora dependency

ikerexxe:
Recently fedora 42 was released and `gawk` was missing as a dependency.


alejandro-colomar (approved):
Thanks!

```
```

Issue #1253, opened 2025-04-29
No error message from `newuidmap` on `sub_uid_open` failure


matthewhughes934:
Reproduction:

1. Make `/etc/subuid` a symlink
2. In your favourite shell run: `newuidmap` with any valid args, e.g. `newuidmap $$ ...`

The process will exit with 1 and no message printed.

`strace` shows the cause of the issue:

```
$ strace --failed-only -e trace=openat -- newuidmap $$
openat(AT_FDCWD, "/etc/subuid", O_RDONLY|O_NOCTTY|O_NONBLOCK|O_NOFOLLOW|O_CLOEXEC) = -1 ELOOP (Too many levels of symbolic links)
+++ exited with 1 +++
```

This is a bit painful to debug, I think it would be helpful if the program could print the message from the error, e.g.

```
newuidmap: cannot open /etc/subuid: Too many levels of symbolic links
```

Here's a naive patch that achieves this (a better solution would be using the standard logging process for this repo, but I didn't look too deep to see what those are)

```patch
diff --git i/src/newuidmap.c w/src/newuidmap.c
index 8deb06f0..e62649d0 100644
--- i/src/newuidmap.c
+++ w/src/newuidmap.c
@@ -139,6 +139,9 @@ int main(int argc, char **argv)
 	}
 
 	if (!sub_uid_open(O_RDONLY)) {
+		fprintf (stderr,
+		         _("%s: cannot open %s: %s\n"),
+		         Prog, sub_uid_dbname (), strerror (errno));
 		return EXIT_FAILURE;
 	}
```

My real-world use case: was running rootless docker (specifically the one packaged here: https://aur.archlinux.org/packages/docker-rootless-extras) with `/etc/subuid` as a symlink, and was left debugging from a log:

            <-- SNIP -->
            Mar 18 20:17:04 desktop dockerd-rootless.sh[14113]: + exec rootlesskit --state-dir=/run/user/1000/dockerd-rootless --net=slirp4netns --mtu=65520 --slirp4netns-sandbox=auto --slirp4netns-seccomp=auto --disable-host-loopback --port-driver=builtin --copy-up=/etc --copy-up=/run --propag>
            Mar 18 20:17:04 desktop dockerd-rootless.sh[14113]: [rootlesskit:parent] error: failed to setup UID/GID map: newuidmap 14124 [0 1000 1 1 165536 65536] failed: : exit status 1
            <-- SNIP -->

[...]

PR #1254, opened 2025-04-29
newuidmap: better error logging on failure

matthewhughes934:
The handling for `sub_uid_open` relies on `commonio_open` preserving `errno`, which it appears to make an effort do, but doesn't explicitly document.

Closes: #1253

### DIFF
--- a/src/newgidmap.c
+++ b/src/newgidmap.c
@@ -190,8 +190,9 @@ int main(int argc, char **argv)
 
 	/* Get the effective uid and effective gid of the target process */
 	if (fstat(proc_dir_fd, &st) < 0) {
-		fprintf(stderr, _("%s: Could not stat directory for process\n"),
-			Prog);
+		fprintf(stderr,
+		        _("%s: Could not stat directory for target process: %s\n"),
+		        Prog, strerror (errno));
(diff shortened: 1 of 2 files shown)


ikerexxe:
The code changes look good, but there are two things to improve. Can you change the `Fixes` in your commit message by `Closes`? That would be more appropriate to use.

On top of that, can you add another commit fixing `newgidmap.c`? The code changes should be the same, or very similar, to the ones you provided for `newuidmap.c`

matthewhughes934:
Thanks for the review

> The code changes look good, but there are two things to improve. Can you change the `Fixes` in your commit message by `Closes`? That would be more appropriate to use.

Done: [5cc516b](https://github.com/shadow-maint/shadow/pull/1254/commits/5cc516b9124980f1f06379624ffbbd4e62f73ee2)

> On top of that, can you add another commit fixing `newgidmap.c`? The code changes should be the same, or very similar, to the ones you provided for `newuidmap.c`

Done: [ba0daae](https://github.com/shadow-maint/shadow/pull/1254/commits/ba0daaee08b83740f5a8fb854300aa131dd5fe1d)

Issue #940, opened 2024-02-07
The --chroot option doesn't work well with SELinux off in-chroot


praiskup:
Consider installing a separate minimal chroot, and then running `useradd --root` (other shadow-utils utility):

```bash
$ sudo dnf5 -y --use-host-config --installroot /tmp/newroot install /bin/sh
$ sudo useradd jdoe --root /tmp/newroot
useradd: failure while writing changes to /etc/passwd
```

I used GDB to detect what was going on. It is the call to `set_selinux_file_context ()` at file-descriptor-closing-time.

The thing is that `set_selinux_file_context()` is called non-conditionally every time, and it asks for `is_selinux_enabled()` - which provides the info about SELinux [on-host, not in-chroot](https://github.com/SELinuxProject/selinux/issues/419):

https://github.com/shadow-maint/shadow/blob/3e59e9613ec40c51c19c7bb5c28468e33a4529d5/lib/selinux.c#L47-L52

A further assumption that the `selabel_open()`, `selabel_lookup_raw()`, etc. in-chroot will work, is wrong.

There are further checks for permissive mode that make shadow utils not fail (in theory):
https://github.com/shadow-maint/shadow/blob/e367d111e513a61495213028e8d4ab2e36775790/lib/selinux.c#L66-L68

But one would have to make sure that the `/tmp/newroot/sys/fs/selinux/enforce` file exists first, and contains 0.

This problem is causing problems to [Mock (RPM chrooted builds)](https://github.com/rpm-software-management/mock/issues/1285).  We could fake the `/sys/fs/selinux/enforce` file for sure in Mock (we do other hacks to make things work), though this isn't the ideal solution;  the mode in-chroot isn't really "permissive", it is supposed to be "disabled". So we should ideally not even attempt to call `selabel_open()`, etc.

While I see this probably isn't a mistake of shadow-utils (but rather chroot-unfriendly linbselinux API), I'm curious what is the best way out of this.  Could we have a new `--no-selinux` option?

Also, note that the manual pages are a bit misleading:

```
       -P, --prefix PREFIX_DIR
           Apply changes to configuration files under the root filesystem
           found under the directory PREFIX_DIR. This option does not chroot
           and is intended for preparing a cross-compilation target. Some
           limitations: NIS and LDAP users/groups are not verified. PAM
           authentication is using the host files. No SELINUX support.
```

[...]

PR #1258, opened 2025-05-05
src/useradd.c: chroot or prefix SELinux file context

ikerexxe:
Do not process SELinux file context during file closure when chroot or
prefix options are selected.

As I'm changing a lot of files I decided to split the changes in a set of
patches to make them easier to understand.

Tests: https://github.com/shadow-maint/shadow/issues/940
Closes: https://github.com/shadow-maint/shadow/issues/940


ikerexxe:
This is blocked by https://github.com/next-actions/pytest-mh/pull/101

ikerexxe:
@praiskup and @pmatilai I have created a [COPR repository](https://copr.fedorainfracloud.org/coprs/ipedrosa/useradd-chroot/) to help test these changes. Do you mind testing them?

By the way, if you need the build in some other distribution let me know.

pmatilai:
Hmm, I'm afraid I don't see any behavior change from this on the rpm case (this on Fedora 41'ish):

> [root@lumikko tmp]# rpm -q shadow-utils
shadow-utils-4.17.4-30test.fc41.x86_64
[root@lumikko tmp]# rm -rf /srv/test; rpm -U --root /srv/test/ setup-2.15.0-8.fc41.noarch.rpm --nodeps --nosignature --define "__systemd_sysusers /usr/lib/rpm/sysusers.sh"
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
groupadd: failure while writing changes to /etc/group
useradd: group '4' does not exist
useradd: group '1' does not exist
useradd: group '2' does not exist
useradd: group '50' does not exist
useradd: group '100' does not exist
useradd: group '0' does not exist
useradd: group '7' does not exist
useradd: group '12' does not exist
useradd: group '65534' does not exist
useradd: group '0' does not exist
useradd: group '0' does not exist
useradd: group '0' does not exist
useradd: group '0' does not exist
[root@lumikko tmp]#

That's how it fails with the stock F41 shadow-utils-4.15.1-12.fc41.x86_64 too.

ikerexxe:
That's expected behaviour as I only proposed the fix for useradd and groupadd binaries, everything else isn't fixed yet.

@alejandro-colomar @hallyn would you mind reviewing the general concept of this PR? You can skip the testing and just review how I propose to handle the propagation of the `process_selinux` flag from the various binaries to the library that handles file management. There are already quite a few commits in this PR and making all the binaries behave the same way is going to increase them even more, so I would like to be sure that this is the best way to solve this problem before proceeding with the rest.

alejandro-colomar:
> That's expected behaviour as I only proposed the fix for useradd and groupadd binaries, everything else isn't fixed yet.
> 
> @alejandro-colomar @hallyn would you mind reviewing the general concept of this PR? You can skip the testing and just review how I propose to handle the propagation of the `process_selinux` flag from the various binaries to the library that handles file management. There are already quite a few commits in this PR and making all the binaries behave the same way is going to increase them even more, so I would like to be sure that this is the best way to solve this problem before proceeding with the rest.

Yep, I'll review.

pmatilai:
Doh, I've missed the update round here. I'll try to retest soon, thanks for looking into this!

src/useradd.c
```diff
@@ -741,11 +750,16 @@ set_defaults(void)
...
+static int get_groups (char *list, struct option_flags *flags)
 {
 	struct group *grp;
 	bool errors = false;
 	int ngroups = 0;
+	bool process_selinux = false;
+
+	if (!flags->chroot && !flags->prefix) {
+		process_selinux = true;
+	}
```
alejandro-colomar: Please don't use braces for a single statement.

alejandro-colomar: I'd write this as

```c
bool  process_selinux;

process_selinux = flags->chroot || flags->prefix;
```

That removes one initialization.  Initialization (as opposed to assignment) is usually dangerous, as it removes the ability of the compiler to find dead code and code paths that don't have the expected results.

It also removes one conditional.  The complexity of source code is significantly affected by the number of conditionals.

src/useradd.c
```diff
@@ -1573,38 +1589,44 @@ static void process_flags (int argc, char **argv)
...
  */
-static void close_files (void)
+static void close_files (struct option_flags *flags)
 {
-	if (pw_close (true) == 0) {
+	bool process_selinux = false;
+
+	if (!flags->chroot && !flags->prefix) {
+		process_selinux = true;
+	}
```
alejandro-colomar: Same here.

src/useradd.c
```diff
@@ -1573,38 +1589,44 @@ static void process_flags (int argc, char **argv)
...
+
+	if (pw_close (process_selinux) == 0) {
```
alejandro-colomar, line 1582: Please don't add spaces between a function name and it's opening parenthesis.  I don't mind keeping the old ones, but please don't add new ones.

src/useradd.c
```diff
@@ -2420,6 +2442,7 @@ int main (int argc, char **argv)
 #endif
 	unsigned long subuid_count = 0;
 	unsigned long subgid_count = 0;
+	struct option_flags flags;
```
alejandro-colomar: Please use two spaces between the type name and the variable name.  This helps distinguish `struct option_flags` from `flags` more easily.

src/useradd.c
alejandro-colomar: Ahh, sorry, it was an already-existing one.  It's just that github formats the diff weirdly.  Ignore my comment.

src/useradd.c
```diff
@@ -2226,11 +2226,13 @@ static void create_home (void)
...
-			_("%s: cannot set SELinux context for home directory %s\n"),
-			Prog, user_home);
-		fail_exit(E_HOMEDIR);
+	if (!flags->chroot && !flags->prefix) {
+		if (set_selinux_file_context(prefix_user_home, S_IFDIR) != 0) {
+			fprintf(stderr,
+				_("%s: cannot set SELinux context for home directory %s\n"),
+				Prog, user_home);
+			fail_exit(E_HOMEDIR);
+		}
```
alejandro-colomar, line 2225: How about doing

```c
bool  process_selinux;

process_selinux = flags->chroot || flags->prefix;
```

like in the previous commit?  That would make these conditionals simpler.

src/useradd.c
```diff
@@ -2311,12 +2313,14 @@ static void create_home (void)
...
-			Prog);
-		fail_exit(E_HOMEDIR);
+	if (!flags->chroot && !flags->prefix) {
+		/* Reset SELinux to create files with default contexts */
+		if (reset_selinux_file_context() != 0) {
+			fprintf(stderr,
+				_("%s: cannot reset SELinux file creation context\n"),
+				Prog);
+			fail_exit(E_HOMEDIR);
+		}
```
alejandro-colomar, line 2313: Same here.

lib/cleanup_group.c
```diff
@@ -199,9 +201,11 @@ void cleanup_unlock_group (MAYBE_UNUSED void *arg)
  *
  * It should be registered after the gshadow file is successfully locked.
  */
-void cleanup_unlock_gshadow (MAYBE_UNUSED void *arg)
+void cleanup_unlock_gshadow (void *process_selinux)
 {
-	if (sgr_unlock (true) == 0) {
+	bool process = *((bool *)process_selinux);
```
alejandro-colomar: Please use a space between a cast and the expression it casts.  Using a consistent style makes it easier to grep for casts.

Alternatively, you could use a temporary pointer:

```c
void
cleanup_unlock_gshadow(void *process_selinux)
{
        const bool  *ps = process_selinux;
        bool        process = *ps;
```

which avoids the cast altogether.

alejandro-colomar:
The changes seem relatively simple.  I ignore SELinux, so I can't review the idea, but the code seems reasonable.

alejandro-colomar:
> The changes seem relatively simple. I ignore SELinux, so I can't review the idea, but the code seems reasonable.

It would be interesting to merge this PR as a proper merge commit instead of a rebase, to keep it organized as a single block of changes, BTW.

src/useradd.c
ikerexxe: In reality, this would be
```
process_selinux = !flags->chroot && !flags->prefix;
```

Updated the code accordingly.

src/useradd.c
ikerexxe: Done

src/useradd.c
ikerexxe: Yeah, makes sense

lib/cleanup_group.c
ikerexxe: I prefer casting in a line. Updated.

src/useradd.c
alejandro-colomar: Ahh, yeah, I was trying to write `!(A || B)`, but I don't know why.  Your version seems more reasonable.  :)

ikerexxe:
This PR is ready for review and testing. The COPR repository has also been updated with the latest changes. @praiskup and @pmatilai do you mind testing?

I've updated all binaries to prevent SELinux file context processing during file closure when `chroot` or `prefix` options are selected. Subordinate IDs library hasn't been updated due to API changes and the low probability of SELinux usage within `chroot` for subIDs, thus in this case the SELinux file context will be processed.

System tests have been removed as this PR already includes significant changes and container compatibility issues prevent proper execution.

hallyn (approved):
I'm ok with this.

I do wonder whether we would be better off using some ambient state (maybe just a simple global) to keep track of the process_selinux bool.  Have it automatically set to true at the start of each (relevant) program, and have the chroot and prefix options, when parsed, set it to false.

It would keep a lot of the function signatures simpler, but I'm not sure the end result is better, so feel free to merge this.

alejandro-colomar:
> I'm ok with this.
> 
> I do wonder whether we would be better off using some ambient state (maybe just a simple global) to keep track of the process_selinux bool. Have it automatically set to true at the start of each (relevant) program, and have the chroot and prefix options, when parsed, set it to false.
> 
> It would keep a lot of the function signatures simpler, but I'm not sure the end result is better, so feel free to merge this.

I prefer passing it locally.  While a static looks simpler, it might complicate things later.  I'm working on removing some statics elsewhere, and I think it would be good to not add more.

alejandro-colomar:
@ikerexxe You need to rebase a few conflicts.

ikerexxe:
> I prefer passing it locally. While a static looks simpler, it might complicate things later. I'm working on removing some statics elsewhere, and I think it would be good to not add more.

I also prefer to pass it locally and add yet another global variable that will complicate things in the future.

> @ikerexxe You need to rebase a few conflicts.

I'll do it during the week.

By the way, @pmatilai @praiskup I'm waiting for you to test this.

praiskup:
> By the way, @pmatilai @praiskup I'm waiting for you to test this.

After or before the rebase?

praiskup:
Before the rebase, great!  It works, thank you.

```
# rpm -q mock shadow-utils
mock-6.3-1.fc42.noarch
shadow-utils-4.18.0-33test.fc42.x86_64

# dnf diff mock
--- /usr/lib/python3.13/site-packages/mockbuild/shadow_utils.py 2025-06-17 20:00:00.000000000 -0400
+++ /usr/lib/python3.13/site-packages/mockbuild/shadow_utils.py 2025-10-07 02:27:50.990522104 -0400
@@ -22,7 +22,7 @@
             # provide this workaround.
             # Tracking upstream bug https://github.com/shadow-maint/shadow/issues/897
             if self.root.config['use_host_shadow_utils']:
-                do_with_status(command + ['--prefix', self.root.make_chroot_path()], raiseExc=not can_fail)
+                do_with_status(command + ['--root', self.root.make_chroot_path()], raiseExc=not can_fail)
             else:
                 self.root.doChroot(command, raiseExc=not can_fail)
                 
# a few testing mock builds succeeded
```

ikerexxe:
> After or before the rebase?

It doesn't really matter because the code has changed quite a bit between shadow 4.18 and master, so I would have to port a few commits to the COPR build before I could apply this branch cleanly.

> Before the rebase, great! It works, thank you.

For my part, these tests would be sufficient. Thank you for checking.

hallyn:
@ikerexxe so we're good to merge?

ikerexxe:
@hallyn yes, I think this is good as it is. Please keep in mind Alejandro's comment:

> It would be interesting to merge this PR as a proper merge commit instead of a rebase, to keep it organized as a single block of changes, BTW.

PR #1260, opened 2025-05-25
lib/sizeof.h: Make sure STRLEN() only accepts string literals

alejandro-colomar:
Link: <https://stackoverflow.com/a/79369560/6872717>


alejandro-colomar:
> What exactly is the benefit of this change? I only see one instance of `STRLEN` in the code and it is very clear that it uses a string. I am not saying that there is no fortification in the code, only that its benefit seems low.

Yeah, the benefit is low.  I also doubted if I should send this patch.  On the other hand, it's trivial to do it.

hallyn (approved):
Hm, well, safety...

I'll approve, please feel free to either merge or close.

PR #1264, opened 2025-06-04
lib/get_pid.c: pid_t is a signed integer

alejandro-colomar:
Print it with "%d".  Also, we need to adapt the size of the allocation. To avoid making mistakes, just make it `PATH_MAX`, and let the compiler optimize that if it's smart enough.


lslebodn:
One should ideally do just one change in the commit. 2 commits would be much better.

Otherwise LGTM

alejandro-colomar:
> One should ideally do just one change in the commit. 2 commits would be much better.
> 
> Otherwise LGTM

Hmmm, yeah, I can split it in two.

alejandro-colomar:
@lslebodn I have taken your comment as an Acked-by.  And have split the PR into two commits.

Thanks!

PR #1265, opened 2025-06-06
Fields678

alejandro-colomar:
@hallyn 

This addresses several issues in <src/newusers.c>.


alejandro-colomar:
Regarding the check for 6 fields, I don't know why the code checks for 6 fields.

The manual page says the fields are:
```
pw_name:pw_passwd:pw_uid:pw_gid:pw_gecos:pw_dir:pw_shell
```

alejandro-colomar:
git-blame(1) says the source code is checking for 6 fields since

```
$ git ref2 45c6603cc
45c6603cc86c (2007-10-07, 2007-10-07; "[svn-upgrade] Integrating new upstream version, shadow (19990709)")
```

alejandro-colomar:
Ughhhh, CodeQL has caught an important bug here.

<https://github.com/shadow-maint/shadow/blob/b9323a0ae33a42a67a42ad95de4ba3365cff6a66/src/newusers.c#L1235-L1237>

This is reading a buffer that we have never initialized.  :(

alejandro-colomar:
Cc: @thesamesam , @ikerexxe , @jubalh

hallyn (commented):
Hm, no, it may be wrong now, but we *do* inte d to read seven fields, the seventh (fields[6]) being the shell to use.

hallyn:
Yeah i think it is correct at the moment. Nfields ends as 6 bc cp is null at bottom of loop after the seventh field is read.

alejandro-colomar:
@hallyn 

I don't understand your comments.

This is the situation as I see it:

-  The array is unnecessarily declared as an array of 8.  I think we agree on the first commit that it should be an array of 7.

-  The documentation says there are 7 fields, and the comments in the source code agree with the documentation.  This was probably true on other Unix systems.

-  However, our code has never worked for 7 fields.  It has always checked for 6 fields, and reported an error for 7 (or 5,4,3,2,1,0).

-  We can fix this by adding support for the 7th field, or we can just assume that there has never been a 7th field, and continue working for 6 fields.  If we add support for the 7th field, we should consider accepting 6 too, since every existing script that uses our tools will be passing 6 currently.

-  Regardless of not having supported the 7th field, we tried to read it, but it was never there, or we would have reported an error earlier.  This means we've had been reading as a string the pw_shell field (7th), which contained garbage.  A dangerous bug.  This is true regardless of our opinion on supporting the 7th field.

alejandro-colomar:
Maybe I should split the PR.

alejandro-colomar:
Nah, see #1267 and #1266

PR #1155, opened 2024-12-13
Add some wrappers for usual loops around strsep(3)

alejandro-colomar:
Suggested-by: @hallyn 

---

Revisions:

<details>
<summary>v2</summary>

-  Workaround two GCC bugs [@uecker , @pinskia]
   <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118044>
   <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118043>

```
$ git range-diff master gh/astrsep2ls astrsep2ls 
 1:  ad66293e !  1:  ccee0abc lib/string/strtok/: strsep2arr(), STRSEP2ARR(): Add APIs
    @@ lib/string/strtok/strsep2arr.h (new)
     +)
     +
     +
    -+ATTR_ACCESS(read_write, 1) ATTR_ACCESS(write_only, 4)
    ++ATTR_ACCESS(read_write, 1) ATTR_ACCESS(write_only, 4, 3)
     +ATTR_STRING(1) ATTR_STRING(2)
     +inline ssize_t strsep2arr(char *s, const char *restrict delim,
     +    size_t n, char *a[restrict n]);
 2:  5f98de6e !  2:  95f4a921 lib/string/strtok/: strsep2ls(), STRSEP2LS(): Add APIs
    @@ lib/string/strtok/strsep2ls.h (new)
     +#define STRSEP2LS(s, delim, ls)  strsep2ls(s, delim, NITEMS(ls), ls)
     +
     +
    -+ATTR_ACCESS(read_write, 1) ATTR_ACCESS(write_only, 4)
    ++ATTR_ACCESS(read_write, 1) ATTR_ACCESS(write_only, 4, 3)
     +ATTR_STRING(1) ATTR_STRING(2)
     +inline ssize_t strsep2ls(char *s, const char *restrict delim,
     +    size_t n, char *ls[restrict n]);
 3:  a5390a9c =  3:  ab168394 lib/: Use STRSEP2LS() instead of its pattern
 4:  4f29f5e7 =  4:  411dd0be lib/port.c: getportent(): Use STRSEP2ARR() instead of its pattern
 5:  519fc865 =  5:  dc232372 lib/: Use strsep2arr() instead of its pattern
 6:  57f481b4 =  6:  a2dd29c8 lib/: Use STRSEP2ARR() instead of its pattern
 7:  70f747cc =  7:  884f9611 lib/string/strtok/: astrsep2ls(): Add function
[...]

PR #1267, opened 2025-06-07
src/newusers.c: Move break to loop condition

alejandro-colomar:
This is a cherry-pick from #1265 (at v3b).

This is a refactor that will help see a bug (and its fix).  It will also help transition to using STRSEP2ARR() in #1155

### DIFF
--- a/src/newusers.c
+++ b/src/newusers.c
@@ -1124,10 +1124,8 @@ int main (int argc, char **argv)
 		 * There MUST be 7 colon separated fields, although the
 		 * values aren't that particular.
 		 */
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
Oops, no, this is bogus.

alejandro-colomar:
Ahhhh, now I see it.  So we parse 7 fields indeed:

nfields is 0,1,2,3,4,5,6, and when we parsed the 7th (nfields is 6) we break out early without advancing nfields, which is why it says 6.  So the second bug is not valid.

PR #1269, opened 2025-06-07
Disable building subids support on systems that don't have the prctl

sthibaul:


alejandro-colomar:
Cc: @thesamesam

lslebodn:
@sthibaul would you be so kind and could you a bit more elaborate about "systems that don't have the prctl" ?

sthibaul:
> @sthibaul would you be so kind and could you a bit more elaborate about "systems that don't have the prctl" ?

prctl is not in posix. AFAIK it exists only on Linux and IRIX, so it's essentially all other systems. The case that matters to me is GNU/Hurd.

configure.ac
```diff
@@ -293,6 +293,7 @@ if test "$enable_man" = "yes"; then
 fi
 AM_CONDITIONAL(ENABLE_REGENERATE_MAN, test "x$enable_man" != "xno")
 
+AC_CHECK_HEADERS(sys/prctl.h, [], [enable_subids="no"])
```
alejandro-colomar: Please quote everything:

```
AC_CHECK_HEADERS([sys/prctl.h], [], [enable_subids="no"])
```

We recently applied a commit to quote everything consistently:
<https://github.com/shadow-maint/shadow/commit/630fb7ece1f731cb7ba85064d3568c9e130c9fbd>

sthibaul: ok, done so

configure.ac
```diff
@@ -293,6 +293,7 @@ if test "$enable_man" = "yes"; then
 fi
 AM_CONDITIONAL(ENABLE_REGENERATE_MAN, test "x$enable_man" != "xno")
 
+AC_CHECK_HEADERS([sys/prctl.h], [], [enable_subids="no"])
```
lslebodn: Let says somebody wants to build shadow with suids ans explicitelly pass `--enable-subordinate-ids=yes` to configure.  But for some reason `sys/prctl.h` is missing. Configure check will silently pass but override the feature to `no`.
```
checking for DocBook XML DTD V4.5 in XML catalog... found
checking for DocBook XSL Stylesheets >= 1.70.1 in XML catalog... found
checking for sys/prctl.h... no
checking for sd_session_get_remote_host in -lsystemd... no
```

Even though configure help says yes, the support would not be built.
And it would be complicated to identify from logs why.
```
sh$ ./configure --help | grep sub -A2
  --enable-subordinate-ids
                          support subordinate ids [default=yes]
  --enable-lastlog        enable lastlog [default=no]
  --enable-logind         enable logind [default=yes]
```

Instead of overriding default to `no`, we should rather fail with useful error message `AC_MSG_ERROR`.

We should inform user to either install missing `sys/prctl.h`  for building SUBUIDS( newgidmap newuidmap) or run configure `--disable-subordinate-ids`

ikerexxe: Yes, let's make this obvious so that the project is set up correctly and explicitly.

sthibaul: Right, done so

configure.ac
```diff
@@ -293,6 +293,15 @@ if test "$enable_man" = "yes"; then
 fi
 AM_CONDITIONAL(ENABLE_REGENERATE_MAN, test "x$enable_man" != "xno")
 
+if test "$enable_subids" != "no"; then
+	AC_CHECK_HEADERS([sys/prctl.h], [], [
+		if test "x$enable_subids" = "xyes"; then
```
lslebodn: We do not need yet another if block. The help says default is `yes`.
```
sh$ ./configure --help | grep sub -A2
  --enable-subordinate-ids
                          support subordinate ids [default=yes]
  --enable-lastlog        enable lastlog [default=no]
  --enable-logind         enable logind [default=yes]
```

and `enable_subids` will have value `no` when configure is is executed with parameter  `--disable-subordinate-ids`

configure.ac
```diff
@@ -293,6 +293,15 @@ if test "$enable_man" = "yes"; then
...
+			AC_MSG_ERROR([Cannot enable support the subordinate IDs on systems which do not have sys/prctl.h])
+		fi
+		enable_subids="no"
```
lslebodn: Please do not override value passed form command line parameter.

It is much better to inform user about optional feature for `support the subordinate IDs`
and let them decide how to solve it. Either to install missing header or to disable feature.
e.g.
```
checking for sys/prctl.h... no
configure: error: Cannot enable support the subordinate IDs on systems which do not have sys/prctl.h
                  The feature is optional and can be disabled by parameter --disable-subordinate-ids

```

lslebodn:
@sthibaul a bit offtopic to your PR.

`prctl` is not used always. It is compiled conditionally if header `<sys/capability.h>` can be included
* https://github.com/shadow-maint/shadow/blob/c44f1e096a19a7d356da5969295393247e61874f/lib/idmapping.c#L132-L150

But it is included always, even though it will not be used
* https://github.com/shadow-maint/shadow/blob/c44f1e096a19a7d356da5969295393247e61874f/lib/idmapping.c#L15-L18

I wonder whether you would be able to compile it  on GNU/Hurd with following diff
```
diff --git a/lib/idmapping.c b/lib/idmapping.c
index e70a037b..3425e1bb 100644
--- a/lib/idmapping.c
+++ b/lib/idmapping.c
@@ -12,8 +12,8 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <strings.h>
-#include <sys/prctl.h>
 #if __has_include(<sys/capability.h>)
+#include <sys/prctl.h>
 # include <sys/capability.h>
 #endif
```

It would be almost a revert of f50a39e8f9a00c1c2b6b26a80448d510c56353bb
```
commit f50a39e8f9a00c1c2b6b26a80448d510c56353bb
Date:   Tue Nov 12 14:42:12 2024 +0100

    lib/idmapping.c: Unconditionally include <sys/prctl.h>
    
    It's a widely available header.
    
diff --git a/lib/idmapping.c b/lib/idmapping.c
index 89c03b10..cef73b4f 100644
--- a/lib/idmapping.c
+++ b/lib/idmapping.c
@@ -12,8 +12,8 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <strings.h>
-#if HAVE_SYS_CAPABILITY_H
 #include <sys/prctl.h>
+#if HAVE_SYS_CAPABILITY_H
 #include <sys/capability.h>
 #endif
```

alejandro-colomar:
> @sthibaul a bit offtopic to your PR.
> 
> `prctl` is not used always. It is compiled conditionally if header `<sys/capability.h>` can be included
> 
>     * https://github.com/shadow-maint/shadow/blob/c44f1e096a19a7d356da5969295393247e61874f/lib/idmapping.c#L132-L150
> 
> 
> But it is included always, even though it will not be used
> 
>     * https://github.com/shadow-maint/shadow/blob/c44f1e096a19a7d356da5969295393247e61874f/lib/idmapping.c#L15-L18
> 
> 
> I wonder whether you would be able to compile it on GNU/Hurd with following diff
> 
> ```
> diff --git a/lib/idmapping.c b/lib/idmapping.c
> index e70a037b..3425e1bb 100644
> --- a/lib/idmapping.c
> +++ b/lib/idmapping.c
> @@ -12,8 +12,8 @@
>  #include <stdlib.h>
>  #include <stdio.h>
>  #include <strings.h>
> -#include <sys/prctl.h>
>  #if __has_include(<sys/capability.h>)
> +#include <sys/prctl.h>
>  # include <sys/capability.h>
>  #endif
> ```
> 
> It would be almost a revert of [f50a39e](https://github.com/shadow-maint/shadow/commit/f50a39e8f9a00c1c2b6b26a80448d510c56353bb)

Shouldn't we use `#if __has_include(<sys/prctl.h>)`?

configure.ac
sthibaul: > Please do not override value passed form command line parameter.

This is exactly like below: `if enable_subids != no` means it's either an explicit yes or no command line. If it's an explicit yes we error out, so the only case where enable_subids=no is done is in the no command line.

> or to disable feature
> can be disabled by parameter --disable-subordinate-ids

It's rather "not require the feature". This case is when the users explicitly asked for --enable-subordinate-ids, so we can only tell them, "well, sorry, no, we cannot enable it as you requested" The feature being optional is also fully known since the user explicitly asked for it.

sthibaul:
> @sthibaul a bit offtopic to your PR.
> 
> `prctl` is not used always. It is compiled conditionally if header `<sys/capability.h>` can be included
> 
>     * https://github.com/shadow-maint/shadow/blob/c44f1e096a19a7d356da5969295393247e61874f/lib/idmapping.c#L132-L150
> 
> 
> But it is included always, even though it will not be used
> 
>     * https://github.com/shadow-maint/shadow/blob/c44f1e096a19a7d356da5969295393247e61874f/lib/idmapping.c#L15-L18
> 
> 
> I wonder whether you would be able to compile it on GNU/Hurd with following diff
> 
> ```
> diff --git a/lib/idmapping.c b/lib/idmapping.c
> index e70a037b..3425e1bb 100644
> --- a/lib/idmapping.c
> +++ b/lib/idmapping.c
> @@ -12,8 +12,8 @@
>  #include <stdlib.h>
>  #include <stdio.h>
>  #include <strings.h>
> -#include <sys/prctl.h>
>  #if __has_include(<sys/capability.h>)
> +#include <sys/prctl.h>
>  # include <sys/capability.h>
>  #endif
> ```

That does make the build work indeed. Not sure it will be future-proof, though.

lib/idmapping.c
```diff
@@ -12,8 +12,8 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <strings.h>
-#include <sys/prctl.h>
 #if __has_include(<sys/capability.h>)
+# include <sys/prctl.h>
 # include <sys/capability.h>
 #endif
```
alejandro-colomar: I think it would be more appropriate to do

```diff
diff --git i/lib/idmapping.c w/lib/idmapping.c
index e70a037b..e55de97f 100644
--- i/lib/idmapping.c
+++ w/lib/idmapping.c
@@ -12,10 +12,12 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <strings.h>
-#include <sys/prctl.h>
 #if __has_include(<sys/capability.h>)
 # include <sys/capability.h>
 #endif
+#if __has_include(<sys/prctl.h>)
+# include <sys/prctl.h>
+#endif
 
 #include "alloc/calloc.h"
 #include "alloc/x/xmalloc.h"
```

lib/defines.h
```diff
@@ -65,6 +65,14 @@
...
 #endif
 
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
+
+#ifndef MAXHOSTNAMELEN
+#define MAXHOSTNAMELEN 64
+#endif
+
```
alejandro-colomar: We already merged this some time ago, no?

<https://github.com/shadow-maint/shadow/pull/1222/files>

sthibaul: Sorry, I don't know how this commit ended up in the branch, I though I had come back to master, but apparently it was lying there. Now gone

configure.ac
alejandro-colomar: I agree that silently proceeding with something different than the explicit desire of the user is morally wrong.  I think a hard error is better, as it allows the user to consciously decide what it wants.

sthibaul: > I agree that silently proceeding with something different than the explicit desire of the user is morally wrong.

Sure, and that's what that new code was doing.

lslebodn:
> That does make the build work indeed. Not sure it will be future-proof, though.

Hopefully one would check git blame  before "cleanup" similar as in f50a39e8f9a00c1c2b6b26a80448d510c56353bb :)

:crossed_fingers: 

```
commit f50a39e8f9a00c1c2b6b26a80448d510c56353bb
Date:   Tue Nov 12 14:42:12 2024 +0100

    lib/idmapping.c: Unconditionally include <sys/prctl.h>
    
    It's a widely available header.
    
diff --git a/lib/idmapping.c b/lib/idmapping.c
index 89c03b10..cef73b4f 100644
--- a/lib/idmapping.c
+++ b/lib/idmapping.c
@@ -12,8 +12,8 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <strings.h>
-#if HAVE_SYS_CAPABILITY_H
 #include <sys/prctl.h>
+#if HAVE_SYS_CAPABILITY_H
 #include <sys/capability.h>
 #endif
 ```

PR #1220, opened 2025-02-19
-Werror=...

alejandro-colomar:
Closes: <https://github.com/shadow-maint/shadow/issues/1218>
Reported-by: @zeha

---

-  Queued after <https://github.com/shadow-maint/shadow/pull/1402>.

---

Revisions:

<details>
<summary>v2</summary>

-  Promote this diagnostic to an error.

```
$ git range-diff master gh/c c
1:  5ba2ba92 = 1:  5ba2ba92 lib/string/strspn/: Add missing const
-:  -------- > 2:  d6e71302 autogen.sh: CFLAGS: Promote -Wdiscarded-qualifiers to an error
```
</details>

<details>
<summary>v3</summary>

-  Turn on `-Werror=all`.
-  Remove unnecessary uses of `[[gnu::unused]]`.  In some cases, we can remove the parameter.  In others, we can't but we can unname it.  That attribute should only be used when the use of a variable is conditionally compiled.

```
$ git range-diff master gh/c c
 1:  5ba2ba92 =  1:  5ba2ba92 lib/string/strspn/: Add missing const
 2:  d6e71302 =  2:  d6e71302 autogen.sh: CFLAGS: Promote -Wdiscarded-qualifiers to an error
 -:  -------- >  3:  5de673ed lib/, src/: Remove unused variables
 -:  -------- >  4:  f0dfbb85 src/: Unname unused parameter of main()
 -:  -------- >  5:  a932f31b src/gpasswd.c: Remove unused parameter $1 of check_perms()
 -:  -------- >  6:  1df3863f autogen.sh: CFLAGS: Promote -Wall to errors
 -:  -------- >  7:  f5624170 lib/, src/: Unname unused parameters in callbacks
 -:  -------- >  8:  56db38b0 lib/, src/: Remove unused parameter $2 of audit_logger()
 -:  -------- >  9:  c7a7513f lib/copydir.c: Remove unused parameter $3 of copy_symlink()
 -:  -------- > 10:  7c0ce754 lib/copydir.c: Remove unused parameter $2 of copy_hardlink()
 -:  -------- > 11:  14ebca70 lib/logind.c: Unname unused function parameter
 -:  -------- > 12:  f93d7e81 lib/: Remove unused parameter $3 of password_check() and propagate
 -:  -------- > 13:  421654e1 lib/obscure.c: Remove unused parameter $1 of palindrome()
 -:  -------- > 14:  503682c1 lib/, src/: Remove unused parameter $3 of passwd_check()
```
</details>

[...]

PR #1271, opened 2025-06-09
[dd] Remove dead code

alejandro-colomar:
These are cherry-picks from #1220

### DIFF
--- a/lib/audit_help.c
+++ b/lib/audit_help.c
@@ -56,13 +56,12 @@ void audit_help_open (void)
  *
  * type - type of message. A list of possible values is available in
  *        "audit-records.h" file.
- * pgname - program's name
  * op  -  operation. "adding user", "changing finger info", "deleting group"
  * name - user's account or group name. If not available use NULL.
  * id  -  uid or gid that the operation is being performed on. This is used
  *	  only when user is NULL.
  */
-void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
+void audit_logger (int type, const char *op,
                    const char *name, unsigned int id,
                    shadow_audit_result result)
 {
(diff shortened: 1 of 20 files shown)


lib/audit_help.c
```diff
@@ -56,13 +56,12 @@ void audit_help_open (void)
...
  * type - type of message. A list of possible values is available in
  *        "audit-records.h" file.
- * pgname - program's name
  * op  -  operation. "adding user", "changing finger info", "deleting group"
  * name - user's account or group name. If not available use NULL.
  * id  -  uid or gid that the operation is being performed on. This is used
  *	  only when user is NULL.
  */
-void audit_logger (int type, MAYBE_UNUSED const char *pgname, const char *op,
+void audit_logger (int type, const char *op,
```
ikerexxe, line 64: I have my concerns regarding this removal, not so much by the removal itself but because `audit_log_acct_message()` provides an argument to set `pgname` and we are specifically ignoring it. That's really weird.

alejandro-colomar: Yup, that was weird.

alejandro-colomar: Feel free to replace that commit by a PR that actually uses the parameter.

alejandro-colomar: I can move that patch to a separate PR if you want.

ikerexxe: I've been checking the git history and this was removed many many years ago. In fact, there's a commit from 2010 that already notices that the parameter isn't used and sets the `unused` attribute.

```
commit 6a2f349ec507327c643f4e20080c522dc5ead527
Author: nekral-guest <nekral-guest@5a98b0ae-9ef6-0310-add3-de5d479b70d7>
Date:   Sat Aug 21 15:22:39 2010 +0000

            * libmisc/audit_help.c (audit_logger):pgname is not used. We let
            audit find out the name of the program.
            * libmisc/audit_help.c: Re-indent.
```

IMO, the approach from this PR is fine and we should remove the unused parameter altogether since nobody complaint in many years

alejandro-colomar: Okay, then I'll close the other PR.  Feel free to merge this one.  Thanks!

PR #1281, opened 2025-07-08
CodeQL fixes

ikerexxe:
Fixes to two weaknesses detected by CodeQL. More information about each issue in the commit message.

### DIFF
--- a/src/chfn.c
+++ b/src/chfn.c
@@ -67,7 +67,7 @@ NORETURN static void fail_exit (int code);
 NORETURN static void usage (int status);
 static bool may_change_field (int);
 static void new_fields (void);
-static char *copy_field (char *, char *, char *);
+static char *copy_field (char *, char *, char *, unsigned long);
 static void process_flags (int argc, char **argv);
 static void check_perms (const struct passwd *pw);
 static void update_gecos (const char *user, char *gecos);
@@ -215,8 +215,10 @@ static void new_fields (void)
  *	in - the current GECOS field
  *	out - where to copy the field to
  *	extra - fields with '=' get copied here
+ *	out_size - output string size
  */
-static char *copy_field (char *in, char *out, char *extra)
+static char *
+copy_field (char *in, char *out, char *extra, unsigned long out_size)
 {
 	char  *next = NULL;
 
@@ -230,6 +232,10 @@ static char *copy_field (char *in, char *out, char *extra)
 			break;
 
 		if (NULL != extra) {
+			if (strlen(extra) + 1 + strlen(f) + 1 > out_size) {
+				return NULL;
+			}
+
 			if (!streq(extra, "")) {
 				strcat (extra, ",");
 			}
@@ -521,28 +527,32 @@ static void get_old_fields (const char *gecos)
 	 * Now get the full name. It is the first comma separated field in
 	 * the GECOS field.
 	 */
-	cp = copy_field (old_gecos, fflg ? NULL : fullnm, slop);
+	cp = copy_field (old_gecos, fflg ? NULL : fullnm, slop,
+			fflg ? 0 : countof(fullnm));
 
 	/*
 	 * Now get the room number. It is the next comma separated field,
 	 * if there is indeed one.
 	 */
 	if (NULL != cp) {
-		cp = copy_field (cp, rflg ? NULL : roomno, slop);
+		cp = copy_field (cp, rflg ? NULL : roomno, slop,
+				rflg ? 0 : countof(roomno));
 	}
 
 	/*
 	 * Now get the work phone number. It is the third field.
 	 */
 	if (NULL != cp) {
-		cp = copy_field (cp, wflg ? NULL : workph, slop);
+		cp = copy_field (cp, wflg ? NULL : workph, slop,
+				wflg ? 0 : countof(workph));
 	}
 
 	/*
 	 * Now get the home phone number. It is the fourth field.
 	 */
 	if (NULL != cp) {
-		cp = copy_field (cp, hflg ? NULL : homeph, slop);
+		cp = copy_field (cp, hflg ? NULL : homeph, slop,
+				hflg ? 0 : countof(homeph));
 	}
 
 	/*


ikerexxe:
CI is failing due to https://github.com/shadow-maint/shadow/blob/master/lib/sizeof.h#L23 and I see you even converted this warning into an error with `-Werror=sizeof-pointer-div` in https://github.com/shadow-maint/shadow/commit/dc8ead618db53fa5f0d30142d720ddb245326ad6. Is this working as expected or are we still missing something from gcc and clang?

src/chfn.c
```diff
@@ -230,11 +231,15 @@ static char *copy_field (char *in, char *out, char *extra)
 			break;
 
 		if (NULL != extra) {
+			if (strlen(extra) + 1 + strlen(f) + 1 > BUFSIZ) {
+				return NULL;
+			}
+
 			if (!streq(extra, "")) {
-				strcat (extra, ",");
+				STRNCAT(extra, ",");
```
alejandro-colomar: STRNCAT() should never be used with string input.  It is for reading from something like a utmp field into a string:

```c
$ grep -rnI 'STRNCAT('
src/logoutd.c:209:			STRNCAT(tty_name, ut->ut_line);
lib/string/strcpy/strncat.h:16:#define STRNCAT(dst, src)  strncat(dst, src, countof(src))
lib/string/strdup/xstrndup.h:21:	STRNCAT(strcpy(XMALLOC(strnlen(s, countof(s)) + 1, char), ""), s)     \
```

You probably want STRTCAT()  (the T is for truncation).  (But I haven't implemented that so far.)

However, here we don't know the destination size, so we need to use strcat(3).  We can't possibly truncate without knowing the boundary.

There's nothing we can do here.

src/chfn.c
```diff
@@ -230,11 +231,15 @@ static char *copy_field (char *in, char *out, char *extra)
 			break;
 
 		if (NULL != extra) {
+			if (strlen(extra) + 1 + strlen(f) + 1 > BUFSIZ) {
```
alejandro-colomar: Why BUFSIZ?

lib/limits.c
```diff
@@ -393,7 +393,7 @@ static int setup_user_limits (const char *uname)
 		 * the last encountered entry for a matching group rules.
 		 * If there is no matching group entry, the default limits rule.
 		 */
-		if (sscanf (buf, "%s%[ACDFIKLMNOPRSTUacdfiklmnoprstu0-9 \t-]",
+		if (sscanf (buf, "%1023s%1023[ACDFIKLMNOPRSTUacdfiklmnoprstu0-9 \t-]",
```
alejandro-colomar: I'm not convinced of this fix.  I'd avoid sscanf(3) as the plague.  I wanted to replace this by proper parsing some time ago, but didn't find time.

I'd recommend something like strsep(3) here.

alejandro-colomar: I'd leave it like that for now, and I can have a look at it later.  I'm a bit busy within the C Committee, glibc, and the kernel at the moment, though.  I'll try to do that after some months (and hopefully, after we've merged some of the changes under review, which will reduce the queue of patches I have here).

alejandro-colomar:
> CI is failing due to https://github.com/shadow-maint/shadow/blob/master/lib/sizeof.h#L23 and I see you even converted this warning into an error with `-Werror=sizeof-pointer-div` in [dc8ead6](https://github.com/shadow-maint/shadow/commit/dc8ead618db53fa5f0d30142d720ddb245326ad6). Is this working as expected or are we still missing something from gcc and clang?

It is expected.  You're misusing STRNCAT().

I think a good way to explain strncpy(3), strncat(3), and strndup(3) would be: "`strn` stands for strings NO".  :)

src/chfn.c
ikerexxe: > STRNCAT() should never be used with string input. It is for reading from something like a utmp field into a string:

Where is this limitation stated? I haven't found any documentation indicating that `STRNCAT()` should only be used with fixed-size inputs, nor that it's unsuitable for general string input.

src/chfn.c
ikerexxe: Because that's the size of the variables that will hold the output values for `copy_field()`. Example: https://github.com/shadow-maint/shadow/blob/master/src/chfn.c#L47

lib/limits.c
ikerexxe: Thanks for the feedback and for the `strsep(3)` suggestion.

I agree that breaking the string into a sequence of tokens with `strsep(3)` makes more sense here, and it's a cleaner approach than `sscanf(3)`. However, implementing that properly will take some time, and I'd like to get this fix in sooner rather than later.

Given your current commitments, I propose we proceed with the current fix for now. Once the `strsep(3)` replacement is ready, we can absolutely review and merge it.

src/chfn.c
alejandro-colomar: > > STRNCAT() should never be used with string input. It is for reading from something like a utmp field into a string:
> 
> Where is this limitation stated? I haven't found any documentation indicating that `STRNCAT()` should only be used with fixed-size inputs,

That's implicitly stated by the fact that it asserts (through countof()) that the input is an array.  

> nor that it's unsuitable for general string input.

The `string_copying(7)` manual page explains that.

```
     strncat(3)
            Catenate the input character sequence, contained in a null‐
            padded fixed‐size buffer, into a destination  string.   The
            programmer  is  responsible  for  allocating a buffer large
            enough.  The return value is useless.

            Do not confuse this function with strncpy(3); they are  not
            related at all.

            stpcpy(mempcpy(dst, src, strnlen(src, NITEMS(src))), "") is
            a faster alternative to this function.
```

The strncat(3) manual page also explains that:

```
$ man strncat | head -n22
strncat(3)              Library Functions Manual             strncat(3)

NAME
     strncat  -  append non‐null bytes from a source array to a string,
     and null‐terminate the result

LIBRARY

SYNOPSIS
     #include <string.h>

     char *strncat(size_t ssize;
                   char *restrict dst, const char src[restrict ssize],
                   size_t ssize);

DESCRIPTION
     This function appends at most ssize non‐null bytes from the  array
     pointed to by src, followed by a null character, to the end of the
     string pointed to by dst.  dst must point to a string contained in
     a buffer that is large enough, that is, the buffer size must be at
     least strlen(dst) + strnlen(src, ssize) + 1.
```

And I'm working on getting the ISO C standard to improve this situation too.

src/chfn.c
alejandro-colomar: That's going to be dangerous.  If we ever change those buffer sizes, this will result in buffer overflows all over the place.  Those buffer sizes should be calculated with countof(), which would require adding parameters to the function.

lib/limits.c
alejandro-colomar: > Thanks for the feedback and for the `strsep(3)` suggestion.
> 
> I agree that breaking the string into a sequence of tokens with `strsep(3)` makes more sense here, and it's a cleaner approach than `sscanf(3)`. However, implementing that properly will take some time, and I'd like to get this fix in sooner rather than later.

IIRC, I looked at this code some time ago, and concluded it was safe.  Since we read from a string contained in `buf`, we can't possibly overflow the other buffers (which have the same size).

So, this is a false positive from CodeQL.  We're not in a hurry to fix it.

> 
> Given your current commitments, I propose we proceed with the current fix for now. Once the `strsep(3)` replacement is ready, we can absolutely review and merge it.

I'd prefer having the false positive around, which will remind us that we need to improve the code.

I'm skeptic of any code that adds complexity instead of removing it, so I'd like to not merge this.

src/chfn.c
ikerexxe: I can change it to calculate the size of the output buffer:
```
if (strlen(extra) + 1 + strlen(f) + 1 > sizeof(out)) {
```

WDYT?

alejandro-colomar: `sizeof(out)` is wrong.  You need `countof(out)`.

`sizeof(out)` is equivalent to `sizeof(char *)`, which is 8.

src/chfn.c
ikerexxe: I've removed this part of the changes

src/chfn.c
ikerexxe: Done

lib/limits.c
ikerexxe: I've removed this part of the changes

src/chfn.c
```diff
@@ -230,6 +230,10 @@ static char *copy_field (char *in, char *out, char *extra)
 			break;
 
 		if (NULL != extra) {
+			if (strlen(extra) + 1 + strlen(f) + 1 > strlen(out)) {
```
alejandro-colomar: `out` is where we want to write.  There need not be a string to begin with, so we can't read it.  What you want is really `countof(out)`.  `strlen(out)` won't help.

ikerexxe: That won't work either. As an example, on a 64-bit system, `sizeof(out)` is 8 bytes (the size of a pointer). `sizeof(out[0])` is 1 byte (`sizeof(char)`). Thus, `countof(my_ptr)` calculates 8 / 1 = 8.

alejandro-colomar: `countof()` is currently implemented with `sizeof()`, yes.  But we will eventually use the one from `<stdcountof.h>`. which is implemented with `_Countof()`.  That will eventually work with array parameters (maybe next year).

For the moment, the consequence is that this is impossible to do.

Unless you pass the array size manually to the function as an extra parameter.

ikerexxe: > Unless you pass the array size manually to the function as an extra parameter.

I think we should move it this way. Let me see if I find some minutes to update the implementation.

src/chfn.c
```diff
@@ -67,7 +67,7 @@ NORETURN static void fail_exit (int code);
 NORETURN static void usage (int status);
 static bool may_change_field (int);
 static void new_fields (void);
-static char *copy_field (char *, char *, char *);
+static char *copy_field (char *, char *, char *, unsigned long);
```
alejandro-colomar, line 70: Please use size_t.  unsigned long is a very weird thing.

src/chfn.c
```diff
@@ -230,6 +232,10 @@ static char *copy_field (char *in, char *out, char *extra)
 			break;
 
 		if (NULL != extra) {
+			if (strlen(extra) + 1 + strlen(f) + 1 > out_size) {
```
alejandro-colomar, line 235: What do those +1 mean?

src/chfn.c
```diff
@@ -230,6 +232,10 @@ static char *copy_field (char *in, char *out, char *extra)
...
+				return NULL;
+			}
+
 			if (!streq(extra, "")) {
 				strcat (extra, ",");
 			}
```
alejandro-colomar, line 241: Why is the check here, if the out_size is the size protecting the `strcpy(out, in)` in line 242?

src/chfn.c
```diff
@@ -521,28 +527,32 @@ static void get_old_fields (const char *gecos)
...
 
 	/*
 	 * Now get the room number. It is the next comma separated field,
 	 * if there is indeed one.
 	 */
 	if (NULL != cp) {
-		cp = copy_field (cp, rflg ? NULL : roomno, slop);
+		cp = copy_field (cp, rflg ? NULL : roomno, slop,
+				rflg ? 0 : countof(roomno));
 	}
```
alejandro-colomar, line 540: What happens if one of these doesn't copy due to not-enough space?  Should that be an error?  Should the program fail?

alejandro-colomar:
I've written a PR with a different apporach: simplify this code, and remove most of it: #1307

alejandro-colomar:
> I've written a PR with a different apporach: simplify this code, and remove most of it: #1307

Since #1307 has been merged, this is unnecessary.  Feel free to reopen if anything from here remains useful.

PR #1283, opened 2025-07-10
configure.ac: Make sure that logind is enabled if requested

Karlson2k:
Currently if configured with --enable-logind, but `libsystemd` is not found, `configure` silently succeed, but `logind` is efficiently disabled. 
With this patch `configure` fails if `logind` is explicitly requested, but `libsystemd` is not found.
If `logind` is not requested explicitly, it is enabled only if found.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -154,7 +154,13 @@ AC_ARG_ENABLE([lastlog],
 AC_ARG_ENABLE([logind],
 	[AS_HELP_STRING([--enable-logind],
 		[enable logind @<:@default=yes if found@:>@])],
-	[enable_logind="${enableval}"],
+	[
+		AS_CASE([${enableval}],
+			[yes],[],
+			[no],[],
+			[AC_MSG_ERROR([bad parameter value for --enable-logind=${enableval}])]
+		)
+	],
 	[enable_logind="yes"]
 )
 
--- a/configure.ac
+++ b/configure.ac
@@ -346,10 +346,14 @@ AM_CONDITIONAL([ENABLE_LASTLOG], [test "x$enable_lastlog" != "xno"])
 AC_SUBST([LIBSYSTEMD])
 if test "X$enable_logind" = "Xyes"; then
 	AC_CHECK_LIB([systemd], [sd_session_get_remote_host],
-		[enable_logind="yes"; [LIBSYSTEMD=-lsystemd];
-		AC_DEFINE([ENABLE_LOGIND], [1],
-			[Define to manage session support with logind.])],
-		[enable_logind="no"])
+		[
+			enable_logind="yes"
+			LIBSYSTEMD=-lsystemd
+			AC_DEFINE([ENABLE_LOGIND], [1], [Define to manage session support with logind.])
+		],[
+			enable_logind="no"
+		]
+	)
 fi
 AM_CONDITIONAL([ENABLE_LOGIND], [test "x$enable_logind" != "xno"])
 
--- a/configure.ac
+++ b/configure.ac
@@ -153,7 +153,7 @@ AC_ARG_ENABLE([lastlog],
 
 AC_ARG_ENABLE([logind],
 	[AS_HELP_STRING([--enable-logind],
-		[enable logind @<:@default=yes@:>@])],
+		[enable logind @<:@default=yes if found@:>@])],
 	[enable_logind="${enableval}"],
 	[enable_logind="yes"]
 )
--- a/configure.ac
+++ b/configure.ac
@@ -158,10 +158,11 @@ AC_ARG_ENABLE([logind],
 		AS_CASE([${enableval}],
 			[yes],[],
 			[no],[],
+			[maybe|auto],[enable_logind="auto"],
 			[AC_MSG_ERROR([bad parameter value for --enable-logind=${enableval}])]
 		)
 	],
-	[enable_logind="yes"]
+	[enable_logind="auto"]
 )
 
 AC_ARG_WITH([audit],
@@ -344,18 +345,20 @@ fi
 AM_CONDITIONAL([ENABLE_LASTLOG], [test "x$enable_lastlog" != "xno"])
 
 AC_SUBST([LIBSYSTEMD])
-if test "X$enable_logind" = "Xyes"; then
+if test "X$enable_logind" != "Xno"; then
 	AC_CHECK_LIB([systemd], [sd_session_get_remote_host],
 		[
 			enable_logind="yes"
 			LIBSYSTEMD=-lsystemd
 			AC_DEFINE([ENABLE_LOGIND], [1], [Define to manage session support with logind.])
 		],[
+			AS_IF([test "X$enable_logind" = "Xyes"],
+				[AC_MSG_ERROR([libsystemd not found, logind cannot be enabled])])
 			enable_logind="no"
 		]
 	)
 fi
-AM_CONDITIONAL([ENABLE_LOGIND], [test "x$enable_logind" != "xno"])
+AM_CONDITIONAL([ENABLE_LOGIND], [test "x$enable_logind" = "xyes"])
 
 AC_CHECK_LIB([crypt], [crypt], [LIBCRYPT=-lcrypt],
 	[AC_MSG_ERROR([crypt() not found])])


configure.ac
```diff
@@ -336,14 +336,17 @@ fi
 AM_CONDITIONAL(ENABLE_LASTLOG, test "x$enable_lastlog" != "xno")
 
 AC_SUBST(LIBSYSTEMD)
-if test "$enable_logind" = "yes"; then
+if test "X$enable_logind" != "Xno"; then
```
alejandro-colomar: Why do we need the `X`?  It's not used in the other tests.

alejandro-colomar: If we want this, I expect this would be global.

Karlson2k: Just a few lines below:
https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L346

It is also used in many tests:
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L217
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L222
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L227
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L245
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L255
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L268
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L294
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L307
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L313
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L336
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L376
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L618
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L624
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L626
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L628
* https://github.com/shadow-maint/shadow/blob/36debf3ccfa33ea9f3b0e562a133a22d1bdc87bd/configure.ac#L654

It not strictly required to use `X` in this test, as this variable must be always defined and non-empty.
Use of such prefix is workaround for broken shells, that may expand such tests to `test  = yes` even when arguments are quoted.
For uniformity, to avoid bugs with future code changes, to avoid analysis whether variable is guaranteed to b e non-empty, it is easy to always use forms like `test "X$some_var" = "Xyes"` everywhere in scripts designed for portability (configure and Makefiles).
I designed this PR to be minimal.

There are other things that could be improved in this configure.
I can make a separate PR with other fixes to keep this PR focused on a particular change.

alejandro-colomar: Yup, please add a separate PR for doing that globally.

Karlson2k: Done: https://github.com/shadow-maint/shadow/pull/1287

Karlson2k:
Updated with rebase after #1287.
Added parameters check for "--enable-logind".

configure.ac
```diff
@@ -338,14 +345,18 @@ fi
 AM_CONDITIONAL([ENABLE_LASTLOG], [test "x$enable_lastlog" != "xno"])
 
 AC_SUBST([LIBSYSTEMD])
-if test "X$enable_logind" = "Xyes"; then
+if test "X$enable_logind" != "Xno"; then
 	AC_CHECK_LIB([systemd], [sd_session_get_remote_host],
-		[enable_logind="yes"; [LIBSYSTEMD=-lsystemd];
+		[enable_logind="yes"
+		LIBSYSTEMD=-lsystemd
```
alejandro-colomar: What's the rationale for this?

Karlson2k: To fix overcomplicated and confusing code.
Second level of quoting specifically around `LIBSYSTEMD=-lsystemd` is not needed, confusing and inconsistent.
When second level of quoting is used, the selected part treated literally, but nothing needs to be treated literally in this part (prevent expansion).
Also, when each assignment has its own line, the shell code is better (in terms of readability).

alejandro-colomar: Please separate those cosmetic changes from the main commit.

Karlson2k: Converted to four commits.

Karlson2k:
@alejandro-colomar Updated as requested.

configure.ac
```diff
@@ -153,9 +153,16 @@ AC_ARG_ENABLE([lastlog],
...
+		[enable logind @<:@default=yes if found@:>@])],
+	[
+		AS_CASE([${enableval}],
+			[yes],[],
+			[no],[],
+			[maybe|auto],[enable_logind="auto"],
+			[AC_MSG_ERROR([bad parameter value for --enable-logind=${enableval}])]
+		)
+	],
+	[enable_logind="auto"]
```
lslebodn, line 165: Ternary logic is too complicated. Sure it is possible.
But the simplest approach for `--enable`/`--disable` options is to allow just `yes` and `no`

One just need to provide useful ERROR message if the feature is enabled by default(preferred)
but just optional e.g.
```
- libsystemd not found, logind cannot be enabled
+ libsystemd not found, logind cannot be enabled. The feature is optional and it can be disabled.
+ Please install missing dependencies or disable it with configure time option --enable-logind
```

Karlson2k: This looks like indirect pushing everyone to `systemd`, while for number of uses, like embedded projects, it still may be not the best choice.
The suggested change will make `configure` larger and less consistent.

Configure already has a number of automatically enabled options. For example:
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L140-L145
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L161-L163
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L164-L166
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L167-L169
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L170-L172
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L173-L175
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L176-L178
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L182-L184

Currently (without this PR) `logind` support is automatic in fact.

Let's keep it this way, but with proper documentation (help message) and proper error message if support is explicitly requested.

alejandro-colomar: > This looks like indirect pushing everyone to `systemd`, while for number of uses, like embedded projects, it still may be not the best choice. The suggested change will make `configure` larger and less consistent.

I certainly wouldn't want to push for systemd, as I use Devuan myself.
@thesamesam , please chime in.  I can't decide anything autotools-related.

lslebodn: > This looks like indirect pushing everyone to systemd, while for number of uses, like embedded projects, it still may be not the best choice.

Sorry I had a typo in the message. I meant
```
- libsystemd not found, logind cannot be enabled
+ libsystemd not found, logind cannot be enabled. The feature is optional and it can be disabled.
+ Please install missing dependencies or disable it with configure time option --disable-logind
```

Current master that default value should be `yes`.
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L154-L159

if we want to change default to `no` that it is a different discussion :-)

> The suggested change will make configure larger and less consistent.
> 
> Configure already has a number of automatically enabled options. For example:

There are many things which are not ideal in the configure.ac You already fix couple of them.
Moreover, there are already examples where defaults are just `yes|no`
There is not much benefit to have more `maybe/auto` defaults

* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L137
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L151
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L181
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L187
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L190
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L193
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L196
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L199
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L202
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L205
* https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L208

Automagic detection in configure and silent override is not ideal;. They just complicate the configure code.
it is better to fail; and let user to resolve issue based on useful error message from the configure.

Karlson2k: > Current master that default value should be `yes`.

Not correct. See [here](https://github.com/shadow-maint/shadow/pull/1283#discussion_r2243595981).

Automagic detection will not push anyone to some preferred (by you or someone else) solution.
Automagic is used in this configure already, so no inconsistency.
Automagic is not nice if specific `yes` or `no` value cannot be enforced. If user-provided value just silently reset to another value (like in current configure), this is a bad behaviour.
This PR corrects it. If user selects explicitly `--enable-logind` then logind will be used or configure will fail. This is correct behaviour.

> Automagic detection in configure and silent override is not ideal;. They just complicate the configure code.

What complication you are talking about?
Let's compare.
The version with automagic:
```autoconf
AC_ARG_ENABLE([logind],
	[AS_HELP_STRING([--enable-logind],
		[enable logind @<:@default=yes if found@:>@])],
	[
		AS_CASE([${enableval}],
			[yes],[],
			[no],[],
			[maybe|auto],[enable_logind="auto"],
			[AC_MSG_ERROR([bad parameter value for --enable-logind=${enableval}])]
		)
	],
	[enable_logind="auto"]
)

AC_SUBST([LIBSYSTEMD])
if test "X$enable_logind" != "Xno"; then
	AC_CHECK_LIB([systemd], [sd_session_get_remote_host],
		[
			enable_logind="yes"
			LIBSYSTEMD=-lsystemd
			AC_DEFINE([ENABLE_LOGIND], [1], [Define to manage session support with logind.])
		],[
			AS_IF([test "X$enable_logind" = "Xyes"], [AC_MSG_ERROR([libsystemd not found, logind cannot be enabled])])
			enable_logind="no"
		]
	)
fi
AM_CONDITIONAL([ENABLE_LOGIND], [test "x$enable_logind" = "xyes"])

```

The version without automagic:
```autoconf
AC_ARG_ENABLE([logind],
	[AS_HELP_STRING([--enable-logind],
		[enable logind @<:@default=yes@:>@])],
	[
		AS_CASE([${enableval}],
			[yes],[],
			[no],[],
			[AC_MSG_ERROR([bad parameter value for --enable-logind=${enableval}])]
		)
	],
	[enable_logind="yes"]
)

AC_SUBST([LIBSYSTEMD])
if test "X$enable_logind" != "Xno"; then
	AC_CHECK_LIB([systemd], [sd_session_get_remote_host],
		[
			enable_logind="yes"
			LIBSYSTEMD=-lsystemd
			AC_DEFINE([ENABLE_LOGIND], [1], [Define to manage session support with logind.])
		],[
			AC_MSG_ERROR([libsystemd not found, logind cannot be enabled. The feature is optional and it can be disabled.])
			AC_MSG_ERROR([Please install missing dependencies or disable it with configure time option --disable-logind])
		]
	)
fi
AM_CONDITIONAL([ENABLE_LOGIND], [test "x$enable_logind" = "xyes"])

```
One line difference is not really huge complication. The difference can be shortened to zero lines by removing `maybe|auto` case.

> it is better to fail; and let user to resolve issue based on useful error message from the configure.

No, it is better to use what is found (with clear report at the end of the configure) if nothing is requested by the user or follow strictly what is requested by the user. This fits perfectly both novice users (no parameters) and package maintainers (explicit `--disable-logind` or `--enable-logind`).

Otherwise the message could be read like "important functionality is missing, install systemd". Novice users will better user automagic instead of some cryptic message. 
Let's not force user's choice.

Karlson2k:
Updated with rebase and better formatting.

configure.ac
Karlson2k: Currently (without this PR) configure works with automagic, but does not fail, if `--enable-logind` is used and `libsystemd` not found. See [this comment](https://github.com/shadow-maint/shadow/pull/1283#issuecomment-3137493493).
Let's not change the default behaviour and just fix non-failure for missing lib, when user provided explicit parameter.

Karlson2k:
For clarity.

Without this PR, current `configure` behaviour:
* no `--enable-logind` / `--disable-logind` parameter:
** if `libsystemd` found -> enable logind
** if `libsystemd` not found -> disable logind
* `--enable-logind` parameter:
** if `libsystemd` found -> enable logind
** if `libsystemd` not found -> disable logind (no error, wrong!)
*  `--disable-logind` parameter:
** `libsystemd` not checked and always disabled.

`configure` behaviour with this PR:
* no `--enable-logind` / `--disable-logind` parameters:
** if `libsystemd` found -> enable logind
** if `libsystemd` not found -> disable logind
* `--enable-logind` parameter:
** if `libsystemd` found -> enable logind
** if `libsystemd` not found -> error
*  `--disable-logind` parameter:
** `libsystemd` not checked and always disabled.

configure.ac
Karlson2k: > Current master that default value should be `yes`.
> 
> * https://github.com/shadow-maint/shadow/blob/1f617f8aa76b27de547adb34952633a7d2c85068/configure.ac#L154-L159

The actual default and behaviour is `yes if found`, but documented incorrectly.

lslebodn: > The actual default and behaviour is yes if found, but documented incorrectly.

Which is bug in implementation.

-1 for introducing complicated autodetection.

Please keep it **simple**  `yes/no` with useful error message for missing deps in case of `yes`

lslebodn: I like your version: `The version without automagic:` https://github.com/shadow-maint/shadow/pull/1283/files#r2243574381
Just a few comments

```
AC_SUBST([LIBSYSTEMD])
if test "X$enable_logind" != "Xno"; then
```
it simpler to read positive conditions; it improves understanding a bit. But it is kind of nitpick
The same as it is in current master `if test "X$enable_logind" = "Xyes"`
```
	AC_CHECK_LIB([systemd], [sd_session_get_remote_host],
		[
			enable_logind="yes"
```
^^ line is not needed; there is already value set by macro `AC_ARG_ENABLE`
```
			LIBSYSTEMD=-lsystemd
			AC_DEFINE([ENABLE_LOGIND], [1], [Define to manage session support with logind.])
		],[
			AC_MSG_ERROR([libsystemd not found, logind cannot be enabled. The feature is optional and it can be disabled.])
			AC_MSG_ERROR([Please install missing dependencies or disable it with configure time option --disable-logind])
```
I have a feeling the 2nd message will not be printed because the configure script would fail for the 1st one.
But I did not test it.

alejandro-colomar: > Automagic detection will not push anyone to some preferred (by you or someone else) solution. Automagic is used in this configure already, so no inconsistency. Automagic is not nice if specific `yes` or `no` value cannot be enforced. If user-provided value just silently reset to another value (like in current configure), this is a bad behaviour. This PR corrects it. If user selects explicitly `--enable-logind` then logind will be used or configure will fail. This is correct behaviour.

I **hate** automagic.  It makes things much more difficult to reason about.

> 
> What complication you are talking about? Let's compare. The version with automagic:
>
> 
> The version without automagic:

Please show diffs for comparing stuff.

> One line difference is not really huge complication. The difference can be shortened to zero lines by removing `maybe|auto` case.
> 
> > it is better to fail; and let user to resolve issue based on useful error message from the configure.
> 
> No, it is better to use what is found (with clear report at the end of the configure) if nothing is requested by the user or follow strictly what is requested by the user. This fits perfectly both novice users (no parameters) and package maintainers (explicit `--disable-logind` or `--enable-logind`).

I hate this kind of behavior.  I think autotools are horrible, and this is one of the reasons.  Maybe it's not autotools that are bad, but how people often use them.

Please don't use automagic here.  If there's existing automagic, let's remove it as much as we can.

> Otherwise the message could be read like "important functionality is missing, install systemd". Novice users will better user automagic instead of some cryptic message. Let's not force user's choice.

lslebodn:
> `configure` behaviour with this PR:
> 
>     * no `--enable-logind` / `--disable-logind` parameters:
>       ** if `libsystemd` found -> enable logind
>       ** if `libsystemd` not found -> disable logind

This is bad default behavior; there can be packaging changes in dependencies.
and something which was previously enabled will be silently disabled; even though `shadow` will not change at all.

Such silent surprises are very complicated to notice for maintainers; especially for features which have fallback implementation but fallback implementation can behave differently as it was in previous version which used "enabled optinal feature"

Karlson2k:
> > `configure` behaviour with this PR:
> > ```
> > * no `--enable-logind` / `--disable-logind` parameters:
> >   ** if `libsystemd` found -> enable logind
> >   ** if `libsystemd` not found -> disable logind
> > ```
> 
> This is bad default behavior; there can be packaging changes in dependencies. and something which was previously enabled will be silently disabled; even though `shadow` will not change at all.

Nope, this is good behaviour when parameters are **omitted**.
No package maintainers are using "no parameters" version. Typically all key parameters are specified explicitly.

> Such silent surprises are very complicated to notice for maintainers; especially for features which have fallback implementation but fallback implementation can behave differently as it was in previous version which used "enabled optinal feature"

Again, **no maintainers** are using absent parameters. Zero surprises.
The PR is actually **fixing** surprises, as currently logind interface can be silently disabled **even if explicitly enabled by `--enable-logind`**.

What use-case is not covered or not controlled by the suggested implementation?
Someone what to give a quick shot - use configure without parameters to adapt to the system.
Package maintainers with need to full control - just specify the required parameters.

alejandro-colomar:
> Nope, this is good behaviour when parameters are **omitted**. No package maintainers are using "no parameters" version. Typically all key parameters are specified explicitly.
> 
> > Such silent surprises are very complicated to notice for maintainers; especially for features which have fallback implementation but fallback implementation can behave differently as it was in previous version which used "enabled optinal feature"
> 
> Again, **no maintainers** are using absent parameters. Zero surprises. The PR is actually **fixing** surprises, as currently logind interface can be silently disabled **even if explicitly enabled by `--enable-logind`**.
> 
> What use-case is not covered or not controlled by the suggested implementation? Someone what to give a quick shot - use configure without parameters to adapt to the system. Package maintainers with need to full control - just specify the required parameters.

When I give a shot at random projects, I really prefer that their configure fails 3 or 4 times for me, telling me to install certain dependencies or pass some flags.  For example, GCC's configure tells me to `--disable-multilib` if I don't have the 32-bit libraries.

I strongly prefer no automagic.

configure.ac
Karlson2k: This is against autotools ideology.
`configure` must automatically adapt the build to the specific system.

`configure` must provide user the full control over key parameters **if user needs such control**.

Without automatic detection the user will have to always specify `--enable-logind` if systemd is used or specify `--disable-logind` if systemd is not used. So first user must detect what is available and then `configure` will repeat the detection.

If no detection needed, then configure code could be removed completely and user may use
`configure LIBSYSTEMD=-lsystemd CPPFLAGS="$CPPFLAGS -DENABLE_LOGIND"` (just need add some `#ifdef` guards to `.c` files) when logind must be used.
Or even simpler:
`make LIBSYSTEMD=-lsystemd CPPFLAGS="$CPPFLAGS -DENABLE_LOGIND"` to enable logind.

The same for other detections. For example, let user define HAVE_STRUCT_UTMPX_UT_HOST macro if  `utmpx.ut_host` if present.
Configure must offload the user from such things.

---
But automagic **must never be enforced**.
This PR corrects enforment of automagic (currently it is enforced!).
This version gives user a choice:

1. I do not care, just build something - use no parameter, logind enabled if detected
2. I need logind - use `--enable-logind`, if logind is not detected then configure fail
3. I do not need logind - use `--disable-logind`, logind is never used

Package maintainers always use the second or the third form.
Users, who just exploring, typically use the first form for a quick try. Let's not raise the entrance border.

Karlson2k: > -1 for introducing complicated autodetection.

It is not an introduction of autodectetion.
The autodetection was already in the code.

Please tell me how the next difference can be "complicated":

```diff
--- a/config1.ac
+++ b/config2.ac
@@ -1,15 +1,14 @@
 AC_ARG_ENABLE([logind],
        [AS_HELP_STRING([--enable-logind],
-               [enable logind @<:@default=yes if found@:>@])],
+               [enable logind @<:@default=yes@:>@])],
        [
                AS_CASE([${enableval}],
                        [yes],[],
                        [no],[],
-                       [maybe|auto],[enable_logind="auto"],
                        [AC_MSG_ERROR([bad parameter value for --enable-logind=${enableval}])]
                )
        ],
-       [enable_logind="auto"]
+       [enable_logind="yes"]
 )

 AC_SUBST([LIBSYSTEMD])
@@ -20,8 +19,8 @@ if test "X$enable_logind" != "Xno"; then
                        LIBSYSTEMD=-lsystemd
                        AC_DEFINE([ENABLE_LOGIND], [1], [Define to manage session support with logind.])
                ],[
-                       AS_IF([test "X$enable_logind" = "Xyes"], [AC_MSG_ERROR([libsystemd not found, logind cannot be enabled])])
-                       enable_logind="no"
+                       AC_MSG_ERROR([libsystemd not found, logind cannot be enabled. The feature is optional and it can be disabled.])
+                       AC_MSG_ERROR([Please install missing dependencies or disable it with configure time option --disable-logind])
                ]
        )
 fi
```

The version **without autodetection** is even larger.

Karlson2k:
> When I give a shot at random projects, I really prefer that their configure fails 3 or 4 times for me, telling me to install certain dependencies or pass some flags. For example, GCC's configure tells me to `--disable-multilib` if I don't have the 32-bit libraries.
> 
> I strongly prefer no automagic.

This is still not a... perfect approach.
If you want to have a control over build, just run `./configure --help` and use the parameters.
With try & fail approach, how you can be sure that the rest of the configure defaults fit your needs?

I strongly against the approach like "User, you must duplicate the configure job for detection. Then configure will re-check whether you, user, did your job properly."

Fail early, at configure stage, if mandatory component is missing is a good approach (but, if I'm not mistaken, you prefer compilation failures later rather then early configure failures).

configure.ac
alejandro-colomar: > This is against autotools ideology. `configure` must automatically adapt the build to the specific system.

Well, then I have good reasons to hate autotools.  The software might be okay, but their ideology is bad.

> `configure` must provide user the full control over key parameters **if user needs such control**.
> 
> Without automatic detection the user will have to always specify `--enable-logind` if systemd is used or specify `--disable-logind` if systemd is not used. So first user must detect what is available and then `configure` will repeat the detection.

Not necessarily.  One of the two could be the default.  But yes, one of them should always be specified.  I think it's a good idea.

> If no detection needed, then configure code could be removed completely and user may use `configure LIBSYSTEMD=-lsystemd CPPFLAGS="$CPPFLAGS -DENABLE_LOGIND"` (just need add some `#ifdef` guards to `.c` files) when logind must be used. Or even simpler: `make LIBSYSTEMD=-lsystemd CPPFLAGS="$CPPFLAGS -DENABLE_LOGIND"` to enable logind.

That's bad, we agree.  There has to be a single knob.

> 
> The same for other detections. For example, let user define HAVE_STRUCT_UTMPX_UT_HOST macro if `utmpx.ut_host` if present. Configure must offload the user from such things.

This is not in the same category, IMO.

alejandro-colomar: > > -1 for introducing complicated autodetection.
> 
> It is not an introduction of autodectetion. The autodetection was already in the code.
> 
> Please tell me how the next difference can be "complicated":
> 
> ```diff
> --- a/config1.ac
> +++ b/config2.ac
> @@ -1,15 +1,14 @@
>  AC_ARG_ENABLE([logind],
>         [AS_HELP_STRING([--enable-logind],
> -               [enable logind @<:@default=yes if found@:>@])],
> +               [enable logind @<:@default=yes@:>@])],
>         [
>                 AS_CASE([${enableval}],
>                         [yes],[],
>                         [no],[],
> -                       [maybe|auto],[enable_logind="auto"],
>                         [AC_MSG_ERROR([bad parameter value for --enable-logind=${enableval}])]
>                 )
>         ],
> -       [enable_logind="auto"]
> +       [enable_logind="yes"]
>  )
> 
>  AC_SUBST([LIBSYSTEMD])
> @@ -20,8 +19,8 @@ if test "X$enable_logind" != "Xno"; then
>                         LIBSYSTEMD=-lsystemd
>                         AC_DEFINE([ENABLE_LOGIND], [1], [Define to manage session support with logind.])
>                 ],[
> -                       AS_IF([test "X$enable_logind" = "Xyes"], [AC_MSG_ERROR([libsystemd not found, logind cannot be enabled])])
> -                       enable_logind="no"
> +                       AC_MSG_ERROR([libsystemd not found, logind cannot be enabled. The feature is optional and it can be disabled.])
> +                       AC_MSG_ERROR([Please install missing dependencies or disable it with configure time option --disable-logind])
>                 ]
>         )
>  fi
> ```
> 
> The version **without autodetection** is even larger.

The version without autodetection is config2.ac, right?  That's one line shorter (and the lines are less complex too).  Anyway, the important thing is the behavior, not the small difference in number of lines

alejandro-colomar:
> > When I give a shot at random projects, I really prefer that their configure fails 3 or 4 times for me, telling me to install certain dependencies or pass some flags. For example, GCC's configure tells me to `--disable-multilib` if I don't have the 32-bit libraries.
> > I strongly prefer no automagic.
> 
> This is still not a... perfect approach. If you want to have a control over build, just run `./configure --help` and use the parameters. With try & fail approach, how you can be sure that the rest of the configure defaults fit your needs?

You're still allowed to check the documentation with the iterative approach.

> I strongly against the approach like "User, you must duplicate the configure job for detection. Then configure will re-check whether you, user, did your job properly."

autotools's approach is akin to completing all system calls after EINTR.  Reporting an error and letting the user run again is simpler.

Fail early, fail hard.

> 
> Fail early, at configure stage, if mandatory component is missing is a good approach (but, if I'm not mistaken, you prefer compilation failures later rather then early configure failures).

If a header file is mandatory, I prefer a compilation error.  That's much simpler, and both work just fine.  Whenever you see "missing `<foo.h>`", you can `apt-file find /usr/include/foo.h` (or your distro equivalent) to find which dependency you need.

Especially in a small project where a build takes a few seconds, there's no big difference in usability.  If compilation took minutes, or hours (as in GCC), it might make sense to fail earlier for some dependencies (and even GCC doesn't fail in configure for some dependencies, and just leave it to the compiler).

Karlson2k:
@alejandro-colomar If you like the approach when configure asks user for parameters instead of detection of available features, then I hope you will like PR #1317.

alejandro-colomar:
Merged <https://github.com/shadow-maint/shadow/pull/1318> instead.

PR #1284, opened 2025-07-10
Implement ability to use both utmp and logind

Karlson2k:
This PR adds ability to use BOTH utmp and logind.
If both enabled, then for the number of active sessions logind used first, and utmp is used as a fallback. The same for "get_session_host": logind tried first and then utmp tried as a fallback.
Other utmp functionality enabled independently from logind.

This patch is needed because modern systemd (at least 256, 257) does not update utmp. Neither pad_systemd, nor logind does not write information to utmp.
This still breaks some software, including basic `logname` command.

With the patch utmp and logind can be enabled/disabled independently, allowing building a package that suits particular needs.
The only limitation is that utmp and logind cannot be simultaneously disabled (at least one option must be enabled).

The updated test has not been check for proper functioning on my machine.

### DIFF
--- a/lib/limits.c
+++ b/lib/limits.c
@@ -119,7 +119,8 @@ set_umask(const char *value)
 /* Counts the number of user logins and check against the limit */
 static int check_logins (const char *name, const char *maxlogins)
 {
-	unsigned long limit, count;
+	unsigned long limit;
+	int exceeded = -1;
 
 	if (str2ul(&limit, maxlogins) == -1) {
 		if (errno == ERANGE) {
@@ -134,9 +135,16 @@ static int check_logins (const char *name, const char *maxlogins)
 		return LOGIN_ERROR_LOGIN;
 	}
 
-	count = active_sessions_count(name, limit);
+#ifdef ENABLE_LOGIND
+	if (exceeded < 0)
+		exceeded = active_sessions_limit_exceeded_lgnd(name, limit);
+#endif /* ENABLE_LOGIND */
+#ifdef ENABLE_UTMP
+	if (exceeded < 0)
+		exceeded = active_sessions_limit_exceeded_utmp(name, limit);
+#endif /* ENABLE_UTMP */
 
-	if (count > limit) {
+	if (exceeded > 0) {
 		SYSLOG ((LOG_WARN,
 		         "Too many logins (max %lu) for %s\n",
 		         limit, name));
--- a/lib/logind.c
+++ b/lib/logind.c
@@ -14,7 +14,7 @@
 
 #include <systemd/sd-login.h>
 
-int get_session_host (char **out)
+int get_session_host_lgnd (char **out)
 {
     char *host = NULL;
     char *session = NULL;
@@ -36,17 +36,17 @@ done:
     return ret;
 }
 
-unsigned long active_sessions_count(const char *name, MAYBE_UNUSED unsigned long limit)
+int active_sessions_limit_exceeded_lgnd(const char *name, unsigned long limit)
 {
     struct passwd *pw;
     unsigned long count = 0;
 
     pw = prefix_getpwnam(name);
     if (pw == NULL) {
-        return 0;
+        return -1;
     }
 
     count = sd_uid_get_sessions(pw->pw_uid, 0, NULL);
 
-    return count;
+	return (count > limit) ? 1 : 0;
 }
--- a/lib/prototypes.h
+++ b/lib/prototypes.h
@@ -463,17 +463,18 @@ extern int user_busy (const char *name, uid_t uid);
 /*
  * Session management: utmp.c or logind.c
  */
+#ifdef ENABLE_UTMP
 
 /**
- * @brief Get host for the current session
+ * @brief Get host for the current session from utmp
  *
  * @param[out] out Host name
  *
  * @return 0 or a positive integer if the host was obtained properly,
  *         another value on error.
  */
-extern int get_session_host (char **out);
-#ifndef ENABLE_LOGIND
+extern int get_session_host_utmp (char **out);
+
 /**
  * @brief Update or create an utmp entry in utmp, wtmp, utmpw, or wtmpx
  *
@@ -498,19 +499,50 @@ extern int update_utmp (const char *user,
 extern void record_failure(const char *failent_user,
                            const char *tty,
                            const char *hostname);
-#endif /* ENABLE_LOGIND */
 
 /**
- * @brief Number of active user sessions
+ * @brief Check if the number of active utmp user sessions has exceeded limit
  *
  * @param[in] name username
  * @param[in] limit maximum number of active sessions
  *
- * @return number of active sessions.
+ * @return 1 if number of active user sessions has exceeded the limit,
+ *         0 if number of active user sessions less or equal the limit,
+ *         -1 on error
  *
  */
-extern unsigned long active_sessions_count(const char *name,
-                                           unsigned long limit);
+extern int active_sessions_limit_exceeded_utmp(const char *name,
+											   unsigned long limit);
+
+#endif /* ENABLE_UTMP */
+#ifdef ENABLE_LOGIND
+
+/**
+ * @brief Get host for the current session from logind
+ *
+ * @param[out] out Host name
+ *
+ * @return 0 or a positive integer if the host was obtained properly,
+ *         another value on error.
+ */
+extern int get_session_host_lgnd (char **out);
+
+/**
+ * @brief Check if the number of active logind user sessions has exceeded limit
+ *
+ * @param[in] name username
+ * @param[in] limit maximum number of active sessions
+ *
+ * @return 1 if number of active user sessions has exceeded the limit,
+ *         0 if number of active user sessions less or equal the limit,
+ *         -1 on error
+ *
+ */
+extern int active_sessions_limit_exceeded_lgnd(const char *name,
+											   unsigned long limit);
+
+#endif /* ENABLE_LOGIND */
+
 
 /* valid.c */
 extern bool valid (const char *, const struct passwd *);
--- a/lib/utmp.c
+++ b/lib/utmp.c
@@ -181,7 +181,7 @@ get_current_utmp(void)
 
 
 int
-get_session_host(char **out)
+get_session_host_utmp(char **out)
 {
 	int           ret = 0;
 	struct utmpx  *ut;
@@ -397,15 +397,14 @@ record_failure(const char *failent_user, const char *tty, const char *hostname)
 	}
 }
 
-
-unsigned long
-active_sessions_count(const char *name, unsigned long limit)
+int
+active_sessions_limit_exceeded_utmp(const char *name, unsigned long limit)
 {
 	struct utmpx   *ut;
 	unsigned long  count = 0;
 
 	setutxent();
-	while ((ut = getutxent()))
+	while ((ut = getutxent()) && (count <= limit))
 	{
 		if (USER_PROCESS != ut->ut_type) {
 			continue;
@@ -417,11 +416,8 @@ active_sessions_count(const char *name, unsigned long limit)
 			continue;
 		}
 		count++;
-		if (count > limit) {
-			break;
-		}
 	}
 	endutxent();
 
-	return count;
+	return (count > limit) ? 1 : 0;
 }
--- a/src/login.c
+++ b/src/login.c
@@ -450,7 +450,7 @@ static /*@observer@*/const char *get_failent_user (/*@returned@*/const char *use
  */
 int main (int argc, char **argv)
 {
-	int            err;
+	int            err = -1;
 	bool           subroot = false;
 	char           **envp = environ;
 	char           *host = NULL;
@@ -504,7 +504,14 @@ int main (int argc, char **argv)
 		exit (1);	/* must be a terminal */
 	}
 
-	err = get_session_host(&host);
+#ifdef ENABLE_LOGIND
+	if (0 != err)
+		err = get_session_host_lgnd(&host);
+#endif /* ENABLE_LOGIND */
+#ifdef ENABLE_UTMP
+	if (0 != err)
+		err = get_session_host_utmp(&host);
+#endif /* ENABLE_UTMP */
 	/*
 	 * Be picky if run by normal users (possible if installed setuid
 	 * root), but not if run by root.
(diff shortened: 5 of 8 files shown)


lib/limits.c
```diff
@@ -134,9 +135,16 @@ static int check_logins (const char *name, const char *maxlogins)
 		return LOGIN_ERROR_LOGIN;
 	}
 
-	count = active_sessions_count(name, limit);
+#ifdef ENABLE_LOGIND
+	if (exceeded < 0)
```
alejandro-colomar, line 139: Isn't this dead code?  The condition is always true, because exceeded is `-1`, so you could do that unconditionally, right?

lib/prototypes.h
```diff
@@ -498,19 +499,50 @@ extern int update_utmp (const char *user,
...
+ * @param[in] name username
+ * @param[in] limit maximum number of active sessions
+ *
+ * @return 1 if number of active user sessions has exceeded the limit,
+ *         0 if number of active user sessions less or equal the limit,
+ *         -1 on error
+ *
+ */
+extern int active_sessions_limit_exceeded_utmp(const char *name,
+											   unsigned long limit);
```
alejandro-colomar, line 515: There is some alignment mistake here.

lib/prototypes.h
```diff
@@ -498,19 +499,50 @@ extern int update_utmp (const char *user,
...
- * @return number of active sessions.
+ * @return 1 if number of active user sessions has exceeded the limit,
+ *         0 if number of active user sessions less or equal the limit,
+ *         -1 on error
  *
  */
-extern unsigned long active_sessions_count(const char *name,
-                                           unsigned long limit);
+extern int active_sessions_limit_exceeded_lgnd(const char *name,
+											   unsigned long limit);
```
alejandro-colomar, line 542: Alignment.  Please use spaces for alignment (tabs for indentation).

alejandro-colomar, line 542: Since this API set is becoming big and complex, it would be interesting to move it to its own header files, "lib/logind.h" and "lib/utmp.h".

lib/utmp.c
```diff
@@ -397,15 +397,14 @@ record_failure(const char *failent_user, const char *tty, const char *hostname)
...
-active_sessions_count(const char *name, unsigned long limit)
+int
+active_sessions_limit_exceeded_utmp(const char *name, unsigned long limit)
 {
 	struct utmpx   *ut;
 	unsigned long  count = 0;
 
 	setutxent();
-	while ((ut = getutxent()))
+	while ((ut = getutxent()) && (count <= limit))
```
alejandro-colomar, line 407: Please don't add redundant parens (the existing ones are for allowing `=`).

lib/utmp.c
```diff
@@ -417,11 +416,8 @@ active_sessions_count(const char *name, unsigned long limit)
 			continue;
 		}
 		count++;
-		if (count > limit) {
-			break;
-		}
```
alejandro-colomar, line 422: Why is this moved to the while condition?  Is it necessary, or is it cosmetic?  If the latter, please do it in a separate commit.

src/login.c
```diff
@@ -504,7 +504,14 @@ int main (int argc, char **argv)
 		exit (1);	/* must be a terminal */
 	}
 
-	err = get_session_host(&host);
+#ifdef ENABLE_LOGIND
+	if (0 != err)
```
alejandro-colomar, line 508: Dead code?

lib/limits.c
```diff
@@ -119,7 +119,8 @@ set_umask(const char *value)
 /* Counts the number of user logins and check against the limit */
 static int check_logins (const char *name, const char *maxlogins)
 {
-	unsigned long limit, count;
+	unsigned long limit;
+	int exceeded = -1;
```
alejandro-colomar, line 123: Please, don't initialize.  Prefer assignment closer to use.  Initialization of variables can hide logic bugs.

lib/limits.c
```diff
@@ -134,9 +135,16 @@ static int check_logins (const char *name, const char *maxlogins)
...
+	if (exceeded < 0)
+		exceeded = active_sessions_limit_exceeded_lgnd(name, limit);
+#endif /* ENABLE_LOGIND */
+#ifdef ENABLE_UTMP
+	if (exceeded < 0)
+		exceeded = active_sessions_limit_exceeded_utmp(name, limit);
+#endif /* ENABLE_UTMP */
 
-	if (count > limit) {
+	if (exceeded > 0) {
```
alejandro-colomar, line 147: What if it's still -1?  Is that possible?

lib/logind.c
```diff
@@ -36,17 +36,17 @@ int get_session_host (char **out)
     return ret;
 }
 
-unsigned long active_sessions_count(const char *name, MAYBE_UNUSED unsigned long limit)
+int active_sessions_limit_exceeded_lgnd(const char *name, unsigned long limit)
```
alejandro-colomar, line 39: `exceeds`, in present, is probably more conventional, no?

ikerexxe:
I have significant concerns about this approach. As you've noted, this change could potentially **break basic operating system software**, and it appears to exacerbate the problem rather than mitigating it.

Currently, the established practice is to choose between `utmp` or `logind` for session management. When distributions decide to transition from one to the other, it typically involves a coordinated change in the build configuration. For example, in Fedora, a change of this magnitude would necessitate a **Fedora System-Wide Change** proposal to ensure proper communication and adaptation across the ecosystem. Other distributions likely have similar mechanisms.

While I'm not privy to the specific reasons behind `systemd`'s shift away from updating `utmp` directly, I believe that the solution should involve proper communication and adaptation from `systemd` to other packages, rather than introducing a dual-system approach that could lead to further fragmentation and instability.

Introducing a scenario where both `utmp` and `logind` can be simultaneously enabled, with a fallback mechanism, seems to complicate the system and potentially lead to unexpected behavior and maintenance challenges down the line. The current "either/or" model, while perhaps requiring more effort for transitions, ensures a clearer and more predictable session management environment.

alejandro-colomar:
Cc: @thesamesam

Karlson2k:
> I have significant concerns about this approach. As you've noted, this change could potentially **break basic operating system software**, and it appears to exacerbate the problem rather than mitigating it.

Actually, the basic system software is already broken.
This approach allows to **choose** the way, how things are handled. It does not enforce to use both `utmp` and `logind`.

> Currently, the established practice is to choose between `utmp` or `logind` for session management. When distributions decide to transition from one to the other, it typically involves a coordinated change in the build configuration. For example, in Fedora, a change of this magnitude would necessitate a **Fedora System-Wide Change** proposal to ensure proper communication and adaptation across the ecosystem. Other distributions likely have similar mechanisms.

And this will continue to work perfectly fine. Fedora still can use `logind`-only mode.

> While I'm not privy to the specific reasons behind `systemd`'s shift away from updating `utmp` directly, I believe that the solution should involve proper communication and adaptation from `systemd` to other packages, rather than introducing a dual-system approach that could lead to further fragmentation and instability.

I cannot see any instability, that can be introduced by this change. 
For example, instead of silently accepting more logins than allowed by limit (as it currently done when number of logins cannot be acquired from systemd), a fallback path is used.

This PR can be extended with `--enable-utmp=update-only` mode, which could be enabled only together with systemd. In such mode utmp could be only updated, but not used as a fallback.

However, I'm sure that fallback is better than nothing when systemd fails.

> Introducing a scenario where both `utmp` and `logind` can be simultaneously enabled, with a fallback mechanism, seems to complicate the system and potentially lead to unexpected behavior and maintenance challenges down the line. The current "either/or" model, while perhaps requiring more effort for transitions, ensures a clearer and more predictable session management environment.

This situation is already broken and must be fixed now.
For example, on Gentoo with current stable systemd (sys-apps/systemd-257.6) console sessions are incorrect: any software relays on `logname` command or `getlogin()` function is not working properly (while the same software in terminal emulators under KDE is fine). The output is always literally `LOGIN` instead of the name of the actual user.  (@thesamesam I decided not to open a Gentoo bug as it is a clear upstream problem).
coreutls (for `logname` command) will not fix this problem:
https://github.com/coreutils/coreutils/blob/8f9fc8f08c10c3b097211f95c6354a85d41f1101/src/logname.c#L72-L74

Glibc uses audit first (disabled on many distributions and has global performance penalty) and then utmp as a fallback:
https://github.com/bminor/glibc/blob/d6c2760ef7f7cdeab912767f04db4b14632fbb5f/sysdeps/unix/sysv/linux/getlogin.c#L34-L38
This is a `getlogin_fd0 ()` implementation:
https://github.com/bminor/glibc/blob/d6c2760ef7f7cdeab912767f04db4b14632fbb5f/sysdeps/unix/getlogin.c

Probably glibc will never be fixed:
https://github.com/thkukuk/utmpx/blob/main/Y2038.md#glibc

In short:
* full transition may take years, while the situation is broken already.
* systemd is still updating utmp: https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#UtmpMode=, but these updates are limited ("user" mode will not work if user is not specified for the service, i.e. `login` must still update utmp otherwise it stuck with `LOGIN` entry for "login process").
* agetty is still updating utmp: https://github.com/util-linux/util-linux/blob/6c87a3ac5ee96194e0e461db24b1bbfe369bdb59/term-utils/agetty.c#L976
* neither `logind` nor `pam_systemd.so` updates utmp.

This looks inconsistent (utmp is updated before firing `login` binary, but not by the login binary, neither during login process) and the best solution is to add build-time option for `utmp` manipulation, like suggested in this PR.
I'd emphasise: this build options **does not enforce** `utmp` updates for these users who do not need it. 

When other software will be fixed, or utmp is completely eliminated, or when one does not want `utmp` updates, support for `utmp` in shadow can be easily disabled at build-time

lib/logind.c
alejandro-colomar: Or I don't know.  I'm not convinced by the function name, but I don't want to strongly oppose it if you want it.

Karlson2k: English is not my native language, so I'm not very picky in English words. :)
I'm fine with renaming it to `active_sessions_limit_exceeds_lgnd`.
Probably it should be renamed to `active_sessions_limit_ok_lgnd` (with reversing the logic).
Select whatever is nice for you and I'll update the code.

alejandro-colomar: Please change it to exceeds for now.  I'm not sure about ok.

(I still don't fully understand this, so we may need to come back to this issue.  Sorry.)

Karlson2k:
@thesamesam PR https://github.com/shadow-maint/shadow/pull/1282 is needed to complete the fix for Gentoo.
See also details in the comment https://github.com/shadow-maint/shadow/pull/1282#issuecomment-3063358096 (when built **without logind**).

alejandro-colomar:
> @thesamesam PR #1282 is needed to complete the fix for Gentoo. See also details in the comment [#1282 (comment)](https://github.com/shadow-maint/shadow/pull/1282#issuecomment-3063358096) (when built **without logind**).

Is there a Gentoo bug ticket?  Could you please link to it?

ikerexxe:
> Actually, the basic system software is already broken. This approach allows to **choose** the way, how things are handled. It does not enforce to use both `utmp` and `logind`.

I must have misunderstood the PR because I understood that the fallback was going to be forced. Since you are making it configurable I see it as fine.

> Probably glibc will never be fixed: https://github.com/thkukuk/utmpx/blob/main/Y2038.md#glibc

From my conversations with their developers it's not going to be fixed, so we will all have to migrate to logind (or some other mechanism) sooner rather than later.

lib/logind.c
```diff
@@ -36,17 +36,17 @@ int get_session_host (char **out)
...
     pw = prefix_getpwnam(name);
     if (pw == NULL) {
-        return 0;
+        return -1;
     }
 
     count = sd_uid_get_sessions(pw->pw_uid, 0, NULL);
 
-    return count;
+	return (count > limit) ? 1 : 0;
```
ikerexxe, line 51: Indentation issue

thesamesam:
(Just saying that I've seen the discussion(s) and I will look properly, but I can't look just yet and it's not the kind of thing I can give a quick take on. Thank you for the ping!)

tests/unit/test_logind.c
```diff
@@ -35,34 +35,52 @@ __wrap_sd_uid_get_sessions(uid_t uid, int require_active, char ***sessions)
  **********************/
 static void test_active_sessions_count_return_ok(void **state)
```
ikerexxe, line 36: Since you are changing the meaning of the test I'd also change the name of the function.

tests/unit/test_logind.c
```diff
@@ -35,34 +35,52 @@ __wrap_sd_uid_get_sessions(uid_t uid, int require_active, char ***sessions)
...
 int main(void)
 {
     const struct CMUnitTest tests[] = {
-        cmocka_unit_test(test_active_sessions_count_return_ok),
-        cmocka_unit_test(test_active_sessions_count_prefix_getpwnam_failure),
+        cmocka_unit_test(test_active_sessions_limit_exceeded_lgnd_return_ok),
+        cmocka_unit_test(\
+        test_active_sessions_limit_exceeded_lgnd_return_exceeded),
+        cmocka_unit_test(\
+        test_active_sessions_limit_exceeded_lgnd_prefix_failure),
```
ikerexxe, line 83: Let's keep this in the same line

lib/utmp.c
```diff
@@ -397,15 +397,14 @@ record_failure(const char *failent_user, const char *tty, const char *hostname)
 	}
 }
 
-
-unsigned long
-active_sessions_count(const char *name, unsigned long limit)
+int
+active_sessions_limit_exceeded_utmp(const char *name, unsigned long limit)
```
ikerexxe, line 401: This function doesn't return `-1` on error paths like `setutxent()` or `getutxent()` failures. Currently, if `setutxent()` fails, it will just iterate through 0 entries and return `0` (not exceeded). If `getutxent()` fails to return anything, it also returns `0`. This means `utmp` might silently fail to check the limit and incorrectly report "not exceeded" if there's an issue with `utmp` file access.

Karlson2k: Yes, good point.
I'll update the code.

tests/unit/test_logind.c
Karlson2k: The line becomes longer then 80 chars.
If it is not a problem, I'll move back to the same line.

alejandro-colomar: In general, I prefer to avoid surpassing the right margin, but in this case, let's keep it in one line.  Tests are special, anyway.

Karlson2k:
@alejandro-colomar Before further updating this PR, let's finish #1292 first, as this PR should be used on top.

tests/unit/test_logind.c
ikerexxe: Yes, let's keep it in the same line

alejandro-colomar:
This needs to be rebased.

alejandro-colomar:
This would need some rebasing.

Karlson2k:
I'll take care about it soon or shortly after the New Year.

PR #1286, opened 2025-07-12
src/login.c: fixed check for "init" process, when PAM is not used

Karlson2k:
When 'login' is started as 'init' process it has PID == 1. If PAM is not used, 'login' does not fork itself and getppid() return 0. To simplify the code, check for own PID before forking which gives correct value ('1' for 'init' process) for both PAM and non-PAM builds.

Initially the code for 'init' mode was introduced to support starting by `init=/bin/login`.
See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=374547


Karlson2k:
Superseded by #1305

PR #1287, opened 2025-07-12
fixes for configure and autotools

Karlson2k:
A set of random configure (and autotools) fixes and improvements.
As requested [here](https://github.com/shadow-maint/shadow/pull/1283#discussion_r2200800045).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,4 +1,5 @@
 ## Process this file with automake to produce Makefile.in
+ACLOCAL_AMFLAGS = -I m4
 
 SUBDIRS = lib
 
--- a/configure.ac
+++ b/configure.ac
@@ -254,13 +254,13 @@ PKG_CHECK_MODULES([CMOCKA], [cmocka], [have_cmocka="yes"],
 	[AC_MSG_WARN([libcmocka not found, cmocka tests will not be built])])
 AM_CONDITIONAL([HAVE_CMOCKA], [test x$have_cmocka = xyes])
 
+AC_ARG_ENABLE([vendordir],
+    [AS_HELP_STRING([--enable-vendordir=DIR], [Directory for distribution provided configuration files])],[],[])
 AC_CHECK_LIB([econf],[econf_readDirs],[LIBECONF="-leconf"],[LIBECONF=""])
 if test -n "$LIBECONF"; then
         AC_DEFINE_UNQUOTED([VENDORDIR], ["$enable_vendordir"],
 					[Directory for distribution provided configuration files])
 	ECONF_CPPFLAGS="-DUSE_ECONF=1"
-	AC_ARG_ENABLE([vendordir],
-		[AS_HELP_STRING([--enable-vendordir=DIR], [Directory for distribution provided configuration files])],[],[])
 fi
 AC_SUBST([ECONF_CPPFLAGS])
 AC_SUBST([LIBECONF])
--- a/configure.ac
+++ b/configure.ac
@@ -345,13 +345,9 @@ if test "$enable_logind" = "yes"; then
 fi
 AM_CONDITIONAL(ENABLE_LOGIND, test "x$enable_logind" != "xno")
 
-AC_SUBST(LIBCRYPT)
 AC_CHECK_LIB(crypt, crypt, [LIBCRYPT=-lcrypt],
 	[AC_MSG_ERROR([crypt() not found])])
-
-AC_SUBST(LIYESCRYPT)
-AC_CHECK_LIB(crypt, crypt, [LIYESCRYPT=-lcrypt],
-	[AC_MSG_ERROR([crypt() not found])])
+AC_SUBST(LIBCRYPT)
 
 AC_SUBST(LIBBSD)
 if test "$with_libbsd" != "no"; then
--- a/configure.ac
+++ b/configure.ac
@@ -20,8 +20,8 @@ AC_SUBST([LIBSUBID_ABI_MICRO], [libsubid_abi_micro])
 AC_SUBST([LIBSUBID_ABI], [libsubid_abi])
 
 dnl Some hacks...
-test "$prefix" = "NONE" && prefix="/usr"
-test "$prefix" = "/usr" && exec_prefix=""
+test "x$prefix" = "xNONE" && prefix="/usr"
+test "X$prefix" = "X/usr" && exec_prefix=""
 
 AC_USE_SYSTEM_EXTENSIONS
 
@@ -207,9 +207,9 @@ AC_ARG_WITH([libbsd],
 	[AS_HELP_STRING([--with-libbsd], [use libbsd support @<:@default=yes if found@:>@])],
 	[with_libbsd=$withval], [with_libbsd=yes])
 
-if test "$with_group_name_max_length" = "no" ; then
+if test "X$with_group_name_max_length" = "Xno" ; then
 	with_group_name_max_length=0
-elif test "$with_group_name_max_length" = "yes" ; then
+elif test "X$with_group_name_max_length" = "Xyes" ; then
 	with_group_name_max_length=32
 fi
 AC_DEFINE_UNQUOTED([GROUP_NAME_MAX_LENGTH], [$with_group_name_max_length], [max group name length])
@@ -218,27 +218,27 @@ GROUP_NAME_MAX_LENGTH="$with_group_name_max_length"
 
 
 AM_CONDITIONAL([USE_SHA_CRYPT], [test "x$with_sha_crypt" = "xyes"])
-if test "$with_sha_crypt" = "yes"; then
+if test "X$with_sha_crypt" = "Xyes"; then
 	AC_DEFINE([USE_SHA_CRYPT], [1], [Define to allow the SHA256 and SHA512 password encryption algorithms])
 fi
 
 AM_CONDITIONAL([USE_BCRYPT], [test "x$with_bcrypt" = "xyes"])
-if test "$with_bcrypt" = "yes"; then
+if test "X$with_bcrypt" = "Xyes"; then
 	AC_DEFINE([USE_BCRYPT], [1], [Define to allow the bcrypt password encryption algorithm])
 fi
 
 AM_CONDITIONAL([USE_YESCRYPT], [test "x$with_yescrypt" = "xyes"])
-if test "$with_yescrypt" = "yes"; then
+if test "X$with_yescrypt" = "Xyes"; then
 	AC_DEFINE([USE_YESCRYPT], [1], [Define to allow the yescrypt password encryption algorithm])
 fi
 
-if test "$with_nscd" = "yes"; then
+if test "X$with_nscd" = "Xyes"; then
 	AC_CHECK_FUNC([posix_spawn],
 	              [AC_DEFINE([USE_NSCD], [1], [Define to support flushing of nscd caches])],
 	              [AC_MSG_ERROR([posix_spawn is needed for nscd support])])
 fi
 
-if test "$with_sssd" = "yes"; then
+if test "X$with_sssd" = "Xyes"; then
 	AC_CHECK_FUNC([posix_spawn],
 	              [AC_DEFINE([USE_SSSD], [1], [Define to support flushing of sssd caches])],
 	              [AC_MSG_ERROR([posix_spawn is needed for sssd support])])
@@ -273,12 +273,12 @@ if test "x$enable_vendordir" != x; then
 fi
 AM_CONDITIONAL([HAVE_VENDORDIR], [test "x$enable_vendordir" != x])
 
-if test "$enable_shadowgrp" = "yes"; then
+if test "X$enable_shadowgrp" = "Xyes"; then
 	AC_DEFINE([SHADOWGRP], [1], [Define to support the shadow group file.])
 fi
 AM_CONDITIONAL([SHADOWGRP], [test "x$enable_shadowgrp" = "xyes"])
 
-if test "$enable_man" = "yes"; then
+if test "X$enable_man" = "Xyes"; then
 	dnl
 	dnl Check for xsltproc
 	dnl
@@ -315,7 +315,7 @@ if test "$enable_subids" != "no"; then
 fi
 AM_CONDITIONAL([ENABLE_SUBIDS], [test "x$enable_subids" != "xno"])
 
-if test "$enable_lastlog" = "yes" && test "$ac_cv_header_lastlog_h" = "yes"; then
+if test "X$enable_lastlog" = "Xyes" && test "X$ac_cv_header_lastlog_h" = "Xyes"; then
 	AC_CACHE_CHECK([for ll_host in struct lastlog],
 		[ac_cv_struct_lastlog_ll_host],
 		[AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <lastlog.h>]],
@@ -325,7 +325,7 @@ if test "$enable_lastlog" = "yes" && test "$ac_cv_header_lastlog_h" = "yes"; the
 		]
 	)
 
-	if test "$ac_cv_struct_lastlog_ll_host" = "yes"; then
+	if test "X$ac_cv_struct_lastlog_ll_host" = "Xyes"; then
 		AC_DEFINE([HAVE_LL_HOST], [1],
 			[Define if struct lastlog has ll_host])
 		AC_DEFINE([ENABLE_LASTLOG], [1], [Define to support lastlog.])
@@ -338,7 +338,7 @@ fi
 AM_CONDITIONAL([ENABLE_LASTLOG], [test "x$enable_lastlog" != "xno"])
 
 AC_SUBST([LIBSYSTEMD])
-if test "$enable_logind" = "yes"; then
+if test "X$enable_logind" = "Xyes"; then
 	AC_CHECK_LIB([systemd], [sd_session_get_remote_host],
 		[enable_logind="yes"; [LIBSYSTEMD=-lsystemd];
 		AC_DEFINE([ENABLE_LOGIND], [1],
@@ -356,7 +356,7 @@ if test "$with_libbsd" != "no"; then
 	AC_SEARCH_LIBS([readpassphrase], [bsd], [], [
 		AC_MSG_ERROR([readpassphrase() is missing, either from libc or libbsd])
 	])
-	AS_IF([test "$ac_cv_search_readpassphrase" = "-lbsd"], [
+	AS_IF([test "X$ac_cv_search_readpassphrase" = "X-lbsd"], [
 		PKG_CHECK_MODULES([LIBBSD], [libbsd-overlay])
 	])
 	dnl Make sure either the libc or libbsd provide the header.
@@ -376,17 +376,17 @@ AM_CONDITIONAL([WITH_LIBBSD], [test x$with_libbsd = xyes])
 AC_SUBST([LIBACL])
 if test "$with_acl" != "no"; then
 	AC_CHECK_HEADERS([acl/libacl.h attr/error_context.h], [acl_header="yes"], [acl_header="no"])
-	if test "$acl_header$with_acl" = "noyes" ; then
+	if test "X$acl_header$with_acl" = "Xnoyes" ; then
 		AC_MSG_ERROR([acl/libacl.h or attr/error_context.h is missing])
-	elif test "$acl_header" = "yes" ; then
+	elif test "X$acl_header" = "Xyes" ; then
 		AC_CHECK_LIB([acl], [perm_copy_file],
 		             [AC_CHECK_LIB([acl], [perm_copy_fd],
 		                           [acl_lib="yes"],
 		                           [acl_lib="no"])],
 		             [acl_lib="no"])
-		if test "$acl_lib$with_acl" = "noyes" ; then
+		if test "X$acl_lib$with_acl" = "Xnoyes" ; then
 			AC_MSG_ERROR([libacl not found])
-		elif test "$acl_lib" = "no" ; then
+		elif test "X$acl_lib" = "Xno" ; then
 			with_acl="no"
 		else
 			AC_DEFINE([WITH_ACL], [1],
@@ -402,17 +402,17 @@ fi
 AC_SUBST([LIBATTR])
 if test "$with_attr" != "no"; then
 	AC_CHECK_HEADERS([attr/libattr.h attr/error_context.h], [attr_header="yes"], [attr_header="no"])
-	if test "$attr_header$with_attr" = "noyes" ; then
+	if test "X$attr_header$with_attr" = "noyes" ; then
 		AC_MSG_ERROR([attr/libattr.h or attr/error_context.h is missing])
-	elif test "$attr_header" = "yes" ; then
+	elif test "X$attr_header" = "Xyes" ; then
 		AC_CHECK_LIB([attr], [attr_copy_file],
 		             [AC_CHECK_LIB([attr], [attr_copy_fd],
 		                           [attr_lib="yes"],
 		                           [attr_lib="no"])],
 		             [attr_lib="no"])
-		if test "$attr_lib$with_attr" = "noyes" ; then
+		if test "X$attr_lib$with_attr" = "Xnoyes" ; then
 			AC_MSG_ERROR([libattr not found])
-		elif test "$attr_lib" = "no" ; then
+		elif test "X$attr_lib" = "Xno" ; then
 			with_attr="no"
 		else
 			AC_DEFINE([WITH_ATTR], [1],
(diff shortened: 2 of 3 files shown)


configure.ac
```diff
@@ -345,13 +345,12 @@ if test "$enable_logind" = "yes"; then
 fi
 AM_CONDITIONAL(ENABLE_LOGIND, test "x$enable_logind" != "xno")
 
-AC_SUBST(LIBCRYPT)
-AC_CHECK_LIB(crypt, crypt, [LIBCRYPT=-lcrypt],
+# FIXME check whether the next variable really used somewhere 
+AC_CHECK_LIB(crypt, crypt,[LIBCRYPT=-lcrypt
```
alejandro-colomar: Why remove the space?

configure.ac
```diff
@@ -587,16 +587,15 @@ if test "$with_libpam" = "yes"; then
...
 
 	AC_MSG_CHECKING(use login and su access checking if PAM not used)
 	AC_MSG_RESULT(no)
 else
 	AC_DEFINE(SU_ACCESS, 1, [Define to support /etc/suauth su access control.])
-	AM_CONDITIONAL(USE_PAM, [false])
 	AC_MSG_CHECKING(use login and su access checking if PAM not used)
 	AC_MSG_RESULT(yes)
 fi
+AM_CONDITIONAL([USE_PAM], [test "X$with_libpam" = "Xyes"])
```
alejandro-colomar: Please document in the commit message why it's incorrect.  I have little knowledge of autotools, so it's not directly obvious to me.

configure.ac
```diff
@@ -1,8 +1,8 @@
 dnl Process this file with autoconf to produce a configure script.
 AC_PREREQ([2.69])
-m4_define([libsubid_abi_major], 5)
-m4_define([libsubid_abi_minor], 0)
-m4_define([libsubid_abi_micro], 0)
+m4_define([libsubid_abi_major], [5])
+m4_define([libsubid_abi_minor], [0])
+m4_define([libsubid_abi_micro], [0])
```
alejandro-colomar, line 5: What's the meaning of `[]` in M4?

configure.ac
```diff
@@ -255,13 +255,13 @@ PKG_CHECK_MODULES([CMOCKA], [cmocka], [have_cmocka="yes"],
...
 
+AC_ARG_ENABLE([vendordir],
+    [AS_HELP_STRING([--enable-vendordir=DIR], [Directory for distribution provided configuration files])],[],[])
 AC_CHECK_LIB([econf],[econf_readDirs],[LIBECONF="-leconf"],[LIBECONF=""])
 if test -n "$LIBECONF"; then
         AC_DEFINE_UNQUOTED([VENDORDIR], ["$enable_vendordir"],
 					[Directory for distribution provided configuration files])
 	ECONF_CPPFLAGS="-DUSE_ECONF=1"
-	AC_ARG_ENABLE([vendordir],
-		[AS_HELP_STRING([--enable-vendordir=DIR], [Directory for distribution provided configuration files])],[],[])
```
alejandro-colomar: Why do we want this?  I don't understand the commit message.

configure.ac
```diff
@@ -298,8 +298,8 @@ if test "$enable_subids" != "no"; then
 	dnl
 	dnl FIXME: check if 32 bit UIDs/GIDs are supported by libc
 	dnl
-	AC_CHECK_SIZEOF([uid_t],[], [#include "sys/types.h"])
-	AC_CHECK_SIZEOF([gid_t],[], [#include "sys/types.h"])
+	AC_CHECK_SIZEOF([uid_t],[], [[#include "sys/types.h"]])
+	AC_CHECK_SIZEOF([gid_t],[], [[#include "sys/types.h"]])
```
alejandro-colomar: Does this mean that the existing code was bogus?  Could you please clarify how this changes?

(Again, I'm an autotools inept here; need a lot of explanation.)

Makefile.am
```diff
@@ -1,4 +1,5 @@
 ## Process this file with automake to produce Makefile.in
+ACLOCAL_AMFLAGS = -I m4
```
alejandro-colomar, line 2: What was the warning?  What does this do?

configure.ac
```diff
@@ -6,8 +6,9 @@ m4_define([libsubid_abi_micro], [0])
 m4_define([libsubid_abi], [libsubid_abi_major.libsubid_abi_minor.libsubid_abi_micro])
 AC_INIT([shadow], [4.18.0], [pkg-shadow-devel@lists.alioth.debian.org], [],
 	[https://github.com/shadow-maint/shadow])
-AM_INIT_AUTOMAKE([1.11 foreign dist-xz subdir-objects tar-pax])
+AC_CONFIG_AUX_DIR([build-aux])
 AC_CONFIG_MACRO_DIRS([m4])
+AM_INIT_AUTOMAKE([1.11 foreign dist-xz subdir-objects tar-pax])
```
alejandro-colomar: Why do we need to move the AM_INIT_AUTOMAKE line?

configure.ac
```diff
@@ -18,8 +18,8 @@ AC_SUBST([LIBSUBID_ABI_MICRO], [libsubid_abi_micro])
 AC_SUBST([LIBSUBID_ABI], [libsubid_abi])
 
 dnl Some hacks...
-test "$prefix" = "NONE" && prefix="/usr"
-test "$prefix" = "/usr" && exec_prefix=""
+test "x$prefix" = "xNONE" && prefix="/usr"
+test "X$prefix" = "X/usr" && exec_prefix=""
```
alejandro-colomar, line 24: It would be interesting to note which shells are broken.

configure.ac
```diff
@@ -563,7 +563,7 @@ if test "$with_libpam" != "no"; then
 	fi
 
 
-	if test "X$pam_lib$pam_headers_found" = "Xyesyes" -a "$pam_conv_function" != "no" ; then
+	if test "X$pam_lib$pam_headers_found" = "Xyesyes" && test "$pam_conv_function" != "no" ; then
```
alejandro-colomar: Thanks!

For this patch (1db760b4057 ("configure: fixed outdated and non-portable 'test' syntax ")):

```
```

configure.ac
```diff
@@ -689,30 +689,28 @@ AC_CONFIG_FILES([
...
-echo "	nscd support:			$with_nscd"
-echo "	sssd support:			$with_sssd"
-echo "	subordinate IDs support:	$enable_subids"
-echo "	enable lastlog:			$enable_lastlog"
-echo "	enable logind:			$enable_logind"
-echo "	use file caps:			$with_fcaps"
-echo "	install su:			$with_su"
-echo "	enabled vendor dir:             $enable_vendordir"
-echo
+AC_MSG_NOTICE([shadow ${PACKAGE_VERSION} has been configured with the following features:
```
alejandro-colomar, line 690: Thanks!  For efe5dbf67a ("configure, final recap: add package version, print to the console and to the log"):

```
```

(But please remove `final recap` from commit message.)

configure.ac
Karlson2k: The space remove is accidental and does not change anything (as parameters are trimmed by M4 anyway).
The key change here is merging of two identical (and duplicated) checks into a single check.

However, looks like both `LIBCRYPT` and `LIYESCRYPT` are not really used anywhere.
Moreover, `LIYESCRYPT` looks totally wrong (`LI` `YES` `CRYPT` ?), but I haven't deeply check the code and in this PR I'm fixing only the obvious problems. Maybe it worth to check git blame.

configure.ac
Karlson2k: See https://www.gnu.org/software/automake/manual/html_node/Usage-of-Conditionals.html#index-AM_005fCONDITIONAL-2
In short: `AM_CONDITIONAL()` must not be used in shell's `if` branches. Instead it must be specified one time only (per conditional variable) with `test something` as a second parameter.

configure.ac
Karlson2k: `[]` is autotools version of M4 quoting.
It controls the order of defines expansion and grouping of parameters.
If you do some crazy thing, like add `m4_define([5],[6, 7])`, the it would make a difference whether parameter is quoted or not. Quoted parameter expanded after expanding the "function" (and `libsubid_abi_major` becomes `6, 7` in later substitutions), while unquoted expansion will add a third parameter (ignored) (and `libsubid_abi_major` becomes `6` in later substitutions).
In this particular case it is not strictly required (unless someone going to do a crazy thing), but it is safer to just **always** quote all the parameters (unless otherwise is required for something tricky). 
It also (arguably) improves readability of the code.

configure.ac
alejandro-colomar: Please check.  It would be worth getting rid of it separately, so this patch gets smaller.  And please git-blame(1) the origin of it, adding a Fixes: tag.

configure.ac
Karlson2k: `AC_ARG_ENABLE()` expands to nothing where it is used, but adds arguments parsing, help message and other related things.
It does not make any sense to put this macro into `if` branch. It may also confuse the reader about how it is used.
In short: `AC_ARG_ENABLE()` must not be used conditionally (accepted by autoconf, but it is incorrect).

configure.ac
alejandro-colomar: Thanks!

configure.ac
Karlson2k: `[[]]` means "use literally, without expansion and substitution".
`#` symbol potentially could be interpreted as a comment. Square brackets (without quoting or in `[]` quoting) replaced with nothing, while in C code they are used as arrays.
In short: always use `[[]]` for the C code unless autoconf macros needs to be expanded in the C code (like `AC_DEFAULT_INCLUDES`).

configure.ac
alejandro-colomar: That link in the commit message would help.  Thanks!

configure.ac
alejandro-colomar: Do you have a guess of why it was there?  It would be interesting to know what the author of that code was trying to do.

Makefile.am
Karlson2k: Run `autoreconf -vi` in the sources root to build autoconf files. You will see libtool warning.
See also: https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Input.html#index-AC_005fCONFIG_005fMACRO_005fDIR-1

Without this flag automake before 1.13 will not work properly. Not really relevant for modern automake, but still produces the warning.

configure.ac
Karlson2k: `AC_CONFIG_AUX_DIR()` must be specified before `AM_INIT_AUTOMAKE()` (otherwise `autoreconf` aborts with error).
For readability it's better to group autoconf settings together, before automake settings.

Makefile.am
alejandro-colomar: Hmmm, yep:

```
libtoolize: Consider adding '-I m4' to ACLOCAL_AMFLAGS in Makefile.am.
```

configure.ac
alejandro-colomar: > `AC_CONFIG_AUX_DIR()` must be specified before `AM_INIT_AUTOMAKE()` (otherwise `autoreconf` aborts with error).

But that doesn't justify moving it after `AD_CONFIG_MACRO_DIRS()`

> For readability it's better to group autoconf settings together, before automake settings.

This does, though.  Thanks!

configure.ac
Karlson2k: Actually, it is still valid for modern shells too, but for another reason.
When variable is (or contains) symbols like `!`, `-z`, `-n` etc. then interpretation could be wrong.
For example, `test "$var" = "yes"` could be expanded to `test -n = yes`. 
By prefixing with "X" (other letters works in the same way) we can make sure that variable value will not be used as an option.
See https://www.gnu.org/software/autoconf/manual/autoconf-2.72/html_node/Limitations-of-Builtins.html

configure.ac
Karlson2k: I think that idea was to use argument `--enable-vendordir` only when `econf` is used. 
But it does not work this way: the set of configure parameters is fixed when `configure` is generated, while the presence of `econf` is checked only when configure is run.

configure.ac
alejandro-colomar: Thanks!

configure.ac
alejandro-colomar: Makes sense.  Thanks!

alejandro-colomar:
In general, please expand the information in the commit messages.  I'd like to be able to read those a year from now and remember why we did what we did.  Consider that the readers of those commit messages (especially myself) may have almost no clue of autotools.

configure.ac
lslebodn: @Karlson2k sure, when an argument is unquoted, M4 will perform macro expansion on that argument before it's passed to the macro (m4_define in this case).

For simple numeric literals like 5 and 0, the practical effect of quoting versus not quoting is often negligible.

Do you really expect  M4 willfind a macro named `5` or `0`.
That would cause an unintended expansion in many other places as well.

Karlson2k: @lslebodn Of course, nobody expects that `5` or `0` will be redefined. This would be crazy, as I mentioned early.
However, it is better to always quote all arguments, unless you really want expansion before use (which is really exceptional for the code in configure.ac).
This unifies calling style and simplifies code writing.

This is also recommended (but not enforced) by autoconf manual:
* https://www.gnu.org/software/autoconf/manual/autoconf-2.72/html_node/Autoconf-Language.html: `Arguments should be enclosed within the quote characters ‘[’ and ‘]’, and be separated by commas.`, `Cautious Autoconf users would keep the quotes`
* https://www.gnu.org/software/autoconf/manual/autoconf-2.72/html_node/One-Macro-Call.html: `while it is acceptable for a configure.ac to avoid unnecessary quotes, it is bad practice for Autoconf macros which must both be more robust and also advocate perfect style.`

Also recommended by M4 manual, which is referenced in autoconf manual:
* https://www.gnu.org/software/m4/manual/html_node/Quoting-Arguments.html#On-Quoting-Arguments-to-macros: `It is common practice to quote all arguments to macros, unless you are sure you want the arguments expanded. `

In short: it is better to quote everything unless there is good reason to skip quoting in some specific case.

lslebodn: It make a sense to quote strings(function names ...) and than you very much for fixing that,
but it does not make much sense to quote integers.

It is simpler to read numbers without quotes. Code is read more than written.

Karlson2k: Readability is better when all arguments are quoted.

When most arguments are quoted, while a few are not quoted it may confuse readers. "Why something is not quoted? Is it something special? What the purpose of **not quoting**?"

It is easier and **better** to quote everything.

And this is what is recommended by autoconf and M4 authors. It worth to follow their recommendations unless there is a strong argument why their recommendations must not be followed.

Karlson2k: With unified quoting style it is easier to maintain the code.
Novice contributors will not have to think "should I quote this particular argument?", "what should be checked to decide, whether it should be quoted?".
Just quote all arguments and it will work fine.

lslebodn: I do not think it is simper to read number with quotes.

Consistency is nice. 
Lets hope your assumption about new contributors will be correct. (I am bit skeptical :crossed_fingers: )

TYVM for changes

Karlson2k: Thank you for your attention and your approval.

alejandro-colomar: As someone inept at autotools, I agree that consistency will help me get it right all the time.  I'll just paste stuff from elsewhere, change it a little bit, and hope it works.  If it is quoted everywhere, I'll have more chances of getting it right.  :)

Karlson2k: > As someone inept at autotools, I agree that consistency will help me get it right all the time. I'll just paste stuff from elsewhere, change it a little bit, and hope it works. If it is quoted everywhere, I'll have more chances of getting it right. :)

Exactly. 
For everyone without deep knowledge of autotools, it is better to see some consistency to start with.
Moreover, quoting is (*almost*) never wrong (for configure.ac), while unquoted parameters could be a problem sometimes.
Note: there some special situations where arguments must NOT be quoted, but you hardly find them in configure. They are more related to autoconf macros. 
All standard autoconf, automake and libtool macros prefer quoted parameters and some may break with unquoted.

Karlson2k:
OK, I need to finish some clean-ups and I'll update this PR.

lslebodn:
> OK, I need to finish some clean-ups and I'll update this PR.

which cleanup do you mean?
Is is something about inconsistency with using `AS_IF` vs manual `if/then/fi` blocks?
* https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Common-Shell-Constructs.html#index-AS_005fIF-1

Maybe it would be better to create follow up changes in different PR

Karlson2k:
> > OK, I need to finish some clean-ups and I'll update this PR.
> 
> which cleanup do you mean? Is is something about inconsistency with using `AS_IF` vs manual `if/then/fi` blocks?

No, this is [requested here](https://github.com/shadow-maint/shadow/pull/1287#discussion_r2203500827) cleanup of unused `LIYESCRYPT`.
Also I need to integrate `Reviewed-by: Alejandro Colomar <alx@kernel.org>` marks

> * https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Common-Shell-Constructs.html#index-AS_005fIF-1
> 
> Maybe it would be better to create follow up changes in different PR

This would be nice.
Also `AS_CASE` could be used instead of `case esac`. And `AS_VAR_IF` may simplify the code and the readability.
But all these things should go to another PR.

configure.ac
lslebodn: That was introduced in https://github.com/shadow-maint/shadow/commit/5cd04d03f94622c12220d4a6352824af081b8531#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R413

I have a feeling the intention was to check whether yescrypt is supported by libcrypt.so. But Following snippet does not do that.
```
AC_SUBST(LIYESCRYPT)
AC_CHECK_LIB(crypt, crypt, [LIYESCRYPT=-lcrypt],
```

question is whether we want to fix "configure time" detection of yescrypt support or to get rid of this "no-op " part in the configure.ac

Karlson2k: Yep.
Erroneous `LIYESCRYPT` (should be `LIBYESCRYPT`?) is not used anywhere.

lslebodn: yescrypt seems to be introduced in libxcrypt 4.2.0
https://github.com/besser82/libxcrypt/blob/e34e0634a7192383dbb4445b6d7568e36baef675/NEWS#L412-L420

It is still the same dynamic library (libcrypt.so) which was checked few lines before `AC_SUBST(LIYESCRYPT)`
One should rather use `AC_RUN_IFELSE` than `AC_CHECK_LIB` or to check something in header files.

But libxcrypt 4.2.0 was released long time ago
* https://github.com/besser82/libxcrypt/releases/tag/v4.2.0.

Just not sure which other implementation of libcryt.so are supported by the project.
Maybe one could completely get rid of conditional support for yescrypt.
But that would be question for maintainers :-)

Karlson2k: Anyway, this particular macro is full duplication of another check + addition of unused substitution. 
Should be removed.

Karlson2k:
@alejandro-colomar Adjusted as requested.

configure.ac
```diff
@@ -345,13 +345,9 @@ if test "$enable_logind" = "yes"; then
...
 AM_CONDITIONAL(ENABLE_LOGIND, test "x$enable_logind" != "xno")
 
-AC_SUBST(LIBCRYPT)
 AC_CHECK_LIB(crypt, crypt, [LIBCRYPT=-lcrypt],
 	[AC_MSG_ERROR([crypt() not found])])
-
-AC_SUBST(LIYESCRYPT)
-AC_CHECK_LIB(crypt, crypt, [LIYESCRYPT=-lcrypt],
-	[AC_MSG_ERROR([crypt() not found])])
+AC_SUBST(LIBCRYPT)
```
alejandro-colomar: For commit b25059ad ("configure: Remove duplicated check and unused Makefile substitution "):

```
```

alejandro-colomar:
Please use two spaces after period in commit messages.  One space is wrong.

See <https://web.archive.org/web/20171217060354/http://www.heracliteanriver.com/?p=324>.

configure.ac
```diff
@@ -584,16 +584,15 @@ if test "$with_libpam" = "yes"; then
...
 
 	AC_MSG_CHECKING(use login and su access checking if PAM not used)
 	AC_MSG_RESULT(no)
 else
 	AC_DEFINE(SU_ACCESS, 1, [Define to support /etc/suauth su access control.])
-	AM_CONDITIONAL(USE_PAM, [false])
 	AC_MSG_CHECKING(use login and su access checking if PAM not used)
 	AC_MSG_RESULT(yes)
 fi
+AM_CONDITIONAL([USE_PAM], [test "X$with_libpam" = "Xyes"])
```
alejandro-colomar: For commit 4db8feeb ("configure: Fix incorrect use of AM_CONDITIONAL "):

```
```

configure.ac
```diff
@@ -55,152 +55,153 @@ AC_CHECK_MEMBERS([struct utmpx.ut_name,
...
 
-AC_CHECK_FUNC(setpgrp)
-AC_CHECK_FUNC(secure_getenv, [AC_DEFINE(HAS_SECURE_GETENV,
-                                        1,
-                                        [Defined to 1 if you have the declaration of 'secure_getenv'])])
+AC_CHECK_FUNC([setpgrp])
+AC_CHECK_FUNC([secure_getenv],
+    [AC_DEFINE([HAS_SECURE_GETENV],
+               [1],
+               [Defined to 1 if you have the declaration of 'secure_getenv'])])
```
alejandro-colomar: Please indent with tabs (as done later in the same file; see below).

Makefile.am
alejandro-colomar, line 2: For commit 8a10fbc5fd ("Makefile.am: Fix libtool warning"):

```
```

configure.ac
```diff
@@ -6,8 +6,9 @@ m4_define([libsubid_abi_micro], [0])
 m4_define([libsubid_abi], [libsubid_abi_major.libsubid_abi_minor.libsubid_abi_micro])
 AC_INIT([shadow], [4.18.0], [pkg-shadow-devel@lists.alioth.debian.org], [],
 	[https://github.com/shadow-maint/shadow])
-AM_INIT_AUTOMAKE([1.11 foreign dist-xz subdir-objects tar-pax])
+AC_CONFIG_AUX_DIR([build-aux])
 AC_CONFIG_MACRO_DIRS([m4])
+AM_INIT_AUTOMAKE([1.11 foreign dist-xz subdir-objects tar-pax])
 AM_SILENT_RULES([yes])
 AC_CONFIG_HEADERS([config.h])
```
alejandro-colomar: AC_CONFIG_HEADERS is below some AM configs.  Should that also be moved for consistency?  Also, if we do that, I think the agroupation of all AC before all AM should be a patch of its own.

configure.ac
alejandro-colomar, line 24: For commit 129b4b7 ("configure: Unify checks for variable values "):

```
```

configure.ac
Karlson2k: Yes, makes sense

Karlson2k: Note: you can move `config.h` to some subdir, like `lib/`.
Then automake will automatically add **build** subdir `lib/` to include paths.

alejandro-colomar: Hmmm, if that's common practice, go ahead.

Karlson2k: I'd prefer to stop adding commits to this PR as autotools can be improved much deeper and this PR will never ends then.
That could be handled in additional PRs.

Karlson2k: > Hmmm, if that's common practice, go ahead.

It is not listed as "recommended" anywhere. Most of the project just use the simplest form, like this project as many just unaware about such options.
Moreover, I'd recommend changing the name, like `shadow_config.h` for clarity.

Karlson2k: Note: the initial part of the list of the next improvements has been [already discussed](https://github.com/shadow-maint/shadow/pull/1287#issuecomment-3078035642).

alejandro-colomar: > > Hmmm, if that's common practice, go ahead.
> 
> It is not listed as "recommended" anywhere. Most of the project just use the simplest form, like this project as many just unaware about such options.

Cc: @thesamesam , any second opinion?

> Moreover, I'd recommend changing the name, like `shadow_config.h` for clarity.

That name would be unconventional, no?  I think it is conventional to call the config just `config.h`.  Cc: @thesamesam

Karlson2k: I'd move this discussion the the new PR or the new issue, as this PR does not change anything related to this header.

alejandro-colomar (approved):
LGTM.  Thanks!

I'll leave it open until tomorrow, in case anyone else has any comments, as I'm not an expert in the build system.  However, all of this looks very reasonable.

For the entire patch set:
```
```

Cc: @hallyn , @ikerexxe , @thesamesam  (in case you have any last words).

PR #1292, opened 2025-07-14
Correct finding UTMP entry

Karlson2k:
This PR fixes wrong utmp entry search, by better following algorithm expected by Linux kernel as described here: https://man7.org/linux/man-pages/man5/utmp.5.html#DESCRIPTION

### DIFF
--- a/lib/utmp.c
+++ b/lib/utmp.c
@@ -3,6 +3,7 @@
  * SPDX-FileCopyrightText: 1996 - 1999, Marek Michałkiewicz
  * SPDX-FileCopyrightText: 2001 - 2005, Tomasz Kłoczko
  * SPDX-FileCopyrightText: 2008 - 2009, Nicolas François
+ * SPDX-FileCopyrightText: 2025, Evgeny Grin (Karlson2k)
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
@@ -138,20 +139,19 @@ err_close:
  * get_current_utmp - return the most probable utmp entry for the current
  *                    session
  *
- *	The utmp file is scanned for an entry with the same process ID.
- *	The line entered by the *getty / telnetd, etc. should also match
- *	the current terminal.
+ *	When an entry is returned by this function, and if the utmpx structure
+ *	has a ut_id field and this field is not empty, then this field
+ *	should be used to update the entry information.
  *
- *	When an entry is returned by get_current_utmp, and if the utmpx
- *	structure has a ut_id field, this field should be used to update
- *	the entry information.
- *
- *	Return NULL if no entries exist in utmp for the current process.
+ *	Return NULL if no entries exist in utmp for the current process or
+ *	            there is an error reading utmp.
  */
 static /*@null@*/ /*@only@*/struct utmpx *
 get_current_utmp(pid_t main_pid)
 {
 	struct utmpx  *ut;
+	struct utmpx  *ut_by_pid  = NULL;
+	struct utmpx  *ut_by_line = NULL;
 
 	setutxent();
 
@@ -161,15 +161,22 @@ get_current_utmp(pid_t main_pid)
 		    && (USER_PROCESS  != ut->ut_type))
 			continue;
 
-		if (   (main_pid == ut->ut_pid)
-		    && ('\0' != ut->ut_id[0])
-		    /* A process may have failed to close an entry
-		     * Check if this entry refers to the current tty */
-		    && is_my_tty(ut->ut_line))
-		{
-			break;
+		if (main_pid == ut->ut_pid) {
+			if (is_my_tty(ut->ut_line))
+				break; /* Perfect match, stop the search */
+
+			if (NULL == ut_by_pid)
+				ut_by_pid = ut;
+
+		} else if (   (NULL == ut_by_line)
+			   && (LOGIN_PROCESS == ut->ut_type) /* Be more picky when matching by 'ut_line' only */
+			   && (is_my_tty(ut->ut_line))) {
+			ut_by_line = ut;
 		}
 	}
+
+	if (NULL == ut)
+		ut = ut_by_pid ?: ut_by_line;
 
 	if (NULL != ut) {
 		struct utmpx  *ut_copy;
@@ -236,7 +243,8 @@ updwtmpx(const char *filename, const struct utmpx *ut)
  *	It accepts an utmp entry in input (ut) to return an entry with
  *	the right ut_id. This is typically an entry returned by
  *	get_current_utmp
- *	If ut is NULL, ut_id will be forged based on the line argument.
+ *	If ut is NULL or ut->ut_id is empty, ut_id will be forged based on
+ *	the line argument.
  *
  *	The ut_host field of the input structure may also be kept, and is
  *	used to define the ut_addr/ut_addr_v6 fields. (if these fields
@@ -275,7 +283,8 @@ prepare_utmp(const char *name, const char *line, const char *host,
 	utent->ut_type = USER_PROCESS;
 	utent->ut_pid = main_pid;
 	STRNCPY(utent->ut_line, line);
-	if (NULL != ut) {
+	if (   (NULL != ut)
+	    && ('\0' != ut->ut_id[0])) {
 		STRNCPY(utent->ut_id, ut->ut_id);
 	} else {
 		STRNCPY(utent->ut_id, strnul(line) - MIN(strlen(line), countof(utent->ut_id)));


lib/logind.c
```diff
@@ -14,13 +14,13 @@
...
 
-int get_session_host (char **out)
+int get_session_host (char **out, bool is_forked)
 {
     char *host = NULL;
     char *session = NULL;
     int ret;
 
-    ret = sd_pid_get_session (getpid(), &session);
+    ret = sd_pid_get_session (is_forked ? getpid() : getpid(), &session);
```
alejandro-colomar: This is probably a typo.

Karlson2k: Right.

lib/logind.c
```diff
@@ -14,13 +14,13 @@
...
 {
     char *host = NULL;
     char *session = NULL;
     int ret;
 
-    ret = sd_pid_get_session (getpid(), &session);
+    ret = sd_pid_get_session (is_forked ? getppid() : getpid(), &session);
     if (ret < 0) {
         return ret;
     }
```
alejandro-colomar: I think it would be simpler if you pass the `pid` to this function directly, instead of passing the boolean.

```c
int
get_session_host(char **out, pid_t pid)
{
    char *host = NULL;
    char *session = NULL;
    int ret;

    ret = sd_pid_get_session (pid, &session);
    if (ret < 0) {
        return ret;
}
```

```c
get_session_host(&host, is_forked ? getppid() : getpid());
```

Karlson2k: Right.
It is better to get PID just one time and save a few syscalls.

lib/utmp.c
```diff
@@ -151,32 +151,35 @@ static /*@null@*/ /*@only@*/struct utmpx *
...
+		if (   (LOGIN_PROCESS != ut->ut_type)
+		    && (USER_PROCESS  != ut->ut_type))
+			continue; /* Match only 'login' and 'user' entries */
+
+		if (main_pid == ut->ut_pid)
+			break; /* Found the entry */
+
+		if (is_my_tty(ut->ut_line))
+			ut_by_line = ut; /* Matched only by 'ut_line' */
 	}
```
alejandro-colomar: Could you please refactor this loop without semantic changes from a while to a for loop in a separate commit, and then apply the semantic changes in a smaller commit that is easier to review?

lib/utmp.c
```diff
@@ -151,32 +151,35 @@ static /*@null@*/ /*@only@*/struct utmpx *
...
 
+	if (NULL == ut)
+		ut = ut_by_line; /* Fallback, *getty may not set correct PID */
+
 	if (NULL != ut) {
-		ret = XMALLOC(1, struct utmpx);
-		memcpy (ret, ut, sizeof (*ret));
+		struct utmpx *copy;
+		copy = (struct utmpx *) XMALLOC(1, struct utmpx);
```
alejandro-colomar: Never write an explicit cast of the return of malloc(3).  That's dangerous.  We have the MALLOC() family of macros precisely to do this for us:

```c
$ grepc XMALLOC .
./lib/alloc/x/xmalloc.h:#define XMALLOC(n, type)                                                      \
	(type *) xmallocarray(n, sizeof(type))                                \
)
```

Karlson2k: :ok_hand:

Karlson2k:
PR has been updated to perform a minimal changes. Extra refactoring was excluded.

lib/utmp.c
```diff
@@ -368,12 +367,13 @@ setutmp(struct utmpx *ut)
...
-update_utmp(const char *user, const char *tty, const char *host)
+update_utmp(const char *user, const char *tty, const char *host,
+            pid_t main_pid)
 {
 	struct utmpx  *utent, *ut;
 
-	utent = get_current_utmp ();
-	ut = prepare_utmp  (user, tty, host, utent);
+	utent = get_current_utmp (main_pid);
+	ut = prepare_utmp  (user, tty, host, utent, main_pid);
```
alejandro-colomar: Please remove the space before the `(`.  I take the opportunity of every change in existing function calls to remove those (since that doesn't complicate the diff in any way).

Karlson2k: OK

lib/utmp.c
```diff
@@ -138,37 +139,47 @@ failtmp(const char *username, const struct utmpx *failent)
...
+		if (   (LOGIN_PROCESS == ut->ut_type)
+		    || (USER_PROCESS  == ut->ut_type)) {
+			if (main_pid == ut->ut_pid) {
+				if (is_my_tty(ut->ut_line))
+					break; /* Found the entry */
+				ut_by_pid = ut;  /* Matched only by 'ut_pid' */
+			} else if (is_my_tty(ut->ut_line))
+				ut_by_line = ut; /* Matched only by 'ut_line' */
 		}
 	}
```
alejandro-colomar: I liked the `for` version with `continue` and `break` more.  This was unreadable, and after this change it's even more unreadable.  :)

Karlson2k: OK, I'll update it and try to guess better what you are looking for.

lib/utmp.c
```diff
@@ -156,24 +156,27 @@ failtmp(const char *username, const struct utmpx *failent)
...
+			if (is_my_tty(ut->ut_line))
+				break; /* Found the entry */
+
+			if (NULL == ut_by_pid)
+				ut_by_pid = ut;  /* Matched only by 'ut_pid' */
+
+		} else if (   (NULL == ut_by_line)
+				   && is_my_tty(ut->ut_line))
+			ut_by_line = ut; /* Matched only by 'ut_line' */
 	}
```
alejandro-colomar: I would prefer the refactor to go before the actual change.  That will result in both a smaller refactor and a smaller change.

Karlson2k:
@alejandro-colomar Rewritten again.

lib/utmp.c
```diff
@@ -173,13 +172,15 @@ get_current_utmp(void)
...
 
 	if (NULL != ut) {
-		ret = XMALLOC(1, struct utmpx);
-		memcpy (ret, ut, sizeof (*ret));
+		struct utmpx *ut_copy;
+		ut_copy = XMALLOC(1, struct utmpx);
+		memcpy(ut_copy, ut, sizeof(*ut));
+		ut = ut_copy;
 	}
```
alejandro-colomar: Hmmm, this sounds like we want memdup().  That would be rewritten simpler if we had it:

```c
ut = XMEMDUP(ut, 1, struct utmpx);
```

where MEMDUP() would be

```c
#define MEMDUP(p, n, T)                                                       \
	_Generic(p, T *: (T *) memdup(p, (n * sizeof(T)) ?: 1))               \
)
```

and memdup() would be:

```c
void *
memdup(const void *p, size_t size)
{
	void  *new;

	new = malloc(size);
	if (new == NULL)
		return NULL;

	return memcpy(new, p, size);
}
```

Not saying you should do it.  It's more a reminder for myself, to check in the future.

alejandro-colomar: Plus, if this sometimes return NULL, I don't see why it uses XMALLOC().  It could pass a malloc(3) failure to the caller as NULL.  (Again, just reminder to self.)

alejandro-colomar: Thanks!  I like this commit.

src/login.c
```diff
@@ -944,7 +946,7 @@ int main (int argc, char **argv)
 			failure (pwd->pw_uid, tty, &faillog);
 		}
 #ifndef ENABLE_LOGIND
-		record_failure(failent_user, tty, hostname);
+		record_failure(failent_user, tty, hostname, initial_pid);
```
alejandro-colomar: I prefer calling getpid() twice, if we can.

alejandro-colomar: Or did the fork(2) call already happen here?

lib/utmp.c
```diff
@@ -160,16 +168,24 @@ get_current_utmp(pid_t main_pid)
...
+		if (main_pid == ut->ut_pid) {
+			if (is_my_tty(ut->ut_line))
+				break; /* Found the entry */
+
+			if (NULL == ut_by_pid)
+				ut_by_pid = ut;  /* Matched only by 'ut_pid' */
+
+		} else if (   (NULL == ut_by_line)
+				   && is_my_tty(ut->ut_line))
+			ut_by_line = ut; /* Matched only by 'ut_line' */
```
alejandro-colomar: Please use braces for both branches.  The GNU style is hard to read.

lib/utmp.c
```diff
@@ -160,16 +168,24 @@ get_current_utmp(pid_t main_pid)
...
-		}
+		if (main_pid == ut->ut_pid) {
+			if (is_my_tty(ut->ut_line))
+				break; /* Found the entry */
+
+			if (NULL == ut_by_pid)
+				ut_by_pid = ut;  /* Matched only by 'ut_pid' */
+
+		} else if (   (NULL == ut_by_line)
+				   && is_my_tty(ut->ut_line))
```
alejandro-colomar: There's some alignment issue here.

Karlson2k: I can't see it, but I'll try to use another form.

src/login.c
Karlson2k: Sorry, I see zero reasons for it.
Record MUST use the initial PID, which is already in variable `initial_pid`. The same variable is used above and below. It should be used here as well at least for three reasons:
* PID is already known. Extra syscall will make binary bigger and slower without any advantage.
* Maintainability of the code. Currently it is not yet forked, but it can be changed later and then it will be broken. While with variable it will remain working correctly.
* Consistency. It is used before, then it is used in the call below in the code.

Karlson2k:
Updated as requested (not counting extra getpid() call which is unsafe in terms of being future-proof).

src/login.c
alejandro-colomar: Okay, let's keep it with the variable, to prevent breaking it if the code is refactored.  Makes sense.

Karlson2k: OK.
The rest is corrected.

Karlson2k:
Updated one more time.
Refined behaviour when `login` started from another `login` session.
There is no description how to process `utmp` entries in such situation as this is not the way how `login` must be started.
However, it is better to add a new `utmp` entry and keep the old entry too instead of overwriting the old entry with new user and PID.
It gives a better tracking of logged users, as `user2` can be logged in the same console as `user1`.
But the main reason as a cleanup: entries are cleaned by PID. The entry from manually started `login` will not be cleaned anyway (as nothing is designed to perform such cleanup), while correctly started `login` entry will be cleared by matching PID.

Karlson2k:
@alejandro-colomar When this will be merged, I can finish #1284

alejandro-colomar:
> @alejandro-colomar When this will be merged, I can finish #1284

I would like to get <https://github.com/shadow-maint/shadow/pull/1296> merged before this one.

Cc: @ikerexxe

Karlson2k:
> > @alejandro-colomar When this will be merged, I can finish #1284
> 
> I would like to get #1296 merged before this one.

This PR and #1296 do not really depend on each other, when require re-base of each other (depending of which one will be merged first).
This is PR ready (and was added early) and #1296 is not yet ready.
Maybe merge this one while updating #1296?

alejandro-colomar:
> > > @alejandro-colomar When this will be merged, I can finish #1284
> > 
> > 
> > I would like to get #1296 merged before this one.
> 
> This PR and #1296 do not really depend on each other, when require re-base of each other (depending of which one will be merged first).

> This is PR ready (and was added early) and #1296 is not yet ready.

This one is ready for review, but #1296 is also ready for review.

> Maybe merge this one while updating #1296?


I'm not convinced by the initial refactors of this PR yet.  I need to check it.  If @ikerexxe / @hallyn merge #1296 first while I review this one, it would be great.

However, I won't arbitrarily delay merging this one.  When I finish reviewing this one, I'll merge it, regardless of #1296.

Karlson2k:
> I'm not convinced by the initial refactors of this PR yet. I need to check it.

It was already rewritten about 9 times to fit your requirements. 
Let me know if further improvements are needed.

> However, I won't arbitrarily delay merging this one. When I finish reviewing this one, I'll merge it, regardless of #1296.

That's nice. Thanks in advance.

lib/utmp.c
```diff
@@ -156,12 +156,14 @@ get_current_utmp(void)
...
+	for (ut = getutxent(); ut != NULL; ut = getutxent()) {
+		if (   (LOGIN_PROCESS != ut->ut_type)
+		    && (USER_PROCESS  != ut->ut_type))
+			continue; /* Look for 'login' or 'user' entries only */
+
 		if (   (   (ut->ut_pid == getpid ())
 		        || (ut->ut_pid == getppid ()))
 		    && ('\0' != ut->ut_id[0])
-		    && (   (LOGIN_PROCESS == ut->ut_type)
-		        || (USER_PROCESS  == ut->ut_type))
```
alejandro-colomar: I like the part of this commit that adds the continue, but prefer to keep the while loop.  (Sorry for going back and forth about this; I didn't have time to look at it carefully last time.)

lib/utmp.c
```diff
@@ -160,8 +160,7 @@ get_current_utmp(void)
 		    && (USER_PROCESS  != ut->ut_type))
 			continue; /* Look for 'login' or 'user' entries only */
 
-		if (   (   (ut->ut_pid == getpid ())
-		        || (ut->ut_pid == getppid ()))
+		if (   (main_pid == ut->ut_pid)
 		    && ('\0' != ut->ut_id[0])
 		    /* A process may have failed to close an entry
 		     * Check if this entry refers to the current tty */
```
alejandro-colomar: Would you mind separating the first 3 commits from the last two in two different PRs?  I'll merge the first three soon, and then we can review the other two separately.

alejandro-colomar: They are logically different changes.

lib/utmp.c
Karlson2k: OK, changed back.

lib/utmp.c
Karlson2k: Just one commit can be separated (utmp entry search behaviour fix), but is does not make sense without fixing which PID is checked.
You asked to separate readability improvements from behavioural changes, so commit could be used only on top of readability changes.
So I'll keep this PR complete (with reordered commits) with major commit last and will add separate PR with extracted commits.

alejandro-colomar: Yeah, you could keep this PR complete, and then add a new PR with the first 3 commits.  I'll apply that one first, and then this one can be rebased on top of it.

Karlson2k: Extracted commits are [here](https://github.com/shadow-maint/shadow/pull/1305)

Karlson2k:
@alejandro-colomar If you merge #1305 first, I'll rebase this one.

Karlson2k:
Updated PR to fix only utmp entry search (align with Linux kernel description).

lib/utmp.c
```diff
@@ -139,19 +140,26 @@ failtmp(const char *username, const struct utmpx *failent)
...
  *
  *	The utmp file is scanned for an entry with the same process ID.
- *	The line entered by the *getty / telnetd, etc. should also match
- *	the current terminal.
+ *	If several entries with matching PID found (due to previous problems
+ *	with utmp file cleanup), then the first entry with matching PID and
+ *	matching 'ut_line' (current terminal) is used or if no entry matches
+ *	'ut_line' then the first entry with matching PID is used.  If no matching
+ *	entry is found, then the entry matched current terminal is selected (for
+ *	the case if *getty has not set the correct PID).
```
alejandro-colomar: Please don't document the algorithm in a comment.  That should go in the commit message.  Source code comments are likely to be wrong/obsolete/detached from reality.

alejandro-colomar: Also, please add a link to the sources for this algorithm in the commit message.

lib/utmp.c
```diff
@@ -161,16 +169,26 @@ get_current_utmp(pid_t main_pid)
...
+
+		} else if (   (NULL == ut_by_line)
+			   && (LOGIN_PROCESS == ut->ut_type) /* Be more picky when matching by 'ut_line' only */
+			   && (is_my_tty(ut->ut_line))) {
+			ut_by_line = ut; /* Matched only by 'ut_line' */
 		}
 	}
 
+	if (NULL == ut)
+		ut = ut_by_pid; /* Match only by PID is enough, *getty may be bypassed */
```
alejandro-colomar: Please call it getty(8) (here and elsewhere).

lib/utmp.c
```diff
@@ -139,19 +140,26 @@ failtmp(const char *username, const struct utmpx *failent)
...
- *	Return NULL if no entries exist in utmp for the current process.
+ *	Return NULL if no entries exist in utmp for the current process or
+ *	            there is an error reading utmp..
  */
 static /*@null@*/ /*@only@*/struct utmpx *
 get_current_utmp(pid_t main_pid)
 {
 	struct utmpx  *ut;
+	struct utmpx  *ut_by_pid  = NULL; /* first entry matched by 'ut_pid' */
+	struct utmpx  *ut_by_line = NULL; /* first entry matched by 'ut_line' */
```
alejandro-colomar: Please remove these comments, as they are redundant.

lib/utmp.c
```diff
@@ -161,16 +169,26 @@ get_current_utmp(pid_t main_pid)
...
+			ut_by_line = ut; /* Matched only by 'ut_line' */
 		}
 	}
 
+	if (NULL == ut)
+		ut = ut_by_pid; /* Match only by PID is enough, *getty may be bypassed */
+
+	if (NULL == ut)
+		ut = ut_by_line; /* Fallback, *getty may not set correct PID */
+
```
alejandro-colomar: This can be rewritten as

```c
ut = ut ?: ut_by_pid ?: ut_by_line;
```

BTW, the comments are vague enough that they're not useful to me (I don't understand them).  And if you make them useful, they'd become more noisy.  Those things belong in the commit message, where you can detail them much better.

lib/utmp.c
```diff
@@ -275,7 +294,7 @@ prepare_utmp(const char *name, const char *line, const char *host,
 	utent->ut_type = USER_PROCESS;
 	utent->ut_pid = main_pid;
 	STRNCPY(utent->ut_line, line);
-	if (NULL != ut) {
+	if ((NULL != ut) && (0 != utent->ut_id[0])) {
```
alejandro-colomar: Please use `'\0'` for the null character.

alejandro-colomar: Also, please don't add redundant parentheses.

Karlson2k: Redundant parentheses are used everywhere in the code when in conditionals, including other part of this commit.
I'll regroup them here to be more consistent.

lib/utmp.c
Karlson2k: OK, I'll replace with Elvis operator but slightly different form, which should be better for readability.

Karlson2k:
@alejandro-colomar Updated as requested, except [this](https://github.com/shadow-maint/shadow/pull/1292#discussion_r2217318129) comment, for which the code is changed differently.

lib/utmp.c
alejandro-colomar: Yes, old code is weird, but we try to avoid writing weird new code.

lib/utmp.c
```diff
@@ -161,16 +161,23 @@ get_current_utmp(pid_t main_pid)
...
+
+		} else if (   (NULL == ut_by_line)
+			   && (LOGIN_PROCESS == ut->ut_type) /* Be more picky when matching by 'ut_line' only */
+			   && (is_my_tty(ut->ut_line))) {
+			ut_by_line = ut;
 		}
 	}
 
+	if (NULL == ut)
+		ut = ut_by_pid ?: ut_by_line;
```
alejandro-colomar, line 179: We can avoid the conditional with a one-liner:

```c
ut = ut ?: ut_by_pid ?: ut_by_line;
```

alejandro-colomar:
```
Updated utmp entry search algorithm to follow GNU/Linux description:
https://man7.org/linux/man-pages/man5/utmp.5.html#DESCRIPTION

An entry is found by looking for matching PID.  If several such entries
found (for example, due to cleanup failure of old entries) then first
entry with both matching PID and matching 'ut_line' (current terminal)
is used.  If not entry has matching 'ut_line' then first entry with
matching PID is used (if getty/init process does not set 'ut_line').

When no single entry is matched by PID, then but at least one entry is
matched current terminal the the first such entry is selected (if getty
does not set correct PID).
```

Could you quote specific parts of utmp(5) to defend the correctness of your claims?

Why is it valid to use an entry with matching PID but not matching ut_line?

Why is it valid to use an entry matching the current terminal but not the PID?

lib/utmp.c
alejandro-colomar: I think this is unambiguous: first, else second, else third.

lib/utmp.c
Karlson2k: "Less lines" != "better code"
"Shorter code" != "better readability".

This one-liner is much harder to read than my version.

lib/utmp.c
Karlson2k: I can only repeat myself.
Without any written guidelines every contributor will have to re-write each PR several times, without a chance to guess expected format at the first try.

lib/utmp.c
Karlson2k: You asked earlier to expand the loop to break/continue version, which has more lines, but better readability.
The same is here.
Unless we have next level of folding, like 
```C
ut ?:= ut_by_pid ?: ut_by_line;
```
my version has better readability.

Karlson2k:
Here is (a shortened and a bit simplified) description of the process from https://man7.org/linux/man-pages/man5/utmp.5.html#DESCRIPTION:

> The first entries ever created result from [init(1)](https://man7.org/linux/man-pages/man1/init.1.html) processing inittab(5).  If no empty record with the needed ut_id can be found, [init(1)](https://man7.org/linux/man-pages/man1/init.1.html) creates a new one.  It sets ut_id from the inittab, ut_pid and ut_time to the current values, and ut_type to INIT_PROCESS.
> 
> getty locates the entry by the PID, changes ut_type to LOGIN_PROCESS, changes ut_time, sets ut_line, and waits for connection to be established.  [login(1)](https://man7.org/linux/man-pages/man1/login.1.html), after a user has been authenticated, changes ut_type to USER_PROCESS, changes ut_time, and sets ut_host and ut_addr.  Depending on getty and [login(1)](https://man7.org/linux/man-pages/man1/login.1.html), records may be located by ut_line instead of the preferable ut_pid.

Old version used entry **only** with **both** matches `ut_pid` and `ut_line`. This was not expected, match by PID is enough.
However, for safety reasons, the new code tries to match both PID and line. (In case if cleanup was not performed correctly).
If perfect match (like the old one) was not found, then the first match by PID is used (this is what described in the documentation).
If no entry matched PID, then used first entry which matched by `ut_line`. (Assuming that no two sessions are possible simultaneously on the same console.) This is also described in the documentation: `records may be located by ut_line instead of the preferable ut_pid.`

However, if user manually started `login` binary after being already logged in the console, previous utmp entry matched by `ut_line` will not be used as it has `USER_PROCESS` type, which is ignored when matching only by `ut_line` (entry must be precisely `LOGIN_PROCESS`). This situation is not described in specification. `login` will create a new entry in such case (the same was before this patch).

lib/utmp.c
Karlson2k: > ```c
> ut ?:= ut_by_pid ?: ut_by_line;
> ```

You may discuss this operator on the C committee.

alejandro-colomar: > > ```c
> > ut ?:= ut_by_pid ?: ut_by_line;
> > ```
> 
> You may discuss this operator on the C committee.

The line with `?: ... ?:` is more uniform than that.  I'd prefer it, even if `?:=` were available.

alejandro-colomar: > You asked earlier to expand the loop to break/continue version, which has more lines, but better readability. The same is here.

Not exactly the same.  In the loop, the conditions were not homogeneous, and spread over several lines.  It was hard to follow the logic.  With consecutive `?` operators, the logic is unambiguous and homogeneous.

Karlson2k: Readability is must better when you see the condition first, like `if (NULL == ut)`.
Especially when combined with the next `if (NULL != ut)`.
The first line tells reader to skip completely assignment, when `ut` has been set.

When line started with `ut = ...` it looks like "ut overwrite". The reader must decipher double Elvis operator and find that value of `ut` may be assigned back.

Anyway, my code is fully correct and fits your code style rules. You may apply your corrections later on top of my patch according to your personal taste.

Karlson2k: @alejandro-colomar Updated with even shorter form and nested Elvis operators

lib/utmp.c
```diff
@@ -161,21 +161,25 @@ get_current_utmp(pid_t main_pid)
...
 		}
 	}
 
-	if (NULL != ut) {
+	if (ut || ut_by_pid || ut_by_line) {
 		struct utmpx  *ut_copy;
 
 		ut_copy = XMALLOC(1, struct utmpx);
-		memcpy(ut_copy, ut, sizeof(*ut));
+		memcpy(ut_copy, ut ?: ut_by_pid ?: ut_by_line, sizeof(*ut));
```
alejandro-colomar: I preferred your previous approach.  Now we have two points where we have split the logic about ut?:ut_by_pid?:ut_by_line.

lib/utmp.c
```diff
@@ -275,7 +280,8 @@ prepare_utmp(const char *name, const char *line, const char *host,
 	utent->ut_type = USER_PROCESS;
 	utent->ut_pid = main_pid;
 	STRNCPY(utent->ut_line, line);
-	if (NULL != ut) {
+	if (   (NULL != ut)
+	    && ('\0' != utent->ut_id[0])) {
```
alejandro-colomar: Why this added check?  Is is related to the other code changes, or is it an unrelated fix?

Karlson2k: This is directly related to other changes in the code.
Before this patch, function `get_current_utmp()` never returned entry with empty `ut_id`. But this was wrong, as other fields could be valid still, even if `ut_id` is empty. For example, it may contain information useful for `get_session_host()`. No reason to filter out such entries.
However, if `ut_id` is empty then `login` must set a valid `ut_id`. This is implemented with this additional check. The related comment is updated above.

lib/utmp.c
Karlson2k: No problem.
I've reverted the latest change.

Karlson2k:
Let me know if any further changes are needed.

alejandro-colomar:
> Let me know if any further changes are needed.

I'm okay with the source code.

However, I'd feel more comfortable if @hallyn and/or @ikerexxe could review the logic.

lib/utmp.c
ikerexxe: The utent variable is assigned and zeroed several lines above, and its value never changes thereafter, so `“\0” != utent->ut_id[0]` is always false. Therefore, the system will incorrectly create a new entry each time.

Karlson2k: Nice catch!
Wrong variable used.
Thank you.

Karlson2k:
> I've been trying to review this patch all week, but every time I started, something else caught my attention. Now I've had a good look at it, and I would say that the change in prepare_utmp() is incorrect. See the inline comment for more details.

Nice catch, thanks!

Corrected.

lib/utmp.c
Karlson2k: Updated with correct variable used in the check.

Karlson2k: Sad (or funny?) that I put correct variable to the doxy comment above, but used wrong variable in the real check.

ikerexxe (approved):
LGTM. Good job!

lib/utmp.c
alejandro-colomar: This is why I think comments are bad.

Karlson2k:
@ikerexxe Thank you!

PR #1293, opened 2025-07-15
lib/utmp.c: align generated "ut_id" with other software

Karlson2k:
This PR aligns generated "ut_id" with ids produced by other modern software, like systemd and libutemper.
For reference: 
* libutemper implementation: https://github.com/altlinux/libutempter/blob/4caa8ab94ff8b207228fa723a89214bf5e929321/libutempter/utempter.c#L99-L103
Note: libutemper is used by many graphical terminal emulator applications (konsole, xterm, others) to deal with utmp file.
* systemd implementation:
use full name of the device (see https://github.com/systemd/systemd/blob/8013beb4a2221680b2c741bac6b3a33fe66406e1/units/getty%40.service.in#L41-L44) or last four chars if name is longer then four chars (see https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#UtmpIdentifier=)

To summarise the idea of this PR.
This PR **does not** change functionality for normal workflow (when `utmp` entry is created before `login` is started).

This PR **does** correct behaviour in non-standard situations, for example, when `login` is started as `init` process, aligning choice of `ut_id` value with modern software (systemd, utemper).

### DIFF
--- a/lib/utmp.c
+++ b/lib/utmp.c
@@ -26,6 +26,7 @@
 #include "alloc/x/xcalloc.h"
 #include "alloc/x/xmalloc.h"
 #include "sizeof.h"
+#include "string/strchr/strnul.h"
 #include "string/strcmp/streq.h"
 #include "string/strcmp/strprefix.h"
 #include "string/strcpy/strncpy.h"
@@ -274,7 +275,7 @@ prepare_utmp(const char *name, const char *line, const char *host,
 	if (NULL != ut) {
 		STRNCPY(utent->ut_id, ut->ut_id);
 	} else {
-		STRNCPY(utent->ut_id, strprefix(line, "tty") ?: line);
+		STRNCPY(utent->ut_id, strnul(line) - MIN(strlen(line), countof(utent->ut_id)));
 	}
 #if defined(HAVE_STRUCT_UTMPX_UT_NAME)
 	STRNCPY(utent->ut_name, name);


lib/utmp.c
```diff
@@ -274,8 +274,14 @@ prepare_utmp(const char *name, const char *line, const char *host,
 	if (NULL != ut) {
 		STRNCPY(utent->ut_id, ut->ut_id);
 	} else {
-		/* XXX - assumes /dev/tty?? */
-		STRNCPY(utent->ut_id, line + 3);
+		size_t line_len, ut_line_maxlen;
+		ut_line_maxlen = countof(utent->ut_id) - 1;
```
alejandro-colomar: `utent->ut_id` is not a string.  We can use the full array for non-zero bytes.

lib/utmp.c
```diff
@@ -274,8 +274,14 @@ prepare_utmp(const char *name, const char *line, const char *host,
...
+		line_len = strlen (line);
+		if (ut_line_maxlen >= line_len)
+			STRNCPY(utent->ut_id, line);
+		else
+			STRNCPY(utent->ut_id, \
+			        line + (line_len - ut_line_maxlen));
```
alejandro-colomar: This seems bogus.  It's truncating from the start, instead of from the end.  Why would we do that?

alejandro-colomar: We should instead error out if the string doesn't fit the buffer (if it's appropriate to fail, which I don't know), or keep truncating as always.

lib/utmp.c
Karlson2k: It is declared as `char ut_id[]`, right next to `char ut_user[]`, `char ut_host[]`, `char ut_line[]` in both Linux (https://man7.org/linux/man-pages/man5/utmp.5.html) and POSIX (https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/utmpx.h.html) (`ut_host` is missing in POSIX).
User, host and line are clearly string.
This is suggesting that `ut_id` is string as well. Otherwise its type would be `short`, `int` or anything, but not an array or chars.

It is processed as a string by both util-linux (see https://github.com/util-linux/util-linux/blob/master/login-utils/utmpdump.c#L120-L126) and glibc (see https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/generic/utmp-equal.h;h=79c4dd13aaa21c6ee73bcf50c5ddb97a36a2bd56;hb=HEAD#l36). Emptiness is indicated by first zero byte, compared by strncmp().

I think it would be a mistake to assume that `ut_id` is four random bytes. This assumption would at least conflict with other base ecosystem code.

Moreover, all over the shadow's code `ut_id` is already used as a string.

Karlson2k:
`ut_id` historically (on GNU/Linux) was identifier from `inittab` file.
See https://manpages.debian.org/unstable/sysvinit-core/inittab.5.en.html

If an entry is being written to utmp file, the file is first checked for entries with the same `ut_id` and if one found, it is replaced with the new/updated one.

It was designed to work in the following way (simplified, see https://man7.org/linux/man-pages/man5/utmp.5.html):
* `init` process reads inittab file.
* `init` process adds utmp file `INIT_PROCESS` entry with `ut_id` from inittab file and `ut_pid` for newly started `*getty` process.
* `*getty` process finds utmp entry by `ut_pid` (and gets `ut_id` as assigned by `init`), sets entry type to `LOGIN_PROCESS`, sets `ut_line` and write utmp entry back. As entry has matching `ut_id`, it replaces old entry, created by `init`. `login` process started with the same PID.
* `login` process finds utmp entry by `ut_pid` (and gets again the same `ut_id` as initially assigned by `init`). After successful user login (authentication), `login` changes utmp entry type to `USER_PROCESS`, sets `ut_user`, `ut_host`, `ut_addr(_v6)`. When `utmp` entry is written, it replaces older `LOGIN_PROCESS` because the same `ut_id` is used.

I skipped extra details, like updates of `ut_time` on each step.

This mechanism allows `init` process to set `ut_id` initially for `*getty` and the same `ut_id` is preserved to the final user process.

However, this description does not provide any requirements for graphical terminal emulators (like `xterm`).
Graphical terminal does not have any `inittab` entries so they have no `ut_id` predefined. They decided to use last four chars from `ut_line` as `ut_id`. I guess the logic was something like: `ut_id` must not be empty (could rise compatibility problems), must be unique (as the entries with the same `ut_id` overwrites each other), and should not be confused with real console entries. As full `pts/1` string does not fit `ut_id` and first four chars (`pts/` would be the same for all terminals and would overwrite each other, they decided to use the unique part.

For graphical terminals **the last** four chars from `ut_line` are used as `ut_id`. See https://github.com/altlinux/libutempter/blob/4caa8ab94ff8b207228fa723a89214bf5e929321/libutempter/utempter.c#L99-L103 (the library used by many graphical terminal emulators).

The same is used for `ssh` logins. I haven't checked telnet logins. (Maybe it was invented first for telnet.)

However, `inittab` now is mostly historical implementation.

Modern implementations, like `systemd`, adapted `utempter` approach: used full name of the device (see https://github.com/systemd/systemd/blob/8013beb4a2221680b2c741bac6b3a33fe66406e1/units/getty%40.service.in#L41-L44) or **last four** chars if name is longer then four chars (see https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#UtmpIdentifier=).
You may see `it_id` assigned by systemd here: https://github.com/shadow-maint/shadow/pull/1282#issuecomment-3063358096. Check the entry with `LOGIN` username.

This PR changes nothing if an entry with `ut_id` exists. The same `ut_id` will be reused as before.
But, if no utmp entry has been created before starting of `login` (should not happen formally, but may happen if `login` is started as `init`, for example), then utmp entry is created with the same name as it would be done by systemd (or graphical terminals). The `ut_id` must be unique and truncation from the start gives less chance of id conflicts.

lib/utmp.c
Karlson2k: > This seems bogus. It's truncating from the start, instead of from the end. Why would we do that?

Because the last four symbols are more unique then first four, when numbers are added at the end.

> We should instead error out if the string doesn't fit the buffer (if it's appropriate to fail, which I don't know), or keep truncating as always.

Definitely not!
A missing utmp entry is a kind of soft error. Nothing that prevents user to login.
`ut_id` does not have to follow `ut_line`. Trimming `ut_line` is just usual way to fill the `ut_id`. Alternatively, just a random strings could be used, but this would not match other utmp file uses.

Check my [comment](https://github.com/shadow-maint/shadow/pull/1293#issuecomment-3077988615) for more details.

alejandro-colomar: It will probably still have collisions, even if they are less likely.  I think it is better to make the collisions more likely, which will make it more evident that there's a problem.

As long as we can't fix the problem, I wouldn't try hiding it.

Cc: @thesamesam

lib/utmp.c
alejandro-colomar: If you look carefully at util-linux and glibc (through the same links you shared), they both correctly treat it as null-padded fixed-width character arrays.  This is different from strings.

ISO C defines a string as:

```
A string is a contiguous sequence of characters terminated by and including the first null character.
```
<https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf#subsection.7.1.1>

And since these things are not guaranteed to be terminated, they are not strings.

Karlson2k: > If you look carefully at util-linux and glibc (through the same links you shared), they both correctly treat it as null-padded fixed-width character arrays. This is different from strings.

[skipped]

> And since these things are not guaranteed to be terminated, they are not strings.

Nope. They used `ut_id` as length-limited string.
Anything after NUL char is ignored.
If it would be just array of characters then NUL (literal zero) would not have any special meaning.
Fixed-width arrays must be compared by `memcmp()`, which always compares all bytes, while currently they are compared by `strncmp()` which stops at the first zero (NUL char) and ignores everything after it.
As I mentioned earlier, `it_id` is considered empty if  first (zero position) char is zero.

In short:
`"12" "\0" "3"` and `"12" "\0" "4"` are equal as length-limited strings, but differ as fixed-size character arrays.

Karlson2k: In other words: fixed length array value is limited by array size.
Limited-length string value is limited by first NUL char, but not longer then fixed length limit.

Karlson2k: Note that `utmpdump` prints `ut_id` as length-limited string, stopping at first literal zero.
And it is printed as (ASCII) string, not as "%02hhx%02hhx%02hhx%02hhx" (four hex numbers).

lib/utmp.c
Karlson2k: The collision would be hardly noticed during normal use.
And missing `utmp` entry is not a "problem". It could be intentional. For example, if `login` is started as `init` process. There is nothing to fix in this situation.

Karlson2k:
To summarise the idea of this PR.
This PR **does not** change functionality for normal workflow (when `utmp` entry is created before `login` is started).

This PR **does** correct behaviour in non-standard situations, for example, when `login` is started as `init` process, aligning choice of `ut_id` value with modern software (systemd, utemper).

lib/utmp.c
Karlson2k: > `utent->ut_id` is not a string. We can use the full array for non-zero bytes.

Yes, full array can be used for non-zero bytes.
But if zero-byte is used, the value is terminated by the first zero byte.

Karlson2k: Yes, this PR can be updated to use four bytes instead of three for `ut_id`.
Thanks for pointing this.

alejandro-colomar: strncmp(3) is used precisely because these are not strings.  If they were strings, they would use strcmp(3).

Cc: @chrisbazley

There's no such thing as a length-limited string in C.  What you're referring to is a null-padded fixed-width character array.  This means the array has a fixed size, but the contents may be narrower than that, padding with null characters until the array is filled.

Such a thing is not a string (again, read the definition of what a string is, in ISO C):

```
A string is a contiguous sequence of characters terminated by and including the first null character.
```

If the trailing null character is optional, then it's not a string.  In fact, GCC provides the [[gnu::nonstring]] attribute for this:
<https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#index-nonstring-variable-attribute>

CC: @fweimer-rh 

Florian, I think glibc should add the nonstring attribute to the utmp(5) structure fields.  I'll file a bug in glibc's bugzilla.

alejandro-colomar: TL;DR: `strn` stands for "strings NO".

lib/utmp.c
alejandro-colomar: Then this should not be a problem in the first place, so let's keep the code simple and not try to truncate from the start of the string, which is significantly more complex.

lib/utmp.c
alejandro-colomar: Ahhh, sorry @fweimer-rh , I missed that it is indeed already used in glibc:

```c
$ grepc utmp /usr/include/
/usr/include/rpcsvc/rusers.h:typedef struct rusers_utmp rusers_utmp;
/usr/include/x86_64-linux-gnu/bits/utmp.h:struct utmp
{
  short int ut_type;		/* Type of login.  */
  pid_t ut_pid;			/* Process ID of login process.  */
  char ut_line[UT_LINESIZE]
    __attribute_nonstring__;	/* Devicename.  */
  char ut_id[4]
    __attribute_nonstring__;	/* Inittab ID.  */
  char ut_user[UT_NAMESIZE]
    __attribute_nonstring__;	/* Username.  */
  char ut_host[UT_HOSTSIZE]
    __attribute_nonstring__;	/* Hostname for remote login.  */
  struct exit_status ut_exit;	/* Exit status of a process marked
				   as DEAD_PROCESS.  */
/* The ut_session and ut_tv fields must be the same size when compiled
   32- and 64-bit.  This allows data files and shared memory to be
   shared between 32- and 64-bit applications.  */
#if __WORDSIZE_TIME64_COMPAT32
  int32_t ut_session;		/* Session ID, used for windowing.  */
  struct
  {
    __uint32_t tv_sec;		/* Seconds.  */
    int32_t tv_usec;		/* Microseconds.  */
  } ut_tv;			/* Time entry was made.  */
#else
  long int ut_session;		/* Session ID, used for windowing.  */
  struct timeval ut_tv;		/* Time entry was made.  */
#endif

  int32_t ut_addr_v6[4];	/* Internet address of remote host.  */
  char __glibc_reserved[20];		/* Reserved for future use.  */
};
```

@Karlson2k , as you can see, glibc marks these fields as non-strings.

Karlson2k: > TL;DR: `strn` stands for "strings NO".

Funny. :smile: 
Of course not.

It means "not more than `n` chars", but it is still about strings.

Function that work with bytes are called mem*. `memcmp()`, `memcpy()`. You may call them "NO string". :)
Function that work with strings without limited length are `strcmp()`, `strcpy()`.
Function that work with strings AND limiting the maximum processed length are `strncmp()`, `strncpy()`.

Array of bytes does not have length. It only has size, which is the number of elements.

See https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/sys/cdefs.h;h=215ff937ee9c8eb85ebd9412c0ff5a93c6809f3f;hb=HEAD#l700:
```C
/* Describes a char array whose address can safely be passed as the first
   argument to strncpy and strncat, as the char array is not necessarily
   a NUL-terminated string.  */
# define __attribute_nonstring__ __attribute__ ((__nonstring__))
```
If something can be terminated with NUL char, it is still a string.

Calling them "array of bytes" is incorrect. For "array" all four bytes would be always used, even if the first byte is zero.

**For "possibly unterminated" strings the tail after first zero byte is always ignored. 
This is a key difference.**

alejandro-colomar: > > TL;DR: `strn` stands for "strings NO".
> 
> Funny. 😄 Of course not.
> 
> It means "not more than `n` chars", but it is still about strings.
> 
> Function that work with bytes are called mem*. `memcmp()`, `memcpy()`. You may call them "NO string". :) Function that work with strings without limited length are `strcmp()`, `strcpy()`. Function that work with strings AND limiting the maximum processed length are `strncmp()`, `strncpy()`.

This is a dangerous misconception.  If you want to use strings, but limiting the maximum processed string (so truncating), you have strlcpy(3) and strlcat(3), or other things.  strn*(3) functions are for nonstrings (things with the `[[gnu::nonstring]]` attribute), which as the name says, they're not strings.

> 
> Array of bytes does not have length. It only has size, which is the number of elements.
> 
> See https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/sys/cdefs.h;h=215ff937ee9c8eb85ebd9412c0ff5a93c6809f3f;hb=HEAD#l700:
> 
> ```c
> /* Describes a char array whose address can safely be passed as the first
>    argument to strncpy and strncat, as the char array is not necessarily
>    a NUL-terminated string.  */
> # define __attribute_nonstring__ __attribute__ ((__nonstring__))
> ```
> 
> If something can be terminated with NUL char, it is still a string.

You may call it string informally, but it is NOT a string.  Again, the definition of a string is specified by ISO C:

```
A string is a contiguous sequence of characters terminated by and including the first null character.
```

That is not subject to opinion.

> Calling them "array of bytes" is incorrect. For "array" all four bytes would be always used, even if the first byte is zero.

I didn't call them "array of bytes".  I call them null-padded (or zero-padded) fixed-width *character* array (not byte array).

> 
> **For "possibly unterminated" strings the tail after first zero byte is always ignored. This is a key difference.**

*Possibly unterminated strings* is an oxymoron.

Karlson2k: Looks like we are talking about the same things, but just calling them differently.
If sequence of chars can be terminated by zero, it is a string. However, there is a (relatively new) concept of string, that may use array completely without zero-termination, or less then full array, with termination.
When you say "something is not a string" (as [here](https://github.com/shadow-maint/shadow/pull/1293#discussion_r2208668771)) without additional clarifications, you can be easily understood incorrectly.
In C language any array has fixed size (possibly defined at run-time) so zero-padded only makes sense if you copy your fixed-size array into something larger. Sometimes size of array is called "array length" but this is incorrect.
The name of the attribute has been chosen for simplicity, but it is not the best choice. See attribute description [here](https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#index-nonstring-variable-attribute):

> nonstring
> The nonstring variable attribute specifies that an object or member declaration with type array of char, signed char, or unsigned char, or pointer to such a type is intended to store character arrays that do not necessarily contain a terminating NUL. This is useful in detecting uses of such arrays or pointers with functions that expect NUL-terminated strings, and to avoid warnings when such an array or pointer is used as an argument to a bounded string manipulation function such as strncpy. For example, without the attribute, GCC will issue a warning for the strncpy call below because it may truncate the copy without appending the terminating NUL character. Using the attribute makes it possible to suppress the warning. However, when the array is declared with the attribute the call to strlen is diagnosed because when the array doesn’t contain a NUL-terminated string the call is undefined. To copy, compare, or search non-string character arrays use the memcpy, memcmp, memchr, and other functions that operate on arrays of bytes. In addition, calling strnlen and strndup with such arrays is safe provided a suitable bound is specified, and not diagnosed.
> 

There is no known good term for this concept.
`Possibly unterminated` string (used in glibc) is not nice.
But just "array" is not nice too. 
When you talking about ASCII data, it is usually called "string". If say "it is not a string, it is array", it is read as "it is not ASCII (readable) data, it is binary data".

Fully correct name would be "a character array with left-aligned ASCII sequence terminated by NUL character or by the end of the array".
Or shorter "sequence of ASCII chars terminated at fixed length or at first NUL char".
From my point of view "length limited string" is pretty close the the last wording.

Your `nul-padded fixed-width character array` is not clear.
Firstly, padded on the left or on the right? Or both sides?
Secondly, it says nothing that NUL-char is not allowed inside the string. Therefore, `'a', '\0', 'b', '\0', '\0', '\0'` and `'a', '\0', 'c', '\0', '\0', '\0'` are different (according to your wording, just like compared by memcmp()), while `strncmp()` would report that they are equal.
You see why calling them "array" is bad? "Fixed-width" means that "all positions are always used".

Be careful to not misguide yourself. By saying "string no" your may treat such entities as suitable for `memcmp()`. And you wordings are about the same, you missing the point that NUL is still terminating such entities (as least you describe them in such way).

padding != termination

Karlson2k:
The PR has been updated based on comments.
Feel free to cherry-pick just the first commit.

lib/utmp.c
alejandro-colomar: > Looks like we are talking about the same things, but just calling them differently.

Yes.  But nomenclature is very important.

> If sequence of chars can be terminated by zero, it is a string.

This wording is incorrect.  *can be terminated* is not appropriate for a string.  *must be terminated* is the appropriate thing for a string.

> However, there is a (relatively new) concept of string, that may use array completely without zero-termination, or less then full array, with termination.

This is certainly not new.  This already existed in Seventh Edition Unix.  If you look at where strncpy(3) is used in V7 Unix, you'll find that it is exclusively used with utmp(5) members (unless I forget some other use cases there, I'm talking from memory).

And you should not call those strings.  They are not strings, according to the standard definition of string.

strncpy(3) was originally (V7 Unix) designed to be used with nonstring utmp(5) members, and any uses with strings are a misuse of the API.

> When you say "something is not a string" (as [here](https://github.com/shadow-maint/shadow/pull/1293#discussion_r2208668771)) without additional clarifications, you can be easily understood incorrectly.

I'd say it's the opposite.  If you say something is a string, you're likely holding it wrong.  Which your first revision of this PR exhibited, as you were leaving the 4th byte with garbage in some cases.

> In C language any array has fixed size (possibly defined at run-time)

If it has (run-time) variable size, it's not fixed, it's variable.  This is the terminology of the C language.

Whether that terminology matches other languages, I don't care.

> so zero-padded only makes sense if you copy your fixed-size array into something larger.

Yes.  For example, if you `strncpy(ut->ut_id, "f", _Countof(ut->ut_od);`.

> Sometimes size of array is called "array length" but this is incorrect.

Agree.  See <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3327.pdf>.

On the other hand, size of array is also incorrect sometimes.

> The name of the attribute has been chosen for simplicity, but it is not the best choice.

I disagree.  It's the best choice, IMO.  It is not a string.

> See attribute description [here](https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#index-nonstring-variable-attribute):
> 
> > nonstring
> > The nonstring variable attribute specifies that an object or member declaration with type array of char, signed char, or unsigned char, or pointer to such a type is intended to store character arrays that do not necessarily contain a terminating NUL. This is useful in detecting uses of such arrays or pointers with functions that expect NUL-terminated strings, and to avoid warnings when such an array or pointer is used as an argument to a bounded string manipulation function such as strncpy. For example, without the attribute, GCC will issue a warning for the strncpy call below because it may truncate the copy without appending the terminating NUL character. Using the attribute makes it possible to suppress the warning. However, when the array is declared with the attribute the call to strlen is diagnosed because when the array doesn’t contain a NUL-terminated string the call is undefined. To copy, compare, or search non-string character arrays use the memcpy, memcmp, memchr, and other functions that operate on arrays of bytes. In addition, calling strnlen and strndup with such arrays is safe provided a suitable bound is specified, and not diagnosed.
> 



> There is no known good term for this concept. `Possibly unterminated` string (used in glibc) is not nice. But just "array" is not nice too.

The term is nonstring, which englobes both

-  zero-padded fixed-width character arrays (those used in utmp(5), which use strncpy(3)), and
-  length-bound character sequences (this is what some projects define as `struct str {size_t len; char *p;};`).

*Possibly unterminated string* (glibc) is wrong, indeed.  That is contradicting, and ambiguous.

> When you talking about ASCII data, it is usually called "string".

This is where you're holding it wrong.  ASCII data is ASCII data.  Not a string.  A string is a string.  ASCII data is ASCII data.  They sometimes go together, sometimes (like here) not.

> If say "it is not a string, it is array", it is read as "it is not ASCII (readable) data, it is binary data".
> 
> Fully correct name would be "a character array with left-aligned ASCII sequence terminated by NUL character or by the end of the array". Or shorter "sequence of ASCII chars terminated at fixed length or at first NUL char". From my point of view "length limited string" is pretty close the the last wording.

It is not a string, and string should not appear anywhere in its name.  That would lead to bugs.  Also, length-limited string is ambiguous: it's not clear whether it is padded, or the length is exact.

Please use `zero-padded fixed-width character array` and `length-bounded character sequence`, for the two different non-string objects.  See the string_copying(7) manual page for more information:
<https://www.man7.org/linux/man-pages/man7/string_copying.7.html>

> 
> Your `nul-padded fixed-width character array` is not clear. Firstly, padded on the left or on the right? Or both sides?

Obviously on the right.

> Secondly, it says nothing that NUL-char is not allowed inside the string. Therefore, `'a', '\0', 'b', '\0', '\0', '\0'` and `'a', '\0', 'c', '\0', '\0', '\0'` are different (according to your wording, just like compared by memcmp()), while `strncmp()` would report that they are equal. You see why calling them "array" is bad? "Fixed-width" means that "all positions are always used".

That's why zero-padding comes in: the padding is for not using them.

> Be careful to not misguide yourself. By saying "string no" your may treat such entities as suitable for `memcmp()`. And you wordings are about the same, you missing the point that NUL is still terminating such entities (as least you describe them in such way).

mem*(3) is for handling bytes.  strn*() is for handling zero-padded character arrays.  Don't handle zero-padded character arrays as bytes.  Don't handle bytes as zero-padded character arrays.

> 
> padding != termination

lib/utmp.c
```diff
@@ -274,8 +274,14 @@ prepare_utmp(const char *name, const char *line, const char *host,
 	if (NULL != ut) {
 		STRNCPY(utent->ut_id, ut->ut_id);
 	} else {
-		/* XXX - assumes /dev/tty?? */
-		STRNCPY(utent->ut_id, line + 3);
+		size_t line_len;
+
+		line_len = strlen (line);
```
alejandro-colomar: Please don't add spaces before `(`.

lib/utmp.c
```diff
@@ -274,8 +274,14 @@ prepare_utmp(const char *name, const char *line, const char *host,
...
+		if (line_len > 3) {
+			STRNCPY(utent->ut_id, line + 3);
+		} else {
+			STRNCPY(utent->ut_id, line);
+		}
```
alejandro-colomar: I assume the contents of the first 3 characters are (usually) `tty`, right?

If so, I propose rewriting this in a more explicit and elegant way:

```c
STRNCPY(utent->ut_id, strprefix(line, "tty") ?: line);
```

lib/utmp.c
Karlson2k: > The term is nonstring, which englobes both
> 
> * zero-padded fixed-width character arrays (those used in utmp(5), which use strncpy(3)), and
> * length-bound character sequences (this is what some projects define as `struct str {size_t len; char *p;};`).

Could you please point to the official definition of the `nonstring`?

> If you say something is a string, you're likely holding it wrong. Which your first revision of this PR exhibited, as you were leaving the 4th byte with garbage in some cases.

My mistake in the first version of this PR was the lack of knowledge that `ut_id` might be not zero-terminated.
You comment that it is not a string pushed me to wrong direction.

> Please use `zero-padded fixed-width character array` and `length-bounded character sequence`, for the two different non-string objects. See the string_copying(7) manual page for more information: https://www.man7.org/linux/man-pages/man7/string_copying.7.html
> 

I see. Looks like this idea (terminology) was invented by Linux kernel team and you are bound to it.
However, it does not make it perfect automatically. And this terminology has a limited use outside Linux kernel.
`Null-padded character sequences` is not universal term and not self-explainable. If you use this term without the reference to its definition, then it not clear. First and the most important: is NUL-char is a char? It must be a char, so sequence like `'a', 'b', NUL, 'c', 'd'` is still the sequence of characters, right? If you add (right) padding to 8 chars, it become `'a', 'b', NUL, 'c', 'd', NUL, NUL, NUL`, correct?
What happens if you compare it with another `Null-padded character sequences` `'a', 'b', NUL, NUL, NUL, NUL, NUL, NUL`, where just two characters 'a' and 'b' are (right-)padded with  NUL?
Both are valid `Null-padded character sequences` and they are different, correct? Why `Null-padded character sequences` function `strncmp()` tells that they are equal?
Number can be left-padded and right-padded. Do not see why one must assume automatic right padding for the "Character sequence" (which is not a string).

Please, do not use non-universal terms, assuming that they will be clear for everyone.

Added: I found the reason why you are so insisting on this term. You invented it. Nice.
But it still does not make it perfect. See my previous explanations. 
The key point: the name does not suggest than any NUL is terminating. The name says only about the padding.
Also: original name of zero character is "NUL", not "null". "Null" is more often used for pointers.

Karlson2k: Sorry, I'll not use this term anyway as I found it incorrect and misleading.

lib/utmp.c
Karlson2k: Good point.
But I'll avoid non-portable skipping of second parameter.

lib/utmp.c
Karlson2k: I still think that simple strlen together with '+' and '-' is not something complex.
What is more important, it aligns behaviour with other modern software.

lib/utmp.c
alejandro-colomar: You mean non-portability of `?:`?  We're discussing standardization of that in the C Committee right now.

And we're using it a lot in this project already:

```

alx@devuan:~/src/shadow/shadow/master$ grep -rn '?:' lib* src/
lib/alloc/reallocf.h:19:	_Generic(p, type *: (type *) reallocarrayf(p, (n) ?: 1, sizeof(type)))\
lib/alloc/reallocf.h:33:	q = reallocarray(p, nmemb ?: 1, size ?: 1);
lib/alloc/realloc.h:16:	_Generic(p, type *: (type *) reallocarray(p, (n) ?: 1, sizeof(type))) \
lib/string/strdup/memdup.h:30:	new = malloc(size ?: 1);
lib/console.c:107:	tty = strprefix(tty, "/dev/") ?: tty;
lib/user_busy.c:208:		tmp_d_name = strprefix(tmp_d_name, ".") ?: tmp_d_name;
lib/limits.c:477:			cp = strprefix(cp, ",") ?: cp;
lib/utmp.c:263:	line = strprefix(line, "/dev/") ?: line;
src/newgrp.c:254:	tty = strprefix(tty, "/dev/") ?: tty;
```

Karlson2k: Correct.
While it is only under discussion and not standardised, I'll not use it.

lib/utmp.c
alejandro-colomar: > > The term is nonstring, which englobes both
> > 
> > * zero-padded fixed-width character arrays (those used in utmp(5), which use strncpy(3)), and
> > * length-bound character sequences (this is what some projects define as `struct str {size_t len; char *p;};`).
> 
> Could you please point to the official definition of the `nonstring`?

In ISO C, there is none at the moment, but we're talking about it, indeed.

In GNU C, where [[gnu::nonstring]] was invented, the definition is here:

<https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#index-nonstring-variable-attribute>

> 
> > If you say something is a string, you're likely holding it wrong. Which your first revision of this PR exhibited, as you were leaving the 4th byte with garbage in some cases.
> 
> My mistake in the first version of this PR was the lack of knowledge that `ut_id` might be not zero-terminated. 

Wherever you see strncpy(3), consider that it doesn't terminate the output, so either the programmer was holding it wrong, or the output need not be zero-terminated.

> You comment that it is not a string pushed me to wrong direction.

My comment was after the PR already existed.

> > Please use `zero-padded fixed-width character array` and `length-bounded character sequence`, for the two different non-string objects. See the string_copying(7) manual page for more information: https://www.man7.org/linux/man-pages/man7/string_copying.7.html
> 
> I see. Looks like this idea (terminology) was invented by Linux kernel team and you are bound to it. However, it does not make it perfect automatically.

It is not necessarily perfect, but at least it avoids using the term 'string' for nonstrings.

> And this terminology has a limited use outside Linux kernel. `Null-padded character sequences` is not universal term and not self-explainable. If you use this term without the reference to its definition, then it not clear. First and the most important: is NUL-char is a char? It must be a char, so sequence like `'a', 'b', NUL, 'c', 'd'` is still the sequence of characters, right? If you add (right) padding to 8 chars, it become `'a', 'b', NUL, 'c', 'd', NUL, NUL, NUL`, correct? What happens if you compare it with another `Null-padded character sequences` `'a', 'b', NUL, NUL, NUL, NUL, NUL, NUL`, where just two characters 'a' and 'b' are (right-)padded with NUL? Both are valid `Null-padded character sequences` and they are different, correct? Why `Null-padded character sequences` function `strncmp()` tells that they are equal?

Because the first NUL delimits the padding, and padding should be ignored.

> Number can be left-padded and right-padded. Do not see why one must assume automatic right padding for the "Character sequence" (which is not a string).
> 
> Please, do not use non-universal terms, assuming that they will be clear for everyone.
> 
> Added: I found the reason why you are so insisting on this term. You invented it. Nice. But it still does not make it perfect. See my previous explanations.

I don't mind which terms you use, as long as you don't use string or anything with string in it.  'string' is the only term with a very strict definition, and it is incompatible with strn*() and utmp(5).

> The key point: the name does not suggest than any NUL is terminating. The name says only about the padding. Also: original name of zero character is "NUL", not "null". "Null" is more often used for pointers.

NUL is the ASCII name for the null character (aka, null byte).  NULL is the name of a null pointer constant.  "Null" isn't anything clear.

lib/utmp.c
alejandro-colomar: We're using it a lot in this project already.  You won't make the code less portable by using it.

alejandro-colomar: I oppose using magic numbers in the code.  `+ 3` and `> 3` are error prone.

Karlson2k: Good point, thanks!

Karlson2k:
Updated again

lib/utmp.c
```diff
@@ -274,8 +274,11 @@ prepare_utmp(const char *name, const char *line, const char *host,
 	if (NULL != ut) {
 		STRNCPY(utent->ut_id, ut->ut_id);
 	} else {
-		/* XXX - assumes /dev/tty?? */
-		STRNCPY(utent->ut_id, line + 3);
+		if (0 == strncmp(line, "tty", STRLEN("tty"))) {
```
alejandro-colomar: We have forbidden use of strncmp(3) (or any strn*() functions) with strings in this project.  Please use strprefix().

lib/utmp.c
```diff
@@ -274,8 +274,11 @@ prepare_utmp(const char *name, const char *line, const char *host,
...
+			STRNCPY(utent->ut_id, line + STRLEN("tty"));
+		} else {
+			STRNCPY(utent->ut_id, line);
+		}
```
alejandro-colomar: Please don't use braces when all branches have one line.

Karlson2k: Does this project have some code rules written somewhere?

alejandro-colomar: Nope.  We more or less follow the Linux kernel coding style, with some exceptions.

Karlson2k: It would be nice to put some references somewhere, for example to `CONTRIBUTING` file.
It hard to follow rules that are not declared nor described.

lib/utmp.c
Karlson2k: It would be nice to put these rules somewhere, like `CONTRIBUTION` file. The you would be able just to point to the file and ask to follow it.
It is impossible to guess which standard C functions can be used.

Karlson2k: If it is common use patter for `strprefix()`, I'd suggest adding `strprefixskip()` and avoid `? : param` each time. This would improve readability too.

lib/utmp.c
alejandro-colomar: Yep, we should eventually do that.  However, we have already a lot of work with the code at the moment, so that will take some time.

Karlson2k: It is always a lot of work with the code. Usually it never ends.
There is no "perfect" time to work on other important things.

Karlson2k:
Updated as requested

lib/utmp.c
alejandro-colomar: > If it is common use patter for `strprefix()`, I'd suggest adding `strprefixskip()` and avoid `? : param` each time.

I've been considering this.  I'm still undecided, but might do it eventually.

> This would improve readability too.

I'm unsure about the improved readability.  The code with `?:` is more explicit about how it works.  Otherwise, you'd need to have a deep understanding on the function, which is why I'm not convinced that it would be a good idea.

> It would be nice to put these rules somewhere, like `CONTRIBUTION` file. The you would be able just to point to the file and ask to follow it. It is impossible to guess which standard C functions can be used.

We haven't finished working on that.  The rules are being changed on the fly, while I research <string.h> APIs.  Once our APIs settle, yes, we should write a document about it.

alejandro-colomar: > It is impossible to guess which standard C functions can be used.

The underlying rule is: "only those that are obviously correct".  :)

I'm researching to determine which are obviously correct, which isn't obvious.

lib/utmp.c
alejandro-colomar: Nah, we're really on a peak.  I have over 400 patches pending review by the other co-maintainers.  Once that queue of patches reduces to two-digits only, I could start writing that.  :)

lib/utmp.c
Karlson2k: > > This would improve readability too.
> 
> I'm unsure about the improved readability. The code with `?:` is more explicit about how it works. Otherwise, you'd need to have a deep understanding on the function, which is why I'm not convinced that it would be a good idea.

For me, without reading the `strprefix()` code it is not obvious that function returns **pointer**  the the char after the prefix. By looking on the name only, I'd think that it is boolean. 
While `strprefixskip()` says more in its name.
However, it is just my view.

> The rules are being changed on the fly, while I research <string.h> APIs.

Yep, I had the same feeling. :smile:

alejandro-colomar: > > > This would improve readability too.
> > 
> > 
> > I'm unsure about the improved readability. The code with `?:` is more explicit about how it works. Otherwise, you'd need to have a deep understanding on the function, which is why I'm not convinced that it would be a good idea.
> 
> For me, without reading the `strprefix()` code it is not obvious that function returns **pointer** the the char after the prefix. By looking on the name only, I'd think that it is boolean. While `strprefixskip()` says more in its name. However, it is just my view.

It's similar to strchr(3), which returns a boolean-like pointer.  It returns a pointer to the found match if found, and NULL otherwise.  This converts implicitly to bool just fine, but is more useful.

Nevertheless, we have discussed this in the C Committee and agreed to split the API in two:

-  `strpfx(s, pfx)` would return `bool`.  It would be equivalent to `(bool) stppfx(s, pfx)`.
-  `stppfx(s, pfx)` would return a pointer.  This is consistent with the `stp*()` functions returning a pointer in the middle of the string.

The main rationale for the separation was for callbacks compatible with streq() and strchr(3), but it is also nice to be able to distinguish where the pointer is converted to bool by the name.

This will be voted by the committee for inclusion:
<https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3612.txt>

Karlson2k:
Rebased, excluded merged commit.

lib/utmp.c
```diff
@@ -274,7 +274,14 @@ prepare_utmp(const char *name, const char *line, const char *host,
...
 	} else {
-		STRNCPY(utent->ut_id, strprefix(line, "tty") ?: line);
+		const size_t ut_line_maxlen = countof(utent->ut_id);
+		size_t line_len;
+
+		line_len = strlen(line);
+		if (line_len > ut_line_maxlen)
+			STRNCPY(utent->ut_id, line + (line_len - ut_line_maxlen));
+		else
+			STRNCPY(utent->ut_id, line);
```
alejandro-colomar: Do you mean that systemd and utemper would use the full `tty3` instead of `3`?

alejandro-colomar: Would you mind providing links for that in the commit message?

alejandro-colomar: We (normally) use C89 declarations.  This means not initializing variables with a function (or function-like) call:

```c
size_t  size, len;

size = countof(utent->ut_id);
len = strlen(line);
if (len <= size)
	STRNCPY(utent->ut_id, line);
else
	STRNCPY(utent->ut_id, line + len - size);
```

alejandro-colomar: Also, prefer short variable names.  The context fills the rest of the information.

<https://doc.cat-v.org/bell_labs/pikestyle>

Karlson2k: Yes, exactly.
On the system with KDE started, three terminals running, one terminal has been closed, one SSH incoming connection and logged in TTY3 console :
```console
 $ utmpdump /run/utmp
Utmp dump of /run/utmp
[2] [00000] [~~  ] [reboot  ] [~           ] [6.15.6-gentoo-k2k-special10] [0.0.0.0        ] [2025-07-12T12:58:44,620331+00:00]
[7] [00803] [    ] [k2k     ] [tty2        ] [:0                  ] [0.0.0.0        ] [2025-07-12T12:58:46,947344+00:00]
[7] [00907] [ts/0] [k2k     ] [pts/0       ] [:0                  ] [0.0.0.0        ] [2025-07-12T12:58:47,659494+00:00]
[7] [323681] [ts/2] [k2k     ] [pts/2       ] [:0                  ] [0.0.0.0        ] [2025-07-16T09:55:57,739677+00:00]
[7] [01570] [ts/1] [k2k     ] [pts/1       ] [:0                  ] [0.0.0.0        ] [2025-07-12T12:58:48,269989+00:00]
[8] [83944] [ts/5] [k2k     ] [pts/5       ] [                    ] [0.0.0.0        ] [2025-07-12T20:43:21,766341+00:00]
[7] [323776] [ts/4] [k2k     ] [pts/4       ] [127.0.0.1           ] [127.0.0.1      ] [2025-07-16T09:56:01,745642+00:00]
[6] [325414] [tty3] [LOGIN   ] [tty3        ] [                    ] [0.0.0.0        ] [2025-07-16T10:16:27,781377+00:00]
[7] [326020] [3   ] [k2k     ] [tty3        ] [                    ] [0.0.0.0        ] [2025-07-16T10:29:27,296676+00:00]
```
`tty2` was created by SDDM (incorrect, empty `ut_it`).
`pts/0`, `pts/2`, `pts/1`, `pts/5` entries were created by `konsole` (KDE's terminal emulator)
`pts/4` was created by SSH.
First `tty3` entry (with "LOGIN") was created by systemd
Last `tty3` entry was created by unpatched `login` (no fork/PID patches yet). 

You see, if terminal emulators would use just four chars, they all would be just `pts/` and overwrite each other.

alejandro-colomar: Sounds reasonable.  Thanks!

alejandro-colomar: It would be interesting to have `STRNCPYTAIL()`:

```c
#define STRNCPYTAIL(dst, src)  strncpytail(countof(dst), dst, src)

char *
strncpytail(size_t dsize, char *dst, const char *src)
{
	size_t  len;

	len = strlen(src);
	if (len <= dsize)
		strncpy(dst, src, dsize);
	else
		strncpy(dst, src + len - size, dsize);

	return dst;
}
```

Karlson2k: C89 allows initialisation at declaration.
Moreover, it is the only way to used `const` variables.
The only limitations is the static variables cannot be initialised by non-constants (like function calls).

However, if you want to separate declarations from initialisation, it's fine, except you are losing ability to use `const` (which have higher chance to be better optimised by compiler).

alejandro-colomar: > C89 allows initialisation at declaration.

Hmm, did it?  I didn't remember.

Anyway, it's ugly.  We don't do that.

> Moreover, it is the only way to used `const` variables.

We usually avoid `const` variables (but not pointers to const).

> The only limitations is the static variables cannot be initialised by non-constants (like function calls).
> 
> However, if you want to separate declarations from initialisation, it's fine, except you are losing ability to use `const` (which have higher chance to be better optimised by compiler).

Nah, the compiler should be able to optimize everything just fine, unless it's a 30-yr-old compiler.

Karlson2k: > It would be interesting to have `STRNCPYTAIL()`:

Yes, it would be nice to add it to your local lib with various chars processing functions.
I'm sure that you'll find other uses for it later.

alejandro-colomar: > It would be interesting to have `STRNCPYTAIL()`:
I'm thinking we could get rid of the conditional (or actually hide it) with MIN():

```c
char *
strncpytail(size_t dsize; char dst[dsize], const char *src, size_t dsize)
{
	return strncpy(dst, strnul(src) - MIN(strlen(src), dsize), dsize);
}
```

alejandro-colomar: Would you mind using strnul() and MIN() in the patch?  This would make it a one-liner.

I can add the API later (unless you also want to add it, which would be fine).

Karlson2k: > I'm thinking we could get rid of the conditional (or actually hide it) with MIN():
> 
> ```c
> char *
> strncpytail(size_t dsize, char *dst, const char *src)
> {
> 	return strncpy(dst, strnul(src) - MIN(strlen(src), dsize), dsize);
> }
> ```
`MIN()` will evaluate parameters twice, which is not always optimised by compiler.

alejandro-colomar: > > I'm thinking we could get rid of the conditional (or actually hide it) with MIN():
> > ```c
> > char *
> > strncpytail(size_t dsize, char *dst, const char *src)
> > {
> > 	return strncpy(dst, strnul(src) - MIN(strlen(src), dsize), dsize);
> > }
> > ```
> 
> `MIN()` will evaluate parameters twice, which is not always optimised by compiler.

The only problem with double evaluation is side effects, but here we know there are no side effects.

Performance shouldn't be an issue: compilers are smart enough these days.  And by using MIN(), we're removing an entire call to strncpy(3), so we can add an strlen(3) call.

alejandro-colomar: Also, strlen("pts/0"); is going to be blazing fast.  I wouldn't worry even if the compiler was dumb enough to call it twice.

Karlson2k: The sizes of binaries are still different. 
However, clang makes my version shorter, while gcc makes your version shorter..
See https://godbolt.org/z/Kxfsreb54
But, sure, no performance issues anyway.

Karlson2k:
Updated exactly as suggested.

lib/utmp.c
alejandro-colomar: Hmmm, in my experiment, I've written

```c
alx@devuan:~/tmp$ cat strncpytail1.c 
#include <string.h>
#include <sys/param.h>

char *
strncpytail(size_t dsize, char *dst, const char *src)
{
	return strncpy(dst, src + strlen(src) - MIN(strlen(src), dsize), dsize);
}
```
```c
alx@devuan:~/tmp$ cat strncpytail2.c 
#include <string.h>

char *
strncpytail(size_t dsize, char *dst, const char *src)
{
	size_t  len;

	len = strlen(src);
	if (len <= dsize)
		strncpy(dst, src, dsize);
	else
		strncpy(dst, src + len - dsize, dsize);

	return dst;
}
```

Then built with
```sh
$ clang -Wall -Wextra -S -Os strncpytail1.c
$ clang -Wall -Wextra -S -Os strncpytail2.c
```
(and also tried gcc, and -O3 (with both clang and gcc).

My results are that *2.c results in larger code under both -O3 and -Os with GCC, and almost equivalent code under both -O3 and -Os in Clang.

I've always compared with
```sh
$ diff -u strncpytail*.s
```

lib/utmp.c
```diff
@@ -32,6 +32,7 @@
 #include "string/strcpy/strtcpy.h"
 #include "string/strdup/xstrdup.h"
 #include "string/strdup/xstrndup.h"
+#include "string/strchr/strnul.h"
```
alejandro-colomar: Please insert the include alphabetically.  The one for strnul should go before the one for streq.

alejandro-colomar (approved):
Thanks!  LGTM.

```
```

alejandro-colomar:
> This PR aligns generated "ut_id" with ids produced by other modern software, like systemd and libutemper. For reference:
> 
>     * libutemper implementation: https://github.com/altlinux/libutempter/blob/4caa8ab94ff8b207228fa723a89214bf5e929321/libutempter/utempter.c#L99-L103
>       Note: libutemper is used by many graphical terminal emulator applications (konsole, xterm, others) to deal with utmp file.
> 
>     * systemd implementation:
>       use full name of the device (see https://github.com/systemd/systemd/blob/8013beb4a2221680b2c741bac6b3a33fe66406e1/units/getty%40.service.in#L41-L44) or last four chars if name is longer then four chars (see https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#UtmpIdentifier=)
> 
> 
> To summarise the idea of this PR. This PR **does not** change functionality for normal workflow (when `utmp` entry is created before `login` is started).
> 
> This PR **does** correct behaviour in non-standard situations, for example, when `login` is started as `init` process, aligning choice of `ut_id` value with modern software (systemd, utemper).

Please include those two links in the commit message, if you don't mind.

lib/utmp.c
Karlson2k: A few extra optimisations could be added.
Second version of the code:
```C
char *
strncpytail(char *dst, const char *src, size_t dsize)
{
	size_t  len;

	len = strlen(src);
	if (len <= dsize)
		return memcpy(dst, src, dsize);

	return strncpy(dst, src + len - dsize, dsize);
}
```
1. Return value, returned by strncpy. At least it gives less cleanup, but can be optimised even better (see below).
2. Reorder function parameters. If functions have the same signature, compiler may use "tail optimisation" (for example on x86 `call something, ret` replaced with `jmp something`)
3. When size if know, `memcpy()` could be **MUCH** faster then `strncpy()`. The code is larger, but also is much faster.

The second optimisation could be applied to your version as well:
```C
char *
strncpytail(char *dst, const char *src, size_t dsize)
{
	return strncpy(dst, src + strlen(src) - MIN(strlen(src), dsize), dsize);
}

#define STRNCPYTAIL(dst, src) strncpytail((dst), (src), countof((dst))
```
Try it.

Most distributions use `-O2` as `-O3` is know to produce more problems and sometimes results in performance degradation (usually due to larger code that does not fit CPU cache), so I always try with `-O2`, sometimes followed by `-Oz` and `-O3`.
Therefore `-O3` optimisation (which still could be faster) is not relevant for real life use-cases.

Sometimes I use `-fverbose-asm`, but it can be too noisy.

Karlson2k:
@alejandro-colomar Includes sorted, links added.

alejandro-colomar:
Thanks!

PR #1294, opened 2025-07-15
*/: Fix including <config.h> as system header

Karlson2k:
"config.h" is a locally generated header. It must be included as '#include "config.h"'.
It is already included correctly in some sources files. This commit unifies the way how "config.h" is included.
Examples of current correct includes:
* https://github.com/shadow-maint/shadow/blob/9195f3807dfaa2062bcb8f79c61efc1e1f6dba46/lib/defines.h#L7
* https://github.com/shadow-maint/shadow/blob/9195f3807dfaa2062bcb8f79c61efc1e1f6dba46/lib/attr.h#L5

### DIFF
--- a/lib/addgrps.c
+++ b/lib/addgrps.c
@@ -6,7 +6,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 
-#include <config.h>
+#include "config.h"
 
 #if !defined(USE_PAM)
 
--- a/lib/adds.c
+++ b/lib/adds.c
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 
-#include <config.h>
+#include "config.h"
 
 #include "adds.h"
 
--- a/lib/adds.h
+++ b/lib/adds.h
@@ -6,7 +6,7 @@
 #define SHADOW_INCLUDE_LIB_ADDS_H_
 
 
-#include <config.h>
+#include "config.h"
 
 #include <errno.h>
 #include <limits.h>
--- a/lib/age.c
+++ b/lib/age.c
@@ -7,7 +7,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-#include <config.h>
+#include "config.h"
 
 #include <sys/types.h>
 #include <stdio.h>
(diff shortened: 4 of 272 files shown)


alejandro-colomar:
Was this a scripted change?  Would you mind saying which script you ran exactly?  Thanks!

Karlson2k:
I used "search and replace in files" in my default IDE Eclipse CDT.
However, simple commands, like `sed -i -e 's@^#include <config\.h>$@#include "config.h"@' */*.{c,h} */*/*.{c,h} */*/*/*.{c,h} */*/*/*/*.{c,h}` or `find . -type f \( -name '*.c' -o -name '*.h' \) -exec sed -i -e 's@^#include <config\.h>$@#include "config.h"@'  {} +` can do the same (assuming your sed supports non-portable `-i`, like GNU sed).

alejandro-colomar:
> I used "search and replace in files" in my default IDE Eclipse CDT. However, simple commands, like `sed -i -e 's@^#include <config\.h>$@#include "config.h"@' */*.{c,h} */*/*.{c,h} */*/*/*.{c,h} */*/*/*/*.{c,h}` or `find . -type f \( -name '*.c' -o -name '*.h' \) -exec sed -i -e 's@^#include <config\.h>$@#include "config.h"@' {} +` can do the same (assuming your sed supports non-portable `-i`, like GNU sed).

Yup, I'd have done

```sh
$ find . -type f | grep '\.[ch]$' | xargs sed -i 's/^#include <config\.h>$/#include "config.h"/';
```

It is always good to know how scripted changes have been applied exactly, in case it misses some cases accidentally.

Karlson2k:
> Yup, I'd have done
> 
> ```shell
> $ find . -type f | grep '\.[ch]$' | xargs sed -i 's/^#include <config\.h>$/#include "config.h"/';
> ```
> It is always good to know how scripted changes have been applied exactly, in case it misses some cases accidentally.

Sed command can be further improved to really catch everything:
`sed -i -Ee 's/^([  ]*)#([  ]*)include <config\.h>/#\2include "config.h/'`, where `[  ]` should be `[` `space` `tab` `]`.
:smile:

PR #1296, opened 2025-07-16
Add and use memdup(), MEMDUP()

alejandro-colomar:
See also: <https://github.com/shadow-maint/shadow/pull/1292#discussion_r2208039003>

Cc: @Karlson2k


lib/string/strdup/memdup.h
```diff
@@ -0,0 +1,38 @@
...
+#include "config.h"
+
+#include <stddef.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "attr.h"
+
+
+#define MEMDUP(p, n, T)  _Generic(p, T *: (T *) memdup(p, n * sizeof(T)))
```
Karlson2k: Not checked possible overflow, when `n > sizeof(T)` is larger than `SIZE_MAX`

Karlson2k: The check could be moved to `memdup()` function.

lib/string/strdup/memdup.h
```diff
@@ -0,0 +1,38 @@
...
+
+
+ATTR_MALLOC(free)
+inline void *memdup(const void *p, size_t size);
+
+
+inline void *
+memdup(const void *p, size_t size)
```
Karlson2k:
```C
inline void *
memdup(const void *p, size_t n, size_t size)
{
	size_t alloc_size;
	void  *new;

    
	if (0 == size)
		size = 1;
      
	alloc_size = n * size;
	
	if (alloc_size / n != size)
		return NULL; /* overflow */
	new = malloc(alloc_size);
	if (new == NULL)
		return NULL;

	return memcpy(new, p, alloc_size);
}
```

lib/string/strdup/memdup.h
```diff
@@ -0,0 +1,38 @@
...
+{
+	void  *new;
+
+	new = malloc(size ?: 1);
```
Karlson2k: Should just `NULL` be returned if requested copy of zero items?
Enforcing copy of single item looks dangerous.

Karlson2k: Skipping second parameter for ternary operator is GNU extension and non-portable.

lib/string/strdup/memdup.h
```diff
@@ -0,0 +1,38 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+#include "attr.h"
+
+
+#define MEMDUP(p, n, T)  _Generic(p, T *: (T *) memdup(p, n * sizeof(T)))
+
+
+ATTR_MALLOC(free)
```
Karlson2k: `ATTR_MALLOC(free)` is not enough.
This expands to `[[gnu::malloc(free)]]` and means that returned value must be de-allocated by specific function.
It must be marked with `[[gnu::malloc]]` as well, which means "returned pointer is unique and does not alias any other pointers". These two forms are independent from each other.
See https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-malloc-function-attribute

Karlson2k: Note: clang supports these attributes too, `ATTR_MALLOC` expands to nothing with clang.
See https://clang.llvm.org/docs/AttributeReference.html#malloc

lib/string/strdup/memdup.h
Karlson2k: Compiler will optimise overflow check and will use resulting processor flag instead of performing division.
Alternatively [__builtin_mul_overflow_p](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html#index-_005f_005fbuiltin_005fmul_005foverflow_005fp) can be used, but it makes code less readable, produce exactly the same binary result and makes it non-portable.

Karlson2k: :white_check_mark:  Already corrected.

Karlson2k:
I still recommend adding attribute `malloc` (in form without destructor). When used without `-fno-strict-aliasing` (like this project) it improves optimisations.

Karlson2k:
See https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-malloc-function-attribute

> Attribute malloc indicates that a function is malloc-like, i.e., that the pointer P returned by the function cannot alias any other pointer valid when the function returns, and moreover no pointers to valid objects occur in any storage addressed by P. In addition, GCC predicts that a function with the attribute returns non-null in most cases.
> 
> **Independently**, the form of the attribute with one or two arguments associates deallocator as a suitable deallocation function for pointers returned from the malloc-like function. ptr-index denotes the positional argument to which when the pointer is passed in calls to deallocator has the effect of deallocating it.

> For example, besides stating that the functions return pointers that do not alias any others, the following declarations make fclose a suitable deallocator for pointers returned from all functions except popen, and pclose as the only suitable deallocator for pointers returned from popen. The deallocator functions must be declared before they can be referenced in the attribute.
> ```C
> int fclose (FILE*);
> int pclose (FILE*);
> 
> __attribute__ ((malloc, malloc (fclose, 1)))
>   FILE* fdopen (int, const char*);
> __attribute__ ((malloc, malloc (fclose, 1)))
>   FILE* fopen (const char*, const char*);
> __attribute__ ((malloc, malloc (fclose, 1)))
>   FILE* fmemopen(void *, size_t, const char *);
> __attribute__ ((malloc, malloc (pclose, 1)))
>   FILE* popen (const char*, const char*);
> __attribute__ ((malloc, malloc (fclose, 1)))
>   FILE* tmpfile (void);
> ```

alejandro-colomar:
> I still recommend adding attribute `malloc` (in form without destructor). When used without `-fno-strict-aliasing` (like this project) it improves optimisations.

I think by having this function declared `inline` the compiler should be able to realize about any possible optimizations, since it transparently sees the malloc(3) call.

I don't like using attributes for telling the compiler that it can optimize.  That's error-prone.  I prefer allowing it to see more of the code, so it can reason by itself.

Karlson2k:
> I think by having this function declared `inline` the compiler should be able to realize about any possible optimizations, since it transparently sees the malloc(3) call.

Then you do not need current `ATTR_MALLOC(free)` as compiler may see `malloc()` call.
It looks inconsistent that function must have two attributes, but uses only one of them.

alejandro-colomar:
> > I think by having this function declared `inline` the compiler should be able to realize about any possible optimizations, since it transparently sees the malloc(3) call.
> 
> Then you do not need current `ATTR_MALLOC(free)` as compiler may see `malloc()` call. It looks inconsistent that function must have two attributes, but uses only one of them.

The `ATTR_MALLOC(free)` attribute is not for optimization, but for safety.  It allows the compiler to warn when we forget to free(3).  There, redundancy increases safety.

Karlson2k (approved):

lib/string/strdup/memdup.h
alejandro-colomar: I've removed `n`, as I expect it to always be 1.

However, in my first implementation, I had meditated about this, and I decided to not protect against overflow for a reason:

This is copying an existing object.  If that object exists, it must have been allocated successfully before (or created anyhow).  If the object exists, the multiplication shouldn't overflow.

lib/string/strdup/memdup.h
alejandro-colomar: Considering that `n` is the real number of elements of the original object we're copying, it couldn't possibly overflow.  The object must exist, and thus have a size that is representable in size_t.

lib/string/strdup/memdup.h
alejandro-colomar: > Should just `NULL` be returned if requested copy of zero items?

Zero-size objects are just fine.  malloc(0) returns non-null.  memdup() of such an object should produce a new zero-sized object.

alejandro-colomar: > Enforcing copy of single item looks dangerous.

memdup() doesn't enfore a single item.  It is MEMDUP() that enforces it.

Do you know of any project that copies a number of elements different than one in any case?  It would be interesting to see when that would be useful.  If so, I'd bring back the `n` argument to the MEMDUP() macro.

lib/string/strdup/memdup.h
alejandro-colomar: Is this a recent addition?  I don't remember why I did that, but I guess it was because some Clang version I tried didn't support [[gnu::malloc(free)]].

Cc: @AaronBallman

alejandro-colomar: > `ATTR_MALLOC(free)` is not enough. This expands to `[[gnu::malloc(free)]]` and means that returned value must be de-allocated by specific function. It must be marked with `[[gnu::malloc]]` as well, which means "returned pointer is unique and does not alias any other pointers". These two forms are independent from each other. See https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-malloc-function-attribute

As I said in another thread, the variant without arguments is not interesting to me.  `inline` gives about as much information to the compiler to optimize.

I want the variant with arguments for safety reasons, where redundancy enforces safety.

Karlson2k: With recent compilers you may use form

(somewhere globally):
```C
#ifdef __has_attribute
#  define SHADOW_CC_HAS_ATTR(attr) __has_attribute(attr)
#else
#  define SHADOW_CC_HAS_ATTR(attr) 0 /* not supported */
#endif
```

And in the file:
```C
#if SHADOW_CC_HAS_ATTR(gnu::malloc)
#  define ATTR_MALLOC(deallocator)     [[gnu::malloc(deallocator)]]
#else
#  define ATTR_MALLOC(deallocator)  /* empty */
#endif
```

See https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html
https://clang.llvm.org/docs/LanguageExtensions.html#has-attribute

The you will need not to track supported compiler versions.

alejandro-colomar: What's the difference between __has_attribute and __has_c_attribute?

alejandro-colomar: Hmmm, I need __has_c_attribute.

On the other hand, we have an issue: because GCC created two attributes with the same name, we can't use __has_c_attribute to determine availability of one of them.

Karlson2k: This can be further improved:
```C
#ifdef __has_c_attribute
#  define SHADOW_CC_HAS_C_ATTR(attr) __has_c_attribute(attr)
#  define SHADOW_C_ATTR(attr)  [[attr]]
#else
#  define SHADOW_CC_HAS_C_ATTR(attr) 0 /* not supported */
#  define SHADOW_C_ATTR(attr)  /* empty */
#endif

#ifdef __has_attribute
#  define SHADOW_CC_HAS_GNU_ATTR(attr) __has_attribute(attr)
#  define SHADOW_GNU_ATTR(attr)  __attribute__ ((attr))
#else
#  define SHADOW_CC_HAS_GNU_ATTR(attr) 0 /* not supported */
#  define SHADOW_GNU_ATTR(attr)  /* empty */
#endif
```

```C
#if SHADOW_CC_HAS_C_ATTR(gnu::malloc)
#  define ATTR_MALLOC(deallocator)     SHADOW_C_ATTR(gnu::malloc(deallocator))
#elif  SHADOW_CC_HAS_GNU_ATTR(malloc)
#  define ATTR_MALLOC(deallocator)     SHADOW_GNU_ATTR(malloc(deallocator))
#else
#  define ATTR_MALLOC(deallocator)  /* empty */
#endif
```

Both GCC and clang always support either both forms `malloc(descructor)` and `malloc` or neither of them. No problem with detecting just one form.

alejandro-colomar: Are you sure?  I seem to remember that my problem was that Clang first implemented the argument-less version, and later the one with arguments.

Which versions of Clang introduced support for the attribute in each form?

Karlson2k: You are right.
Two attributes will be introduced only in clang version 21 or 22.
I usually check for attributes support in "configure" and do not care about versions.

Karlson2k: For GCC it works.
However, GCC first added support for `malloc` attribute (both forms) and significantly later added support for `__has_attribute`

AaronBallman: > What's the difference between __has_attribute and __has_c_attribute?

`__has_attribute` reports back whether the attribute is supported with an `__attribute__` spelling and `__has_c_attribute` reports back whether the attribute is supported with an `[[]]` spelling in C mode.

> Two attributes will be introduced only in clang version 21 or 22.

Clang has supported `__attribute__((malloc))` since at least Clang 3.1. Clang doesn't support the version with an argument in any version of Clang: https://godbolt.org/z/MbPPTe437 (our docs are not doing a good job of making that clear, sorry!)

alejandro-colomar: @AaronBallman 

> > What's the difference between __has_attribute and __has_c_attribute?
> 
> `__has_attribute` reports back whether the attribute is supported with an `__attribute__` spelling and `__has_c_attribute` reports back whether the attribute is supported with an `[[]]` spelling in C mode.

Thanks!

> > Two attributes will be introduced only in clang version 21 or 22.
> 
> Clang has supported `__attribute__((malloc))` since at least Clang 3.1. Clang doesn't support the version with an argument in any version of Clang: https://godbolt.org/z/MbPPTe437 (our docs are not doing a good job of making that clear, sorry!)

If you ever implement the two-argument version, would you mind giving it a different name?  Having two completely different attributes under the same name is too bad.  You could call it `[[gnu::alloc()]]` (or `[[clang::alloc()]]`).  Hopefully, that would give GCC an incentive to fix this issue.

AaronBallman: If we implemented the two argument version, I suspect it would only be for GCC compatibility reasons, so picking a different name wouldn't help much with that. Generally, we try to change the result of what the `__has_` feature test macro returns. e.g., `__has_attribute(malloc) == 1` means the no-argument version and `__has_attribute(malloc) == 2` means the no-argument and the two-argument version. But that's an easy thing to miss for vendor attributes and requires additional coordination between implementations to make it a portable approach.

alejandro-colomar: > If we implemented the two argument version, I suspect it would only be for GCC compatibility reasons,

I use this attribute extensively, as it improves static analysis.  I strongly recommend adding it, not just for compatibility with GCC.

> so picking a different name wouldn't help much with that.

If you'd do that, I'd ask GCC to add an alias for their attribute with a different name, for compatibility with Clang.

> Generally, we try to change the result of what the `__has_` feature test macro returns. e.g., `__has_attribute(malloc) == 1` means the no-argument version and `__has_attribute(malloc) == 2` means the no-argument and the two-argument version. But that's an easy thing to miss for vendor attributes and requires additional coordination between implementations to make it a portable approach.

Hmm, that could work.

Karlson2k:
Hard to review this PR as it has many not fully related changes.
The changes related to `get_current_utmp()` are fine.

alejandro-colomar:
> Hard to review this PR as it has many not fully related changes.

That's because this is queued after another PR.

Here's how it looks in my computer:

```
| * G 1cf10a6c6557 (gh/memdup) lib/utmp.c: get_current_utmp(): Use memdup_T() instead o>
| * G fa7692fd563c lib/utmp.c: get_current_utmp(): Don't exit(3) from library code
| * G e4f14282f695 lib/string/strdup/: memdup{,_T}(): Add APIs
| * G 2b2ad57fb4a4 (gh/T) lib/search/: lsearch_T(): Don't return anything
| * G 8de9d4e4a83e lib/: Rename XCALLOC() => xcalloc_T()
| * G 2d28653a905d lib/: Rename CALLOC() => calloc_T()
| * G 1fc1ebc87dd8 lib/: Rename XMALLOC() => xmalloc_T()
| * G f2a96d60b5ea lib/: Rename MALLOC() => malloc_T()
| * G d9e33b125ea9 lib/: Rename XREALLOC() => xrealloc_T()
| * G be0cd1060289 lib/: Rename REALLOC() => realloc_T()
| * G b11b780c0aad lib/, src/: Rename REALLOCF() => reallocf_T()
| * G 8622877a0981 lib/alloc/: REALLOC[F](): Move _Generic(3) to separate line
| * G 435059189086 lib/: Use compound literals to avoid casts
| * G 666f7a4e958f lib/alloc/: reallocarray[f]_(): Add helper macros to handle n?:1
| * G 8557266d399c lib/{alloc,search}/: Use typeas() to add support for arbitrary types
| * G c8b88d8f3aa0 lib/sizeof.h: typeas(): Add macro
| * G 0bb83a0bc1cf lib/search/: Split APIs
| * G 0ae8aac83929 lib/search/: Simplify CMP()
| * G 1725bb81ec19 lib/search/, lib/, src/: Add a type parameter to the type-safe macros
| * G b4044b01d2c7 lib/: Use a consistent name for macro arguments representing a type >
|/  
```

Once we merge the other branch, this will be small.

> The changes related to `get_current_utmp()` are fine.

Thanks!

alejandro-colomar:
@Karlson2k , this is now ready.  All dependencies have been merged.

Karlson2k:
Looks good!

alejandro-colomar:
Cc: @kees

alejandro-colomar:
> I'm hesitant to approve a new API for `utmp`.

I'd like to insist.  It's a simple API, and I seem to remember having found some other places where we may use them in the future (but need other refactors before I can use it there).

> The project is currently working toward deprecating utmp entirely due to its lack of Y2038 support (see #674). Adding a new interface that is only used in one place adds maintenance overhead for a subsystem that is being phased out

As long as we support utmp, I'd like to make the code readable.  And the maintenance overhead of memdup_T() should be trivial: it's a small thin wrapper, and we have other similar ones.  The improved readability should pay for the maintenance overhead of memdup_T().

ikerexxe:
I'm still wary of adding to a obsolete subsystem, but I won't block on it if you feel strongly. I'll continue the review on the logic itself

lib/utmp.c
```diff
@@ -182,13 +183,8 @@ get_current_utmp(pid_t main_pid)
...
 		ut = ut_by_pid ?: ut_by_line;
 
-	if (NULL != ut) {
-		struct utmpx  *ut_copy;
-
-		ut_copy = xmalloc_T(1, struct utmpx);
-		memcpy(ut_copy, ut, sizeof(*ut));
-		ut = ut_copy;
-	}
+	if (NULL != ut)
```
ikerexxe, line 186: Since you are touching this part of the code move this NULL check to `memdup_T_()`. This will help us avoid any NULL pointer dereference and it will make the code more readable by avoiding to check for NULL every time we call `memdup_T_()`

lib/string/strdup/memdup.h
```diff
@@ -0,0 +1,29 @@
...
+
+
+#include "config.h"
+
+#include <stddef.h>
+
+#include "alloc/malloc.h"
+
+
+// memdup_T - memory duplicate type-safe
```
ikerexxe, line 16: Add documentation explaining this API.  This information could go in this file or in `lib/string/README` as you did for other APIs

lib/utmp.c
ikerexxe, line 171: I think I found another instance where this API can be utilized

ikerexxe, line 178: I think I found another instance where this API can be utilized

lib/utmp.c
alejandro-colomar: I prefer not having null checks in generic APIs unless they really accept a null pointer as input.

The design of memdup() is that it should duplicate some memory that should certainly exist.  It is to memcpy(3) like strdup(3) is to strcpy(3).  For consistency with strdup(3), memdup() should not accept null pointers as input.

alejandro-colomar: By accepting a null pointer, we'd complicate the logic, which itself would complicate static analysis.

lib/string/strdup/memdup.h
alejandro-colomar: Thanks!  Done.  (I'll push soon.)

lib/utmp.c
alejandro-colomar: The problem with those places is that they exit on failure, and replacing them by memdup_T() would require adding error-handling code.

I have other plans for this code:
<https://github.com/shadow-maint/shadow/pull/1296>

In that PR, I'll refactor the logic to not need the duplication at all.
That PR depends on this one, BTW.

alejandro-colomar: Simlarly here, the x is problematic.  I'll get rid of this code soon, anyway.

alejandro-colomar:
I've pushed a version with the README updated.  Cc: @ikerexxe

lib/utmp.c
ikerexxe: Ok, that makes sense

ikerexxe (approved):
LGTM

PR #1297, opened 2025-07-16
lib/utmp.c: prepare_utmp(): Fix buffer overrun when 'line' isn't "tty*"

alejandro-colomar:
Sometimes, line may be a string of length less than 3, in which case the
    +3 was jumping somewhere after the terminating null byte.  This is
    probably not easy to exploit, as we only read at most 4 bytes, but this
    was theoretically quite bad.

Reported-by: @Karlson2k

### DIFF
--- a/lib/utmp.c
+++ b/lib/utmp.c
@@ -271,12 +271,11 @@ prepare_utmp(const char *name, const char *line, const char *host,
 	utent->ut_type = USER_PROCESS;
 	utent->ut_pid = getpid ();
 	STRNCPY(utent->ut_line, line);
-	if (NULL != ut) {
+	if (NULL != ut)
 		STRNCPY(utent->ut_id, ut->ut_id);
-	} else {
-		/* XXX - assumes /dev/tty?? */
-		STRNCPY(utent->ut_id, line + 3);
-	}
+	else
+		STRNCPY(utent->ut_id, strprefix(line, "tty") ?: line);
+
 #if defined(HAVE_STRUCT_UTMPX_UT_NAME)
 	STRNCPY(utent->ut_name, name);
 #endif


alejandro-colomar:
@Karlson2k  Do you have a system where this bug is reproducible?

Karlson2k:
> @Karlson2k Do you have a system where this bug is reproducible?

No.
It is a hypothetical possibility.
However, it is not very hard to modify other software to use shorter strings, but this assumes that user already has full control on the system.

It would be nice to pick my [patch](https://github.com/shadow-maint/shadow/pull/1293) for this issue.

alejandro-colomar:
> > @Karlson2k Do you have a system where this bug is reproducible?
> 
> No. It is a hypothetical possibility. However, it is not very hard to modify other software to use shorter strings, but this assumes that user already has full control on the system.
> 
> It would be nice to pick my [patch](https://github.com/shadow-maint/shadow/pull/1293) for this issue.

If you write it with `?:`, I'd pick your patch.  I'm just opposed to being excessively verbose where `?:` will just work.

Also, please open a separate PR for the second patch.  We can only merge via github, which doesn't allow picking.

Karlson2k:
> If you write it with `?:`, I'd pick your patch. I'm just opposed to being excessively verbose where `?:` will just work.
> 
> Also, please open a separate PR for the second patch. We can only merge via github, which doesn't allow picking.

Done.
Regarding merging: manual cherry picking locally + manual PR closing always works. :wink:

alejandro-colomar:
> > If you write it with `?:`, I'd pick your patch. I'm just opposed to being excessively verbose where `?:` will just work.
> > Also, please open a separate PR for the second patch. We can only merge via github, which doesn't allow picking.
> 
> Done. Regarding merging: manual cherry picking locally + manual PR closing always works. 😉

Nope; we can't push manually to `master`.  The branch is protected.  (We could change that, but we'd need to agree on that.)

alejandro-colomar:
Merged @Karlson2k 's equivalent patch in #1301 .

PR #1148, opened 2024-12-08
seprintf() and stpecpy() improvements.

alejandro-colomar:
Cc: @uecker 

---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase

```
$ git range-diff alx/master..gh/snprintf_ master..snprintf_ 
1:  6a656139 = 1:  99aef5b4 lib/subordinateio.c: get_owner_id(): Use snprintf_() instead of its pattern
```
</details>

<details>
<summary>v1c</summary>

-  Rebase

```
$ git range-diff alx/master..gh/snprintf_ master..snprintf_ 
1:  99aef5b4 = 1:  1fad167c lib/subordinateio.c: get_owner_id(): Use snprintf_() instead of its pattern
```
</details>

<details>
<summary>v1d</summary>

-  Rebase

```
$ git range-diff master..gh/snprintf_ shadow/master..snprintf_ 
1:  1fad167c = 1:  242ba524 lib/subordinateio.c: get_owner_id(): Use snprintf_() instead of its pattern
```
</details>

<details>
<summary>v1e</summary>

-  Rebase

```
$ git range-diff master..gh/snprintf_ shadow/master..snprintf_ 
1:  242ba524 = 1:  1c546325 lib/subordinateio.c: get_owner_id(): Use snprintf_() instead of its pattern
```
</details>

<details>
<summary>v2</summary>

-  Redesign stpecpy() and stpeprintf().

```
$ git range-diff shadow/master gh/snprintf_ snprintf_ 
1:  1c546325 = 1:  1c546325 lib/subordinateio.c: get_owner_id(): Use snprintf_() instead of its pattern
-:  -------- > 2:  edda7ae9 lib/string/: Redesign stpecpy() and stpeprintf()
```
</details>

<details>
<summary>v3</summary>

-  Also set errno from snprintf_(), and rename it to stprintf().

```
$ git range-diff master gh/snprintf_ snprintf_ 
1:  1c546325 = 1:  1c546325 lib/subordinateio.c: get_owner_id(): Use snprintf_() instead of its pattern
-:  -------- > 2:  593e767b lib/string/sprintf/snprintf.h: vsnprintf_(): Set errno = E2BIG on truncation
-:  -------- > 3:  8a5852a9 lib/string/sprintf/: snprintf_(): Use int instead of size_t in $2
2:  edda7ae9 = 4:  62d3f05c lib/string/: Redesign stpecpy() and stpeprintf()
[...]

PR #1300, opened 2025-07-17
lib/string/: stpecpy(), stpeprintf(): Return NULL (and set E2BIG) on truncation

alejandro-colomar:
This makes it easier to check truncation.


alejandro-colomar:
#1148 instead.

PR #1301, opened 2025-07-17
lib/utmp.c: Fixed generated strings for "ut_id"

Karlson2k:
When no "ut_id" is provided by previous utmp file entry, "login" must generate its own "ut_id". Historically tty number  was used for it. However, if current device name is three characters or shorter, then empty "ut_id" is produced or even garbage is copied from uninitialised part of the "line".
This commit fixes it.
This patch uses unportable C extension Elvis operator as it is already used everywhere in the code.

Extracted from #1293


alejandro-colomar (approved):
Thanks!

```
```

PR #1303, opened 2025-07-18
While-loop readability improvements.

alejandro-colomar:
---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase

```
$ git rd
1:  49e5b01a ! 1:  682e6ae8 lib/, src/: Reorder while() conditions for safety
    @@ Metadata
      ## Commit message ##
         lib/, src/: Reorder while() conditions for safety
     
    -    In conditions that check for NULL after assignment, it's safer to put
    -    the NULL check first, as a mistake would result in a compiler error,
    -    as opposed to unwanted implicit conversions.  It's also more readable.
    +    In conditions that perform simple assignment (=) before comparison,
    +    it's safer to put the comparison first, as a mistake would result in a
    +    compiler error, as opposed to assigning something incorrect.
    +    It's also more readable, IMO.
     
```
</details>

### DIFF
--- a/lib/copydir.c
+++ b/lib/copydir.c
@@ -310,7 +310,7 @@ static int copy_tree_impl (const struct path_info *src, const struct path_info *
 		dst_orig = dst->full_path;
 		set_orig = true;
 	}
-	while ((0 == err) && (ent = readdir (dir)) != NULL) {
+	while (0 == err && NULL != (ent = readdir(dir))) {
 		char              *src_name = NULL;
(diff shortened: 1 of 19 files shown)


hallyn (approved):
You're showing your prejudice against left-handed people :)

kidding.

alejandro-colomar:
> You're showing your prejudice against left-handed people :)
> 
> kidding.

:)

PR #1305, opened 2025-07-18
Fixes related to use of forked PID

Karlson2k:
Number of fixes around using PID of forked or parent process.

A subset of commits from #1292, as requested [here](https://github.com/shadow-maint/shadow/pull/1292#discussion_r2215911542).

PR #1306, opened 2025-07-18
Refactor utmp entry search

Karlson2k:
Readability improvements only, without functional changes.

A subset from #1305 

@alejandro-colomar

### DIFF
--- a/lib/utmp.c
+++ b/lib/utmp.c
@@ -158,11 +158,13 @@ get_current_utmp(void)
 
 	/* First, try to find a valid utmp entry for this process.  */
 	while ((ut = getutxent()) != NULL) {
+		if (   (LOGIN_PROCESS != ut->ut_type)
+		    && (USER_PROCESS  != ut->ut_type))
+			continue;
+
 		if (   (   (ut->ut_pid == getpid ())
 		        || (ut->ut_pid == getppid ()))
 		    && ('\0' != ut->ut_id[0])
-		    && (   (LOGIN_PROCESS == ut->ut_type)
-		        || (USER_PROCESS  == ut->ut_type))
 		    /* A process may have failed to close an entry
 		     * Check if this entry refers to the current tty */
 		    && is_my_tty(ut->ut_line))
--- a/lib/utmp.c
+++ b/lib/utmp.c
@@ -152,7 +152,6 @@ static /*@null@*/ /*@only@*/struct utmpx *
 get_current_utmp(void)
 {
 	struct utmpx  *ut;
-	struct utmpx  *ret = NULL;
 
 	setutxent();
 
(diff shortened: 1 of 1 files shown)


lib/utmp.c
```diff
@@ -158,11 +158,13 @@ get_current_utmp(void)
 
 	/* First, try to find a valid utmp entry for this process.  */
 	while ((ut = getutxent()) != NULL) {
+		if (   (LOGIN_PROCESS != ut->ut_type)
+		    && (USER_PROCESS  != ut->ut_type))
+			continue; /* Look for 'login' or 'user' entries only */
```
alejandro-colomar: Please remove this comment, as it seems redundant with the code.

lib/utmp.c
```diff
@@ -151,8 +151,7 @@ failtmp(const char *username, const struct utmpx *failent)
 static /*@null@*/ /*@only@*/struct utmpx *
 get_current_utmp(void)
 {
-	struct utmpx  *ut;
-	struct utmpx  *ret = NULL;
+	struct utmpx *ut;
```
alejandro-colomar: Please keep the two spaces after the base type.

lib/utmp.c
```diff
@@ -174,13 +173,15 @@ get_current_utmp(void)
 	}
 
 	if (NULL != ut) {
-		ret = XMALLOC(1, struct utmpx);
-		memcpy (ret, ut, sizeof (*ret));
+		struct utmpx *ut_copy;
```
alejandro-colomar: Please use two spaces after the base type.

alejandro-colomar: Also, a blank after it (before the code) would be nice.  (Although, since I'm going to replace that soon with MEMDUP(), I don't mind too much about this block of code.)

lib/utmp.c
Karlson2k: > Please use two spaces after the base type.

Never saw such requirements.
Could you be specific?
When two spaces after base type should be used in variables definitions?
When point is used?
When base type is aggregate?
When both (pointer to an aggregate type)?

alejandro-colomar: I inherited those rules from NGINX Unit, which itself has rules similar to NGINX.

The NGINX rules are documented here: <https://nginx.org/en/docs/dev/development_guide.html#code_style>
(Search for 'Variables'.)

The difference in the NGINX Unit project (and shadow) is that the `*` is aligned as if it were part of the name

alejandro-colomar:
Merged as part of #1305

lib/utmp.c
Karlson2k: It would be nice to copy at least this short explanation to some file/document.

alejandro-colomar: Yup, eventually.

PR #1309, opened 2025-07-22
lib/: Use simple assignment instead of memcpy(3)

alejandro-colomar:
memcpy(3) is overkill, and much more dangerous than simple assignment. Simple assignment adds type safety, and removes any possibility of specifying a wrong size.


lib/utmp.c
```diff
@@ -318,16 +318,10 @@ prepare_utmp(const char *name, const char *line, const char *host,
 				struct sockaddr_in *sa =
 					(struct sockaddr_in *) info->ai_addr;
 # if defined(HAVE_STRUCT_UTMPX_UT_ADDR)
-				memcpy (&(utent->ut_addr),
-				        &(sa->sin_addr),
-				        MIN(sizeof(utent->ut_addr),
-				            sizeof(sa->sin_addr)));
+				utent->ut_addr = sa->sin_addr.s_addr;
```
hallyn, line 321: Looked over this for quite awhile.  The UTMPX_UT_ADDR{,_V6} #defines really make this uglier...  Are these required to support musl?

Anyway, I agree it's an improvement, thanks.

alejandro-colomar: I think I have some patch removing some of those.  I'll check.

alejandro-colomar: Thank you!  You've merged quite a lot of patches today.  :-)

PR #1317, opened 2025-07-31
Make sure that logind is enabled if requested, make --enable-logind/--disable-logind mandatory

Karlson2k:
This is alternative version of #1283

This PR:
* Fix silent disable of logind if `--enable-logind` is requested, but libsystemd is missing
* Make mandatory one of --enable-logind/--disable-logind configure parameters. If none is specified, then `configure` stops with error and asks user to provide one of them.

@alejandro-colomar The code does not assume anything. It asks user to specify explicitly, whether logind is needed or not.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -18,6 +18,7 @@ CFLAGS="$CFLAGS -Wno-unknown-warning-option"
 "$(dirname "$0")"/configure \
 	CFLAGS="$CFLAGS" \
 	--enable-lastlog \
+	--disable-logind \
 	--enable-man \
 	--enable-maintainer-mode \
 	--enable-shared \
--- a/configure.ac
+++ b/configure.ac
@@ -153,7 +153,7 @@ AC_ARG_ENABLE([lastlog],
 
 AC_ARG_ENABLE([logind],
 	[AS_HELP_STRING([--enable-logind],
-		[enable logind @<:@default=yes if found@:>@])],
+		[enable logind support])],
 	[
 		AS_CASE([${enableval}],
 			[yes],[],
@@ -161,7 +161,9 @@ AC_ARG_ENABLE([logind],
 			[AC_MSG_ERROR([bad parameter value for --enable-logind=${enableval}])]
 		)
 	],
-	[enable_logind="yes"]
+	[
+		AC_MSG_ERROR([neither --enable-logind nor --disable-logind specified. One of these two parameters must be set explicitly.])
+	]
 )
 
 AC_ARG_WITH([audit],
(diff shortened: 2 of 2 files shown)


alejandro-colomar:
>* Make mandatory one of --enable-logind/--disable-logind configure parameters. If none is specified, then `configure` stops with error and asks user to provide one of them.

I don't like this.  A default behavior for the most common case would be appropriate.

Karlson2k:
> > * Make mandatory one of --enable-logind/--disable-logind configure parameters. If none is specified, then `configure` stops with error and asks user to provide one of them.
> 
> I don't like this. A default behavior for the most common case would be appropriate.

I thought [you like the opposite](https://github.com/shadow-maint/shadow/pull/1283#issuecomment-3139181827).

alejandro-colomar:
> > > * Make mandatory one of --enable-logind/--disable-logind configure parameters. If none is specified, then `configure` stops with error and asks user to provide one of them.
> > 
> > 
> > I don't like this. A default behavior for the most common case would be appropriate.
> 
> I thought [you like the opposite](https://github.com/shadow-maint/shadow/pull/1283#issuecomment-3139181827).

That's not the opposite.  One thing is automagic use of dependencies where available (or automagic guessing of the user in general), which I don't like.  Another is having defaults for user decisions; this is reasonable and good.  As long as the defaults are hardcoded, the user can expect them to be consistent.

lslebodn:
> One thing is automagic use of dependencies where available (or automagic guessing of the user in general), which I don't like.

Fully agree; 

>  Another is having defaults for user decisions; this is reasonable and good. As long as the defaults are hardcoded, the user can expect them to be consistent.

@alejandro-colomar What does it mean in this case? Are you fine with `enable_logind="yes"` by default?
or we want to switch default to `enable_logind="no"` and let maintainer to explicitly enable support for logind with `--enable-logind`

alejandro-colomar:
> > Another is having defaults for user decisions; this is reasonable and good. As long as the defaults are hardcoded, the user can expect them to be consistent.
> 
> @alejandro-colomar What does it mean in this case? Are you fine with `enable_logind="yes"` by default? or we want to switch default to `enable_logind="no"` and let maintainer to explicitly enable support for logind with `--enable-logind`

As co-maintainer, I don't care about which default value we want.  Both are arguably valid.  Maybe as co-maintainer I would maintain the status-quo, unless there are reasons to change it.

If you ask me, I would choose the one that works in non-systemd, just because I don't use systemd, but that's just my opinion as user, not as co-maintainer.

Karlson2k:
I'll update the PR with default "yes", just because the absence will be detected and user will be asked to set the parameter or install library. 
With default "no" user will never be asked for anything. 
@alejandro-colomar ok?

alejandro-colomar:
> I'll update the PR with default "yes", just because the absence will be detected and user will be asked to set the parameter or install library. With default "no" user will never be asked for anything. @alejandro-colomar ok?

Yes, that sounds reasonable.  Thanks!

Karlson2k:
The second alternative version is [here](https://github.com/shadow-maint/shadow/pull/1318).

alejandro-colomar:
Merged <https://github.com/shadow-maint/shadow/pull/1318> instead.

PR #1319, opened 2025-08-02
configure.ac: Propagate ./configure flags to 'distcheck'

alejandro-colomar:
'make distcheck' runs ./configure (among other things).  That command should inherit the flags passed on the command line, as they are the flags necessary to build in the current system.

This also allows testing different configurations in the same system. Until now, we only tested the default automagic configuration.  With this change, one can test the same default automagic configuration, by not passing any flags to ./configure, or they can test more specific configurations, by passing flags.

Cc: @lslebodn, @Karlson2k


alejandro-colomar:
Does this allow any further clanup?  (Such as not needing to install dependencies on CI?)

Karlson2k:
There is a missing part.
You need to replace the AM_DISTCHECK_CONFIGURE_FLAGS assignment in the top-level Makefile.ac:
```automake
AM_DISTCHECK_CONFIGURE_FLAGS = @ac_configure_args@
```

Edit: No, you used AM_DISTCHECK_CONFIGURE_FLAGS directly.
Then you need to remove AM_DISTCHECK_CONFIGURE_FLAGS assignment from the top-level Makefile.ac

Karlson2k:
> Does this allow any further clanup? (Such as not needing to install dependencies on CI?)

I'd do the opposite: remove "--disable-logind" from Ubuntu -based CI. Then it will be build-tested with `logind` integration, which is the most common scenario currently.

Karlson2k:
The code is valid.

However, for the records, I still think that automatic "configure" flags should be enough to build the package if minimal required dependencies are available.

lslebodn:
> I still think that automatic "configure" flags should be enough to build the package
 if minimal required dependencies are available.
 
I fully agree with 1st part, not about `minimal required dependencies`
 
> This also allows testing different configurations in the same system. Until now, we only tested the default automagic configuration. 

A nitpick; we do not need `make distcheck` for testing different configuration.
`git clean -fdx && autoreconf -if && ./configure ${parametes here}  && make check` is enough.
and moreover testing would also depend on installed dependencies.

Anyway, `make dist(check)` is mostly for maintainer to ensure all necessary files will be in tarball before the release.
So users can run standard commands
* tar xzf <filename>.tar.gz
* cd <directory>
* ./configure #parameters here
* make # with BUILD/LINK flags
* sudo make install

And because it is meant for maintainer you should ask other maintainers what kind of preferences they have.
I do not have any preference about this PR.

Karlson2k:
> Anyway, `make dist(check)` is mostly for maintainer to ensure all necessary files will be in tarball before the release. So users can run standard commands

Right. `make distcheck`, `make installcheck` are for maintainers. To test for possible breakage. 
> 
> * tar xzf .tar.gz
> * cd
> * ./configure #parameters here
> * make # with BUILD/LINK flags
> * sudo make install

A small correction.
CC, CFLAGS, LDFLAGS etc. should be used with configure for better detection.
`./configure --some-parsms CC=somecc CFLAGS="--fsomeflags" && make && sudo make install`

alejandro-colomar:
Hi,

On Sat, Aug 02, 2025 at 01:12:24PM +0000, lslebodn wrote:
> lslebodn left a comment (shadow-maint/shadow#1319)
> 
> > I still think that automatic "configure" flags should be enough to build the package
>  if minimal required dependencies are available.
>  
> I fully agree with 1st part, not about `minimal required dependencies`
>  
> > This also allows testing different configurations in the same system. Until now, we only tested the default automagic configuration. 
> 
> A nitpick; we do not need `make distcheck` for testing different configuration.
> `git clean -fdx && autoreconf -if && ./configure ${parametes here}  && make check` is enough.

make distcheck includes more testing, such as installcheck (I don't know
if we have that here).

But yeah, we could do that similarly with installcheck.

On the other hand, 'distcheck' is a test-everything, so I often do
'make distcheck' as a way to test everything, instead of running the
separate commands.  It's handy.

> and moreover testing would also depend on installed dependencies.

That has always been true.

> Anyway, `make dist(check)` is mostly for maintainer to ensure all necessary files will be in tarball before the release.
> So users can run standard commands
> * tar xzf <filename>.tar.gz
> * cd <directory>
> * ./configure
> * make
> * sudo make install
> 
> And because it is meant for maintainer you should ask other maintainer what kind of preferences they have.

Yup, they'll have to merge it, anyway.  :)

> I do not have any preference about this PR.

Thanks!


Have a lovely day!
Alex

-- 

Karlson2k:
Just as a reminder: the line `AC_SUBST([AM_DISTCHECK_CONFIGURE_FLAGS], ["$ac_configure_args"])` will add `AM_DISTCHECK_CONFIGURE_FLAGS` to every Makefile generated by `configure`, while the top level Makefile is enough.

Alternatively, as I [mentioned earlier](https://github.com/shadow-maint/shadow/pull/1318#issuecomment-3145943055), a purer solution would be lines in configure.ac:
```autoconf
AC_SUBST([ac_configure_args])
AM_SUBST_NOTMAKE([ac_configure_args])
```
and in the top Makefile:
```automake
AM_DISTCHECK_CONFIGURE_FLAGS = @ac_configure_args@
```
This will add a single line only to the top Makefile.

However, if a pure Makefiles are not a goal (which makes sense as they are clattered anyway), the current version will work fine.

alejandro-colomar:
> Just as a reminder: the line `AC_SUBST([AM_DISTCHECK_CONFIGURE_FLAGS], ["$ac_configure_args"])` will add `AM_DISTCHECK_CONFIGURE_FLAGS` to every Makefile generated by `configure`, while the top level Makefile is enough.
> 
> Alternatively, as I [mentioned earlier](https://github.com/shadow-maint/shadow/pull/1318#issuecomment-3145943055), a purer solution would be lines in configure.ac:
> 
> ```m4
> AC_SUBST([ac_configure_args])
> AM_SUBST_NOTMAKE([ac_configure_args])
> ```
> 
> and in the top Makefile:
> 
> ```
> AM_DISTCHECK_CONFIGURE_FLAGS = @ac_configure_args@
> ```
> 
> This will add a single line only to the top Makefile.
> 
> However, if a pure Makefiles are not a goal (which makes sense as they are clattered anyway), the current version will work fine.

Since makefiles are generated, I don't mind how much crap they contain. I prefer a slim config file.

hallyn (approved):

PR #1322, opened 2025-08-09
lib/, src/: Remove dead calls to umask(2)

alejandro-colomar:
We're calling [f]chmod(2) right afterwards to set the mode, so it makes aboslutely no sense to set a umask(2) for calling fopen(3).

Not only it is dead code, it might even be dangerous, by creating a window of time where anyone can write to these files.

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -245,11 +245,8 @@ static /*@null@*/ /*@dependent@*/FILE *fopen_set_perms (
 	const struct stat *sb)
 {
 	FILE *fp;
-	mode_t mask;
 
-	mask = umask (0777);
 	fp = fopen (name, mode);
-	(void) umask (mask);
 	if (NULL == fp) {
 		return NULL;
 	}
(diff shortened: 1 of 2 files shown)


hallyn:
Hm, I disagree: by setting umask to 0777, we ensure the files have 000 perms to begin with.  If we undo that, then we *introduce* a race until we chmod.  Right?

Now, this is nevertheless old code.  If we want to improve it, how about passing the mode to open(), and then using fdopen to get the FILE*?

alejandro-colomar:
> Hm, I disagree: by setting umask to 0777, we ensure the files have 000 perms to begin with. If we undo that, then we _introduce_ a race until we chmod. Right?

Ahhh, I'm not used to using umask(2).  I was reading it as having 777 perms, instead of umask.  You're right, indeed.

> 
> Now, this is nevertheless old code. If we want to improve it, how about passing the mode to open(), and then using fdopen to get the FILE*?

Yup, could make sense.  I'll think about it.

PR #1323, opened 2025-08-09
lib/commonio.c: create_backup(): Remove dead call to fseek(3)

alejandro-colomar:
fopen(, "w") already guarantees that the stream is positioned at the beginning of the file.

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -289,13 +289,12 @@ static int create_backup (const char *backup, FILE * fp)
 
 	/* TODO: faster copy, not one-char-at-a-time.  --marekm */
 	c = 0;
-	if (fseek (fp, 0, SEEK_SET) == 0) {
-		while ((c = getc (fp)) != EOF) {
-			if (putc (c, bkfp) == EOF) {
-				break;
-			}
+	while ((c = getc (fp)) != EOF) {
+		if (putc (c, bkfp) == EOF) {
+			break;
 		}
 	}
+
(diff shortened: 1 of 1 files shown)


hallyn:
> fopen(, "w") already guarantees that the stream is positioned at the beginning of the file.

Sure, but there is no guarantee that this fp was just opened.

So this patch appears wrong.

(Please re-open if I'm misunderstanding)

alejandro-colomar:
There's the line

```c
bkfp = fopen_set_perms (backup, "w", &sb);
```

just above.  Unless I'm misreading something, that unconditionally does an `fopen(, "w");`.

```c
$ sed -n '242,+30p' lib/commonio.c
static /*@null@*/ /*@dependent@*/FILE *fopen_set_perms (
	const char *name,
	const char *mode,
	const struct stat *sb)
{
	FILE *fp;
	mode_t mask;

	mask = umask (0777);
	fp = fopen (name, mode);
	(void) umask (mask);
	if (NULL == fp) {
		return NULL;
	}

	if (fchown (fileno (fp), sb->st_uid, sb->st_gid) != 0) {
		goto fail;
	}
	if (fchmod (fileno (fp), sb->st_mode & 0664) != 0) {
		goto fail;
	}

	return fp;

      fail:
	(void) fclose (fp);
	/* fopen_set_perms is used for intermediate files */
	(void) unlink (name);
	return NULL;
}

```

alejandro-colomar:
Ahhhh, sorry!  I was thinking we were seeking the backup, but the seek is on the original file.  Indeed, I misread.

PR #1324, opened 2025-08-09
lib/commonio.c: Parenthesize calls that look like libc APIs

alejandro-colomar:
libc is allowed to provide any functions as macros (7.1.4p1 in C23).

This means that libc is allowed to provide a free(3) macro, which could look like

	#define free(p)  __free(p)

And that would be expanded by the preprocessor in our code, turning our structure members into some code that won't work (or even worse, it might misbehave).

So, fix this undefined behavior.

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -328,9 +328,8 @@ static void free_linked_list (struct commonio_db *db)
 
 		free (p->line);
 
-		if (NULL != p->eptr) {
-			db->ops->free (p->eptr);
-		}
+		if (NULL != p->eptr)
+			(db->ops->free)(p->eptr);
 
 		free (p);
 	}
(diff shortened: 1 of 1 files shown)


hallyn:
Does any libc do this?

alejandro-colomar:
> Does any libc do this?

I'm experimenting with an `_Optional` qualifier that is being debated within the C Committee.
<https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3422.pdf>

I've installed the following header (written by the committee member that authored the proposal of `_Optional`) in my `/usr/local/include/`:
<https://github.com/chrisbazley/CBUtilLib/blob/main/Optional.h>

It wraps libc APIs to add the qualifier where appropriate.  And it indeed does this by necessity.

Cc: @chrisbazley

alejandro-colomar:
@hallyn 

How about changing the names of the structure members to be prefixed with `io_` (or something else), similar to `struct tm` members being `tm_*`?

PR #1325, opened 2025-08-09
lib/commonio.h: Use [[gnu::copy()]] and typeof() for function pointers compatible with libc

alejandro-colomar:


alejandro-colomar:
We'll remove these structure members in the other PR, so it doesn't make sense to improve them.  I've closed this.

PR #1326, opened 2025-08-10
lib/utmp.c: Fix use of last utmp entry instead of patrial-match entry

Karlson2k:
This is a correction for PR #1292.


Karlson2k:
@ikerexxe Please, take a look on it.

lib/utmp.c
```diff
@@ -162,33 +162,34 @@ get_current_utmp(pid_t main_pid)
...
+			has_tty_match = is_my_tty(ut->ut_line);
+
+			if (NULL == ut_copy) {
+				ut_copy = XMALLOC(1, struct utmpx);
+				*ut_copy = *ut;
+			} else if (   has_tty_match
+			           || ut_copy_is_by_line_only)
+				*ut_copy = *ut;
+
+			if (has_tty_match)
```
alejandro-colomar: Please separate refactors from the actual fix into two separate commits.

lib/utmp.c
```diff
@@ -162,33 +162,34 @@ get_current_utmp(pid_t main_pid)
...
-		ut = ut_by_pid ?: ut_by_line;
-
-	if (NULL != ut) {
-		struct utmpx  *ut_copy;
-
-		ut_copy = XMALLOC(1, struct utmpx);
-		memcpy(ut_copy, ut, sizeof(*ut));
-		ut = ut_copy;
-	}
-
```
alejandro-colomar: I'm thinking it might be better to first revert to a known-good state.  I'm not happy fixing fixes.

Karlson2k: It was known-bad state before.
The behaviour in the situation when old code worked correctly has not been changed. The current code works the same.

lib/utmp.c
Karlson2k: It could be hard as deep copy needs more allocations so it would be refactors anyway.
However, I may try to fully reproduce the old behaviour as the first commit.

Karlson2k: I wasn't really hard. Updated.

Karlson2k:
Updated with separated fix and refactors.

The fix is really simple. Can be used alone, without refactors.

lib/utmp.c
Karlson2k: To make it clearer, I may add a revert as the first commit.

Karlson2k:
Added an alternative version #1327 with full revert as the first commit.

lib/utmp.c
Karlson2k: Added as an [alternative PR](https://github.com/shadow-maint/shadow/pull/1327).

lib/utmp.c
```diff
@@ -165,13 +165,16 @@ get_current_utmp(pid_t main_pid)
 			if (is_my_tty(ut->ut_line))
 				break; /* Perfect match, stop the search */
 
-			if (NULL == ut_by_pid)
-				ut_by_pid = ut;
+			if (NULL == ut_by_pid) {
+				ut_by_pid = XMALLOC(1, struct utmpx);
+				*ut_by_pid = *ut;
+			}
```
alejandro-colomar, line 171: Okay, this seems much more reasonable as a patch.  We can live with it without reverting.

Karlson2k: Yes, this is much easier to read.

lib/utmp.c
```diff
@@ -150,8 +150,8 @@ static /*@null@*/ /*@only@*/struct utmpx *
 get_current_utmp(pid_t main_pid)
 {
 	struct utmpx  *ut;
-	struct utmpx  *ut_by_pid  = NULL;
-	struct utmpx  *ut_by_line = NULL;
+	struct utmpx  *ut_copy = NULL;
+	bool          ut_copy_is_by_line_only = false;
```
alejandro-colomar: Please keep the optimization for a separate PR.  I'll take the fix, as it seems an easy one.  However, the optimization seems to add non-trivial complexity to the code, and I prefer simpler and slower code, in general, as long as the "slowness" is controlled.

lib/utmp.c
Karlson2k: Maybe replace `*ut_by_pid = *ut;` with `memcpy(ut_by_pid, ut, sizeof(*ut));` for consistency?
The second form is used below, when making a full copy.

lib/utmp.c
alejandro-colomar: Thanks a lot!

lib/utmp.c
Karlson2k: OK, I'll put it to the separate PR.
The other drawback that this version uses more allocation, however it should not be a problem as they are small.

lib/utmp.c
alejandro-colomar: Nah, in fact I have a PR for moving the other way around: transforming some memcpy(3) calls to simple assignment.
<https://github.com/shadow-maint/shadow/pull/1309>

In this case, I'll use MEMDUP() once we merge <https://github.com/shadow-maint/shadow/pull/1296>.

lib/utmp.c
Karlson2k: A separate PR https://github.com/shadow-maint/shadow/pull/1328 with optimisations added.

lib/utmp.c
Karlson2k: OK, I will not change it as you are going to replace with your new macro anyway.

lib/utmp.c
```diff
@@ -186,6 +189,8 @@ get_current_utmp(pid_t main_pid)
 		ut = ut_copy;
 	}
 
+	free (ut_by_line);
+	free (ut_by_pid);
```
alejandro-colomar: Please remove the white spaces.

Karlson2k: Done.

Karlson2k:
Thank you!
Sorry for missing this in the initial PR.
I'm glad that I spotted it before the next release.

alejandro-colomar:
> Thank you! Sorry for missing this in the initial PR. I'm glad that I spotted it before the next release.

No problem!  Mistakes happen to everyone.  :)

PR #1327, opened 2025-08-10
Utmp entry correction alt1

Karlson2k:
This is a correction for PR #1292.

This is an alternative version of PR #1326. This PR includes a full revert  as the first commit.

Last two commits could be squashed into one commit.


alejandro-colomar:
We merged the alternative implementation without reverting.

PR #1328, opened 2025-08-10
lib/utmp.c: Fix use of last utmp entry instead of patrial-match entry + Optimisations

Karlson2k:
This is a correction for PR #1292 (and is an alternative version of PR https://github.com/shadow-maint/shadow/pull/1326; this PR includes some optimisations.)

### DIFF
--- a/lib/utmp.c
+++ b/lib/utmp.c
@@ -150,8 +150,8 @@ static /*@null@*/ /*@only@*/struct utmpx *
 get_current_utmp(pid_t main_pid)
 {
 	struct utmpx  *ut;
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
You'll need to rebase.

alejandro-colomar:
But anyway, I think I'll say no.  It makes the code unreadable, so unless you come up with an idea that makes it readable, I don't think we'll merge it.

Karlson2k:
Rebased.

I agree that it is hard to read this code.
However, let's decide something first with [this idea](https://github.com/shadow-maint/shadow/pull/1329#issuecomment-3172668202), which can notable change the code.

PR #1290, opened 2025-07-14
Clean up lib/utmp.c and lib/logind.c

alejandro-colomar:
Cc: @Karlson2k , @ikerexxe 

---

Revisions:

<details>
<summary>v2</summary>

-  Use ENOSYS instead of ENOTSUP  [@ikerexxe ]

```
$ git range-diff shadow/master gh/logind_utmp logind_utmp 
 1:  bfb57d2b =  1:  bfb57d2b lib/logind.c: Fix style
 2:  738ee5d9 =  2:  738ee5d9 lib/logind.c: Remove dead initializations
 3:  00a50759 =  3:  00a50759 lib/: Move utmp.c and logind.c prototypes to new "session_management.h" header
 4:  d3ae3cf1 =  4:  d3ae3cf1 lib/logind.c: get_session_host(): free() earlier, to avoid a goto
 5:  98425add =  5:  98425add lib/logind.c: Wrap libsystemd function in our wrapper get_session()
 6:  7c58ee42 =  6:  7c58ee42 lib/logind.c: Wrap libsystemd function in our wrapper session_get_remote_host()
 7:  7f0fee3e =  7:  7f0fee3e lib/utmp.c: get_session_host(): Fix memory leak
 8:  99d5669b !  8:  c47b572f lib/, src/: get_session_host(): Return the host
    @@ lib/utmp.c: get_current_utmp(void)
      #else
     -  *out = NULL;
     -  ret = -2;
    -+  errno = ENOTSUP;
    ++  errno = ENOSYS;
      #endif
     +  host = NULL;
     +out:
 9:  528997fb =  9:  4fcf0a3d lib/string/strdup/: STRNDUP(): Add macro
10:  2361b195 = 10:  f6b6ec9d lib/utmp.c: get_session_host(): Don't exit from library code
```
</details>

<details>
<summary>v3</summary>

-  Avoid a goto [@hallyn ]

```
$ git range-diff shadow/master gh/logind_utmp logind_utmp 
 1:  bfb57d2b =  1:  bfb57d2b lib/logind.c: Fix style
 2:  738ee5d9 =  2:  738ee5d9 lib/logind.c: Remove dead initializations
 3:  00a50759 =  3:  00a50759 lib/: Move utmp.c and logind.c prototypes to new "session_management.h" header
 4:  d3ae3cf1 =  4:  d3ae3cf1 lib/logind.c: get_session_host(): free() earlier, to avoid a goto
[...]

PR #1329, opened 2025-08-10
Fixed memory leak in UTMP processing

Karlson2k:
Obvious bug. Nothing to comment.

### DIFF
--- a/lib/utmp.c
+++ b/lib/utmp.c
@@ -21,6 +21,7 @@
 #include <sys/socket.h>
 #include <netdb.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <fcntl.h>
 
--- a/lib/utmp.c
+++ b/lib/utmp.c
@@ -26,6 +26,7 @@
 
 #include "alloc/x/xcalloc.h"
 #include "alloc/x/xmalloc.h"
+#include "attr.h"
 #include "sizeof.h"
 #include "string/strchr/strnul.h"
 #include "string/strcmp/streq.h"
@@ -146,6 +147,7 @@ err_close:
  *	Return NULL if no entries exist in utmp for the current process or
  *	            there is an error reading utmp.
  */
+ATTR_MALLOC(free)
 static /*@null@*/ /*@only@*/struct utmpx *
 get_current_utmp(pid_t main_pid)
 {
--- a/lib/utmp.c
+++ b/lib/utmp.c
@@ -203,7 +203,6 @@ get_session_host(char **out, pid_t main_pid)
 #if defined(HAVE_STRUCT_UTMPX_UT_HOST)
 	if ((ut != NULL) && (ut->ut_host[0] != '\0')) {
 		*out = XSTRNDUP(ut->ut_host);
-		free (ut);
 	} else {
 		*out = NULL;
 		ret = -2;
@@ -212,6 +211,7 @@ get_session_host(char **out, pid_t main_pid)
 	*out = NULL;
 	ret = -2;
 #endif
+	free(ut);
 
 	return ret;
 }


Karlson2k:
Seems to be a copy of https://github.com/shadow-maint/shadow/pull/1290/commits/a3667fdfd293270ab14a61657d8af2598f4d3364 from PR #1290

lib/utmp.c
```diff
@@ -203,7 +203,6 @@ get_session_host(char **out, pid_t main_pid)
 #if defined(HAVE_STRUCT_UTMPX_UT_HOST)
 	if ((ut != NULL) && (ut->ut_host[0] != '\0')) {
 		*out = XSTRNDUP(ut->ut_host);
-		free (ut);
```
alejandro-colomar, line 206: This is being fixed already: <https://github.com/shadow-maint/shadow/pull/1290/commits/a3667fdfd293270ab14a61657d8af2598f4d3364>

Karlson2k: Yep, noticed this too after making of this PR.
There is a second commit in this PR.

Karlson2k:
Related idea.
Instead of all these allocations/freeings, would it be better to use a static buffer for utmpx entry in `get_current_utmp()`?
The size of the entry is not large and suits quite well for a static buffer.
`get_current_utmp()` would still return a pointer, but there would be no need to free it.
This approach, however, will make `get_current_utmp()` thread-unsafe, but the underlying `getutxent()` is not thread-safe anyway (even now, if `get_current_utmp()` is called in two threads simultaneously, the result will be incorrect).

The same approach is used in glibc.

lib/utmp.c
Karlson2k: Probably it worth to merge the fix separately from the larger PR as it looks like independent?

alejandro-colomar:
> Related idea. Instead of all these allocations/freeings, would it be better to use a static buffer for utmpx entry in `get_current_utmp()`? The size of the entry is not large and suits quite well for a static buffer. `get_current_utmp()` would still return a pointer, but there would be no need to free it. This approach, however, will make `get_current_utmp()` thread-unsafe, but the underlying `getutxent()` is not thread-safe anyway (even now, if `get_current_utmp()` is called in two threads simultaneously, the result will be incorrect).
> 
> The same approach is used in glibc.

I prefer avoiding `static` in general.  It makes it more difficult to reason about the code.  It's like global variables, after all.

lib/utmp.c
alejandro-colomar: Maybe;  I'll think about it.

Karlson2k:
> I prefer avoiding `static` in general. It makes it more difficult to reason about the code. It's like global variables, after all.

I personally prefer avoiding global variables (as the scope is not controlled), but a small static buffer inside the function is perfectly fine (the scope is fully controlled), it can make code better, faster, smaller, easier-to-read and safer at the same time.
However, here it is up to you. I'll keep allocations.

lib/utmp.c
alejandro-colomar: I like this patch set.  It adds `[[gnu::malloc(free)]]`, which I didn't, and that helps prevent this bug in the future.  I'll merge it.  Thanks!

alejandro-colomar (approved):
```
```

lib/utmp.c
```diff
@@ -146,6 +146,7 @@ failtmp(const char *username, const struct utmpx *failent)
  *	Return NULL if no entries exist in utmp for the current process or
  *	            there is an error reading utmp.
  */
+ATTR_MALLOC(free)
```
alejandro-colomar, line 151: You should include `"attr.h"` (right before `"sizeof.h"`) for this.

alejandro-colomar: And also `<stdlib.h>`, right after `<stdio.h>`, BTW, for `free`.

lib/utmp.c
Karlson2k: You are welcome.
I'll update it based on your comments.

lib/utmp.c
Karlson2k: > You should include `"attr.h"` (right before `"sizeof.h"`) for this.

It is included automatically by "alloc/x/xcalloc.h", "alloc/x/xmalloc.h" and other internal headers.  However, I'll add it if you want explicit include.

Karlson2k: > And also `<stdlib.h>`, right after `<stdio.h>`, BTW, for `free`.

It is included by "alloc/x/xmalloc.h" (and probably others).

Karlson2k:
Updated as requested.
However, I suggest to use PR #1332 instead.

lib/utmp.c
alejandro-colomar: Yup, I want explicit includes.  See <https://include-what-you-use.org/>.

alejandro-colomar:
> Updated as requested. However, I suggest to use PR #1332 instead.

Thanks!  I merged #1332.

PR #1331, opened 2025-08-10
Refactor of utmp entry search + optimisations

Karlson2k:
The code avoids multiple allocations (maximum one allocation instead of maximum three allocations).
Also some minor optimisations added.
Hope the code is readable enough.


alejandro-colomar:
I don't find this code readable either.  I'll close this.  I've written something better based on your rightful concerns about malloc(3) being overkill for temporary copies: <https://github.com/shadow-maint/shadow/pull/1333>

PR #1338, opened 2025-08-19
lib/string/README: Add guidelines for using strings

alejandro-colomar:
Suggested-by: @ikerexxe 
Suggested-by: @hallyn 
Suggested-by: @Karlson2k
Suggested-by: @lslebodn

Here, I've covered also some of the functions I haven't added yet.

<details>
<summary>v2</summary>

-  Address some issues reported by @ikerexxe .

```
$ git range-diff master gh/strreadme strreadme 
1:  d6d127b6 ! 1:  cc66977f lib/string/README: Add guidelines for using strings
    @@ lib/string/README (new)
     +   (almost) exclusively.
     +
     +
    -+Forbidden libc functions:
    -+=========================
    ++Don't use some libc functions without Really Good Reasons:
    ++==========================================================
     +
     +<stdio.h>
     +    asprintf(3)
    @@ lib/string/README (new)
     +    strcmp(3)
     +  Use streq() instead.
     +  The return value of strcmp(3) is confusing.
    -+  strcmp(3) would be legitimate as a callback to a search
    -+  function, but we don't need that here.
    ++  strcmp(3) would be legitimate for sorting strings.
     +
     +    strcasecmp(3)
     +  Use strcaseeq() instead.
    @@ lib/string/README (new)
     +    strncat(3)
     +  Use strndup(3) or strndupa(3) instead.
     +  strncat(3) is legitimate for catenating a prefix of a string
    -+  after an existing string, but we don't need that here.
    ++  after an existing string.
     +
     +    strlcpy(3)
     +  Use strtcpy() instead.
    @@ lib/string/README (new)
     +Specific guidelines:
     +====================
     +
    -+ctype/
    ++ctype/ - Character classification functions
    ++
     +    strchrisascii/
     +  These functions return true
     +  if the string has any characters that belong to a category.
    @@ lib/string/README (new)
     +    strtoascii/
     +  These functions translate all characters in a string.
     +
    -+memset/
    ++memset/ - Memory zeroing utilities
    ++
     +    memzero()
     +  Synonym of explicit_bzero(3).
     +    MEMZERO()
    @@ lib/string/README (new)
     +    strzero()
     +  Like memzero(), but takes a string.
     +
    -+strchr/
    ++strchr/ - Character search and counting
    ++
     +    strchrcnt()
     +  Count the number of occurrences of a given character in a string.
     +
    @@ lib/string/README (new)
     +    strnul()
     +  Return a pointer to the terminating null byte.  (s + strlen(s))
     +
    -+strcmp/
    ++strstr/ - String search
    ++
    ++  s/
    ++    stppfx()  // Current name: strprefix()
    ++  Return a pointer to the end of the prefix, or NULL if not found.
    ++
    ++    stpsfx()  // Unimplemented
    ++  Return a pointer to the suffix, or NULL if not found.
    ++
    ++strcmp/ - String comparison
    ++
     +  s/
     +    streq()
     +  Return true if the strings are equal.
    @@ lib/string/README (new)
     +    strpfx()  // Unimplemented
     +  Return true if the string starts with a prefix.
     +
    -+    stppfx()  // Current name: strprefix()
    -+  Like strpfx(), but return a pointer to the end of the prefix.
    -+
     +    strsfx()  // Unimplemented
     +  Return true if the string ends with a suffix.
     +
    -+    stpsfx()  // Unimplemented
    -+  Like strsfx(), but return a pointer to the suffix.
    -+
     +    strcaseeq()
     +  Like streq(), but ignore upper-/lower-case.
     +
    @@ lib/string/README (new)
     +
     +  n/
     +    strneq()  // Unimplemented
    -+  Return true if a nonstring is equal to a string.
    ++  Return true if a [[gnu::nonstring]] is equal to a string.
     +    STRNEQ()  // Unimplemented
     +  Like strneq(), but takes an array.
     +
     +    strnpfx()  // Unimplemented
    -+  Return true if a nonstring starts with a prefix.
    ++  Return true if a [[gnu::nonstring]] starts with a prefix.
     +    STRNPFX()  // Unimplemented
     +  Like strnpfx(), but takes an array.
     +
    -+strdup/
    ++strdup/ - Memory duplication
    ++
     +  s/
     +    strndupa(3)
    -+  Create a new string (in stack storage) from a nonstring.
    ++  Create a new string (in stack) from a [[gnu::nonstring]].
     +    STRNDUPA()
     +  Like strndupa(3), but takes an array.
     +
    @@ lib/string/README (new)
     +    MEMDUP()
     +  Like memdup(), but with type-safety checks.
     +
    -+strcpy/
    ++strcpy/ - String copying
    ++
     +  n/
     +    STRNCPY()
     +  Like strncpy(3), but takes an array.
    @@ lib/string/README (new)
     +    MEMCPY()
     +  Like memcpy(3), but takes two arrays.
     +
    -+sprintf/
    ++sprintf/ - Formatted string creation
    ++
     +    aprintf()
     +  sprintf(3) variant that allocates.  It has better calling
     +  conventions than asprintf(3).
    @@ lib/string/README (new)
     +  Similar to stprintf(), but takes a pointer to the end instead of
     +  a size.  This makes it safer for chaining several calls.
     +
    -+strspn/
    ++strspn/ - String span searching
    ++
     +  Naming conventions:
     +  -  'r': reverse (search from the end).
     +  -  'c': complement (negate the second argument).
    @@ lib/string/README (new)
     +  This is rarely useful.  It was useful for implementing
     +  basename().
     +
    -+strtok/
    ++strsep/ - String separation
    ++
     +    stpsep()
     +  Similar to strsep(3), but swap the input pointer with the return
     +  value.  It writes a null byte at the first delimiter found, and
```
</details>

<details>
<summary>v3</summary>

-  Reword with suggestions from @ikerexxe.

```
$ git range-diff master gh/strreadme strreadme 
1:  cc66977f ! 1:  374c8fc5 lib/string/README: Add guidelines for using strings
    @@ lib/string/README (new)
     +Specific guidelines:
     +====================
     +
    -+ctype/ - Character classification functions
    ++ctype/ - Character classification and conversion functions
     +
     +    strchrisascii/
     +  These functions return true
    -+  if the string has any characters that belong to a category.
    ++  if the string has any characters that
    ++  belong to the category specified in the function name.
     +
     +    strisascii/
     +  These functions return true
    -+  if all of the characters of the string belong to a category
    ++  if all of the characters of the string
    ++  belong to the category specified in the function name
     +  and the string is not an empty string.
     +
     +    strtoascii/
```
</details>

<details>
<summary>v3b</summary>

-  wfix

```
$ git rd
1:  374c8fc5 ! 1:  5616a0b0 lib/string/README: Add guidelines for using strings
    @@ lib/string/README (new)
     +    strtoascii/
     +  These functions translate all characters in a string.
     +
    -+memset/ - Memory zeroing utilities
    ++memset/ - Memory zeroing
     +
     +    memzero()
     +  Synonym of explicit_bzero(3).
```
</details>

<details>
<summary>v4</summary>

-  Reword, as discussed with @ikerexxe .

```
$ git rd 
1:  5616a0b0 ! 1:  cd11ed92 lib/string/README: Add guidelines for using strings
    @@ lib/string/README (new)
     +<stdio.h>
     +    asprintf(3)
     +  Use aprintf() instead.
    -+  It is difficult to handle errors after asprintf(3).  Also, it
    -+  makes it more difficult for static analyzers to check that calls
    -+  to it free(3) memory correctly.
    ++  It is difficult to handle errors after asprintf(3).
    ++  Also, it makes it more difficult for static analyzers to check
    ++  that memory is later free(3)d appropriately.
     +
     +<string.h>
     +    snprintf(3)
    @@ lib/string/README (new)
     +
     +  s/
     +    stppfx()  // Current name: strprefix()
    -+  Return a pointer to the end of the prefix, or NULL if not found.
    ++  Return a pointer to the end of the prefix,
    ++  or NULL if not found.
     +
     +    stpsfx()  // Unimplemented
    -+  Return a pointer to the suffix, or NULL if not found.
    ++  Return a pointer to the beginning of the suffix,
    ++  or NULL if not found.
     +
     +strcmp/ - String comparison
     +
    @@ lib/string/README (new)
     +sprintf/ - Formatted string creation
     +
     +    aprintf()
    -+  sprintf(3) variant that allocates.  It has better calling
    -+  conventions than asprintf(3).
    ++  sprintf(3) variant that allocates.
    ++  It has better interface than asprintf(3).
     +
     +    stprintf()  // Current name: snprintf_()
     +  snprintf(3) wrapper that reports truncation with -1.
```
</details>

<details>
<summary>v5</summary>

-  Address comments from @hallyn.

```
$ git rd 
1:  cd11ed92 ! 1:  09c2912f lib/string/README: Add guidelines for using strings
    @@ lib/string/README (new)
     +
     +Specific guidelines:
     +====================
    ++  Under lib/string/ we provide a set of functions to manipulate
    ++  strings, separated in subdirectories by utility type.  In this
    ++  section, we provide a broad overview.
     +
     +ctype/ - Character classification and conversion functions
     +
     +    strchrisascii/
    -+  These functions return true
    ++  The functions defined under this directory
    ++  return true
     +  if the string has any characters that
     +  belong to the category specified in the function name.
     +
     +    strisascii/
    -+  These functions return true
    ++  The functions defined under this directory
    ++  return true
     +  if all of the characters of the string
     +  belong to the category specified in the function name
     +  and the string is not an empty string.
     +
     +    strtoascii/
    -+  These functions translate all characters in a string.
    ++  The functions defined under this directory
    ++  translate all characters in a string.
     +
     +memset/ - Memory zeroing
     +
```
</details>

### DIFF
--- a/lib/string/README
+++ b/lib/string/README
@@ -1,0 +1,278 @@
+General guidelines:
+===================
+
+-  If there's an upper-case macro that wraps a function, use the macro
+   if possible.  These use macro magic to add safety.
+
+-  x*() functions wrap a function of the same name without the 'x'.
+   These wrappers exit on error instead of returning NULL.
+   They simplify program code.  Avoid them in library code.
+
+-  strn*() functions are forbidden for use with strings.  Consider strn
+   a mnemonic for [[gnu::nonstring]].  We use them with utmp(5) members
+   (almost) exclusively.
+
+
+Don't use some libc functions without Really Good Reasons:
+==========================================================
+
+<stdio.h>
+    asprintf(3)
+	Use aprintf() instead.
+	It is difficult to handle errors after asprintf(3).
+	Also, it makes it more difficult for static analyzers to check
+	that memory is later free(3)d appropriately.
+
+<string.h>
+    snprintf(3)
+	Use stprintf() or seprintf() instead.
+	It is difficult to check for truncation after snprintf(3).
+
+    strcmp(3)
+	Use streq() instead.
+	The return value of strcmp(3) is confusing.
+	strcmp(3) would be legitimate for sorting strings.
+
+    strcasecmp(3)
+	Use strcaseeq() instead.
+
+    strncat(3)
+	Use strndup(3) or strndupa(3) instead.
+	strncat(3) is legitimate for catenating a prefix of a string
+	after an existing string.
+
+    strlcpy(3)
+	Use strtcpy() instead.
+	strlcpy(3) is vulnerable to DoS.
+	Also, it is difficult to check for truncation after strlcpy(3).
+
+    strlcat(3)
+	Use strtcat() or stpecpy() instead.
+	strlcat(3) is vulnerable to DoS.
+	Also, it is difficult to check for truncation after strlcat(3).
+
+
+Specific guidelines:
+====================
+	Under lib/string/ we provide a set of functions to manipulate
+	strings, separated in subdirectories by utility type.  In this
+	section, we provide a broad overview.
+
+ctype/ - Character classification and conversion functions
+
+    strchrisascii/
+	The functions defined under this directory
+	return true
+	if the string has any characters that
+	belong to the category specified in the function name.
+
+    strisascii/
+	The functions defined under this directory
+	return true
+	if all of the characters of the string
+	belong to the category specified in the function name
+	and the string is not an empty string.
+
+    strtoascii/
+	The functions defined under this directory
+	translate all characters in a string.
+
+memset/ - Memory zeroing
+
+    memzero()
+	Synonym of explicit_bzero(3).
+    MEMZERO()
+	Like memzero(), but takes an array.
+
+    strzero()
+	Like memzero(), but takes a string.
+
+strchr/ - Character search and counting
+
+    strchrcnt()
+	Count the number of occurrences of a given character in a string.
+
+    strchrscnt()
+	Count the number of occurrences of several characters.
+
+    strnul()
+	Return a pointer to the terminating null byte.  (s + strlen(s))
+
+strstr/ - String search
+
+  s/
+    stppfx()  // Current name: strprefix()
+	Return a pointer to the end of the prefix,
+	or NULL if not found.
+
+    stpsfx()  // Unimplemented
+	Return a pointer to the beginning of the suffix,
+	or NULL if not found.
+
+strcmp/ - String comparison
+
+  s/
+    streq()
+	Return true if the strings are equal.
+
+    strpfx()  // Unimplemented
+	Return true if the string starts with a prefix.
+
+    strsfx()  // Unimplemented
+	Return true if the string ends with a suffix.
+
+    strcaseeq()
+	Like streq(), but ignore upper-/lower-case.
+
+    strcasepfx()  // Unimplemented
+	Like strpfx(), but ignore upper-/lower-case.
+
+    strcasesfx()  // Unimplemented
+	Like strsfx(), but ignore upper-/lower-case.
+
+  n/
+    strneq()  // Unimplemented
+	Return true if a [[gnu::nonstring]] is equal to a string.
+    STRNEQ()  // Unimplemented
+	Like strneq(), but takes an array.
+
+    strnpfx()  // Unimplemented
+	Return true if a [[gnu::nonstring]] starts with a prefix.
+    STRNPFX()  // Unimplemented
+	Like strnpfx(), but takes an array.
+
+strdup/ - Memory duplication
+
+  s/
+    strndupa(3)
+	Create a new string (in stack) from a [[gnu::nonstring]].
+    STRNDUPA()
+	Like strndupa(3), but takes an array.
+
+    STRNDUP()
+	Like strndup(3), but takes an array.
+
+  m/
+    memdup()
+	Duplicate an object.  Returns a pointer to the dup.
+    MEMDUP()
+	Like memdup(), but with type-safety checks.
+
+strcpy/ - String copying
+
+  n/
+    STRNCPY()
+	Like strncpy(3), but takes an array.
+	Use it *exclusively* for copying from a string into a utmp(5)
+	member.
+
+    strncpytail()
+	Like strncpy(), but when truncating, the tail of the string is
+	kept instead of the beginning.  This is useful for ut_id.
+    STRNCPYTAIL()
+	Like strncpytail, but takes an array.
+
+    STRNCAT()  // To be removed
+	Do NOT use.  I'll remove it soon.
+
+  s/
+    strtcpy()
+	Copy from a string into another string with truncation.
+	This is what the Linux kernel calls strscpy().
+	If you need more than one call to form a string,
+	use stpecpy() instead.
+    STRTCPY()
+	Like strtcpy(), but takes an array.
+
+    strtcat()  // Unimplemented
+	Catenate a string after an existing string, with truncation.
+	Rarely useful.  Consider using stpecpy() instead.
+    STRTCAT()  // Unimplemented
+	Like strtcat(), but takes an array.
+
+    stpecpy()
+	Similar to strtcpy(), but takes a pointer to the end instead of
+	a size.  This makes it safer for chaining several calls.
+
+  m/
+    MEMCPY()
+	Like memcpy(3), but takes two arrays.
+
+sprintf/ - Formatted string creation
+
+    aprintf()
+	sprintf(3) variant that allocates.
+	It has better interface than asprintf(3).
+
+    stprintf()  // Current name: snprintf_()
+	snprintf(3) wrapper that reports truncation with -1.
+	If you need more than one call to form a string,
+	use seprintf() instead.
+    STPRINTF()  // Current name: SNPRINTF()
+	Like stprintf(), but takes an array.
+
+    seprintf()  // Current name: stpeprintf())
+	Similar to stprintf(), but takes a pointer to the end instead of
+	a size.  This makes it safer for chaining several calls.
+
+strspn/ - String span searching
+
+	Naming conventions:
+	-  'r': reverse (search from the end).
+	-  'c': complement (negate the second argument).
+	-  'stp': Return a pointer instead of a length.
+
+    stpspn()
+	Like strspn(3), but return a pointer instead of an offset.
+	It is useful for skipping leading white space.
+
+    strrspn()
+	Use stprspn() instead.
+	Return the offset of a suffix segment consisting only of
+	characters in 'accept'.
+    stprspn()
+	Like strrspn(), but return a pointer instead of an offset.
+	It is useful for removing trailing white space.
+
+    strrcspn()
+	Use stprcspn() instead.
+	Return the offset of a suffix segment consisting only of
+	characters not in 'reject'.
+    stprcspn()
+	Like strrcspn(), but return a pointer instead of an offset.
+	This is rarely useful.  It was useful for implementing
+	basename().
+
+strsep/ - String separation
+
+    stpsep()
+	Similar to strsep(3), but swap the input pointer with the return
+	value.  It writes a null byte at the first delimiter found, and
+	returns a pointer to one past that null byte.  If no delimiters
+	are found, it returns NULL.
+	It is very useful after fgets(3), for removing the '\n', while
+	checking for its presence.  In some cases, it's also useful as
+	a replacement for strsep(3) with a more convenient calling
+	convention, but this depends on the surrounding code.
+
+    strsep2arr()
+	Loop around strsep(3) to fill an array of buffers with all
+	the delimited strings that have been split.
+	Useful when the number of delimiter-separated fields is known.
+    STRSEP2ARR()
+	Like strsep2arr(), but take an array, and simplify error checks.
+
+    strsep2ls()
+	Similar to strsep2arr(), but terminate the list with a null
+	pointer.
+	Useful when the number of delimiter-separated fields is unknown.
+    STRSEP2LS()
+	Like strsep2ls(), but take an array.
+
+    astrsep2ls()
+	Variant of strsep2ls() that allocates the array of strings.
+	(But the strings themselves are not duplicated.)
+
+strftime.h
+    STRFTIME()
+	Like strftime(3), but takes an array.


lib/string/README
```diff
@@ -0,0 +1,257 @@
...
+-  x*() functions wrap a function of the same name without the 'x'.
+   These wrappers exit on error instead of returning NULL.
+   They simplify program code.  Avoid them in library code.
+
+-  strn*() functions are forbidden for use with strings.  Consider strn
+   a mnemonic for [[gnu::nonstring]].  We use them with utmp(5) members
+   (almost) exclusively.
+
+
+Forbidden libc functions:
```
ikerexxe: Forbidden sounds too harsh; I would say not recommended

lib/string/README
```diff
@@ -0,0 +1,257 @@
...
+<string.h>
+    snprintf(3)
+	Use stprintf() or seprintf() instead.
+	It is difficult to check for truncation after snprintf(3).
+
+    strcmp(3)
+	Use streq() instead.
+	The return value of strcmp(3) is confusing.
+	strcmp(3) would be legitimate as a callback to a search
+	function, but we don't need that here.
```
ikerexxe: s/but we don't need that here/but we don't usually need that functionality in shadow

This applies in several other places of this document.

lib/string/README
```diff
@@ -0,0 +1,257 @@
...
+	Also, it is difficult to check for truncation after strlcat(3).
+
+
+Specific guidelines:
+====================
+
+ctype/
+    strchrisascii/
+	These functions return true
+	if the string has any characters that belong to a category.
```
ikerexxe: Which category? After reading the code I assume you mean control character, but I think it's necessary to mention it here

lib/string/README
```diff
@@ -0,0 +1,257 @@
...
+
+    strisascii/
+	These functions return true
+	if all of the characters of the string belong to a category
```
ikerexxe: s/to a category/to the specified category

lib/string/README
```diff
@@ -0,0 +1,257 @@
...
+strcmp/
+  s/
+    streq()
+	Return true if the strings are equal.
+
+    strpfx()  // Unimplemented
+	Return true if the string starts with a prefix.
+
+    stppfx()  // Current name: strprefix()
+	Like strpfx(), but return a pointer to the end of the prefix.
```
ikerexxe: I would explain this function as follows:
Similar to strpfx(), but it returns a pointer to the end of the prefix instead of a boolean.

lib/string/README
```diff
@@ -0,0 +1,257 @@
...
+
+    strsfx()  // Unimplemented
+	Return true if the string ends with a suffix.
+
+    stpsfx()  // Unimplemented
+	Like strsfx(), but return a pointer to the suffix.
```
ikerexxe: Same comment as `stppfx`

lib/string/README
```diff
@@ -0,0 +1,257 @@
...
+
+    strcasepfx()  // Unimplemented
+	Like strpfx(), but ignore upper-/lower-case.
+
+    strcasesfx()  // Unimplemented
+	Like strsfx(), but ignore upper-/lower-case.
+
+  n/
+    strneq()  // Unimplemented
+	Return true if a nonstring is equal to a string.
```
ikerexxe: What do you mean by nonstring? I assume you mean a non-null-terminated string. If so, can you define nonstring at the beginning of the document?

lib/string/README
```diff
@@ -0,0 +1,257 @@
...
+	a size.  This makes it safer for chaining several calls.
+
+  m/
+    MEMCPY()
+	Like memcpy(3), but takes two arrays.
+
+sprintf/
+    aprintf()
+	sprintf(3) variant that allocates.  It has better calling
+	conventions than asprintf(3).
```
ikerexxe: What do you mean by calling conventions?

lib/string/README
```diff
@@ -0,0 +1,257 @@
...
+	-  'r': reverse (search from the end).
+	-  'c': complement (negate the second argument).
+	-  'stp': Return a pointer instead of a length.
+
+    stpspn()
+	Like strspn(3), but return a pointer instead of an offset.
+	It is useful for skipping leading white space.
+
+    strrspn()
+	Use stprspn() instead.
```
ikerexxe, line 230: Does this mean that you'll remove it?

lib/string/README
alejandro-colomar: Hmmm, I could say that, but I really mean forbidden.  Some functions have a legitimate use case, such as strcmp(3), although we don't need them here.  Other functions are really never-to-be-used, such as asprintf(3).

I guess I'll relax that wording; it doesn't hurt.

lib/string/README
alejandro-colomar: It's not a "usually" here; we have no code doing that at all.

But I guess I can change that, so that this guideline applies to other projects too.

lib/string/README
alejandro-colomar: It depends on the function.  strchriscntrl() determines if there are any control characters.  If we had any other function, it'd depend on the function name.  That's why the description is generic for any functions under that directory.

lib/string/README
alejandro-colomar: Hmmm, yup; that's better wording.

lib/string/README
alejandro-colomar: I used that wording so that it fits in one line.  I was trying to balance information and a wall of text.  Do you really prefer that if it means using two lines?

lib/string/README
alejandro-colomar: I didn't put an explicit definition, but there's a reference at the top:

```
$ head -n15 lib/string/README  
General guidelines:
===================

-  If there's an upper-case macro that wraps a function, use the macro
   if possible.  These use macro magic to add safety.

-  x*() functions wrap a function of the same name without the 'x'.
   These wrappers exit on error instead of returning NULL.
   They simplify program code.  Avoid them in library code.

-  strn*() functions are forbidden for use with strings.  Consider strn
   a mnemonic for [[gnu::nonstring]].  We use them with utmp(5) members
   (almost) exclusively.

```

See the mention of `[[gnu::nonstring]]`. Maybe I could use `[[gnu::nonstring]]` everywhere instead of `nonstring`.

lib/string/README
alejandro-colomar: The parameters and return value.

lib/string/README
alejandro-colomar: No; it means strrspn() is there just to be able to implement stprspn(), but it's not meant to be used other than that.

alejandro-colomar:
> Thank you for putting this together, it is a great step forward in understanding all existing functions and the roadmap you are proposing.
> 
> I added several comments inline, but I have some more general comments.
> 
> I miss a table of contents or some explanation for what each category does. My proposal would be to use the following:
> 
> ```
> - ctype/: Character classification functions
> - memset/: Memory zeroing utilities
> - strchr/: Character search and counting
> - strcmp/: String comparison with various modes
> - strdup/: Memory duplication utilities
> - strcpy/: Safe string copying variants
> - sprintf/: Formatted string creation
> - strspn/: String span operations
> - strtok/: String tokenization utilities
> ```

Those titles sound good (I'll amend some a little bit).

But I'd prefer to not write a table of contents explicitly,
as it will likely get out of sync with the contents below.
However, you can create the table of contents with grep(1):

```sh
$ cat lib/string/README | grep '^\S'
General guidelines:
===================
-  If there's an upper-case macro that wraps a function, use the macro
-  x*() functions wrap a function of the same name without the 'x'.
-  strn*() functions are forbidden for use with strings.  Consider strn
Forbidden libc functions:
=========================
<stdio.h>
<string.h>
Specific guidelines:
====================
ctype/
memset/
strchr/
strcmp/
strdup/
strcpy/
sprintf/
strspn/
strtok/
strftime.h
```

Maybe I could add the information after the directory name.

> 
> Furthermore, I find it somewhat difficult to understand the functions that begin with `Like *(3)`, since in most cases I need to first check `man *(3)` and then return to this document to understand the difference. Perhaps it is just me, or perhaps there are other people who have difficulty retaining this type of information and changing the context. I really do not know.

With that, I essentially mean that those are libc functions.  You don't need to read the manual page, as you'll usually remember what the libc functions do, I expect.

lib/string/README
alejandro-colomar: Fixed, by using `[[gnu::nonstring]]` everywhere.

lib/string/README
alejandro-colomar: I've separated them into a different category: strstr/.
Now, they have their own wording without referring to the others.

lib/string/README
alejandro-colomar: I separated these into a separate category (which means I'll eventually move them to a new directory).  The wording is now entirely different.

lib/string/README
alejandro-colomar: Fixed.

lib/string/README
alejandro-colomar: I've reworded that.

lib/string/README
alejandro-colomar: I've just removed those parts.  It's up to the programmer to justify the use of such functions, anyway.

lib/string/README
alejandro-colomar: I've reworded this to be more explicit.

alejandro-colomar:
I've significantly reworded the document.  Please re-check.

Here's the "summary" that you can get now with grep(1):

```
$ cat lib/string/README | grep '^\S'
General guidelines:
===================
-  If there's an upper-case macro that wraps a function, use the macro
-  x*() functions wrap a function of the same name without the 'x'.
-  strn*() functions are forbidden for use with strings.  Consider strn
Don't use some libc functions without Really Good Reasons:
==========================================================
<stdio.h>
<string.h>
Specific guidelines:
====================
ctype/ - Character classification and conversion functions
memset/ - Memory zeroing utilities
strchr/ - Character search and counting
strstr/ - String search
strcmp/ - String comparison
strdup/ - Memory duplication
strcpy/ - String copying
sprintf/ - Formatted string creation
strspn/ - String span searching
strsep/ - String separation
strftime.h
```

lib/string/README
ikerexxe: Ok, this sounds better

lib/string/README
ikerexxe: Previous wording was very generic, thank you for updating it

lib/string/README
ikerexxe: Good, sounds better now

lib/string/README
```diff
@@ -0,0 +1,270 @@
...
+	Return a pointer to the terminating null byte.  (s + strlen(s))
+
+strstr/ - String search
+
+  s/
+    stppfx()  // Current name: strprefix()
+	Return a pointer to the end of the prefix, or NULL if not found.
+
+    stpsfx()  // Unimplemented
+	Return a pointer to the suffix, or NULL if not found.
```
ikerexxe: Beginning or end of the suffix? That may cause some confusion

lib/string/README
ikerexxe: This is more obvious now, and people can check that definition to understand it

lib/string/README
ikerexxe: I see `calling conventions` is a little bit more complex than what you mentioned in the previous comment. Do you think it'd be more sensible to use interface instead?

lib/string/README
ikerexxe: In that case we should hide this by making the function static in the same file as `stprspn()`

lib/string/README
```diff
@@ -0,0 +1,270 @@
...
+
+Don't use some libc functions without Really Good Reasons:
+==========================================================
+
+<stdio.h>
+    asprintf(3)
+	Use aprintf() instead.
+	It is difficult to handle errors after asprintf(3).  Also, it
+	makes it more difficult for static analyzers to check that calls
+	to it free(3) memory correctly.
```
ikerexxe: s/to it free(3)/to free(3)

ikerexxe:
> Maybe I could add the information after the directory name.

That sounds good! I believe this information will be useful to other developers.

> With that, I essentially mean that those are libc functions. You don't need to read the manual page, as you'll usually remember what the libc functions do, I expect.

For some I do, for others I wish.

lib/string/README
alejandro-colomar: Yup, I should clarify.

lib/string/README
alejandro-colomar: > In that case we should hide this by making the function static in the same file as stprspn()

stpspn() can only be implemented as a macro, so it's in a header file, which means we need strrspn() to be an extern function.

alejandro-colomar: > s/to it free(3)/to free(3)

I agree this reads weirdly, but the `it` is correct there.  I'll reword to avoid this weirdness.

alejandro-colomar: > I see calling conventions is a little bit more complex than what you mentioned in the previous comment. Do you think it'd be more sensible to use interface instead?

Agree.  Fixed.

lib/string/README
alejandro-colomar: Yup.  Thanks!

lib/string/README
alejandro-colomar: We can't.  stpspn() is a macro, and thus needs to call an extern function.

ikerexxe:
We have corrected a few errors, and I believe this has improved the comprehensibility of the documentation.

Since @hallyn and I requested this documentation, I would like to wait for an initial review from him before continuing with my own.

alejandro-colomar:
> We have corrected a few errors, and I believe this has improved the comprehensibility of the documentation.
> 
> Since @hallyn and I requested this documentation, I would like to wait for an initial review from him before continuing with my own.

Feel free to continue with your review while we wait for @hallyn .

hallyn:
On Mon, Sep 15, 2025 at 05:04:03AM -0700, Alejandro Colomar wrote:
> alejandro-colomar left a comment (shadow-maint/shadow#1338)
> 
> > We have corrected a few errors, and I believe this has improved the comprehensibility of the documentation.
> > 
> > Since @hallyn and I requested this documentation, I would like to wait for an initial review from him before continuing with my own.
> 
> Feel free to continue with your review while we wait for @hallyn .

I'll take a look this week.

alejandro-colomar:
gnulib has adopted some of these APIs:

- streq(), memeq()

We're discussing the adoption of some more.

lib/string/README
```diff
@@ -0,0 +1,272 @@
...
+
+    strlcat(3)
+	Use strtcat() or stpecpy() instead.
+	strlcat(3) is vulnerable to DoS.
+	Also, it is difficult to check for truncation after strlcat(3).
+
+
+Specific guidelines:
+====================
+
```
hallyn, line 60: Let's have a brief introduction to this section, something like

```
Under lib/string/ we provide a set of helpful functions, separated by utility
type.  In this section we provide a broad overview.
```

lib/string/README
```diff
@@ -0,0 +1,272 @@
...
+ctype/ - Character classification and conversion functions
+
+    strchrisascii/
+	These functions return true
```
hallyn: "Functions defined under this directory return true"

And similarly for each section that isn't a single function like memsory().

hallyn:
Thanks, @alejandro-colomar , this looks great.

lib/string/README
alejandro-colomar: Done; I think.

lib/string/README
alejandro-colomar: Done.

hallyn (approved):

alejandro-colomar:
@hallyn Thanks!

@ikerexxe , I'll wait for the rest of your review.

PR #1340, opened 2025-08-21
Expand system test framework

ikerexxe:
Expand the system test framework to provide python bindings for `newusers`, `groupmems` and `newgrp` binaries. Also include at least one test case example for each of them. Additional changes has been made to the way we generate the test environment (a.k.a. containers) so that it contains everything needed to run these tests.

### DIFF
--- a/tests/system/tests/test_newgrp.py
+++ b/tests/system/tests/test_newgrp.py
@@ -1,0 +1,34 @@
+"""
+Test newgrp
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from framework.roles.shadow import Shadow
+from framework.topology import KnownTopology
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_newgrp__change_to_new_group(shadow: Shadow):
+    """
+    :title: Change to a new group using newgrp
+    :setup:
+        1. Create test user and group
+        2. Add user to the group
+    :steps:
+        1. Use newgrp to change to the new group
+        2. Check that the group change worked
+    :expectedresults:
+        1. newgrp command succeeds
+        2. Current group ID matches the new group
+    :customerscenario: False
+    """
+    shadow.useradd("tuser")
+    shadow.groupadd("tgroup")
+    shadow.usermod("-a -G tgroup tuser")
+
+    cmd, gid = shadow.newgrp("tgroup", run_as="tuser")
+    assert cmd.rc == 0, "newgrp command should succeed"
+    assert gid == 1001, f"Current GID should be {1001}, got {gid}"


ikerexxe:
@danlavu can you review this PR?

danlavu:
Ack, will do it today.

share/ansible/roles/ci_run/tasks/alpine.yml
```diff
@@ -22,6 +22,12 @@
       - pkgconf
     state: present
 
+- name: Make sure expect is found
```
danlavu, line 25: It looks like you can use the apk ansible module and set the`state: present`, to ensure the package is installed, if it's not, it will then install.

tests/system/framework/roles/shadow.py
```diff
@@ -140,3 +144,126 @@ def chage(self, *args) -> ProcessResult:
         self.host.discard_file("/etc/shadow")
 
         return cmd
+
+    def newusers(self, *args, users_data: str = None) -> ProcessResult:
```
danlavu: This will work, but pylint will throw an error. ```user_data: str | None = None```

tests/system/framework/roles/shadow.py
```diff
@@ -140,3 +144,126 @@ def chage(self, *args) -> ProcessResult:
...
+        """
+        Update or create new users in batch.
+
+        Updates or creates multiple users by reading account information in passwd
+        format. If `users_data` is provided, it's passed via stdin and takes
+        precedence; otherwise, the command reads from a file specified in `args`.
```
danlavu, line 156: Do you want to conform to SSSD coding guidelines? If you do, missing docstring details, param, type, return, etc. Applies to all methods. Up to you.

tests/system/tests/test_groupmems.py
```diff
@@ -0,0 +1,44 @@
...
+from framework.roles.shadow import Shadow
+from framework.topology import KnownTopology
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_groupmems__add_user_as_root(shadow: Shadow):
+    """
+    :title: Add user to group as root user
+    :setup:
+        1. Create a test user
```
danlavu: I prefer to generalize the steps, making it easier to understand what the test is doing at a higher level. Something like, 

```
:setup:
1. Create test user and group
:steps:
1. Add user to group using groupmem as root
2. Check group and gshadow entry
:expectedresults:
1. User is added to group
2. Group values, memberships, and gshadow entry values are correct
```

Not much an issue here, but with test_newusers__create_users_from_stdin, those are a lot of steps. Consider it.

tests/system/tests/test_newgrp.py
```diff
@@ -0,0 +1,35 @@
...
+from framework.roles.shadow import Shadow
+from framework.topology import KnownTopology
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_newgrp__change_to_new_group(shadow: Shadow):
+    """
+    :title: Change to a new group using newgrp
+    :setup:
+        1. Create a test user
```
danlavu: Same suggestion.

tests/system/tests/test_newusers.py
```diff
@@ -0,0 +1,215 @@
...
+from framework.roles.shadow import Shadow
+from framework.topology import KnownTopology
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_newusers__create_users_from_stdin(shadow: Shadow):
+    """
+    :title: Create multiple users from stdin
+    :setup:
+        1. Prepare user data
```
danlavu: Is there any less clarity in the following steps?

```
:setup:
1. Prepare and create user data from stdin
:steps:
1. Check user(s) passwd, shadow, group, gshadow, and home folder
:expectedresults:
1. All user(s) passwd, shadow, group, gshadow, and home folders are correct
```

Maybe that's too condensed. 
```
:setup:
1. Prepare and create user data from stdin
:steps:
1. Check the first user's passwd, shadow, group, gshadow, and home folder
2. Check second user's passwd, shadow, group, gshadow, and home folder
:expectedresults:
1. First user's passwd, shadow, group, gshadow, and home folders are correct
2. Second user's passwd, shadow, group, gshadow, and home folders are correct
```

I blame the sssd test transformation for my particularity about this. A suggestion, up to you.

tests/system/tests/test_newusers.py
```diff
@@ -0,0 +1,215 @@
...
+
+    assert shadow.fs.exists("/home/tuser2"), "tuser2 home folder should exist"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_newusers__create_users_from_file(shadow: Shadow):
+    """
+    :title: Create multiple users using file input
+    :setup:
+        1. Create temporary file with user data in passwd format
```
danlavu: Same suggestion.

tests/system/framework/roles/shadow.py
danlavu: I know the command is ```newusers``` I would consider renaming the method to something like ```user_batch_add```, I'm on the fence about this suggestion, it didn't make sense until I found out that the command is newusers, where there would be no mistaking what the method does.

tests/system/framework/roles/shadow.py
```diff
@@ -140,3 +144,126 @@ def chage(self, *args) -> ProcessResult:
...
+        else:
+            self.logger.info(f'Administer {args_dict["name"]} group membership on {self.host.hostname}')
+            cmd = self.host.conn.run(f"echo '{password}' | su - {user} -c 'groupmems {args[0]}'", log_level=ProcessLogLevel.Error)
+
+        self.host.discard_file("/etc/group")
+        self.host.discard_file("/etc/gshadow")
+
+        return cmd
+
+    def newgrp(self, *args, user: str = "root") -> Tuple[ProcessResult, int]:
```
danlavu: Same thing, like newusers, before I knew this was named after the command, a good name would be ```group_tty_mod``` or ```group_ssh_mod```. Just a thought.

danlavu (commented):
A lot of suggestions and preferences. Other than that, it looks good.

share/ansible/roles/ci_run/tasks/alpine.yml
ikerexxe: This is a little bit more complicated than that.
Among other things, Alpine uses `ash` instead of `bash` as the shell and has a different folder division for the system. That makes things hard for us as we need to run the tests through `bash` and Some tools aren't installed in the same folder location and we need to make them available somehow. Another option that I considered was to add the other location to the `PATH`.

tests/system/framework/roles/shadow.py
ikerexxe: Good catch. I still need to enable the linting.

tests/system/framework/roles/shadow.py
ikerexxe: No, not really. For the moment I'll keep this as it is.

tests/system/tests/test_groupmems.py
ikerexxe: mmm I like this idea. I haven't used the exact wording that you mentioned, but it's almost the same

tests/system/tests/test_newgrp.py
ikerexxe: I have updated this one too, but I have the feeling that it can be improved. Feedback is much appreciated.

tests/system/tests/test_newusers.py
ikerexxe: I like. Done!

tests/system/tests/test_newusers.py
ikerexxe: Done

tests/system/framework/roles/shadow.py
ikerexxe: No, this needs to be `newusers` as that's the command shadow is providing and the well-know name.

Thanks for the suggestion though.

tests/system/framework/roles/shadow.py
ikerexxe: This too has to be `newusers` as that's the command shadow is providing.

share/ansible/roles/ci_run/tasks/alpine.yml
danlavu: I'm not familiar with ASH. I do understand the challenge, I'm a ZSH person myself. In general, my preference is to set the PATH for the process/user, leaving the file system as vanilla as possible. Just as a testing philosophy. It is not important here, it may apply somewhere self down the road.

danlavu:
I do not have permissions to approve. The PR looks good.

share/ansible/roles/ci_run/tasks/alpine.yml
ikerexxe: I also like that approach, but we may need to change several other things in how pytest-mh works to make that work. For the moment I'm happy with just having the softlink there.

ikerexxe:
> I do not have permissions to approve. The PR looks good.

Thank you for the review!

ikerexxe:
@alejandro-colomar this one is also ready for merging

hallyn (approved):

PR #1341, opened 2025-08-24
newusers: allow not passing a password

terceiro:
A possible use case for this is wanting to add subuid/subgid entries for an existing user. This change makes it possible to pass `username::::::` to newusers; the empty password will be ignored an everything else will be done. Currently this fails miserably, as PAM errors out on a empty password.


alejandro-colomar:
For consistency with the handling of the rest of the fields, it seems appropriate, as we already keep the other fields intact if they're passed empty to newusers(8).

What's the current behavior?  Is it rejected, or is the password removed (so, passwordless user)?

src/newusers.c
```diff
@@ -1207,7 +1207,7 @@ int main (int argc, char **argv)
 		usernames[nusers-1] = xstrdup(fields[0]);
 		passwords[nusers-1] = xstrdup(fields[1]);
 #endif				/* USE_PAM */
-		if (add_passwd (&newpw, fields[1]) != 0) {
+		if (!streq(fields[1], "") && add_passwd (&newpw, fields[1]) != 0) {
```
alejandro-colomar: Would you mind removing the space in `add_passwd (` while you touch this line?

src/newusers.c
```diff
@@ -1326,6 +1326,9 @@ int main (int argc, char **argv)
 #ifdef USE_PAM
 	/* Now update the passwords using PAM */
 	for (size_t i = 0; i < nusers; i++) {
+		if (streq("", passwords[i])) {
+			continue;
+		}
```
alejandro-colomar: Please don't use braces for this.

alejandro-colomar: Also, please put the `""` on the right (second argument).

terceiro:
> What's the current behavior?  Is it rejected, or is the password
> removed (so, passwordless user)?

It fails like this:

```
$ echo foobar:::::/tmp:/bin/bash | sudo newusers 
No password has been supplied.
No password has been supplied.
No password has been supplied.
newusers: (user foobar) pam_chauthtok() failed, error:
Authentication token manipulation error
newusers: (line 1, user foobar) password not changed
```

The use entry *is* created in /etc/passwd, but the account cannot be
logged into because there is no password. root can sudo into it.

terceiro:
I have pushd an updated patch addressing your comments. Thanks for the review!

alejandro-colomar:
What I'm wondering is if we might be breaking someone by doing this.

So, could someone be doing this and relying on the fact that the user is unable to login, and maybe later enable the users manually?

As I said, I think it's okay.

Cc: @hallyn, @ikerexxe

man/newusers.8.xml
```diff
@@ -107,8 +107,8 @@
 	</term>
 	<listitem>
 	  <para>
-	    This field will be encrypted and used as the new value of the
-	    encrypted password.
+	    If this field it not empty, it will be encrypted and used as the new
```
hallyn: trivial typo: if this field *is* not empty

terceiro:
On Mon, Aug 25, 2025 at 11:14:41AM -0700, Serge Hallyn wrote:
> @hallyn commented on this pull request.
> 
> Sorry, can you explain exactly how you would use this to add subuids?

```
$ echo username:::::: | newusers
```

> > @@ -107,8 +107,8 @@
>  	</term>
>  	<listitem>
>  	  <para>
> -	    This field will be encrypted and used as the new value of the
> -	    encrypted password.
> +	    If this field it not empty, it will be encrypted and used as the new
> 
> trivial typo: if this field *is* not empty

Thanks, fixed.

man/newusers.8.xml
```diff
@@ -107,8 +107,8 @@
 	</term>
 	<listitem>
 	  <para>
-	    This field will be encrypted and used as the new value of the
-	    encrypted password.
+	    If this field is not empty, it will be encrypted and used as the new
+	    value of the encrypted password.
```
alejandro-colomar: Please use semantic newlines for new documentation code.  See man-pages(7):

```
$ man man-pages | sed -n '/Use semantic newlines/,/^$/p'
   Use semantic newlines
       In  the  source of a manual page, new sentences should be started on new
       lines, long sentences should be split into lines at clause breaks  (com‐
       mas, semicolons, colons, and so on), and long clauses should be split at
       phrase  boundaries.   This convention, sometimes known as "semantic new‐
       lines", makes it easier to see the effect of patches, which often  oper‐
       ate at the level of individual sentences, clauses, or phrases.

```

alejandro-colomar: TL;DR:  Break after the comma (and also before 'and').

hallyn:
Oh, not a specific range. Got it.

man/newusers.8.xml
terceiro: Done. But note that the rest of the manpage does not seem to be following that.

alejandro-colomar: Yep, I need to fix existing code.

alejandro-colomar (approved):
LGTM.  I'll leave it to Serge for the final decision.

hallyn (approved):

Issue #1339, opened 2025-08-20
`list_owner_ranges` duplicates ranges when username equals user ID


tolusha:
```
	while (NULL != (range = commonio_next(db))) {
		if (streq(range->owner, owner)) {
			if (!append_range(&ranges, range, count++)) {
				free(ranges);
				ranges = NULL;
				count = -1;
				goto out;
			}
		}

		// Let's also compare with the ID
		if (have_owner_id == true && streq(range->owner, id)) {
			if (!append_range(&ranges, range, count++)) {
				free(ranges);
				ranges = NULL;
				count = -1;
				goto out;
			}
		}
	}
```

[1] https://github.com/shadow-maint/shadow/blob/master/lib/subordinateio.c#L887-L906

alejandro-colomar:
This seems a bug from 3ec32f9975f2 `(2022-07-20; "subordinateio: also compare the owner ID")`.

I also see some dead code being introduced in that commit.  I'll write some fixes for that.

Thanks!

alejandro-colomar:
Could you please show a reproducer?

tolusha:
Hello @alejandro-colomar 
Thank you for your response.
Unfortunately it is hard to provide a short steps to reproduce.
In my scenario, it includes OpenShift + DevSpaces + podman.
I just debugged the podman + underlying runtime libraries to find the root of my issue.

ikerexxe:
Based on the scenario, it seems you're using a Fedora-based system. On these systems, [all-numeric usernames aren't allowed](https://access.redhat.com/solutions/3103631), which suggests another piece of software in the stack might be causing the issue.

Could you provide more details on how this happened? For instance, how was the fully numeric username get created?

tolusha:
@ikerexxe 
Hello,
You are correct — I am using an RHEL-based image.
In fact, CRI-O automatically adds the user into the container [1].
Essentially, the username will match the user ID, which is taken from the runAsUser container security context [2].


[1] https://github.com/cri-o/cri-o/pull/2022/files
[2] https://github.com/cri-o/cri-o/blob/51ce3dd659becb9d31c2752cf11c4411e3b18c22/server/container_create.go#L365

alejandro-colomar:
@tolusha 

Did you accidentally close this?

tolusha:
Yes, sorry for that, reopened

ikerexxe:
[...]

PR #1345, opened 2025-08-25
lib/subordinateio.c: list_owner_ranges(): Fix duplicate range when username matches ID

alejandro-colomar:
Fixes: 3ec32f9975f2 (2022-07-20; "subordinateio: also compare the owner ID")
Closes: <https://github.com/shadow-maint/shadow/issues/1339>
Link: <https://github.com/cri-o/cri-o/issues/9416>
Reported-by: @tolusha

### DIFF
--- a/lib/subordinateio.c
+++ b/lib/subordinateio.c
@@ -885,17 +885,9 @@ int list_owner_ranges(const char *owner, enum subid_type id_type, struct subid_r
 
 	commonio_rewind(db);
 	while (NULL != (range = commonio_next(db))) {
-		if (streq(range->owner, owner)) {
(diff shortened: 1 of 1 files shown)


ikerexxe:
This approach makes the code more reliable, but before approving I'd like to understand the scenario that led to this point.

alejandro-colomar:
> This approach makes the code more reliable, but before approving I'd like to understand the scenario that led to this point.

Agree;  I prefer knowing exactly how it happened.

tolusha:
Hello.
I added a comment
https://github.com/shadow-maint/shadow/issues/1339#issuecomment-3232388867

PR #1350, opened 2025-09-02
Subuid v2

AndersBlomdell:


lib/subordinateio.c
```diff
@@ -685,6 +686,24 @@ uid_t sub_uid_find_free_range(uid_t min, uid_t max, unsigned long count)
 	return start == ULONG_MAX ? (uid_t) -1 : start;
 }
 
+bool want_subuid_file(void)
```
alejandro-colomar, line 689: Please include in the commit message the reason for the rename.

src/newusers.c
```diff
@@ -1077,8 +1077,8 @@ int main (int argc, char **argv)
 	is_shadow_grp = sgr_file_present ();
 #endif
 #ifdef ENABLE_SUBIDS
-	is_sub_uid = sub_uid_file_present () && !rflg;
-	is_sub_gid = sub_gid_file_present () && !rflg;
+	is_sub_uid = want_subuid_file () && sub_uid_file_present () && !rflg;
+	is_sub_gid = want_subgid_file () && sub_gid_file_present () && !rflg;
```
alejandro-colomar, line 1081: Please remove the space between function names and `(` in the lines you've touched.  This helps slowly get rid of that space.

lib/subordinateio.c
AndersBlomdell: Mostly because we might want_subuids even when we don't want_subuid_file, but it's just my 5¢

src/newusers.c
AndersBlomdell: Should I fix all related instances in the lines I touch? Even though it doesn't show I try to do minimal changes!

lib/subordinateio.c
alejandro-colomar: Hmmm, sounds reasonable, I guess.

src/newusers.c
alejandro-colomar: Hmmm, as long as you don't introduce more of them, you can leave them untouched.

On the other hand, both git and github have flags to ignore whitespace changes in diffs, so it won't hurt.

So, act as you prefer.  But please don't write new code with such white space.  :)

AndersBlomdell:
Version 3 of patch ready https://github.com/shadow-maint/shadow/pull/1351

alejandro-colomar:
You don't need to close Pull Requests for each version.  You can push --force.

PR #1355, opened 2025-09-12
Use available libc macros for paths, instead of creating our own.

alejandro-colomar:
This reduces build system code, as well as some definitions in our source code.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -98,8 +98,6 @@ AC_CACHE_CHECK([location of faillog/lastlog/wtmp], [shadow_cv_logdir],
 		break
 	fi
 done])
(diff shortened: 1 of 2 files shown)


alejandro-colomar:
> Do you know if all UNIX variants using this project contain those definitions?

I guess the only systems using this project are really musl and glibc.  If all checks have passed, it means they have them.

And the worst thing that could happen if someone wants to port this project to some other system where it's not available, it's that they would have a compiler error and would need to define it themselves, which is rather trivial.

FWIW, I checked NetBSD and it provides them (in a different header).  I assume that the other BSDs also have these macros.

alejandro-colomar:
Ahh, I probably forgot to open it.  Done.  Thanks for the reminder!

PR #1357, opened 2025-09-12
Expand system test framework: passwd

ikerexxe:
Expand the system test framework to provide python bindings for `passwd`, `chpasswd`, `gpasswd`, `chgpasswd` and `vipw` binaries. Also include at least one test case example for each of them. Additional changes has been made to the way we generate the test environment (a.k.a. containers) so that it contains everything needed to run these tests.

PR #1358, opened 2025-09-12
src/chgpasswd.c: fix handling of files without trailing newlines

ikerexxe:
Fix chgpasswd to properly handle input files that don't end with a newline, similar to how chpasswd handles this case. Previously, chgpasswd would incorrectly report "line too long" errors when processing the last line of a file that didn't end with a newline character.

The fix adds proper end-of-file detection and long line handling logic consistent with chpasswd implementation.

I discovered this issue while developing `test_chgpasswd__change_passwords_from_file` from https://github.com/shadow-maint/shadow/pull/1357.

### DIFF
--- a/src/chgpasswd.c
+++ b/src/chgpasswd.c
@@ -462,12 +462,22 @@ int main (int argc, char **argv)
 	 * file (gshadow or group) and the password changed.
 	 */
 	while (fgets (buf, (int) sizeof buf, stdin) != NULL) {
+		char *cp;
 		line++;
 		if (stpsep(buf, "\n") == NULL) {
-			fprintf (stderr, _("%s: line %jd: line too long\n"),
-			         Prog, line);
-			errors = true;
-			continue;
+			if (feof (stdin) == 0) {
+				// Drop all remaining characters on this line.
+				while (fgets (buf, sizeof buf, stdin) != NULL) {
+					cp = strchr (buf, '\n');
+					if (cp != NULL) {
+						break;
+					}
+				}
+				fprintf (stderr, _("%s: line %jd: line too long\n"),
+				         Prog, line);
+				errors = true;
+				continue;
+			}
 		}
 
 		/*


alejandro-colomar:
> Fix chgpasswd to properly handle input files that don't end with a newline, similar to how chpasswd handles this case. Previously, chgpasswd would incorrectly report "line too long" errors when processing the last line of a file that didn't end with a newline character.

I don't like handling files without a trailing newline.  Those files are not POSIX conforming, and we should reject them.
<https://pubs.opengroup.org/onlinepubs/9799919799.2024edition/basedefs/V1_chap03.html#tag_03_387>

I strongly oppose this change, as it increases the complexity of the code unnecessarily.

Instead files should be fixed to have a trailing line, which is simple to fix.

> 
> The fix adds proper end-of-file detection and long line handling logic consistent with chpasswd implementation.

I want to remove that code, actually.

> 
> I discovered this issue while developing `test_chgpasswd__change_passwords_from_file` from #1357.

src/chgpasswd.c
```diff
@@ -462,12 +462,22 @@ int main (int argc, char **argv)
...
+					cp = strchr (buf, '\n');
+					if (cp != NULL) {
+						break;
+					}
+				}
+				fprintf (stderr, _("%s: line %jd: line too long\n"),
+				         Prog, line);
+				errors = true;
+				continue;
+			}
```
alejandro-colomar, line 480: I oppose such a complexity increase.

We should self-restrict to POSIX-valid files.

src/chgpasswd.c
```diff
@@ -462,12 +462,22 @@ int main (int argc, char **argv)
...
 	 */
 	while (fgets (buf, (int) sizeof buf, stdin) != NULL) {
+		char *cp;
 		line++;
 		if (stpsep(buf, "\n") == NULL) {
-			fprintf (stderr, _("%s: line %jd: line too long\n"),
-			         Prog, line);
-			errors = true;
-			continue;
+			if (feof (stdin) == 0) {
```
alejandro-colomar, line 468: BTW, feof(3)'s return value is meant to be used as a boolean.  Comparing to integer 0 is weird.  :)

src/chgpasswd.c
```diff
@@ -462,12 +462,22 @@ int main (int argc, char **argv)
...
+				// Drop all remaining characters on this line.
+				while (fgets (buf, sizeof buf, stdin) != NULL) {
+					cp = strchr (buf, '\n');
+					if (cp != NULL) {
```
alejandro-colomar, line 472: This variable is unnecessary; you could do

```c
if (strchr(buf, '\n'))
    break;
```

ikerexxe:
> I don't like handling files without a trailing newline. Those files are not POSIX conforming, and we should reject them. https://pubs.opengroup.org/onlinepubs/9799919799.2024edition/basedefs/V1_chap03.html#tag_03_387
> 
> I strongly oppose this change, as it increases the complexity of the code unnecessarily.

After reading this definition, I agree with you. I proceed to close the PR.

alejandro-colomar:
> > I don't like handling files without a trailing newline. Those files are not POSIX conforming, and we should reject them. https://pubs.opengroup.org/onlinepubs/9799919799.2024edition/basedefs/V1_chap03.html#tag_03_387
> > I strongly oppose this change, as it increases the complexity of the code unnecessarily.
> 
> After reading this definition, I agree with you. I proceed to close the PR.

Thanks!  :-)

PR #1366, opened 2025-10-03
A couple of patches around SSSD integration

alexey-tikhonov:
- fix a bug introduced in 59e5eef38f899cf00925f0117e3173566b14a5fc
- don't trigger useless cache flush

### DIFF
--- a/lib/sssd.c
+++ b/lib/sssd.c
@@ -32,6 +32,10 @@ sssd_flush_cache(int dbflags)
 	const char   *spawnedEnv[] = {NULL};
 	struct stat  sb;
 
+	if ((dbflags & (SSSD_DB_PASSWD|SSSD_DB_GROUP)) == 0)
+	    /* Neither passwd nor group, nothing to do */
+	    return 0;
+
 	rv = stat(cmd, &sb);
 	if (rv == -1 && errno == ENOENT)
 		return 0;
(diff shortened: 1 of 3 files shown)


lib/sssd.c
```diff
@@ -43,15 +47,10 @@ sssd_flush_cache(int dbflags)
...
+		p = stpcpy(p, "U");
 	if (dbflags & SSSD_DB_GROUP)
-		stpcpy(p, "G");
+		p = stpcpy(p, "G");
 
-	if (streq(p, "")) {
-		/* Neither passwd nor group, nothing to do */
-		free(sss_cache_args);
-		return 0;
-	}
```
alejandro-colomar, line 54: Oops, my commit was very wrong.  Thanks for the fix!

I should probably issue a stable release 4.18.1 with this fix.

Would you mind reformatting the commit message to refer to the fixed commit this way?:

```
Fixes: 59e5eef38f89 (2024-07-03; "contrib, lib/, src/, tests/: Use stpcpy(3) instead of its pattern")
```

Also:

```
```
(this is only for the first commit.)

alejandro-colomar (approved):
The first commit LGTM.

I don't know the code touched in the second commit, so please @ikerexxe review that, although it also looks obviously good, but just to make sure.

Thanks!

ikerexxe:
I'm fine with the changes.

@alexey-tikhonov do you mind taking into account Alejandro's concern?

lib/sssd.c
alexey-tikhonov: Done.

alexey-tikhonov: > I should probably issue a stable release 4.18.1 with this fix.

JFTR: I don't know if you have other reasons for a release, but this fix is hardly worth it.

Only shadow-utils builds configured `--with-sssd` used on a systems running SSSD 'files provider' are affected. Likely there are very few such systems in the wild.

alejandro-colomar: Okay.  Thanks!  Are there distros that enable that?

alejandro-colomar: I guess I'll not do a stable release just for this then.  If anyone is interested in it, they can ask for it, though, and then I'll do it.

alejandro-colomar:
Thanks!

lib/sssd.c
alexey-tikhonov: > Are there distros that enable that?

Of what I'm aware of, I guess CentOS Stream 8 and 9. But I'm pretty sure @ikerexxe won't rebase those anyway.

ikerexxe: No, we won't get this in CentOS Stream 8 and 9

PR #1395, opened 2025-12-02
Add typeas() and use it

alejandro-colomar:
typeas() is safer than typeof() in that it requires that the input is a type name.

And it is more generally useful than `(T *){}` in that it accepts any type; T* doesn't work with array types nor function types.

(This PR also does a few other changes that are needed before we can use typeas(), and which also make these APIs more readable, safer, and more useful.)

---

-  This is a subset of <https://github.com/shadow-maint/shadow/pull/1388> at v8c.

---

Revisions:

<details>
<summary>v1b</summary>

-  Fix commit message.  That paragraph was true in the past, but not anymore.

```
$ git rd 
1:  5413a2a7e = 1:  5413a2a7e lib/: Use a consistent name for macro arguments representing a type name
2:  5de573ad6 ! 2:  c8c014c9b lib/search/, lib/, src/: Add a type parameter to the type-safe macros
    @@ Commit message
         This also makes the code more readable, by having the type explicitly
         stated at call site.
     
    -    There's a slight difference: 'void *' will now be acceptable input.
    -    That shouldn't be an issue.  'void *' is used rarely, so it would be
    -    rare to accidentally pass it where it doesn't belong.  Also, elsewhere
    -    in the language, a void * converts to anything, so making one or two
    -    APIs safe against 'void *' wouldn't change much.
    -
     
      ## lib/addgrps.c ##
3:  fe626d787 = 3:  fea19948c lib/search/: Simplify CMP()
4:  948fb0bb7 = 4:  597e54089 lib/search/: Split APIs
5:  ff3ca8afa = 5:  9e44c0575 lib/sizeof.h: typeas(): Add macro
6:  8e3450183 = 6:  d52ae9d7e lib/{alloc,search}/: Use typeas() to add support for arbitrary types
```
</details>

<details>
<summary>v1c</summary>

-  Rebase

```
$ git rd 
1:  5413a2a7e = 1:  4f95f69bc lib/: Use a consistent name for macro arguments representing a type name
2:  c8c014c9b = 2:  a10273be8 lib/search/, lib/, src/: Add a type parameter to the type-safe macros
3:  fea19948c = 3:  01e68361b lib/search/: Simplify CMP()
4:  597e54089 = 4:  822f7da29 lib/search/: Split APIs
5:  9e44c0575 = 5:  34c1f823a lib/sizeof.h: typeas(): Add macro
6:  d52ae9d7e = 6:  9dd442b84 lib/{alloc,search}/: Use typeas() to add support for arbitrary types
```
</details>

### DIFF
--- a/lib/search/l/lfind.h
+++ b/lib/search/l/lfind.h
@@ -12,10 +12,12 @@
 #include <stddef.h>
 
 #include "search/cmp/cmp.h"
+#include "sizeof.h"
 
 
 // lfind_T - linear find type-safe
-#define lfind_T(T, k, a, n, cmp)                                      \
+#define lfind_T(T, ...)            lfind_T_(typeas(T), __VA_ARGS__)
+#define lfind_T_(T, k, a, n, cmp)                                     \
 ({                                                                    \
 	_Generic(k, T *: 0, const T *: 0);                            \
 	_Generic(a, T *: 0, const T *: 0);                            \
--- a/lib/alloc/calloc.h
+++ b/lib/alloc/calloc.h
@@ -11,9 +11,11 @@
 #include <stdlib.h>
 
 #include "exit_if_null.h"
+#include "sizeof.h"
 
 
-#define CALLOC(n, T)                                                  \
+#define CALLOC(n, T)   CALLOC_(n, typeas(T))
+#define CALLOC_(n, T)                                                 \
 (                                                                     \
 	(T *) calloc(n, sizeof(T))                                    \
 )
--- a/lib/alloc/malloc.h
+++ b/lib/alloc/malloc.h
@@ -12,9 +12,11 @@
 
 #include "attr.h"
 #include "exit_if_null.h"
+#include "sizeof.h"
 
 
-#define MALLOC(n, T)                                                  \
+#define MALLOC(n, T)   MALLOC_(n, typeas(T))
+#define MALLOC_(n, T)                                                 \
 (                                                                     \
 	(T *) mallocarray(n, sizeof(T))                               \
 )
--- a/lib/alloc/realloc.h
+++ b/lib/alloc/realloc.h
@@ -11,9 +11,11 @@
 #include <stdlib.h>
 
 #include "exit_if_null.h"
+#include "sizeof.h"
 
 
-#define REALLOC(p, n, T)                                              \
+#define REALLOC(p, n, T)   REALLOC_(p, n, typeas(T))
+#define REALLOC_(p, n, T)                                             \
 (                                                                     \
 	_Generic(p, T *: (T *) reallocarray(p, (n) ?: 1, sizeof(T)))  \
 )
--- a/lib/alloc/reallocf.h
+++ b/lib/alloc/reallocf.h
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2023-2024, Alejandro Colomar <alx@kernel.org>
+// SPDX-FileCopyrightText: 2023-2025, Alejandro Colomar <alx@kernel.org>
 // SPDX-License-Identifier: BSD-3-Clause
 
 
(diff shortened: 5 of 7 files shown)


ikerexxe:
Reopening since this was incorrectly closed

hallyn:
Holy cow.  Github closed the issue that was in the Description?  I didn't even
think to manually update that...

alejandro-colomar:
> Holy cow. Github closed the issue that was in the Description? I didn't even think to manually update that...

It's even more funny.  I've heard people had closed other projects' issues by having a Closes tag that linked to another repo.  Github is quite funny.

lib/search/l/lfind.h
```diff
@@ -12,19 +12,13 @@
...
-#define LFIND(k, a, n)                                                \
+#define LFIND(T, k, a, n)                                             \
 ({                                                                    \
-	__auto_type  k_ = k;                                          \
-	__auto_type  a_ = a;                                          \
-                                                                      \
-	static_assert(is_same_typeof(k_, a_), "");                    \
-                                                                      \
-	(typeof(k_)) lfind_(k_, a_, n, sizeof(*k_), CMP(typeof(k_))); \
+	_Generic(k, T *: 0, const T *: 0);                            \
```
hallyn, line 22: Do I understand correctly that the _Generic() here is to make sure, by not
having a default, that if k is neither T * nor const T *, then compilation
will fail?

lib/search/cmp/cmp.h
```diff
@@ -12,10 +12,10 @@
 #define CMP(T)                                                        \
 (                                                                     \
 	_Generic((T) 0,                                               \
-		int *:            cmp_int,                            \
-		long *:           cmp_long,                           \
-		unsigned int *:   cmp_uint,                           \
-		unsigned long *:  cmp_ulong                           \
+		int:            cmp_int,                              \
```
hallyn, line 15: Hm, why does this work without changing cmp_int() to accept an int instead of an int* ?

I'm sure it does, but I'm not getting it...

lib/search/l/lfind.h
alejandro-colomar: Yup, that's exactly what it does.  :)

lib/search/cmp/cmp.h
alejandro-colomar: The change is

```diff
diff --git a/lib/search/cmp/cmp.h b/lib/search/cmp/cmp.h
index 2f8822119..49443f673 100644
--- a/lib/search/cmp/cmp.h
+++ b/lib/search/cmp/cmp.h
@@ -12,10 +12,10 @@
 #define CMP(T)                                                        \
 (                                                                     \
        _Generic((T) 0,                                               \
-               int *:            cmp_int,                            \
-               long *:           cmp_long,                           \
-               unsigned int *:   cmp_uint,                           \
-               unsigned long *:  cmp_ulong                           \
+               int:            cmp_int,                              \
+               long:           cmp_long,                             \
+               unsigned int:   cmp_uint,                             \
+               unsigned long:  cmp_ulong                             \
        )                                                             \
 )
 
diff --git a/lib/search/l/lfind.h b/lib/search/l/lfind.h
index eb102dc55..acc61ae4c 100644
--- a/lib/search/l/lfind.h
+++ b/lib/search/l/lfind.h
@@ -18,7 +18,7 @@
 ({                                                                    \
        _Generic(k, T *: 0, const T *: 0);                            \
        _Generic(a, T *: 0, const T *: 0);                            \
-       (T *) lfind_(k, a, n, sizeof(T), CMP(T *));                   \
+       (T *) lfind_(k, a, n, sizeof(T), CMP(T));                     \
 })
 
 
...
```

What we do is change the selector of `_Generic(3)`.  However, it's used as a callback in the same function (think of `CMP(...)` as a macro that expands to just `cmp_int` within the lfind_() call).  There, nothing changed.

What does change is the type selector we need to pass to CMP(), but it still expands to the same thing, and is used as a callback in the same function, so it should remain the same function.

hallyn:
On Fri, Dec 05, 2025 at 06:29:17AM -0800, Alejandro Colomar wrote:
> What we do is change the selector of _Generic(3).  However, it's used as a callback in the same function (think of a macro that expands to just `cmp_int` within the lfind_() call).  There, nothing changed.

oh, I see - it's a step more abstract than the way I was
processing it.  Thank you.

hallyn (approved):

PR #1396, opened 2025-12-04
useradd: fix uninitialized flags causing aarch64 failure

ikerexxe:
Initialize option_flags structure to prevent garbage memory values in `flags.chroot` and `flags.prefix` fields. Uninitialized memory caused architecture-specific failures where process_selinux evaluation differed between x86_64 and aarch64, leading to `pw_close()` failures when SELinux contexts weren't properly managed.

Fixes: c0c9485d (2025-04-25; "src/useradd.c: chroot or prefix SELinux file context")
Link: <https://bodhi.fedoraproject.org/updates/FEDORA-2025-3d835cfb15>

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2440,7 +2440,10 @@ int main (int argc, char **argv)
 #endif
 	unsigned long subuid_count = 0;
 	unsigned long subgid_count = 0;
-	struct option_flags  flags;
+	struct option_flags  flags = {
+		.chroot = false,
+		.prefix = false,
+	};
 	bool process_selinux;
 
 	log_set_progname(Prog);


ikerexxe:
As an additional comment, I wonder if we should do the same in all definitions of `struct option_flags`, as this bug probably also occurs with other binaries.

src/useradd.c
```diff
@@ -2440,7 +2440,7 @@ int main (int argc, char **argv)
 #endif
 	unsigned long subuid_count = 0;
 	unsigned long subgid_count = 0;
-	struct option_flags  flags;
+	struct option_flags  flags = {0};
```
alejandro-colomar: Please use `{}` instead of `{0}`.  It's a more appropriate initializer for structures.

alejandro-colomar: However, I don't like initializing blindly.  I'd prefer explicitly stating the values of the members:

```c
struct option_flags  flags = {
	.chroot = false;
	.prefix = false;
};
```

I'm assuming those are the defaults you want, right?

ikerexxe: Yes, that's what I wanted. Fixed in the new version

alejandro-colomar:
BTW, the commit that this fixes touched many more files.  Did you check if any other files may have similar issues?

ikerexxe:
This issue was detected by an automated test and I think that only tests `useradd`. Probably other binaries are also affected, that's why I wrote:

> As an additional comment, I wonder if we should do the same in all definitions of `struct option_flags`, as this bug probably also occurs with other binaries.

I wanted to have the fix ready for `useradd` before delving into that, but you merged the changes quite fast.

WDYT? Should we also make similar changes in other binaries?

alejandro-colomar:
> This issue was detected by an automated test and I think that only tests `useradd`. Probably other binaries are also affected, that's why I wrote:
> 
> > As an additional comment, I wonder if we should do the same in all definitions of `struct option_flags`, as this bug probably also occurs with other binaries.
> 
> I wanted to have the fix ready for `useradd` before delving into that, but you merged the changes quite fast.

Well, the fix was obviously good, so I merged it.  :)

> 
> WDYT? Should we also make similar changes in other binaries?

Yup, I've checked the source code, and the bug affects all binaries.  I've pushed a fix in #1400 .

PR #1400, opened 2025-12-05
src/: Fix uninitialized flags, and use const appropriately

alejandro-colomar:
Fixes: c0c9485d9a5a (2025-10-07; "src/useradd.c: chroot or prefix SELinux file context")
Link: <https://github.com/shadow-maint/shadow/pull/1396>


hallyn (approved):
LGTM.  Will leave it for Iker to approve and merge, in case I'm missing something.

Yeah, if they were in BSS they'd be 0/false, but these are not, so...

PR #1406, opened 2025-12-06
lib/chkname.c, src/: Strictly disallow the baddest names

alejandro-colomar:
Some names are bad, and some names are really bad.  '--badname' should
    only allow the mildly bad ones, which we can handle.  Some names are too
    bad, and it's not possible to deal with them.  Reject them
    unconditionally.
    
    -  A leading '-' is too dangerous.  It breaks things like execve(2), and
       almost every command.
    
    -  Spaces are used for delimiting lists of users and groups.
    
    -  '"' is special in many languages, including the shell.  Having it in
       user names would be unnecessarily dangerous.
    
    -  '#' is used for delimiting comments in several of our config files.
       Having it in usernames could result in incorrect configuration files.
    
    -  ',' is used for delimiting lists of users and groups.
    
    -  '/' is used for delimiting files, and thus could result in incorrect
       handling of users and groups.
    
    -  ':' is the main delimiter in /etc/shadow and /etc/passwd.
    
    -  ';' is special in many languages, including the shell.  Having it in
       user names would be unnecessarily dangerous.
    
    There are other characters that we should disallow, but they need more
    research to make sure we don't introduce regressions.  This set should
    be less problematic.

Acked-by: @zeha
Acked-by: @stoeckmann
Cc: @zugschlus
Cc: @ikerexxe 
Cc: @hallyn

---

-  This is a subset of <https://github.com/shadow-maint/shadow/pull/1158> at v14.

---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase
-  Reviewed-by @ikerexxe , @zeha 

```
$ git rd 
1:  1d2c5ea5e ! 1:  e3d835320 lib/chkname.c, src/: Strictly disallow really bad names
    @@ Commit message
         research to make sure we don't introduce regressions.  This set should
         be less problematic.
     
    -    Acked-by: Chris Hofstaedtler <zeha@debian.org>
         Acked-by: Tobias Stoeckmann <tobias@stoeckmann.org>
    +    Reviewed-by: Iker Pedrosa <ipedrosa@redhat.com>
    +    Reviewed-by: Chris Hofstaedtler <zeha@debian.org>
    -    Cc: Iker Pedrosa <ipedrosa@redhat.com>
     
```
</details>

<details>
<summary>v2</summary>

-  Ban `'`.

```
$ git rd 
1:  e3d835320 ! 1:  266308458 lib/chkname.c, src/: Strictly disallow really bad names
    @@ Commit message
         -  '#' is used for delimiting comments in several of our config files.
            Having it in usernames could result in incorrect configuration files.
     
    +    -  "'" is special in many languages, including the shell.  Having it in
    +       user names would be unnecessarily dangerous.
    +
         -  ',' is used for delimiting lists of users and groups.
     
         -  '/' is used for delimiting files, and thus could result in incorrect
    @@ lib/chkname.c: login_name_max_size(void)
     +   || streq(name, ".")
     +   || streq(name, "..")
     +   || strspn(name, "-")
    -+   || strpbrk(name, " \"#,/:;")
    ++   || strpbrk(name, " \"#',/:;")
     +   || strchriscntrl(name)
     +   || strisdigit(name))
     +  {
```
</details>

<details>
<summary>v2b</summary>

-  Rebase

```
$ git rd 
1:  266308458 ! 1:  defdaa950 lib/chkname.c, src/: Strictly disallow really bad names
    @@ lib/chkname.c: login_name_max_size(void)
                return true;
        }
     @@ lib/chkname.c: is_valid_name(const char *name)
    -          *
    -          * as a non-POSIX, extension, allow "$" as the last char for
    -          * sake of Samba 3.x "add machine script"
    --         *
    --         * Also do not allow fully numeric names or just "." or "..".
    -          */
    +    *
    +    * as a non-POSIX, extension, allow "$" as the last char for
    +    * sake of Samba 3.x "add machine script"
    +-   *
    +-   * Also do not allow fully numeric names or just "." or "..".
    +    */
      
     -  if (strisdigit(name)) {
     -          errno = EINVAL;
```
</details>

### DIFF
--- a/lib/chkname.c
+++ b/lib/chkname.c
@@ -13,7 +13,8 @@
  *   true  - OK
  *   false - bad name
  * errors:
- *   EINVAL	Invalid name characters or sequences
+ *   EINVAL	Invalid name
+ *   EILSEQ	Invalid name character sequence (acceptable with --badname)
  *   EOVERFLOW	Name longer than maximum size
  */
 
@@ -27,12 +28,15 @@
 #include <limits.h>
 #include <stdbool.h>
 #include <stddef.h>
+#include <string.h>
 #include <unistd.h>
 
 #include "defines.h"
 #include "chkname.h"
+#include "string/ctype/strchrisascii/strchriscntrl.h"
 #include "string/ctype/strisascii/strisdigit.h"
 #include "string/strcmp/streq.h"
+#include "string/strcmp/strcaseeq.h"
 
 
 #ifndef  LOGIN_NAME_MAX
@@ -59,6 +63,18 @@ login_name_max_size(void)
 static bool
 is_valid_name(const char *name)
 {
+	if (streq(name, "")
+	 || streq(name, ".")
+	 || streq(name, "..")
+	 || strspn(name, "-")
+	 || strpbrk(name, " \"#',/:;")
+	 || strchriscntrl(name)
+	 || strisdigit(name))
+	{
+		errno = EINVAL;
+		return false;
+	}
+
 	if (allow_bad_names) {
 		return true;
 	}
@@ -69,25 +85,15 @@ is_valid_name(const char *name)
 	 *
 	 * as a non-POSIX, extension, allow "$" as the last char for
 	 * sake of Samba 3.x "add machine script"
-	 *
-	 * Also do not allow fully numeric names or just "." or "..".
 	 */
 
-	if (strisdigit(name)) {
-		errno = EINVAL;
-		return false;
-	}
-
-	if (streq(name, "") ||
-	    streq(name, ".") ||
-	    streq(name, "..") ||
-	    !((*name >= 'a' && *name <= 'z') ||
+	if (!((*name >= 'a' && *name <= 'z') ||
 	      (*name >= 'A' && *name <= 'Z') ||
 	      (*name >= '0' && *name <= '9') ||
 	      *name == '_' ||
 	      *name == '.'))
 	{
-		errno = EINVAL;
+		errno = EILSEQ;
 		return false;
 	}
 
@@ -101,7 +107,7 @@ is_valid_name(const char *name)
 		      streq(name, "$")
 		     ))
 		{
-			errno = EINVAL;
+			errno = EILSEQ;
 			return false;
 		}
 	}
--- a/src/newusers.c
+++ b/src/newusers.c
@@ -398,7 +398,7 @@ static int add_user (const char *name, uid_t uid, gid_t gid)
 
 	/* Check if this is a valid user name */
 	if (!is_valid_user_name(name)) {
-		if (errno == EINVAL) {
+		if (errno == EILSEQ) {
 			fprintf(stderr,
 			        _("%s: invalid user name '%s': use --badname to ignore\n"),
 			        Prog, name);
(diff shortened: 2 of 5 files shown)


alejandro-colomar:
This list of characters should be regression-free, I think.

I'll leave the remaining ones for a separate PR, for after the release.

alejandro-colomar:
> Can you update the documentation? It'd be nice to align the man page documents with the existing behaviour

The manual pages are still compatible with this change.  They just say:

```
     -b, --badname
         Allow names that do not conform to standards.
```

Which is still true.  We just allow less names with --badname after this change, but that text is still valid.

ikerexxe:
You're absolutely right. My apologies, I was mistakenly focused on the specific character set mentioned in the **CAVEATS** section of the `useradd` man page, and overlooked that the description for `--badname` is still technically correct. No documentation update is needed for that specific flag.

zeha:
LGTM, thanks.

alejandro-colomar:
I'm going to add `'` to the ban, because it is as dangerous as `"`, for the same exact reasons.  I'll keep the Reviewed-by tags, as I don't expect a different opinion about it.

ikerexxe (approved):

alejandro-colomar:
Thanks!  @hallyn , do you feel comfortable with this patch?

hallyn (approved):

hallyn:
Thanks.

collinfunk:
This seems like a poor decision to me since you can just edit `/etc/passwd` and create a home directory manually to achieve the same thing.

I hope you do not consider further characters.

I have been considering adding tests to GNU coreutils which break implementations that assume UTF-8 encoding. For example, having `\xc2` in a user name which is fine in some encodings:

```
$ LC_ALL=en_US.ISO-8859-1 printf '\xC2\n' | iconv --from-code=ISO-8859-1 --to-code=UTF-8 
Â
```

alejandro-colomar:
> This seems like a poor decision to me since you can just edit `/etc/passwd` and create a home directory manually to achieve the same thing.

Most of these characters either already failed in some cases (for example, `:` would have failed when counting the number of fields in /etc/shadow), or didn't fail but resulted in incorrect behavior (which would be worse).  Spaces, tabs, and newlines are of course also terribly dangerous in user names.  Etc.

> I hope you do not consider further characters.

I'll investigate the source code, in case there are other characters that would result in shadow misbehaving already, such as possibly `.` and `@`.

> 
> I have been considering adding tests to GNU coreutils which break implementations that assume UTF-8 encoding. For example, having `\xc2` in a user name which is fine in some encodings:
> 
> ```
> $ LC_ALL=en_US.ISO-8859-1 printf '\xC2\n' | iconv --from-code=ISO-8859-1 --to-code=UTF-8 
> Â
> ```

We don't have in mind anything about UTF8.  We're concerned about `[[:cntrl:]]`, `[[:space:]]`, and `[[:punct:]]`.

collinfunk:
> I'll investigate the source code, in case there are other characters that would result in shadow misbehaving already, such as possibly . and @.

[...]

> We don't have in mind anything about UTF8. We're concerned about [[:cntrl:]], [[:space:]], and [[:punct:]].

I see. That sounds reasonable. I think my idea of "future characters" was a bit more dramatic than yours. :)

Thanks for the quick response.

PR #1412, opened 2025-12-07
updated Dutch translation

Frans-Spiesschaert:
updated po/nl.po based upon .pot file from 2025-12-04


alejandro-colomar (approved):
I don't know any Dutch, but this seems reasonable.  Thanks!

PR #1414, opened 2025-12-09
src/chsh.c: add argument for fail_exit()

ikerexxe:
Fixes: 374ca05c282f (2025-10-07; "src/chsh.c: SELinux file context for fail_exit()")

### DIFF
--- a/src/chsh.c
+++ b/src/chsh.c
@@ -67,8 +67,8 @@ static bool pw_locked = false;
 NORETURN static void fail_exit (int code, bool process_selinux);
 NORETURN static void usage (int status);
 static void new_fields (void);
-static bool shell_is_listed (const char *);
-static bool is_restricted_shell (const char *);
+static bool shell_is_listed (const char *, bool process_selinux);
+static bool is_restricted_shell (const char *, bool process_selinux);
 static void process_flags (int argc, char **argv, struct option_flags *flags);
 static void check_perms(const struct passwd *pw, const struct option_flags *flags);
 static void update_shell (const char *user, char *loginsh,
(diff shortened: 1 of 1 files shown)


ikerexxe:
Done

src/chsh.c
```diff
@@ -198,7 +198,7 @@ static bool shell_is_listed (const char *sh)
 
 #else /* without HAVE_VENDORDIR */
 
-static bool shell_is_listed (const char *sh)
+static bool shell_is_listed (const char *sh, MAYBE_UNUSED bool process_selinux)
 {
 	bool found = false;
 	char *cp;
```
alejandro-colomar, line 204: Why is this marked as MAYBE_UNUSED?   I see it unconditionally unused.

alejandro-colomar: This should be an unnamed parameter.

See for example

```diff
commit 3d39ab73f00fdfdc099543fdbe8143f77d9c0d26
Author: Alejandro Colomar <alx@kernel.org>
Date:   Fri Feb 21 06:46:53 2025 +0100

    lib/logind.c: Unname unused function parameter
    
    [Serge: preference to keep the names as a comment, but ok]
    Reviewed-by: Serge Hallyn <serge@hallyn.com>
    Signed-off-by: Alejandro Colomar <alx@kernel.org>

diff --git a/lib/logind.c b/lib/logind.c
index f4b187e5b..b1ed632bb 100644
--- a/lib/logind.c
+++ b/lib/logind.c
@@ -36,7 +36,7 @@ done:
     return ret;
 }
 
-unsigned long active_sessions_count(const char *name, MAYBE_UNUSED unsigned long limit)
+unsigned long active_sessions_count(const char *name, unsigned long)
 {
     struct passwd *pw;
     unsigned long count = 0;
```

hallyn (approved):
IMO the signatures should be the same everywhere, so I think it would be better to add MAYBE_UNUSED in the prototype and the other #ifdef branch as well.  But I know alx will hate that.

I'm also fine with merging as is.  Someone give me the latin phrase
for "when noone is happy, everyone is happy"

ikerexxe:
I'm going with the unnamed parameter approach. It seems like the less complex one while still maintaining a similar signature

alejandro-colomar:
Thanks!  I'll merge.

PR #1418, opened 2025-12-12
Add Global Config Option for creating User BTRFS Subvolumes in $HOME

### LINKED ISSUE #1162: Useradd Feature Request: add config option for btrfs subvolumes
Having the useradd cli flag to create btrfs subvolumes for homedirs is a great feature! This is a proposal to add the feature of allowing a variable in `/etc/default/useradd` to do this by default for users/distros that want to. The default would remain false for compatibility.

silverhadch:
Fixes #1162

### DIFF
--- a/man/useradd.8.xml
+++ b/man/useradd.8.xml
@@ -136,6 +136,33 @@
       </varlistentry>
       <varlistentry>
 	<term>
+	  <option>--btrfs-subvolume-home</option>
+	</term>
+	<listitem>
+	  <para>
+	    Create the user's home directory as a Btrfs subvolume.
+	  </para>
+	  <para>
+	    If this option is not specified,
+		<command>useradd</command> will follow the default behavior
+		defined by the <option>BTRFS_SUBVOLUME_HOME</option> variable
+		in <filename>/etc/default/useradd</filename>.
+	    If this variable is not set, the default value is no.
+	  </para>
+	  <para>
+	    When the <option>--btrfs-subvolume-home</option> command-line option
+	    is specified,
+		a Btrfs subvolume is created
+		regardless of any configuration file settings.
+	  </para>
+	  <para>
+	    Note: this feature works only if the underlying filesystem supports
+	    Btrfs subvolumes.
+	  </para>
+	</listitem>
+      </varlistentry>
+      <varlistentry>
+	<term>
 	  <option>-c</option>, <option>--comment</option>&nbsp;<replaceable>COMMENT</replaceable>
 	</term>
 	<listitem>
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -114,6 +114,7 @@ static const char *def_shell = "/bin/bash";
 static const char *def_template = SKEL_DIR;
 static const char *def_usrtemplate = USRSKELDIR;
 static const char *def_create_mail_spool = "yes";
+static const char *def_btrfs_subvolume_home = "no";
 static const char *def_log_init = "yes";
 
 static long def_inactive = -1;
@@ -222,6 +223,7 @@ static bool home_added = false;
 #define DSKEL			"SKEL"
 #define DUSRSKEL		"USRSKEL"
 #define DCREATE_MAIL_SPOOL	"CREATE_MAIL_SPOOL"
+#define DBTRFS_SUBVOLUME_HOME	"BTRFS_SUBVOLUME_HOME"
 #define DLOG_INIT		"LOG_INIT"
 
 /* local function prototypes */
@@ -456,6 +458,7 @@ get_defaults(const struct option_flags *flags)
 				def_usrtemplate = xstrdup(ccp);
 			}
 		}
+
 		/*
 		 * Create by default user mail spool or not ?
 		 */
@@ -464,6 +467,15 @@ get_defaults(const struct option_flags *flags)
 				ccp = "no";
 
 			def_create_mail_spool = xstrdup(ccp);
+		}
+
+		/*
+		 * Create home directories as Btrfs subvolumes by default?
+		 */
+		else if (streq(buf, DBTRFS_SUBVOLUME_HOME)) {
+			if (streq(ccp, ""))
+				ccp = "no";
+			def_btrfs_subvolume_home = xstrdup(ccp);
 		}
 
 		/*
@@ -500,6 +512,7 @@ static void show_defaults (void)
 	printf ("SKEL=%s\n", def_template);
 	printf ("USRSKEL=%s\n", def_usrtemplate);
 	printf ("CREATE_MAIL_SPOOL=%s\n", def_create_mail_spool);
+	printf ("BTRFS_SUBVOLUME_HOME=%s\n", def_btrfs_subvolume_home);
 	printf ("LOG_INIT=%s\n", def_log_init);
 }
 
@@ -523,6 +536,7 @@ set_defaults(void)
 	bool  out_skel = false;
 	bool  out_usrskel = false;
 	bool  out_create_mail_spool = false;
+	bool  out_btrfs_subvolume_home = false;
 	bool  out_log_init = false;
 	char  buf[1024];
 	char  *new_file = NULL;
@@ -639,6 +653,11 @@ set_defaults(void)
 			        DCREATE_MAIL_SPOOL "=%s\n",
 			        def_create_mail_spool);
 			out_create_mail_spool = true;
+		} else if (!out_btrfs_subvolume_home && streq(buf, DBTRFS_SUBVOLUME_HOME)) {
+			fprintf(ofp,
+			        DBTRFS_SUBVOLUME_HOME "=%s\n",
+			        def_btrfs_subvolume_home);
+			out_btrfs_subvolume_home = true;
 		} else if (!out_log_init && streq(buf, DLOG_INIT)) {
 			fprintf(ofp, DLOG_INIT "=%s\n", def_log_init);
 			out_log_init = true;
@@ -673,6 +692,8 @@ set_defaults(void)
 
 	if (!out_create_mail_spool)
 		fprintf (ofp, DCREATE_MAIL_SPOOL "=%s\n", def_create_mail_spool);
+	if (!out_btrfs_subvolume_home)
+		fprintf (ofp, DBTRFS_SUBVOLUME_HOME "=%s\n", def_btrfs_subvolume_home);
 	if (!out_log_init)
 		fprintf (ofp, DLOG_INIT "=%s\n", def_log_init);
 	/*
@@ -1430,6 +1451,9 @@ static void process_flags (int argc, char **argv, struct option_flags *flags)
 			anyflag = true;
 		}
 	}
+
+	if (!subvolflg && strcaseeq(def_btrfs_subvolume_home, "yes"))
+		subvolflg = true;
 
 	if (!gflg && !Nflg && !Uflg) {
 		/* Get the settings from login.defs */


Conan-Kudo:
Can you squash this into a single commit? And format the commit message properly?

Something like `src/useradd: Support config for creating home dirs as Btrfs subvolumes`

Conan-Kudo (approved):

alejandro-colomar:
Please add a `Closes: ` tag in the commit message for the github issue it closes.

silverhadch:
> Please add a `Closes: ` tag in the commit message for the github issue it closes.

Done.

src/useradd.c
```diff
@@ -1431,6 +1445,10 @@ static void process_flags (int argc, char **argv, struct option_flags *flags)
 		}
 	}
 
+	if (!subvolflg && strcaseeq(def_btrfs_subvolume_home, "yes")) {
+		subvolflg = true;
+	}
+
```
alejandro-colomar: Please, don't use braces in this `if`.

silverhadch: Resolved.

src/useradd.c
```diff
@@ -475,6 +477,9 @@ get_defaults(const struct option_flags *flags)
 
 			def_log_init = xstrdup(ccp);
 		}
+		else if (streq(buf, DBTRFS_SUBVOLUME_HOME)) {
+			def_btrfs_subvolume_home = xstrdup(ccp);
+		}
```
Conan-Kudo: Shouldn't this declare the default?

```suggestion
		else if (streq(buf, DBTRFS_SUBVOLUME_HOME)) {
			if (streq(ccp, ""))
				ccp = "no";
			def_btrfs_subvolume_home = xstrdup(ccp);
		}
```

silverhadch: Yeah I actually forgot to add the check.

Conan-Kudo:
This needs squashing again. 😄

src/useradd.c
silverhadch: I let the value be overwritten no matter what even if empty. Thanks for noticing.

Conan-Kudo:
It looks like this project requires DCO signoffs, so you'll want to amend the commit with `-s` and also add my signoff:

```
Signed-off-by: Neal Gompa <ngompa@velocitylimitless.com>
```

Also, can you please update the co-authored-by to: `Co-authored-by: Neal Gompa <ngompa@velocitylimitless.com>`?

src/useradd.c
```diff
@@ -1431,6 +1447,10 @@ static void process_flags (int argc, char **argv, struct option_flags *flags)
 		}
 	}
 
+	if (!subvolflg && strcaseeq(def_btrfs_subvolume_home, "yes"))
+		subvolflg = true;
+	
+
```
alejandro-colomar, line 1457: Please don't use 2 blank lines; just one.

src/useradd.c
```diff
@@ -116,6 +116,7 @@ static const char *def_usrtemplate = USRSKELDIR;
 static const char *def_create_mail_spool = "yes";
 static const char *def_log_init = "yes";
 
+static const char *def_btrfs_subvolume_home = "no";
 static long def_inactive = -1;
 static const char *def_expire = "";
```
alejandro-colomar, line 122: Is there any reason to put it precisely in that line?

The order seems inconsistent with the order used for other additions in this patch.

silverhadch: Neal already knows this but I was batteling the formating for like a day and messed up the structuring a bit by copy pasting my old code snippets back into the proper formating. Sorry.

silverhadch: This should fix it. Also on line 461 I noticed the comment is directly without a new line attached to the end of the Code Block when fixing my PRs Structure. I think thats intended?

silverhadch:
@alejandro-colomar 
Also I added the small comment to match the rest of the else if blocks.

src/useradd.c
alejandro-colomar: > Neal already knows this but I was batteling the formating for like a day and messed up the structuring a bit by copy pasting my old code snippets back into the proper formating. Sorry.

Thanks!  No problem.

> This should fix it. Also on line 461 I noticed the comment is directly without a new line attached to the end of the Code Block when fixing my PRs Structure. I think thats intended?

Nope; that seems accidental.  Thanks for catching that.  :)

silverhadch: Should I add the new line? :)

src/useradd.c
```diff
@@ -466,6 +468,15 @@ get_defaults(const struct option_flags *flags)
...
 
+		/*
+		 * Create home directories as Btrfs subvolumes by default?
+		 */
+		else if (streq(buf, DBTRFS_SUBVOLUME_HOME)) {
+			if (streq(ccp, ""))
+				ccp = "no";
+			def_btrfs_subvolume_home = xstrdup(ccp);
+		}
+
```
alejandro-colomar, line 480: We're in a freeze for the 4.19 release, which will happen around Christmas.

silverhadch:
Thanks for your time and help with polishing and reviewing the PR :)
You , too @Conan-Kudo !

src/useradd.c
alejandro-colomar: Yep.  (I see you added it already, right?)

alejandro-colomar:
BTW, I see that there's no documentation in the manual page for the command line flag.  It's independent of this, but since it is related, you may be interested in adding some documentation for the command line flag in a separate PR.

silverhadch:
Yeah I could do that in a week since then I will have a lot of free time.

src/useradd.c
silverhadch: Now I added the new line.

alejandro-colomar: Ahh, I thought I had seen it; now it's there.  Thanks!

man/login.defs.d/BTRFS_SUBVOLUME_HOME.xml
```diff
@@ -0,0 +1,22 @@
...
+      as Btrfs subvolumes instead of regular directories.
+    </para>
+    <para>
+      If set to <replaceable>yes</replaceable>, and the underlying
+      filesystem supports Btrfs subvolumes, the user's home directory
+      will be created as a separate Btrfs subvolume.
+    </para>
+    <para>
+      When the <option>--btrfs-subvolume-home</option> command-line option
+      is specified, a Btrfs subvolume is created regardless of this setting.
```
alejandro-colomar: Please use semantic newlines.

```
$ MANWIDTH=72 man man-pages| sed -n '/Use semantic newlines/,/^$/p'
   Use semantic newlines
     In the source of a manual page, new sentences should be started on
     new lines, long sentences should be split  into  lines  at  clause
     breaks  (commas,  semicolons, colons, and so on), and long clauses
     should be split at phrase boundaries.  This convention,  sometimes
     known as "semantic newlines", makes it easier to see the effect of
     patches, which often operate at the level of individual sentences,
     clauses, or phrases.

```

man/login.defs.d/BTRFS_SUBVOLUME_HOME.xml
```diff
@@ -0,0 +1,22 @@
+<!--
+   SPDX-FileCopyrightText: 2025, Hadi Chokr
+   SPDX-License-Identifier: BSD-3-Clause
+-->
+<varlistentry>
+  <term><option>BTRFS_SUBVOLUME_HOME</option> (boolean)</term>
```
ikerexxe: We should also indicate the default value. I don't think it's common in this project, but I think it's a good idea.

man/useradd.8.xml
```diff
@@ -134,6 +135,22 @@
...
+	</term>
+	<listitem>
+	  <para>
+	    Create the user's home directory as a Btrfs subvolume.
+	  </para>
+	  <para>
+	    If this option is not specified, <command>useradd</command>
+	    will follow the default behavior defined by the
+	    <option>BTRFS_SUBVOLUME_HOME</option> variable in
+	    <filename>/etc/default/useradd</filename>.
```
ikerexxe: I'm assuming you meant `/etc/login.defs`

man/login.defs.d/BTRFS_SUBVOLUME_HOME.xml
alejandro-colomar: Agree.  See tcp(7) for existing manual pages that do this.  We could follow that syntax.

For example:

```
tcp_abort_on_overflow (Boolean; default: disabled; since Linux 2.4)
```

silverhadch:
Thanks for the review. I will apply the sugesstions in the weekend since I am busy.

man/useradd.8.xml
```diff
@@ -134,6 +135,22 @@
...
+	</term>
+	<listitem>
+	  <para>
+	    Create the user's home directory as a Btrfs subvolume.
+	  </para>
+	  <para>
+	    If this option is not specified, <command>useradd</command>
+	    will follow the default behavior defined by the
+	    <option>BTRFS_SUBVOLUME_HOME</option> variable in
+	    <filename>/etc/login.defs</filename>.
```
alejandro-colomar: Humm, now that you changed this, I realized that something is wrong.  The code you've added implements it as part of /etc/default/useradd, not /etc/login.defs, AFAICS.

Conan-Kudo: Which is what is expected, right? The doc just needs adjustment.

alejandro-colomar: Most likely, yes.

silverhadch: Yeah I am kinda confused, Its an useradd only option. Should i just change it to /etc/default/useradd in the doc and thats it?

alejandro-colomar: Check how it's done in the documentation of other /etc/default/useradd configuration variables (IIRC, they are poorly dodumented; maybe there's no documentation at all).  I don't know from the top of my head.  If you need some help, I'll check when I have some time.

silverhadch: >  maybe there's no documentation at all

They seem to be all specified inside of useradd.8.xml so there isnt an seperate documentation of the varaibles. So should I just get rid of the file in the login.defs.d part in man and have it be at its commandline option instead thats how its done in the rest of the varaiables.

silverhadch:
> You should also add an explanation for this new option in the man pages, similar to what `CREATE_HOME` or `DEFAULT_HOME` have in `man/login.defs.d/`.
> 
> In addition, if you feel comfortable you could also create a test case for this new option.
> 
> Finally, as Alejandro has already mentioned, we are about to release a new version and we are focusing on stabilizing it, so probably this will get merged after that.

Wouldn't that require the thing running the test to be on BTRFS and /tmp which I have seen in the test being the Home Directory also being tmpfs and not BTRFS. Technically I could do an BTRFS Loopdevice Image and use that but that seems to be exetremly complex in contrast to other tests.

man/useradd.8.xml
alejandro-colomar: Yup, that seems correct.

alejandro-colomar:
> > You should also add an explanation for this new option in the man pages, similar to what `CREATE_HOME` or `DEFAULT_HOME` have in `man/login.defs.d/`.
> > In addition, if you feel comfortable you could also create a test case for this new option.
> > Finally, as Alejandro has already mentioned, we are about to release a new version and we are focusing on stabilizing it, so probably this will get merged after that.
> 
> Wouldn't that require the thing running the test to be on BTRFS and /tmp which I have seen in the test being the Home Directory also being tmpfs and not BTRFS. Technically I could do an BTRFS Loopdevice Image and use that but that seems to be exetremly complex in contrast to other tests.

Being so complex, I'd say just test it on your own computer, and if it passes, that's fine.  Please include in the commit message the shell session with which you tested it, so that one can reproduce the test in their computer.

Be aware that we might accidentally break something eventually, since we don't have regression tests, so it would be good if you (plural you, the users of this feature), report when it breaks.  I don't use this feature myself, so can't know if it breaks.

If you eventually want to write a test, that would of course be fine too.

silverhadch:
> > > You should also add an explanation for this new option in the man pages, similar to what `CREATE_HOME` or `DEFAULT_HOME` have in `man/login.defs.d/`.
> > > In addition, if you feel comfortable you could also create a test case for this new option.
> > > Finally, as Alejandro has already mentioned, we are about to release a new version and we are focusing on stabilizing it, so probably this will get merged after that.
> > 
> > 
> > Wouldn't that require the thing running the test to be on BTRFS and /tmp which I have seen in the test being the Home Directory also being tmpfs and not BTRFS. Technically I could do an BTRFS Loopdevice Image and use that but that seems to be exetremly complex in contrast to other tests.
> 
> Being so complex, I'd say just test it on your own computer, and if it passes, that's fine. Please include in the commit message the shell session with which you tested it, so that one can reproduce the test in their computer.
> 
> Be aware that we might accidentally break something eventually, since we don't have regression tests, so it would be good if you (plural you, the users of this feature), report when it breaks. I don't use this feature myself, so can't know if it breaks.
> 
> If you eventually want to write a test, that would of course be fine too.

Ahhh I will probably figure something out.

man/useradd.8.xml
silverhadch: Done.

silverhadch: Nevermind. The fornatting Issues have returned.

silverhadch: NOW. The formatting is repaired again.

silverhadch:
> > > You should also add an explanation for this new option in the man pages, similar to what `CREATE_HOME` or `DEFAULT_HOME` have in `man/login.defs.d/`.
> > > In addition, if you feel comfortable you could also create a test case for this new option.
> > > Finally, as Alejandro has already mentioned, we are about to release a new version and we are focusing on stabilizing it, so probably this will get merged after that.
> > 
> > 
> > Wouldn't that require the thing running the test to be on BTRFS and /tmp which I have seen in the test being the Home Directory also being tmpfs and not BTRFS. Technically I could do an BTRFS Loopdevice Image and use that but that seems to be exetremly complex in contrast to other tests.
> 
> Being so complex, I'd say just test it on your own computer, and if it passes, that's fine. Please include in the commit message the shell session with which you tested it, so that one can reproduce the test in their computer.
> 
> Be aware that we might accidentally break something eventually, since we don't have regression tests, so it would be good if you (plural you, the users of this feature), report when it breaks. I don't use this feature myself, so can't know if it breaks.
> 
> If you eventually want to write a test, that would of course be fine too.

Well we want to use it in a Distro and Fedora has an proposal etc. So eventually there will be enough Users that will show up when something breaks.

alejandro-colomar:
> > > > You should also add an explanation for this new option in the man pages, similar to what `CREATE_HOME` or `DEFAULT_HOME` have in `man/login.defs.d/`.
> > > > In addition, if you feel comfortable you could also create a test case for this new option.
> > > > Finally, as Alejandro has already mentioned, we are about to release a new version and we are focusing on stabilizing it, so probably this will get merged after that.
> > > 
> > > 
> > > Wouldn't that require the thing running the test to be on BTRFS and /tmp which I have seen in the test being the Home Directory also being tmpfs and not BTRFS. Technically I could do an BTRFS Loopdevice Image and use that but that seems to be exetremly complex in contrast to other tests.
> > 
> > 
> > Being so complex, I'd say just test it on your own computer, and if it passes, that's fine. Please include in the commit message the shell session with which you tested it, so that one can reproduce the test in their computer.
> > Be aware that we might accidentally break something eventually, since we don't have regression tests, so it would be good if you (plural you, the users of this feature), report when it breaks. I don't use this feature myself, so can't know if it breaks.
> > If you eventually want to write a test, that would of course be fine too.
> 
> Well we want to use it in a Distro and Fedora has an proposal etc. So eventually there will be enough Users that will show up when something breaks.

That sounds good.  Thanks!

man/useradd.8.xml
```diff
@@ -134,6 +134,32 @@
...
+	    <filename>/etc/default/useradd</filename>.
+	    If this variable is not set, the default value is no.
+	  </para>
+	  <para>
+	    When the <option>--btrfs-subvolume-home</option> command-line option
+	    is specified, a Btrfs subvolume is created regardless of any
+	    configuration file settings.
+	  </para>
+	  <para>
+	    Note: This feature only works if the underlying filesystem supports
```
alejandro-colomar: Please, use lowercase after colon.

man/useradd.8.xml
```diff
@@ -134,6 +134,32 @@
...
+	  <para>
+	    Create the user's home directory as a Btrfs subvolume.
+	  </para>
+	  <para>
+	    If this option is not specified, <command>useradd</command>
+	    will follow the default behavior defined by the
+	    <option>BTRFS_SUBVOLUME_HOME</option> variable in
+	    <filename>/etc/default/useradd</filename>.
+	    If this variable is not set, the default value is no.
+	  </para>
```
alejandro-colomar, line 151: I'd change the break lines of this paragraph as:

```
If this option is not specified,
<command>useradd</command> will follow the default behavior
defined by the <option>BTRFS_SUBVOLUME_HOME</option> variable
in <filename>/etc/default/useradd</filename>.
If this variable is not set, the default value is no.
```

man/useradd.8.xml
```diff
@@ -134,6 +134,32 @@
...
+	  <para>
+	    When the <option>--btrfs-subvolume-home</option> command-line option
+	    is specified, a Btrfs subvolume is created regardless of any
+	    configuration file settings.
```
alejandro-colomar: I'd change the line breaks as:

```
When the <option>--btrfs-subvolume-home</option> command-line option
is specified,
a Btrfs subvolume is created
regardless of any configuration file settings.
```

man/useradd.8.xml
alejandro-colomar: s/only works/works only/

alejandro-colomar:
> As a last note, you should guard the code with `#ifdef WITH_BTRFS`, just like with the command-line option

Would that change anything in observable behavior?

alejandro-colomar:
> > As a last note, you should guard the code with `#ifdef WITH_BTRFS`, just like with the command-line option
> 
> Would that change anything in observable behavior?

I think this code could be fine without wrapping in cpp(1) conditionals.  Please revise, as I'm not 100% sure, though.

silverhadch:
> Thanks for the update in the documentation. Now I see why we can't create a specific file for this option.
> 
> As a last note, you should guard the code with `#ifdef WITH_BTRFS`, just like with the command-line option

Pretty sure having the variable set to true will have zero effects. The Code in this PR only adds the ability to define it in /etc/default/useradd instead of only an cdmline argument. If with BTRFS is disabled the config has zero effect, no?

silverhadch:
@ikerexxe @alejandro-colomar
I just tested it. It has zero effect if you set the variable when you dont compile with BTRFS Support. It will simply create an regular user. create_home() will simply ignore the variable even if it is set since it cant do anything with it when compiled without BTRFS Support.

ikerexxe:
To avoid giving users the false impression that btrfs subvolumes are being created, we should conditionally compile the section. If the feature is compiled out, the associated help text and logic won't be visible to the user

Conan-Kudo:
That's probably out of scope for this PR, since it was never like that before.

alejandro-colomar:
> To avoid giving users the false impression that btrfs subvolumes are being created, we should conditionally compile the section.

On the other hand, there's <https://www.kernel.org/doc/html/latest/process/coding-style.html#conditional-compilation>, which recommends avoiding conditional compilation as much as possible.

> If the feature is compiled out, the associated help text and logic won't be visible to the user

By visible text you mean with strings(1)?
<https://www.man7.org/linux/man-pages/man1/strings.1.html>

Conan-Kudo:
Also, if it's not on btrfs, it's going to do nothing anyway. So I'm not sure conditional compilation is really warranted?

Conan-Kudo:
Now that 4.19 is released, can we get this merged? @ikerexxe, it'd be great to get this backported into Fedora as well.

alejandro-colomar:
> Now that 4.19 is released, can we get this merged? @ikerexxe, it'd be great to get this backported into Fedora as well.

The next release will be in 2.5 months, so unless you really need to backport it, you could wait 2.5 months.  Although it's up to Fedora (and @ikerexxe ) to decide, of course.

Conan-Kudo:
Unfortunately, the next release will be [after Fedora Linux 44 Final Freeze](https://fedorapeople.org/groups/schedule/f-44/f-44-key-tasks.html), so I'd rather have it backported now so we'd have it available for development in Fedora Linux 44.

ikerexxe:
@Conan-Kudo I'm fine with backporting this change to Fedora 44 before the Final Freeze.

Is this part of https://fedoraproject.org/wiki/Changes/BtrfsBootForCloud? Shouldn't there be a ticket for shadow under this umbrella?

Conan-Kudo:
> @Conan-Kudo I'm fine with backporting this change to Fedora 44 before the Final Freeze.
> 
> Is this part of https://fedoraproject.org/wiki/Changes/BtrfsBootForCloud? Shouldn't there be a ticket for shadow under this umbrella?

No, this is for https://pagure.io/fedora-btrfs/project/issue/50

We want to be able to do some testing of the feature before preparing a Change document for this.

ikerexxe:
> No, this is for https://pagure.io/fedora-btrfs/project/issue/50
> 
> We want to be able to do some testing of the feature before preparing a Change document for this.

Ok, then I'll backport it once shadow is rebased. Expect this to happen in the following days

Issue #887, opened 2023-12-30
Inconsistent password aging across tools


stoeckmann:
Some tools in shadow assume that a maximum time before password change of 10.000 (and sometimes above) is equal to "unlimited." Such a limit is set by `newusers` if `PASS_MAX_DAYS` is not set in /etc/login.defs:
```
# newusers <<< user:pass:1234:1234::/home/user:/bin/bash
# grep user /etc/passwd /etc/shadow
/etc/passwd:user:x:1234:1234::/home/user:/bin/bash
/etc/shadow:user:...:19721:0:10000:7:::
# chage -l user
Last password change                                    : Dec 30, 2023
Password expires                                        : never
Password inactive                                       : never
Account expires                                         : never
Minimum number of days between password change          : 0
Maximum number of days between password change          : 10000
Number of days of warning before password expires       : 7
```

According to this output, the password will never expire.

Let's adjust the inactivity timeout and set the last password change far into the past:
```
# chage -I 100 user
# chage -d 1 user
# chage -l user
Last password change                                    : Jan 02, 1970
Password expires                                        : never
Password inactive                                       : never
Account expires                                         : never
Minimum number of days between password change          : 0
Maximum number of days between password change          : 10000
Number of days of warning before password expires       : 7
# grep ^user /etc/shadow
/etc/passwd:user:x:1234:1234::/home/user:/bin/bash
/etc/shadow:user:...:1:0:10000:7:100::
```

According to this output, the password will still never expire.

But if you try to log in, the user account is expired and locked.

alejandro-colomar:
Do you propose any patch?

stoeckmann:
> Do you propose any patch?

It takes a patch series, since this `10000 means infinity` is scattered across the source tree. But let's start small. I've opened a pull request to sync at least the tools so no more `10000` defaults are added to new files.

PR #1426, opened 2025-12-15
lib/pwd2spwd.c: Avoid unneeded (magic) values

stoeckmann:
The `pwd2spwd` function is used within `passwd` to temporarily hold the newly entered password for a user and the user name. All other fields are unused. Set them to so called "uninitialized" values and avoid magic values to make it much easier to review that these values have no special meaning.

Helps https://github.com/shadow-maint/shadow/issues/887 by removing another occurrence of the magic number `10000` from source tree.

### DIFF
--- a/lib/pwd2spwd.c
+++ b/lib/pwd2spwd.c
@@ -12,7 +12,6 @@
 #ident "$Id$"
 
 #include <sys/types.h>
-#include "prototypes.h"
 #include "defines.h"
 #include <pwd.h>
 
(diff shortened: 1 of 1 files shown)


lib/pwd2spwd.c
```diff
@@ -34,22 +33,13 @@ struct spwd *pwd_to_spwd (const struct passwd *pw)
...
-		 * change */
-		sp.sp_lstchg = -1;
-	}
-
 	/*
 	 * These fields have no corresponding information in the password
 	 * file.  They are set to uninitialized values.
 	 */
+	sp.sp_lstchg = -1;
+	sp.sp_min = 0;
```
alejandro-colomar, line 41: Interesting.  I think we should also set it to 0 in

<https://github.com/shadow-maint/shadow/blob/f973c406c52a3562e746579579b24bfcba412158/lib/shadow/shadow/sgetspent.c#L83-L86>

PR #1429, opened 2025-12-15
passwd: Fix dead lock with other tools (no PAM)

stoeckmann:
The `passwd` tool locks shadow first, then passwd. The passwd file is only locked after shadow if the user entry is not found in shadow file. This means, the following conditions must be met:

1. /etc/shadow exists
2. A user exists in /etc/passwd but not in /etc/shadow

Internally, passwd would call `update_shadow` first and then falls back to `update_noshadow`. The former locks shadow, the latter locks passwd.

Take for example the tool `pwconv`, which locks passwd first and then shadow (most tools use this order). If both tools run at the same time, the following dead lock may occur:

1. `passwd` tool locks shadow
2. `pwconv` tool locks passwd
3. `passwd` tool tries to lock passwd and is blocked
4. `pwconv` tool tries to lock shadow and is blocked

In order to fix this, let `passwd` behave the same way: Lock passwd unconditionally first, then shadow. In most cases, the locked passwd file is not necessary. Locking shadow first, then unlocking it and locking passwd and then shadow as fallback seems too error prone.

An alternative would be to not adjust a user entry if /etc/shadow exists but no user entry exists there.

Since we talk about the setuid `passwd` tool here, I hope this gets a lot of review before being merged ...

### DIFF
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -488,18 +488,18 @@ NORETURN
 static void
 fail_exit (int status, bool process_selinux)
 {
-	if (pw_locked) {
-		if (pw_unlock (process_selinux) == 0) {
-			(void) fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, pw_dbname ());
-			SYSLOG ((LOG_ERR, "failed to unlock %s", pw_dbname ()));
-			/* continue */
-		}
-	}
-
 	if (spw_locked) {
 		if (spw_unlock (process_selinux) == 0) {
 			(void) fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, spw_dbname ());
 			SYSLOG ((LOG_ERR, "failed to unlock %s", spw_dbname ()));
+			/* continue */
+		}
+	}
+
+	if (pw_locked) {
+		if (pw_unlock (process_selinux) == 0) {
+			(void) fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, pw_dbname ());
+			SYSLOG ((LOG_ERR, "failed to unlock %s", pw_dbname ()));
 			/* continue */
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/passwd.c
```diff
@@ -136,10 +136,10 @@ static void check_password (const struct passwd *, const struct spwd *);
 static /*@observer@*/const char *pw_status (const char *);
 static void print_status (const struct passwd *);
 NORETURN static void fail_exit (int, bool);
-NORETURN static void oom (bool process_selinux);
+NORETURN static void oom (bool);
 static char *update_crypt_pw (char *, bool);
-static void update_noshadow(const struct option_flags *flags);
-static void update_shadow(const struct option_flags *flags);
+static void update_noshadow (const struct option_flags *);
+static void update_shadow (const struct option_flags *);
```
alejandro-colomar: I would prefer not adding whitespace before parentheses.  Rather, feel free to remove the existing ones.

src/passwd.c
```diff
@@ -555,13 +555,10 @@ static char *update_crypt_pw (char *cp, bool process_selinux)
...
 
 
-static void update_noshadow (const struct option_flags *flags)
+static void update_noshadow (bool process_selinux)
 {
 	const struct passwd *pw;
 	struct passwd *npw;
-	bool process_selinux;
-
-	process_selinux = !flags->chroot && !flags->prefix;
```
alejandro-colomar: Thanks!  The second patch LGTM.  For the 2nd patch only:

```
```

Would you mind sending that in a separate PR, so that I can merge it before reviewing the 3rd and 4th patches?

src/passwd.c
```diff
@@ -564,7 +564,7 @@ static void update_noshadow (bool process_selinux)
 		(void) fprintf (stderr,
 		                _("%s: cannot lock %s; try again later.\n"),
 		                Prog, pw_dbname ());
-		exit (E_PWDBUSY);
+		fail_exit (E_PWDBUSY, process_selinux);
```
alejandro-colomar: This should include

```
Fixes: 45c6603cc86c (2007-10-07; "[svn-upgrade] Integrating new upstream version, shadow (19990709)")
```

alejandro-colomar: Can you please also include a reproducer in the commit message, if possible?

src/passwd.c
stoeckmann: > Would you mind sending that in a separate PR, so that I can merge it before reviewing the 3rd and 4th patches?

See this PR: https://github.com/shadow-maint/shadow/pull/1431

alejandro-colomar:
Please rebase.

src/passwd.c
stoeckmann: Adjusted the commit message. Reproduce is missing (have none so far).

src/passwd.c
```diff
@@ -515,6 +515,87 @@ oom (bool process_selinux)
...
+			fail_exit (E_FAILURE, process_selinux);
+		}
+	}
+}
+
+/*
+ * close_files - close all of the files that were opened
+ *
+ *	close_files() closes all of the files that were opened for this
+ *	user. This causes a possibly modified entry to be written out.
```
alejandro-colomar: Please use two spaces after period.

alejandro-colomar: See also: <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/patches/description#n44>

src/passwd.c
```diff
@@ -515,6 +515,87 @@ oom (bool process_selinux)
...
+	}
+	pw_locked = true;
+	if (pw_open (O_CREAT | O_RDWR) == 0) {
+		(void) fprintf (stderr,
+		                _("%s: cannot open %s\n"),
+		                Prog, pw_dbname ());
+		SYSLOG ((LOG_WARN, "cannot open %s", pw_dbname ()));
+		fail_exit (E_MISSING, process_selinux);
+	}
+	if (spw_file_present ()) {
```
alejandro-colomar: If you invert this conditional, and write it as

```c
if (!spw_file_present())
	return;
```

then the code below will have the same indentation as before, which will make it easier to review the patch with `--color-moved`.

src/passwd.c
stoeckmann: Adjusted

src/passwd.c
stoeckmann: Adjusted

src/passwd.c
```diff
@@ -1117,11 +1132,9 @@ main(int argc, char **argv)
...
 		exit (E_NOPERM);
 	}
-	if (spw_file_present ()) {
-		update_shadow (process_selinux);
-	} else {
-		update_noshadow (process_selinux);
-	}
+	open_files (process_selinux);
+	update_shadow (process_selinux);
+	close_files (process_selinux);
```
alejandro-colomar: Please remove the spaces here.

src/passwd.c
```diff
@@ -592,55 +660,17 @@ static void update_noshadow(bool process_selinux)
...
-	spw_locked = true;
-	if (spw_open (O_CREAT | O_RDWR) == 0) {
-		(void) fprintf (stderr,
-		                _("%s: cannot open %s\n"),
-		                Prog, spw_dbname ());
-		SYSLOG ((LOG_WARN, "cannot open %s", spw_dbname ()));
-		fail_exit (E_FAILURE, process_selinux);
-	}
-	sp = spw_locate (name);
+	sp = spw_locked ? spw_locate (name) : NULL;
```
alejandro-colomar: Please remove the space here.

alejandro-colomar (approved):
LGTM.  Thanks!

```
```

(independent of the minor whitespace changes)

src/passwd.c
stoeckmann: Adjusted

src/passwd.c
stoeckmann: Adjusted

stoeckmann:
> ```
> ```

Added to commit messages

PR #1430, opened 2025-12-15
passwd: Fix TOCTOU race condition (no PAM)

stoeckmann:
The `passwd` tool checks if the password of a user may be changed before locking the passwd/shadow files. This leaves a time window to perform the same action twice (e.g. circumventing PASS_MIN_DAYS limit) or to circumvent a locked password by an administrator.

Perform the check after the lock again. This keeps the behavior as it is today for a user and also prevents the race condition.

Again, since this is `passwd`, a very close review is appreciated!

How to reproduce:

Scenario 1, PASS_MIN_DAYS:

1. Set minimum password change time of a user to 1 day
```
# chage -m 1 user
```
2. As user, start password changing in terminal 1 and 2
```
t1 $ passwd
Changing password for user
Old password: <enter old password here>
Enter the new password (minimum of 5 characters)
Please use a combination of upper and lower case letters and numbers.
New password: <stop typing here, no enter!>
```

```
t2 $ passwd
Changing password for user
Old password: <enter old password here>
Enter the new password (minimum of 5 characters)
Please use a combination of upper and lower case letters and numbers.
New password: <stop typing here, no enter!>
```

3. Finish password changing in terminal 1
4. Finish password changing in terminal 2

Both commands succeed, but one should be blocked.

Scenario 2, Prevent Account Locking:

1. Let user start changing the password
```
$ passwd
Changing password for user
Old password: <enter old password here>
Enter the new password (minimum of 5 characters)
Please use a combination of upper and lower case letters and numbers.
New password: <stop typing here, no enter!>
```
2. Lock user account as root
```
# passwd -l user
```
3. Finish password changing as user

The account is unlocked again.

### DIFF
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -649,6 +649,7 @@ static void update_noshadow(bool process_selinux)
 		                Prog, name, pw_dbname ());
 		fail_exit (E_NOPERM, process_selinux);
 	}
+	check_password(pw, pwd_to_spwd(pw), process_selinux);
 	npw = __pw_dup (pw);
 	if (NULL == npw) {
 		oom (process_selinux);
@@ -664,6 +665,7 @@ static void update_noshadow(bool process_selinux)
 
 static void update_shadow(bool process_selinux)
 {
+	const struct passwd pw = { .pw_passwd = SHADOW_PASSWD_STRING };
 	const struct spwd *sp;
 	struct spwd *nsp;
 
@@ -673,6 +675,7 @@ static void update_shadow(bool process_selinux)
 		update_noshadow (process_selinux);
 		return;
 	}
+	check_password(&pw, sp, process_selinux);
 	nsp = __spw_dup (sp);
 	if (NULL == nsp) {
 		oom (process_selinux);


src/passwd.c
```diff
@@ -437,11 +439,14 @@ static void check_password (const struct passwd *pw, const struct spwd *sp)
 			(void) fprintf (stderr,
 			                _("The password for %s cannot be changed yet.\n"),
 			                pw->pw_name);
-			SYSLOG ((LOG_WARN, "now < minimum age for '%s'", pw->pw_name));
+			SYSLOG ((LOG_WARN, "now < minimum age for '%s'", sp->sp_namp));
```
alejandro-colomar, line 440: This seems to be unrelated to this commit message.  Please use separate commits for typo fixes.

alejandro-colomar: Oh, sorry, I misread.

stoeckmann: Yeah, I've thought about making an extra commit for this one, but it would be hard to explain in a commit message why I would want to do that. Let me know what you think.

alejandro-colomar: In general, the more you split, the easier it is to review.  (With some exceptions, of course.)  If you feel you can split it, I'd prefer it split.  Thanks!

src/passwd.c
```diff
@@ -393,14 +393,15 @@ static void check_password (const struct passwd *pw, const struct spwd *sp)
 	 * PAM) was specified, do nothing. --marekm
 	 */
 	if (kflg && (0 == exp_status)) {
-		exit (E_SUCCESS);
+		errno = E_SUCCESS;
+		return -1;
 	}
```
alejandro-colomar, line 397: I think I'd prefer `fail_exit`ing here directly instead of returning.  That would allow us to not use `errno`.

In the future, I want to remove this feature entirely.

src/passwd.c
```diff
@@ -649,6 +657,8 @@ static void update_shadow(const struct option_flags *flags)
 		spw_locked = false;
 		return;
 	}
+	if (check_password (&pw, sp) < 0)
```
alejandro-colomar: Please use `== -1`, as it's more explicit (and thus readable).

src/passwd.c
```diff
@@ -614,6 +621,7 @@ static void update_noshadow(const struct option_flags *flags)
 
 static void update_shadow(const struct option_flags *flags)
 {
+	const struct passwd pw = { .pw_passwd = SHADOW_PASSWD_STRING, 0 };
```
alejandro-colomar: I would rather use `const struct passwd pw = { .pw_passwd = SHADOW_PASSWD_STRING };`

src/passwd.c
stoeckmann: This requires to pass in `process_selinux` flag, which I tried to avoid. Adjusted.

src/passwd.c
stoeckmann: Function is void again due to calling exit directly in case of error.

alejandro-colomar:
This seems to need rebasing after we merged the other PR.

stoeckmann:
> This seems to need rebasing after we merged the other PR.

Rebased

src/passwd.c
```diff
@@ -437,7 +437,7 @@ static void check_password (const struct passwd *pw, const struct spwd *sp)
 			(void) fprintf (stderr,
 			                _("The password for %s cannot be changed yet.\n"),
 			                pw->pw_name);
-			SYSLOG ((LOG_WARN, "now < minimum age for '%s'", pw->pw_name));
+			SYSLOG ((LOG_WARN, "now < minimum age for '%s'", sp->sp_namp));
```
alejandro-colomar, line 440: For commit 1:

```
```

src/passwd.c
```diff
@@ -393,7 +393,7 @@ static void check_password (const struct passwd *pw, const struct spwd *sp)
 	 * PAM) was specified, do nothing. --marekm
 	 */
 	if (kflg && (0 == exp_status)) {
-		exit (E_SUCCESS);
+		fail_exit(E_SUCCESS, process_selinux);
```
alejandro-colomar, line 396: Should the fail_exit changes go in a separate commit?

src/passwd.c
```diff
@@ -439,7 +439,7 @@ static void check_password (const struct passwd *pw, const struct spwd *sp)
 			                pw->pw_name);
```
alejandro-colomar: We should also initialize `.pw_name`.  Otherwise, we'll have a null pointer there.

alejandro-colomar: On the other hand, you may want to change that to use spwd instead.

stoeckmann: Nice spotting! Indeed.

src/passwd.c
stoeckmann: Adjusted

src/passwd.c
alejandro-colomar: Actually, I think you had already done that.  I just forgot that commit 1 was in this same PR.  :-)

PR #1431, opened 2025-12-15
passwd: Pass process_selinux flag directly

stoeckmann:
No need to re-evaluate option_flags in functions. Unifies checks and simplifies code.

Reviewed-by: @alejandro-colomar


alejandro-colomar:
Thanks!

PR #1436, opened 2025-12-19
Ignore IDE related directories and files

stoeckmann:
Using clangd as language server with Code OSS adds three additional filesystem entries. Add them to the list if ignored files so that they will never be merged.

Having any form of IDE is helpful with these additional layers of code around bare library and system calls. The combination of clangd and (vs)code is a rather popular one, so this shouldn't be too specific to my use case. In fact, that's how I ended up with this quite useful combination.

In case somebody else wants to be able to jump directly to function definitions and get warnings from clangd (which should be fixed in other pull requests):

Instructions based on Arch Linux
```
$ pacman -S bear clang code # install https://github.com/rizsotto/Bear and llvm-clang which should contain clangd
<install Kyling Clangd extension in code>
$ cd shadow # stay in top level of source directory
$ CC=clang ./configure
$ bear -- make
```

The bear wrapper around make catches all relevant command line arguments passed to the compiler in `compile_commands.json` (due to clangd I chose to compile with clang as well) and stores temporary data in `.cache` for faster symbol lookups. Any further options or plugins which might be added for debugging or run options are stored in `.vscode` by Code OSS.

Teaser of what clangd with code will do for us:

code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,9 +5,11 @@ lib*.a
 *.la
 *.mo
 *.gmo
+.cache
 .deps
 .libs
 .dirstamp
+.vscode
 
 *.patch
 *.rej
@@ -21,6 +23,7 @@ test-driver
 /ABOUT-NLS
 /aclocal.m4
 /autom4te.cache
+/compile_commands.json
 /config.cache
 /config.h
 /config.h.in


jubalh:
Wouldn't it make more sense to add this to your global gitignore file? Otherwise each project a person contributes to needs to accumulate all kinds of editor/ide/tools specific files and directories in its gitignore.

Personally I believe it makes more sense if developer X who uses tooling1 adds the corresponding files to their global gitignore. Developer Y who use tooling2 does the same.

These files are based on personal preferences and not on the project and the specific mandatory project tooling itself.

stoeckmann:
> These files are based on personal preferences and not on the project and the specific mandatory project tooling itself.

Yeah, true. Except for the `*.patch` entry and hardly for `*.orig` and `*.rej` lines either. But I guess it makes sense to keep it the way it is.

alejandro-colomar:
> The combination of clangd and (vs)code is a rather popular one, so this shouldn't be too specific to my use case. In fact, that's how I ended up with this quite useful combination.
> 
> In case somebody else wants to be able to jump directly to function definitions and get warnings from clangd (which should be fixed in other pull requests):

You may be interested in grepc(1).  It's a PCRE2-based script that finds function definitions (and more) in files (or stdin, if no files specified), and prints them to stdout.

```c
$ grepc -n error_acl .
./lib/copydir.c:101:format_attr(printf, 2, 3)
static void
error_acl(struct error_context *, const char *fmt, ...)
{
	va_list ap;
	FILE *shadow_logfd = log_get_logfd();

	/* ignore the case when destination does not support ACLs
	 * or extended attributes */
	if (ENOTSUP == errno) {
		errno = 0;
		return;
	}

	va_start (ap, fmt);
	(void) fprintf (shadow_logfd, _("%s: "), log_get_progname());
	if (vfprintf (shadow_logfd, fmt, ap) != 0) {
		(void) fputs (_(": "), shadow_logfd);
	}
	(void) fprintf(shadow_logfd, "%s\n", strerrno());
	va_end (ap);
}
```

It will be released as part of the next release of the Linux man-pages project, and distros will likely package it soon.

> 
> Instructions based on Arch Linux
> 
> ```
> $ pacman -S bear clang code # install https://github.com/rizsotto/Bear and llvm-clang which should contain clangd
> <install Kyling Clangd extension in code>
> $ cd shadow # stay in top level of source directory
> $ CC=clang ./configure
> $ bear -- make
> ```
> 
> The bear wrapper around make catches all relevant command line arguments passed to the compiler in `compile_commands.json` (due to clangd I chose to compile with clang as well) and stores temporary data in `.cache` for faster symbol lookups. Any further options or plugins which might be added for debugging or run options are stored in `.vscode` by Code OSS.

grepc(1) has no state at all; no cache; no index.  It runs grep(1) --actually, pcre2grep(1)-- on every call, and it's really fast.

```sh
$ time grepc -n error_acl . >/dev/null

real	0m0.180s
user	0m0.145s
sys	0m0.064s
```

If you want to try it before it is packaged for your distro, you can install it from <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/>, with `make -R -j4 && sudo make -R -j4 install-bin`.

PR #1437, opened 2025-12-19
Fix documentation style

stoeckmann:
Always start a sentence with lowercase letter after `Note: `. This unifies all occurrences.
Also fix style passwd's manual page.

No functional change.

### DIFF
--- a/man/passwd.5.xml
+++ b/man/passwd.5.xml
@@ -147,7 +147,7 @@
       <varlistentry>
 	<term><filename>/etc/shadow</filename></term>
 	<listitem>
-	  <para>optional encrypted password file</para>
+	  <para>Optional encrypted password file.</para>
 	</listitem>
       </varlistentry>
       <varlistentry>


doc/HOWTO
```diff
@@ -1,4 +1,4 @@
-[ Note: the installation instructions in this document are somewhat
+[ Note: The installation instructions in this document are somewhat
```
alejandro-colomar: I checked to see what's more common in our current documentation, and lowercase after `:` seems to be the most common:

```
$ find | grep '[1-9]\.xml$' | xargs grep -h ': [a-zA-Z]\+' | grep -v -e SPDX | grep -o ': [a-zA-Z]\+' | sort | uniq -c
      1 : In
      9 : NIS
      3 : can
      1 : file
      2 : group
      2 : if
      2 : invalid
      1 : it
     12 : max
      1 : maximum
      5 : no
      1 : pam
      1 : process
      1 : specified
      1 : strings
      1 : success
      1 : testuser
      1 : the
```

Or did you find a different pattern?

Also, at least in Spanish, the rule is that after colon you start with lowercase (<https://www.rae.es/dpd/dos%20puntos>).  I ignore if English has different rules.

stoeckmann: I've used `[^Identifier|%s| ]: [A-Z]` to reduce the amount of false positives and then checked manually (for uppercase, compared to lowercase). I've included source and tests as well.

Looks like British English also recommends to continue with lowercase, so I'm fine with that. In either way, keep it unified.

ikerexxe: This uppercase nonsense comes from marketing & AI. If you speak to people who have really studied the English language, they will tell you that after a colon, lowercase letters are used (with a few exceptions).

stoeckmann: > This uppercase nonsense comes from marketing & AI. If you speak to people who have really studied the English language, they will tell you that after a colon, lowercase letters are used (with a few exceptions).

And from people with German background. :D

PR #1439, opened 2025-12-19
PASS_MAX_LEN.xml: Explain PASS_MAX_LEN use cases

stoeckmann:
The `PASS_MAX_LEN` is effectively only used for DES. Do not describe it in a way that makes it sound like `MD_CRYPT_ENAB=yes` is required to disable it. Any other `ENCRYPT_METHOD` disables it as well.

Also, even for DES, `PASS_MAX_LEN` requires `OBSCURE_CHECKS_ENAB` to have any effect.

Even more, `PASS_MIN_LEN` and `PASS_MAX_LEN` are only used for user passwords. Group passwords are not checked.

Note: All of this is actually true even if compiled with PAM if command line arguments change root. But if compiled with PAM support, this section is not added to manual pages... Since this is true for some more files, it's not part of this commit.

Clarifies use cases spotted in https://github.com/shadow-maint/shadow/issues/886 a bit.

### DIFF
--- a/etc/login.defs
+++ b/etc/login.defs
@@ -276,7 +276,7 @@ PASS_ALWAYS_WARN	yes
 #
 # Number of significant characters in the password for crypt().
 # Default is 8, don't change unless your crypt() is better.
-# Ignored if MD5_CRYPT_ENAB set to "yes".
+# Only used for DES encryption algorithm.
 #
 #PASS_MAX_LEN		8
 


alejandro-colomar:
BTW, do you have any plans to remove support for DES?  If not, I'll have a look at it myself.

stoeckmann:
> BTW, do you have any plans to remove support for DES? If not, I'll have a look at it myself.

Not planing to do it on my own, but greatly looking forward to it. I've thought about merging all these if-else blocks to figure out which method should be used into a single function. It's horrible and so easy to miss something... Removing DES would be quite easy then.

man/login.defs.d/PASS_MAX_LEN.xml
```diff
@@ -10,10 +10,26 @@
...
       Number of significant characters in the password for crypt().
+    </para>
+    <para>
       <option>PASS_MAX_LEN</option> is 8 by default. Don't change unless
-      your crypt() is better. This is ignored if
+      your crypt() is better.
+    </para>
+    <para>
+      <option>PASS_MAX_LEN</option> is only used for DES. It is ignored
+      if an encryption algorithm other than DES is used, either because
```
alejandro-colomar: Please use semantic newlines for new text.  See man-pages(7):

```
$ man man-pages | sed -n '/Use semantic newlines/,/^$/p'
   Use semantic newlines

```

etc/login.defs
```diff
@@ -276,7 +276,7 @@ PASS_ALWAYS_WARN	yes
 #
 # Number of significant characters in the password for crypt().
 # Default is 8, don't change unless your crypt() is better.
-# Ignored if MD5_CRYPT_ENAB set to "yes".
+# Only used for DES encryption algorithm.
```
alejandro-colomar, line 279: Would you mind clarifying in the commit message how this is only enabled for DES?  The source code isn't very explicit to me.  It would be helpful to have some analysis written in the commit message, for verifying.

stoeckmann: Links added to lib/obscure.c in commit message (and which line number is relevant for which aspect) and mentioned that gpasswd.c does not call obscure at all.

man/login.defs.d/PASS_MAX_LEN.xml
stoeckmann: Adjusted (and again to include the duplicated code part of src/passwd.c for output as well)

PR #1440, opened 2025-12-19
passwd: Clarify password strength check comment

stoeckmann:
- The total number of password change tries can be configured
- Except min length, password strength checks can be disabled
- Even the root user can have password strength checks...
- ... except in some cases (stdin, command line arguments)

In general, this code does not run for PAM, except root directory is modified through command line arguments by root user.

Helps https://github.com/shadow-maint/shadow/issues/886


alejandro-colomar (approved):
Thanks!  I might have removed the comment, but fixing it is better than nothing.

PR #1441, opened 2025-12-19
passwd: Fix wrong handling of PASS_MIN_LEN (no PAM)

stoeckmann:
If `PASS_MIN_LEN` is not defined in `/etc/login.defs`, then `passwd` diverges between its output (minimum of 5 characters) and actual check (minimum of 0 characters).

This could be easily fixed in the output case, but it makes much more sense to unify the retrieval of minimum and maximum allowed lengths. For this, I have introduced `obscure_get_range` which returns actual range of allowed password length. The latter also depends on encryption method and if the obscure checks are actually enabled or not.

Fixes issues seen in https://github.com/shadow-maint/shadow/issues/886 and also the edge case when `PASS_MIN_LEN` is set to `-1`. A cast to size_t would turn every password too short.

### DIFF
--- a/lib/obscure.c
+++ b/lib/obscure.c
@@ -115,17 +115,17 @@ static /*@observer@*//*@null@*/const char *obscure_msg (
 	/*@notnull@*/const char *old,
 	/*@notnull@*/const char *new)
 {
-	size_t maxlen, oldlen, newlen;
-	int minlen;
+	int maxlen, minlen;
+	size_t oldlen, newlen;
 	char *new1, *old1;
 	const char *msg;
-	const char *result;
 
 	oldlen = strlen (old);
 	newlen = strlen (new);
 
-	minlen = getdef_num ("PASS_MIN_LEN", 0);
-	if (minlen != -1 && newlen < (size_t) minlen) {
+	obscure_get_range(&minlen, &maxlen);
+
+	if (newlen < (size_t) minlen) {
 		return _("too short");
 	}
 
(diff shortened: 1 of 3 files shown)


lib/obscure.c
```diff
@@ -116,14 +116,16 @@ static /*@observer@*//*@null@*/const char *obscure_msg (
...
 	char *new1, *old1;
 	const char *msg;
 	const char *result;
 
 	oldlen = strlen (old);
 	newlen = strlen (new);
 
-	if (newlen < (size_t) getdef_num ("PASS_MIN_LEN", 0)) {
+	minlen = getdef_num ("PASS_MIN_LEN", 0);
+	if (minlen >= 0 && newlen < (size_t) minlen) {
```
alejandro-colomar: Is -2 also allowed?  Otherwise, I'd find `!= -1` more readable by being more explicit.

lib/obscure.c
```diff
@@ -115,17 +115,17 @@ static /*@observer@*//*@null@*/const char *obscure_msg (
 	/*@notnull@*/const char *old,
 	/*@notnull@*/const char *new)
 {
-	size_t maxlen, oldlen, newlen;
```
alejandro-colomar, line 118: Could you please include in the commit message a reproducer of the inconsistency that we're fixing?

lib/obscure.c
stoeckmann: Adjusted

lib/obscure.c
stoeckmann: Adjusted as well

lib/obscure.c
```diff
@@ -116,14 +116,16 @@ static /*@observer@*//*@null@*/const char *obscure_msg (
...
 	char *new1, *old1;
 	const char *msg;
 	const char *result;
 
 	oldlen = strlen (old);
 	newlen = strlen (new);
 
-	if (newlen < (size_t) getdef_num ("PASS_MIN_LEN", 0)) {
+	minlen = getdef_num ("PASS_MIN_LEN", 0);
+	if (minlen != -1 && newlen < (size_t) minlen) {
```
alejandro-colomar: For patch 1:

```
```

Thanks!

lib/obscure.c
```diff
@@ -209,3 +190,50 @@ obscure(const char *old, const char *new)
...
+ * obscure_get_range - retrieve min and max password lengths
+ *
+ *  Returns minimum and maximum allowed lengths of a password
+ *  to pass obscure checks.
+ */
+void
+obscure_get_range(int *minlen, int *maxlen)
+{
+	const char *method;
+	int val;
```
alejandro-colomar: Please align as

```c
	int          val;
	const char  *method;
```

lib/obscure.c
```diff
@@ -209,3 +190,50 @@ obscure(const char *old, const char *new)
...
+
+	/* Minimum length is 0, even if -1 is configured. */
+	val = getdef_num("PASS_MIN_LEN", 0);
+	*minlen = val < 0 ? 0 : val;
```
alejandro-colomar: Please compare explicitly to ==-1.

lib/obscure.c
stoeckmann: Adjusted

lib/obscure.c
stoeckmann: Adjusted

PR #1443, opened 2025-12-19
src/chgpasswd.c: match chpasswd behaviour for EOF handling

ikerexxe:
chgpasswd was incorrectly reporting "line too long" errors for the last line of input files that didn't end with a newline character. This differed from chpasswd, which gracefully handles missing trailing newlines when at EOF.

Updated chgpasswd.c to check feof(stdin) before reporting line length errors, consistent with chpasswd.c behavior.

The inconsistency was found while developing https://github.com/shadow-maint/shadow/pull/1357

### DIFF
--- a/src/chgpasswd.c
+++ b/src/chgpasswd.c
@@ -479,10 +479,18 @@ int main (int argc, char **argv)
 	while (fgets(buf, sizeof(buf), stdin) != NULL) {
 		line++;
 		if (stpsep(buf, "\n") == NULL) {
-			fprintf (stderr, _("%s: line %jd: line too long\n"),
-			         Prog, line);
-			errors = true;
-			continue;
+			if (feof (stdin) == 0) {
+				// Drop all remaining characters on this line.
+				while (fgets(buf, sizeof(buf), stdin) != NULL) {
+					if (strchr(buf, '\n'))
+						break;
+				}
+
+				fprintf (stderr, _("%s: line %jd: line too long\n"),
+				         Prog, line);
+				errors = true;
+				continue;
+			}
 		}
 
 		/*


src/chgpasswd.c
```diff
@@ -479,10 +479,18 @@ int main (int argc, char **argv)
...
+					if (strchr(buf, '\n'))
+						break;
+				}
+
+				fprintf (stderr, _("%s: line %jd: line too long\n"),
+				         Prog, line);
+				errors = true;
+				continue;
+			}
 		}
```
alejandro-colomar, line 494: I'd rather go the other way around.  Make chpasswd(8) reject non-text files.

POSIX says <https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap03.html#tag_03_387>

```
3.387 Text File

A file that contains characters organized into zero or more lines.
The lines do not contain NUL characters and none can exceed
{LINE_MAX} bytes in length, including the <newline> character.
Although POSIX.1-2024 does not distinguish between text files and
binary files (see the ISO C standard), many utilities only produce
predictable or meaningful output when operating on text files.
The standard utilities that have such restrictions always specify
"text files" in their STDIN or INPUT FILES sections.
```

If the input is not a text file, the behavior will not be predictable or meaningful, per the standard.

That keeps our source code simpler, and thus with less bugs.

BTW, there's an interesting story from the C Committee about this:

A C program that doesn't end in a newline has always had undefined behavior (C89 .. C23).  A few months ago, the committee discussed how to fix this, because UB was unnecessary.  We had two options: allow it and behave as if there was a trailing newline, or require a compiler diagnostic.  I voted for the latter, which would be consistent with what POSIX requires.  However, the majority of the committee voted for allowing it.  A few months after that, someone realized that allowing it had bad consequences in some corner cases, and added more UB than it had removed, so we need to fix this.
Here's a proposal we'll vote soon: <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3633.txt>.

TL;DR: Keep it simple.

alejandro-colomar:
-  This is a duplicate of <https://github.com/shadow-maint/shadow/pull/1358>.

ikerexxe:
As I was implementing this, it seemed familiar, as if I had done it before, but I thought I had dreamed it :facepalm:

PR #1445, opened 2025-12-20
newgrp, sg: Open audit socket after check_fds

stoeckmann:
The check_fds function is supposed to ensure that fds 0, 1, and 2 are opened in a well-defined state, i.e. either they are already connected to supposed input/output files or will be connected to /dev/null if not.

Opening the audit socket before checking the fds allows the audit socket to get one of these numbers.

Avoid this by opening the audit socket after the check.

In general, this check is already covered by system libraries, but this proof of concept works for root user. Note the different states of the file descriptor 2.

In bash or another shell that interprets `2>&-` as closing stderr with shadow + audit support, e.g. Arch Linux:
```
sg bin 'ls -l /proc/self/fd'
sg bin 'ls -l /proc/self/fd' 2>/dev/null
sg bin 'ls -l /proc/self/fd' 2>&-
```

### DIFF
--- a/src/gpasswd.c
+++ b/src/gpasswd.c
@@ -885,10 +885,6 @@ int main (int argc, char **argv)
 	struct passwd *pw = NULL;
 	struct option_flags  flags = {.chroot = false};
 
-#ifdef WITH_AUDIT
-	audit_help_open ();
-#endif
-
 	sanitize_env ();
 	check_fds ();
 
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
In the commit message for commit 1:

> Not the different states of the
file descriptor 2.

Did you mean s/Not/Note/?

stoeckmann:
> Did you mean s/Not/Note/?

Thanks, yes. Fixed

PR #1446, opened 2025-12-23
Preserve original error messages for logging

stoeckmann:
Different error messages based on errno may be logged on stderr and syslog, because printing error messages on stderr may modify errno, which is reevaluated for syslog messages.

Keep the error message in its own variable to preserve the original value.

While at it, fixed two error messages.

Proof of Concept (based on groupadd):

1. Create a target file which cannot be written to anymore (Quota would be easier but this can be reproduced without such a feature):
```
# create temporary directory
BASEDIR=$(mktemp -d)

# create a very small filesystem
dd if=/dev/zero of=$BASEDIR/mini.img bs=1024 count=1024
mkfs.ext2 $BASEDIR/mini.img

# mount filesystem and fill it up completely
sudo losetup -f $BASEDIR/mini.img
mkdir $BASEDIR/mnt
sudo mount /dev/loop0 $BASEDIR/mnt
sudo install -o $(id -u) /dev/null $BASEDIR/mnt/output
dd if=/dev/zero of=$BASEDIR/mnt/output
```

2. Prepare a reproducible `/etc/group` error, e.g. by having a directory instead of a file
```
# create /etc/group as directory to trigger an error
mkdir -p $BASEDIR/etc/group
```

3. Try to add a group and make sure that stderr logging triggers yet another error
```
# try to create a group and trigger another error when writing to stderr
groupadd -P $BASEDIR test 2>>$BASEDIR/mnt/output
```

4. Check the journal (either through `journalctl` or by open the log files, depending on system)
```
journalctl
```

You should be able to find a line like this:
```
cannot open /tmp/tmp.tWa7uKWuEf//etc/group: No space left on device
```

This is clearly a wrong error message. The problem is that /etc/group is a directory. The "No space left on device" part happens because fprintf itself failed, overwriting errno and thus the subsequent `strerrno()` result.

### DIFF
--- a/lib/selinux.c
+++ b/lib/selinux.c
@@ -191,12 +191,14 @@ int check_selinux_permit (const char *perm_name)
 	selinux_set_callback (SELINUX_CB_LOG, (union selinux_callback) { .func_log = selinux_log_cb });
 
 	if (getprevcon_raw (&user_context_raw) != 0) {
+		const char *errmsg = strerrno();
+
 		fprintf (shadow_logfd,
 		    _("%s: can not get previous SELinux process context: %s\n"),
-		    shadow_progname, strerrno());
+		    shadow_progname, errmsg);
 		SYSLOG ((LOG_WARN,
 		    "can not get previous SELinux process context: %s",
-		    strerrno()));
+		    errmsg));
 		return (security_getenforce () != 0);
 	}
 
--- a/src/userdel.c
+++ b/src/userdel.c
@@ -776,10 +776,12 @@ static bool remove_mailbox (void)
 			free(mailfile);
 			return 0;
 		} else {
+			const char *errmsg = strerrno();
+
 			fprintf (stderr,
 			         _("%s: warning: can't remove %s: %s\n"),
-			        Prog, mailfile, strerrno());
-			SYSLOG((LOG_ERR, "Cannot remove %s: %s", mailfile, strerrno()));
+			        Prog, mailfile, errmsg);
+			SYSLOG((LOG_ERR, "Cannot remove %s: %s", mailfile, errmsg));
 #ifdef WITH_AUDIT
 			audit_logger (AUDIT_DEL_USER,
 			              "delete-mail-file",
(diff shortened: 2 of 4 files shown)


lib/run_part.c
```diff
@@ -42,7 +42,7 @@ static int run_part(char *script_path, const char *name, const char *action)
 		return (wait_status);
 	}
 
-	fprintf(shadow_logfd, "waitpid: %s\n", strerrno());
+	fprintf(shadow_logfd, "wait: %s\n", strerrno());
```
alejandro-colomar, line 45: For patch 1:

```
```

src/userdel.c
```diff
@@ -820,7 +820,7 @@ static bool remove_mailbox (void)
 		fprintf (stderr,
 		         _("%s: %s not owned by %s, not removing\n"),
 		         Prog, mailfile, user_name);
-		SYSLOG((LOG_ERR, "%s not owned by %s, not removed", mailfile, strerrno()));
+		SYSLOG((LOG_ERR, "%s not owned by %s, not removed", mailfile, user_name));
```
alejandro-colomar, line 827: Thanks!  This matches the last patch in <https://github.com/shadow-maint/shadow/pull/1289>, <https://github.com/shadow-maint/shadow/pull/1289/commits/c1b93d2ecbad12488b2c8ddf8dfbc5b901916cde>.

For patch 2:

```
```

lib/selinux.c
```diff
@@ -191,12 +191,14 @@ int check_selinux_permit (const char *perm_name)
...
+		const char *errmsg = strerrno();
+
 		fprintf (shadow_logfd,
 		    _("%s: can not get previous SELinux process context: %s\n"),
-		    shadow_progname, strerrno());
+		    shadow_progname, errmsg);
 		SYSLOG ((LOG_WARN,
 		    "can not get previous SELinux process context: %s",
-		    strerrno()));
+		    errmsg));
```
alejandro-colomar, line 201: I don't like this patch.  Instead, I think we should develop an API that preserves errno.

I thought of having errno preservation in fprinte() and SYSLOGE() macros, and be able to call them sequentially.

I'm going to NAK this patch (commit 3) at the moment.

alejandro-colomar: Feel free to send patches 1 and 2 in a separate PR, and I'll merge them now.

Let's keep this PR open, for discussion in the future.

stoeckmann: Opened https://github.com/shadow-maint/shadow/pull/1448 for this

PR #1447, opened 2025-12-23
chage: Remove unneeded xstrdup calls

stoeckmann:
Duplicating name and hash is not needed here, because duplication occurs in spw_update. You can detect the small memory leak with tools like valgrind.

More importantly though, if xstrdup fails, it calls exit. The update_age function is in the "criticial section" between open_files and close_files, though. Correct error handling would require fail_exit to release the held locks.

### DIFF
--- a/src/chage.c
+++ b/src/chage.c
@@ -32,7 +32,6 @@
 #include "string/sprintf/snprintf.h"
 #include "string/strcmp/streq.h"
 #include "string/strcpy/strtcpy.h"
-#include "string/strdup/strdup.h"
 #include "string/strerrno.h"
 #include "string/strftime.h"
 #include "time/day_to_str.h"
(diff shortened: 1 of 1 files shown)


src/chage.c
```diff
@@ -613,8 +612,8 @@ static void update_age (/*@null@*/const struct spwd *sp,
 		struct passwd pwent = *pw;
 
 		memzero(&spwent, sizeof(spwent));
-		spwent.sp_namp = xstrdup (pwent.pw_name);
-		spwent.sp_pwdp = xstrdup (pwent.pw_passwd);
+		spwent.sp_namp = pwent.pw_name;
+		spwent.sp_pwdp = pwent.pw_passwd;
 		spwent.sp_flag = SHADOW_SP_FLAG_UNSET;
```
alejandro-colomar, line 618: Thanks!  I was suspecting this duplication was indeed redundant, but didn't have the energy to follow through all the layers of functions.

alejandro-colomar: You may want to mention in the commit message where excatly this happens.

If I'm reading it correctly, `update_age()` calls `spw_update()`, which calls `commonio_update()`, which calls `db->ops->cio_dup()`, which does the actual duplication.

PR #1449, opened 2025-12-26
src/chgpasswd.c: fix handling of files without trailing newlines

ikerexxe:
The chgpasswd command was incorrectly reporting "missing new password" when processing files with a trailing newline. This happened because the code was missing an EOF check.

Reproduction steps:
$ groupadd tgroup1
$ groupadd tgroup2
$ cat /tmp/test_chgpasswd.txt
tgroup1:Secret123
tgroup2:Secret456

$ #The file contains a trailing newline
$ chgpasswd < /tmp/test_chgpasswd.txt
chgpasswd: line 3: missing new password
chgpasswd: error detected, changes ignored

### DIFF
--- a/src/chgpasswd.c
+++ b/src/chgpasswd.c
@@ -479,10 +479,12 @@ int main (int argc, char **argv)
 	while (fgets(buf, sizeof(buf), stdin) != NULL) {
 		line++;
 		if (stpsep(buf, "\n") == NULL) {
-			fprintf (stderr, _("%s: line %jd: line too long\n"),
-			         Prog, line);
-			errors = true;
-			continue;
+			if (feof(stdin) == 0) {
+				fprintf (stderr, _("%s: line %jd: line too long\n"),
+				         Prog, line);
+				errors = true;
+				continue;
+			}
 		}
 
 		/*


alejandro-colomar:
From that reproducer, what I observe is a spurious empty line, not a file without trailing newline.

That is, the file ends with 2 new-line characters instead of just 1.

alejandro-colomar:
> From that reproducer, what I observe is a spurious empty line, not a file without trailing newline.
> 
> That is, the file ends with 2 new-line characters instead of just 1.

The proposed patch shouldn't change the behavior of the reproducer you shared, I think.

Instead, we should either:

-  continue rejecting blank lines.
-  add code to accept and ignore blank lines.

src/chgpasswd.c
```diff
@@ -479,10 +479,12 @@ int main (int argc, char **argv)
...
-			         Prog, line);
-			errors = true;
-			continue;
+			if (feof(stdin) == 0) {
+				fprintf (stderr, _("%s: line %jd: line too long\n"),
+				         Prog, line);
+				errors = true;
+				continue;
+			}
 		}
```
alejandro-colomar, line 488: Because this code runs only when a newline is **not** found, it won't change the behavior in the shown reproducer, which has the opposite problem: too many newlines are present.

alejandro-colomar:
> The chgpasswd command was incorrectly reporting "missing new password" when processing files with a trailing newline. This happened because the code was missing an EOF check.
> 
> Reproduction steps: $ groupadd tgroup1 $ groupadd tgroup2 $ cat /tmp/test_chgpasswd.txt tgroup1:Secret123 tgroup2:Secret456
> 
> $ #The file contains a trailing newline $ chgpasswd < /tmp/test_chgpasswd.txt chgpasswd: line 3: missing new password chgpasswd: error detected, changes ignored

Do you have a use case for files with blank lines?

ikerexxe:
It turns out that I was creating the file incorrectly :facepalm:

PR #1450, opened 2025-12-26
Test groupmod -u with user list

ikerexxe:
Add comprehensive test for the groupmod -U option when provided with a
list of users to set group membership. This test verifies:

- Setting initial group membership with multiple users
- Proper membership verification in both group and gshadow entries
- Updating group membership by modifying the user list
- Correct handling of membership changes in group databases

### DIFF
--- a/tests/system/tests/test_groupmod.py
+++ b/tests/system/tests/test_groupmod.py
@@ -69,3 +69,63 @@ def test_groupmod__u_option_empty_string_clears_members(shadow: Shadow):
         assert gshadow_entry is not None, "Group should be found"
         assert gshadow_entry.name == "tgroup", "Incorrect groupname"
         assert not gshadow_entry.members, "Group should have no members"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_groupmod__u_option_with_user_list(shadow: Shadow):
+    """
+    :title: Test groupmod -U option with user list to set group membership
+    :setup:
+        1. Create three test users
+        2. Create test group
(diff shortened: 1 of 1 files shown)


tests/system/framework/utils/tools.py
```diff
@@ -389,7 +389,7 @@ def __init__(
         name: str | None,
         password: str | None,
         administrators: str | None,
-        members: str | None,
+        members: list[str],
```
alejandro-colomar, line 392: Would you mind clarifying in the commit message if this is a stylistic fix or a behavioral fix?

alejandro-colomar: Also, would you mind adding a Fixes: trailer field for the commit it fixes?

hallyn (approved):
Agreed with alx, a bit more explanation on patch 1 would be appreciated.

tests/system/framework/utils/tools.py
ikerexxe: Done

ikerexxe:
I discovered an additional problem with `administrators` member from `GShadowEntry`, so I also fixed that one in the first commit. Please review it again before approving.

hallyn:
@alejandro-colomar you ok with merging this before the release, or will that throw a wrench into testing you've already done?

alejandro-colomar:
> @alejandro-colomar you ok with merging this before the release, or will that throw a wrench into testing you've already done?

I'm okay with it.  :)

PR #1457, opened 2025-12-27
etc/, lib/, man/, src/, tests/: Remove support for MD5

alejandro-colomar:
Cc: @andreboscatto, @ikerexxe , @thkukuk 

---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase

```
$ git rd 
1:  c43358172d02 = 1:  9ce4e4a1952e */: Remove support for MD5
2:  62bde435342b = 2:  22bcab38b2db lib/chkhash.c: is_valid_hash(): Update minimum hash lenght
```
</details>

<details>
<summary>v1c</summary>

-  tfix in commit message

```
$ git rd -U0
1:  9ce4e4a1952e = 1:  9ce4e4a1952e */: Remove support for MD5
2:  22bcab38b2db ! 2:  3aaee35e2501 lib/chkhash.c: is_valid_hash(): Update minimum hash lenght
    @@ Commit message
    -    lib/chkhash.c: is_valid_hash(): Update minimum hash lenght
    +    lib/chkhash.c: is_valid_hash(): Update minimum hash length
```
</details>

<details>
<summary>v2</summary>

-  Rebase after #1688 .
-  Remove chkhash() tests that test MD5.

```
$ git range-diff master..gh/nomd5 t..nomd5 
1:  9ce4e4a1952e ! 1:  62c8dd744a7b */: Remove support for MD5
    @@ tests/system/etc/login.defs: CHFN_RESTRICT               rwh
      # Note: If you use PAM, it is recommended to use a value consistent with
      # the PAM modules configuration.
     
    + ## tests/unit/test_chkhash.c ##
    +@@ tests/unit/test_chkhash.c: test_is_valid_hash_ok_sha256(MAYBE_UNUSED void ** _1)
    + }
    + 
    + 
    +-static void
    +-test_is_valid_hash_ok_md5(MAYBE_UNUSED void ** _1)
    +-{
    +-  // Basic MD5 hash: $1$ + salt + $ + 22 character hash
    +-  assert_true(is_valid_hash("$1$salt$abcdefghijklmnopqrstuv"));
    +-
    +-  // MD5 with maximum salt length (8 characters)
    +-  assert_true(is_valid_hash("$1$maxsalt8$abcdefghijklmnopqrstuv"));
    +-}
    +-
    +-
    + static void
    + test_is_valid_hash_ok_special(MAYBE_UNUSED void ** _1)
    + {
    +@@ tests/unit/test_chkhash.c: main(void)
    +         cmocka_unit_test(test_is_valid_hash_ok_bcrypt),
    +         cmocka_unit_test(test_is_valid_hash_ok_sha512),
    +         cmocka_unit_test(test_is_valid_hash_ok_sha256),
    +-        cmocka_unit_test(test_is_valid_hash_ok_md5),
    +         cmocka_unit_test(test_is_valid_hash_ok_special),
    +         cmocka_unit_test(test_is_valid_hash_edge_salt_chars),
    +         cmocka_unit_test(test_is_valid_hash_edge_account_locks),
    +
      ## tests/usertools/chpasswd-PAM/06_chpasswd_usage/data/usage.out ##
     @@
      Usage: chpasswd [options]
2:  3aaee35e2501 = 2:  1366ea9fa258 lib/chkhash.c: is_valid_hash(): Update minimum hash length
```
</details>

<details>
<summary>v3</summary>

-  Rebase
-  Remove another test that uses MD5.

```
$ git range-diff 62c8dd744a7b^..gh/nomd5 t..nomd5 
1:  62c8dd744a7b ! 1:  ec177ecad9e4 */: Remove support for MD5
    @@ tests/unit/test_chkhash.c: test_is_valid_hash_ok_sha256(MAYBE_UNUSED void ** _1)
      static void
      test_is_valid_hash_ok_special(MAYBE_UNUSED void ** _1)
      {
    +@@ tests/unit/test_chkhash.c: test_is_valid_hash_edge_account_locks(MAYBE_UNUSED void ** _1)
    +   // Complex ! prefix scenarios with various hash types should work
    +   assert_true(is_valid_hash("!$2a$12$abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ."));
    +   assert_true(is_valid_hash("!$5$salt$abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQ"));
    +-  assert_true(is_valid_hash("!$1$salt$abcdefghijklmnopqrstuv"));
    + 
    +   // But invalid hashes with ! prefix should still fail
    +   assert_false(is_valid_hash("!invalid"));
     @@ tests/unit/test_chkhash.c: main(void)
              cmocka_unit_test(test_is_valid_hash_ok_bcrypt),
              cmocka_unit_test(test_is_valid_hash_ok_sha512),
2:  1366ea9fa258 = 2:  8e04224e8786 lib/chkhash.c: is_valid_hash(): Update minimum hash length
```
</details>

<details>
<summary>v3b</summary>

-  Rebase

```
$ git rd 
1:  8924a1d658f6 < -:  ------------ tests/unit/test_chkhash.c: Remove tests for DES
2:  5dbeb20153a8 < -:  ------------ man/login.defs.5.xml: Remove references to PASS_MAX_LEN
3:  ec177ecad9e4 = 1:  e8afd8264524 */: Remove support for MD5
4:  8e04224e8786 = 2:  0047f931cfa2 lib/chkhash.c: is_valid_hash(): Update minimum hash length
```
</details>


floppym:
Does this affect validation of existing MD5 hashes? (in login for example)

I assume this would only prevent new MD5 hashes from being generated, but I want to double-check.

alejandro-colomar:
> Does this affect validation of existing MD5 hashes? (in login for example)
> 
> I assume this would only prevent new MD5 hashes from being generated, but I want to double-check.

As far as I can see, yes.  It is up to crypt(3) to decide whether MD5 is accepted for verifying existing hashes; we have no say in it.  This logic is in `pw_encrypt()`.

Do you want to preserve the ability to verify MD5 passwords, or is it just curiosity?

floppym:
> > Does this affect validation of existing MD5 hashes? (in login for example)
> > I assume this would only prevent new MD5 hashes from being generated, but I want to double-check.
> 
> As far as I can see, yes. It is up to crypt(3) to decide whether MD5 is accepted for verifying existing hashes; we have no say in it. This logic is in `pw_encrypt()`.

Yeah, that's what I figured.

> Do you want to preserve the ability to verify MD5 passwords, or is it just curiosity?

I think it would be good maintain compatibility when there's very little maintenance cost.

If we break existing passwords that might require some communication with end users to warn them.

alejandro-colomar:
> If we break existing passwords that might require some communication with end users to warn them.

I think we should still communicate with end users to warn them.  And they should consider moving away from MD5, as it might eventually stop working.

floppym:
> And they should consider moving away from MD5, as it might eventually stop working.

I suspect most users have already migrated away from a settings perspective. However, I believe this is only effective if all users change their password after the hash setting has been changed.

A more robust approach would be to rehash the password on successful login when the existing hash is using an obsolete method like MD5. I don't think shadow or PAM currently do that, but I could be mistaken.

thkukuk:
> A more robust approach would be to rehash the password on successful login when the existing hash is using an obsolete method like MD5. I don't think shadow or PAM currently do that, but I could be mistaken.

Write a small script which checks which hash a password is using, and if the hash is weak, run `passwd -e <account>`.
And the user is forced to change the password at next login.

floppym:
Would be nice to have it happen automatically, without forcing a password change.

alejandro-colomar:
> > A more robust approach would be to rehash the password on successful login when the existing hash is using an obsolete method like MD5. I don't think shadow or PAM currently do that, but I could be mistaken.
> 
> Write a small script which checks which hash a password is using, and if the hash is weak, run `passwd -e <account>`. And the user is forced to change the password at next login.

Agree.

Having more complexity in shadow utils would be dangerous.  I'd favour an approach with a script that checks weak passwords, and expires them immediately, keeping the usual shadow utils simple.

alejandro-colomar:
> Would be nice to have it happen automatically, without forcing a password change.

Would you trust a password that has been hashed with MD5?

I think we shouldn't silently hide the fact that a password was less safe in the past, and thus might have been exposed.

A user would be able to re-enter the same password, anyway, because having different hashing algorithms, passwd(1) wouldn't be able to detect that they're the same.

floppym:
> I think we shouldn't silently hide the fact that a password was less safe in the past, and thus might have been exposed.

Fair enough.

> A user would be able to re-enter the same password, anyway, because having different hashing algorithms, passwd(1) wouldn't be able to detect that they're the same.

Not true: passwd could hash the new password using MD5 to check it against the old one. In fact, it usually asks the user to enter their old password before being allowed to enter a new password.

floppym:
> Having more complexity in shadow utils would be dangerous. I'd favour an approach with a script that checks weak passwords, and expires them immediately, keeping the usual shadow utils simple.

Yeah, I guess it makes sense to keep the shadow implementation simple. Might be a better fit for PAM.

alejandro-colomar:
> > I think we shouldn't silently hide the fact that a password was less safe in the past, and thus might have been exposed.
> 
> Fair enough.
> 
> > A user would be able to re-enter the same password, anyway, because having different hashing algorithms, passwd(1) wouldn't be able to detect that they're the same.
> 
> Not true: passwd could hash the new password using MD5 to check it against the old one. In fact, it usually asks the user to enter their old password before being allowed to enter a new password.

Yup, we could if we wanted to.  However, we don't (and I don't want to do it).  :)

lib/salt.c
```diff
@@ -365,19 +352,13 @@ static /*@observer@*/const char *gensalt (size_t salt_size)
...
-		if (NULL == method) {
-			method = getdef_bool ("MD5_CRYPT_ENAB") ? "MD5" : "DES";
-		}
-	}
+	method = meth ?: getdef_str("ENCRYPT_METHOD") ?: "SHA512";
 
-	if (streq(method, "MD5")) {
-		MAGNUM(result, '1');
-		salt_len = MD5_CRYPT_SALT_SIZE;
-		rounds = 0;
```
alejandro-colomar, line 364: Actually, this will break support for checking existing MD5 passwords, because without the salt, we wouldn't be able to hash the introduced password to compare against the shadow(5) hash.  I suppose.  I'm not too experienced with this part of the code, so you'd need to test.

I'm willing to fully remove support for MD5, FWIW.

Cc: @floppym

floppym: When validating password, we pass the existing password hash in as the salt parameter and crypt() grabs the salt value from there. See the valid() function in lib/valid.c.

I think it should be ok to remove the code from lib/salt.c.

alejandro-colomar: It could be.  I don't care enough to check, though.  If anyone cares about MD5, I guess they should test it.  :)

floppym: I tested this branch locally with an account having an md5crypt password in /etc/shadow.
I was able to login using the `login` command and change the password using `passwd`.
The old password was accepted by the `passwd` command. and the md5crypt hash was replaced with a sha512crypt hash.

alejandro-colomar: Thanks!!

hallyn (approved):

PR #1458, opened 2025-12-28
sg: Fix usage and manual page

stoeckmann:
The usage message of `sg` and its manual page contain different information. Synchronize both and document what `sg`'s options do.

### DIFF
--- a/man/sg.1.xml
+++ b/man/sg.1.xml
@@ -45,13 +45,15 @@
   <refsynopsisdiv id='synopsis'>
     <cmdsynopsis>
       <command>sg</command>
-      <arg choice='opt'>- </arg>
+      <arg choice='opt'>
+        <replaceable>-</replaceable>
+      </arg>
       <arg choice='plain'>
         <replaceable>group</replaceable>
       </arg>
       <arg choice='opt'>
-	<arg choice='opt'>-c </arg>
-	command
+        <arg choice='opt'><replaceable>-c</replaceable></arg>
+        <replaceable>command</replaceable>
       </arg>
     </cmdsynopsis>
   </refsynopsisdiv>
@@ -71,6 +73,37 @@
       <command>sg</command> command you are returned to your previous group
       ID.
     </para>
+  </refsect1>
+
+  <refsect1 id='options'>
+    <title>OPTIONS</title>
+    <para>
+      The options which apply to the <command>sg</command> command are:
+    </para>
+    <variablelist remap='IP'>
+      <varlistentry>
+	<term><option>-</option>, <option>-l</option></term>
+	<listitem>
+	  <para>
+	    Start the shell as a login shell.
+	  </para>
+	</listitem>
+      </varlistentry>
+      <varlistentry>
+	<term>
+	  <option>-c</option>
+	</term>
+	<listitem>
+	  <para>
+	    Specify a command that will be invoked by the shell using its
+	    <option>-c</option>.
+	  </para>
+	  <para>
+	    This is the default; for backward compatibility.
+	  </para>
+	</listitem>
+      </varlistentry>
+    </variablelist>
   </refsect1>
 
   <refsect1 id='configuration'>
--- a/src/newgrp.c
+++ b/src/newgrp.c
@@ -72,7 +72,7 @@ static void usage (void)
 	if (is_newgrp) {
 		(void) fputs (_("Usage: newgrp [-] [group]\n"), stderr);
 	} else {
-		(void) fputs (_("Usage: sg group [[-c] command]\n"), stderr);
+		(void) fputs (_("Usage: sg [-] group [[-c] command]\n"), stderr);
 	}
 }
 
@@ -465,8 +465,8 @@ int main (int argc, char **argv)
 	 * The valid syntax are
 	 *      newgrp [-] [groupid]
 	 *      newgrp [-l] [groupid]
-	 *      sg [-]
-	 *      sg [-] groupid [[-c command]
+	 *      sg [-] groupid [[-c] command]
+	 *      sg [-l] groupid [[-c] command]
 	 */
 	if (   (argc > 0)
 	    && (   streq(argv[0], "-")
@@ -499,8 +499,7 @@ int main (int argc, char **argv)
 
 			/*
 			 * Skip -c if specified so both forms work:
-			 * "sg group -c command" (as in the man page) or
-			 * "sg group command" (as in the usage message).
+			 * "sg group -c command" or "sg group command".
 			 */
 			if ((argc > 1) && streq(argv[0], "-c")) {
 				command = argv[1];


man/sg.1.xml
```diff
@@ -64,7 +64,7 @@
       <command>newgrp</command> and <command>sg</command> is that some
       shells treat <command>newgrp</command> specially, replacing themselves
       with a new instance of a shell that <command>newgrp</command> creates.
-      This doesn't happen with <command>sg</command>, so upon exit from a
+      This doesn't happen with <command>sg</command>, so upon exit from an
       <command>sg</command> command you are returned to your previous group
       ID.
```
alejandro-colomar, line 74: I suspect `a sg command` comes from the fact that `su` means "substitute user" or "superuser", and thus `sg` means "substitute group" or "supergroup", and that this would read "a supergroup (or substitute group) command".

I'd keep it as `a`.

man/sg.1.xml
```diff
@@ -45,10 +45,15 @@
...
-	command
+      <arg choice='opt'>
+        <replaceable>-</replaceable>
+      </arg>
+      <arg choice='plain'>
+        <replaceable>group</replaceable>
+      </arg>
+      <arg choice='opt'>
+        <arg choice='opt'><replaceable>-c</replaceable></arg>
+        <replaceable>command</replaceable>
```
alejandro-colomar, line 56: Ughhh, I don't understand how this XML will render in the screen.  Could you paste the before and after, just to see what's changing?

It gets even worse because my system uses sg(1) from util-linux, so I don't have this manual page installed.

alejandro-colomar: I wish we could have actual man(7) source for the manual pages.

man/sg.1.xml
```diff
@@ -70,6 +75,38 @@
...
+	<term>
+	  <option>-c</option>
+	</term>
+	<listitem>
+	  <para>
+	    Specify a command that will be invoked by the shell using its
+	    <option>-c</option>.
+	  </para>
+	  <para>
+	    This is the default. For backward compatibility.
```
alejandro-colomar: This should use `;` or `,` instead of `.`

man/sg.1.xml
```diff
@@ -70,6 +75,38 @@
...
+    <para>
+      The options which apply to the <command>sg</command> command are:
+    </para>
+    <variablelist remap='IP'>
+      <varlistentry>
+	<term><option>-</option>, <option>-l</option></term>
+	<listitem>
+	  <para>
+	    Provide an environment similar to what the user would expect had
+	    the user logged in directly.
```
alejandro-colomar: Please use semantic newlines.  See man-pages(7).

In this case, I'd break the lines as:

```
Provide an environment
similar to what the user would expect
had the user logged in directly.
```

alejandro-colomar: Also, "the user" is not a clear reference, in a command that works for groups.

man/sg.1.xml
stoeckmann: It's important how you pronounce the abbreviation of "sg." And I guess that you pronounce the letters, so it's "an". Quick lookup revealed https://ell.stackexchange.com/questions/343835/a-an-before-acronyms-british-vs-american which also fits my time of learning English as in "a UFO" instead of "an UFO" due to the way we pronounce "UFO".

man/sg.1.xml
stoeckmann: Before:
```
SYNOPSIS

       sg [-] [group [-c ] command]
```

After:
```
SYNOPSIS

       sg [-] group [[-c] command]
```

Retrieved on Arch Linux

man/sg.1.xml
stoeckmann: The current way fits `CHFN_RESTRICT.xml`. Adjusted anyway

man/sg.1.xml
stoeckmann: Adjusted. The `-` option actually is user-specific, i.e. it resets environment variables and calls `chdir` into the home directory of the calling user.

stoeckmann: Maybe "calling user"? I'm open to suggestions.

man/sg.1.xml
alejandro-colomar: I guess it's fine either way, I'll take the patch if you insist.  :)

man/sg.1.xml
alejandro-colomar: Thanks!  Then LGTM.  Although I think it would be good to split the different changes of this commit, and document each one well in the commit message.

man/sg.1.xml
alejandro-colomar: > Maybe "calling user"? I'm open to suggestions.

That sounds better, yes.

man/sg.1.xml
alejandro-colomar: That is, I'd like to see mention for example that `group` is not optional, and that it is independent of `-c command`.

man/sg.1.xml
alejandro-colomar: > The current way fits `CHFN_RESTRICT.xml`.

Yup, there's a lot I'd fix in these pages.  :)

Once I finish with the code --if that ever happens-- I'll have a look at the manual pages.

> Adjusted anyway

Thanks!

man/sg.1.xml
stoeckmann: Adjusted

man/sg.1.xml
stoeckmann: Split the non-optional group part into its own commit.

man/sg.1.xml
```diff
@@ -73,6 +75,39 @@
...
+      The options which apply to the <command>sg</command> command are:
+    </para>
+    <variablelist remap='IP'>
+      <varlistentry>
+	<term><option>-</option>, <option>-l</option></term>
+	<listitem>
+	  <para>
+	    Provide an environment
+	    similar to what the user would expect
+	    had the calling user logged in directly.
```
alejandro-colomar: About `logged in directly`, I'm not sure.  What would `directly` mean here?

How about something simpler?:

```
Start the shell as a login shell.
```

stoeckmann: Adjusted. Since the same sentence occurs in su's manual page but is not taken, dropped the reference from commit message.

src/newgrp.c
```diff
@@ -72,7 +72,7 @@ static void usage (void)
 	if (is_newgrp) {
 		(void) fputs (_("Usage: newgrp [-] [group]\n"), stderr);
 	} else {
-		(void) fputs (_("Usage: sg group [[-c] command]\n"), stderr);
+		(void) fputs (_("Usage: sg [-] group [[-c] command]\n"), stderr);
```
alejandro-colomar, line 75: This should probably be `[-|-l]`, right?  Which could also be expressed as `[-[l]]`, although I think the former would be more readable.

alejandro-colomar: If we had to chose one, I'd document `-l` here.  It's more readable than `-`.

src/newgrp.c
```diff
@@ -465,8 +465,8 @@ int main (int argc, char **argv)
 	 * The valid syntax are
 	 *      newgrp [-] [groupid]
 	 *      newgrp [-l] [groupid]
-	 *      sg [-]
-	 *      sg [-] groupid [[-c command]
+	 *      sg [-] groupid [[-c] command]
+	 *      sg [-l] groupid [[-c] command]
```
alejandro-colomar, line 469: I'd document this as

```
newgrp [-|-l] [groupid]
sg [-|-l] groupid [[-c] command]
```

src/newgrp.c
alejandro-colomar: You could do that in a separate commit.  The current commits are good, IMO.

man/sg.1.xml
```diff
@@ -45,13 +45,15 @@
   <refsynopsisdiv id='synopsis'>
     <cmdsynopsis>
       <command>sg</command>
-      <arg choice='opt'>- </arg>
+      <arg choice='opt'>
+        <replaceable>-</replaceable>
+      </arg>
```
alejandro-colomar, line 50: Do we have a way to document alternatives in the XML manual pages?  `[-|-l]`

stoeckmann:
Closing this PR due to chatter overflow. Based on these suggestions, someone who is actually willing to progress can find a solution. See https://github.com/shadow-maint/shadow/issues/1460

alejandro-colomar:
> Closing this PR due to chatter overflow. Based on these suggestions, someone who is actually willing to progress can find a solution. See #1460

I'm actually willing to progress.  I said I'm okay with the commits as they are.  I'd just like to improve this with a fourth commit that makes it even better.  The documentation, as you've improved it, still doesn't document the actual behavior: the synopsis doesn't say anything about `-l`.  It is much much better than it was, but it is still confusing to a reader.

At least the paragraphs now are understandable, so I'm happy to merge this as is.

alejandro-colomar (approved):

stoeckmann:
> The documentation, as you've improved it, still doesn't document the actual behavior: the synopsis doesn't say anything about `-l`. It is much much better than it was, but it is still confusing to a reader.

I assume that the different styles of usage messages and manual pages across the various shadow tools must be the reason why people stop using shadow where possible because they are confusing to them? Is it documented anywhere how these are supposed to look like or is a mixture of various styles actually what you want (in which case it doesn't make sense to "fix" this very one manual page, since it adds variation)?

> At least the paragraphs now are understandable, so I'm happy to merge this as is.

You might want to adjust the `su` manual page then because it cannot be understandable either. At least, if you actually build and read it as a maintainer, since your distribution rightfully abandons shadow tools where it can.

alejandro-colomar:
> > The documentation, as you've improved it, still doesn't document the actual behavior: the synopsis doesn't say anything about `-l`. It is much much better than it was, but it is still confusing to a reader.
> 
> I assume that the different styles of usage messages and manual pages across the various shadow tools must be the reason why people stop using shadow where possible because they are confusing to them?

I don't know.  All of that predates my involvement in the project (and even in programming), and I'm not a distribution maintainer, so I don't know why distros prefer some binaries from shadow and some from other projects.

> Is it documented anywhere how these are supposed to look like or is a mixture of various styles actually what you want (in which case it doesn't make sense to "fix" this very one manual page, since it adds variation)?

I'm not talking about style here.  I'm talking about undocumented features.  But as we discussed in the other PR, what should probably do is to change the behavior of sg(1) to align with that of su(1).

> > At least the paragraphs now are understandable, so I'm happy to merge this as is.
> 
> You might want to adjust the `su` manual page then because it cannot be understandable either. At least, if you actually build and read it as a maintainer, since your distribution rightfully abandons shadow tools where it can.

I have been wanting to fix the manual pages for a long time, indeed.  It's just that the XML is too painful for me to do it.

I might eventually port them to man(7) source code and improve them from there.

PR #1470, opened 2026-01-01
lib/list.c: Use for loop instead of while loop

alejandro-colomar:
---

Revisions:

<details>
<summary>v2</summary>

-  Use size_t.  [@stoeckmann ]

```
$ git rd 
1:  5595da525 = 1:  5595da525 lib/list.c: Use for loop instead of while loop
-:  --------- > 2:  13e636554 lib/list.c: Use size_t for array indices and sizes
```
</details>

<details>
<summary>v2b</summary>

-  Use size_t earlier.

```
$ git rd 
1:  5595da525 ! 1:  37c85da55 lib/list.c: Use for loop instead of while loop
    @@ lib/list.c: bool is_on_list (char *const *list, const char *member)
      
     -  while (NULL != *list) {
     -          if (streq(*list, member)) {
    -+  for (int i = 0; list[i] != NULL; i++) {
    ++  for (size_t i = 0; list[i] != NULL; i++) {
     +          if (streq(list[i], member))
                        return true;
     -          }
2:  13e636554 ! 2:  fe073f5b9 lib/list.c: Use size_t for array indices and sizes
    @@ lib/list.c: del_list(/*@returned@*/ /*@only@*/char **list, const char *member)
      
        assert (NULL != list);
      
    -@@ lib/list.c: bool is_on_list (char *const *list, const char *member)
    -   assert (NULL != member);
    -   assert (NULL != list);
    - 
    --  for (int i = 0; list[i] != NULL; i++) {
    -+  for (size_t i = 0; list[i] != NULL; i++) {
    -           if (streq(list[i], member))
    -                   return true;
    -   }
```
</details>

<details>
<summary>v2c</summary>

-  Rebase

```
$ git rd 
1:  37c85da5563c = 1:  77b0251d0f48 lib/list.c: Use for loop instead of while loop
2:  fe073f5b9bc3 = 2:  12a411d1ac6b lib/list.c: Use size_t for array indices and sizes
```
</details>

<details>
<summary>v2d</summary>

-  Rebase

```
$ git rd 
1:  77b0251d = 1:  fa821183 lib/list.c: Use for loop instead of while loop
2:  12a411d1 = 2:  711d98c5 lib/list.c: Use size_t for array indices and sizes
```
</details>

<details>
<summary>v2e</summary>

-  Rebase

```
$ git rd 
1:  fa821183 = 1:  d42beada lib/list.c: Use for loop instead of while loop
2:  711d98c5 ! 2:  b829ea2e lib/list.c: Use size_t for array indices and sizes
    @@ lib/list.c
      
      #include "config.h"
      
    --#ident "$Id$"
    --
    --#include <assert.h>
     +#include <stddef.h>
    - 
    ++
      #include "alloc/malloc.h"
      #include "prototypes.h"
    -@@
    - #include "string/strdup/strdup.h"
    - #include "string/strtok/strsep2ls.h"
    - 
    -+#include <assert.h>
    -+
    - 
    - /*
    -  * add_list - add a member to a list of group members
    + #include "defines.h"
     @@
      /*@only@*/char **
      add_list(/*@returned@*/ /*@only@*/char **list, const char *member)
```
</details>

<details>
<summary>v2f</summary>

-  Rebase

```
$ git rd 
1:  d42beada89a3 = 1:  2e94e37bb88a lib/list.c: Use for loop instead of while loop
2:  b829ea2e70fe = 2:  93597ba47a97 lib/list.c: Use size_t for array indices and sizes
```
</details>

<details>
<summary>v2g</summary>

-  Rebase

```
$ git rd 
1:  2e94e37bb88a = 1:  8f06265fe3fc lib/list.c: Use for loop instead of while loop
2:  93597ba47a97 = 2:  aee5d3391da2 lib/list.c: Use size_t for array indices and sizes
```
</details>

### DIFF
--- a/lib/list.c
+++ b/lib/list.c
@@ -8,6 +8,8 @@
 
 #include "config.h"
 
+#include <stddef.h>
+
 #include "alloc/malloc.h"
 #include "prototypes.h"
 #include "defines.h"
@@ -30,8 +32,8 @@
 /*@only@*/char **
 add_list(/*@returned@*/ /*@only@*/char **list, const char *member)
 {
-	int i;
-	char **tmp;
+	size_t  i;
+	char    **tmp;
 
 	assert (NULL != member);
 	assert (NULL != list);
@@ -81,8 +83,8 @@ add_list(/*@returned@*/ /*@only@*/char **list, const char *member)
 /*@only@*/char **
 del_list(/*@returned@*/ /*@only@*/char **list, const char *member)
 {
-	int i, j;
-	char **tmp;
+	size_t  i, j;
+	char    **tmp;
 
 	assert (NULL != member);
 	assert (NULL != list);
@@ -136,8 +138,8 @@ del_list(/*@returned@*/ /*@only@*/char **list, const char *member)
 /*@only@*/char **
 dup_list(char *const *list)
 {
-	int i;
-	char **tmp;
+	size_t  i;
+	char    **tmp;
 
 	assert (NULL != list);
 
--- a/lib/list.c
+++ b/lib/list.c
@@ -145,12 +145,8 @@ dup_list(char *const *list)
 
 	tmp = xmalloc_T(i + 1, char *);
 
-	i = 0;
-	while (NULL != *list) {
-		tmp[i] = xstrdup (*list);
-		i++;
-		list++;
-	}
+	for (i = 0; NULL != list[i]; i++)
+		tmp[i] = xstrdup(list[i]);
 
 	tmp[i] = NULL;
 	return tmp;
@@ -178,11 +174,9 @@ bool is_on_list (char *const *list, const char *member)
 	assert (NULL != member);
 	assert (NULL != list);
 
-	while (NULL != *list) {
-		if (streq(*list, member)) {
+	for (size_t i = 0; list[i] != NULL; i++) {
+		if (streq(list[i], member))
 			return true;
-		}
-		list++;
 	}
 
 	return false;


lib/list.c
```diff
@@ -179,11 +175,9 @@ bool is_on_list (char *const *list, const char *member)
 	assert (NULL != member);
 	assert (NULL != list);
 
-	while (NULL != *list) {
-		if (streq(*list, member)) {
+	for (int i = 0; list[i] != NULL; i++) {
```
stoeckmann: I still add this remark here, even while considering the current price of RAM and the amount of entries it would take:

It is possible to go beyond what int can handle with groups. The `MAX_MEMBERS_PER_GROUP` definition in login.defs can lead to split groups in /etc/group. If you add enough of them, the list of members can exceed `ÌNT_MAX`. Either go with `size_t` just like it's done in groupio.c for this reason or keep the while loop which completely removes the need for an index variable.

And yes, there are many more `int` examples in this file and `unsigned int` ones in groupio.c. But let's not make it worse.

stoeckmann (dismissed):

lib/list.c
alejandro-colomar: Oops, sorry; yep, I did it out of inertia with the other functions in this file, but I agree size_t is what we should be using.

I'll add a second patch using size_t.

Thanks!

lib/list.c
```diff
@@ -179,11 +175,9 @@ bool is_on_list (char *const *list, const char *member)
 	assert (NULL != member);
 	assert (NULL != list);
```
stoeckmann, line 177: Definitely just a remark on this PR (and a bad place to discuss, but probably not worth an issue or PR in shadow at all):

This assert can be triggered with systemd and shadow's `sg` as described here: https://github.com/systemd/systemd/pull/40218

And I think it's something that has to be fixed in systemd, not shadow. Yet, currently shadow code can `abort()` or even worse do a NULL pointer dereference in a setuid binary while `getuid()` is 0. Should we reconsider this?

If we do, obviously it's its own PR.

alejandro-colomar: Hmmm, I'll have a look.  I've opened an issue so we keep a reminder.  Thanks for pointing this out!

stoeckmann (approved):

PR #1473, opened 2026-01-03
Drop libattr dependency

stoeckmann:
Spotted this while rebuilding the Arch Linux shadow package, which gave me the warning that libattr is not needed (even though referred to by PKGBUILD file).

Turns out: It's absolutely correct!

And here's the history of the whole thing:

https://github.com/shadow-maint/shadow/commit/791ebc58dd2573c0825b82202abb14d30b680b0f -- Added `reset_selinux` flag (`useradd` set it to `false`, `usermod` set it to `true`)
https://github.com/shadow-maint/shadow/commit/06eb4e4d76ac7f1ac86e68a89b2dc9be7c7323a2 -- Made both tools set it to `true`
https://github.com/shadow-maint/shadow/commit/74c17c71675a60edaa47b280b0b4e2190ddcaee0 -- Introduced a `false` call in `useradd`
https://github.com/shadow-maint/shadow/commit/48aa12af31c0b72872b411857d03a518a4200a3d -- Made both tools set it to `true` again

So, as long as @ikerexxe agrees, this disabled `libattr` support for a few years and nobody complained. In fact, people complained when it was enabled.

I keep https://github.com/shadow-maint/shadow/commit/786e44f6ddb47648c724a82c7dc74ac692e02dd9 as it is, unless @ikerexxe says that it can be removed now.


lib/copydir.c
```diff
@@ -27,16 +27,11 @@
 #ifdef WITH_SELINUX
 #include <selinux/selinux.h>
 #endif				/* WITH_SELINUX */
-#if defined(WITH_ACL) || defined(WITH_ATTR)
+#if defined(WITH_ACL)
 #include <stdarg.h>
 #include <attr/error_context.h>
```
alejandro-colomar, line 32: Was this a (pre-existing) bug?  Why were we including `<attr/error_context.h>` even if not WITH_ATTR?

stoeckmann: The ACL code uses the `struct error_context` which I've also found in their example code and, since I didn't save the exact location, this quick codesearch gave another example: https://sources.debian.org/src/acl/2.3.2-2/libacl/perm_copy_fd.c?hl=147#L41

stoeckmann: The existence of the header is also properly checked in configure.ac for libacl.

alejandro-colomar: Hmmm, thanks!

alejandro-colomar:
It LGTM.  Do you want to wait for @ikerexxe 's review?  I'd be happy to merge now and expand later if appropriate; the current patch makes sense to me.

stoeckmann:
I'd prefer to wait. I wonder if this actually reveals a bug or a feature that was missed. If it's 100 % sure that we really don't want libattr, then we can merge.

stoeckmann:
> I can still see three references to `LIBATTR` in Makefile.am files. Can you remove them?

Removed. Thanks for careful review! :)

ikerexxe (approved):
LGTM! Thank you for taking care of that too

PR #1471, opened 2026-01-02
Add C-locale variants of isascii(3) APIs, and use them

alejandro-colomar:
Reported-by: @stoeckmann 
Cc: @eggert 

---

Revisions:

<details>
<summary>v2</summary>

-  Reimplement isascii_c() in a way that only evaluates the argument once, which is more robust, and allows implementing them as one-liner macros, more compactly.

```
$ git rd 
 1:  43fa84b9c !  1:  a22f8b4b0 lib/string/ctype/isascii.[ch]: is*_c(): functions
    @@ Metadata
     Author: Alejandro Colomar <alx@kernel.org>
     
      ## Commit message ##
    -    lib/string/ctype/isascii.[ch]: is*_c(): functions
    +    lib/string/ctype/isascii.[ch]: is*_c(): Add APIs
     
         These are like the isascii(3) family of APIs, but use the C locale, as
         the _c suffix hints.
     
    -    These functions behave well with non-casted input, unlike isascii(3).
    +    These macros behave well with non-casted input, unlike isascii(3).
    +
    +    The isascii_c() and iscntrl_c() implementations are different from the
    +    rest because they must return true for '\0'.
     
         Reported-by: Tobias Stoeckmann <tobias@stoeckmann.org>
         Cc: Paul Eggert <eggert@cs.ucla.edu>
    @@ lib/string/ctype/isascii.c (new)
     +#include "config.h"
     +
     +#include "string/ctype/isascii.h"
    -+
    -+#include <stdbool.h>
    -+
    -+
    -+extern inline bool iscntrl_c(unsigned char c);
    -+extern inline bool islower_c(unsigned char c);
    -+extern inline bool isupper_c(unsigned char c);
    -+extern inline bool isdigit_c(unsigned char c);
    -+extern inline bool ispunct_c(unsigned char c);
    -+extern inline bool isspace_c(unsigned char c);
    -+extern inline bool isalpha_c(unsigned char c);
    -+extern inline bool isalnum_c(unsigned char c);
    -+extern inline bool isgraph_c(unsigned char c);
    -+extern inline bool isprint_c(unsigned char c);
    -+extern inline bool isxdigit_c(unsigned char c);
    -+extern inline bool isascii_c(unsigned char c);
     
      ## lib/string/ctype/isascii.h (new) ##
     @@
    @@ lib/string/ctype/isascii.h (new)
     +
     +#include "config.h"
     +
    -+#include <ctype.h>
    -+#include <stdbool.h>
     +#include <string.h>
     +
[...]

PR #1475, opened 2026-01-03
src/newusers.c: A user/group name with a leading digit is valid

### LINKED ISSUE #1474: Incorrect use of isdigit(3) where strisdigit() is wanted
https://github.com/shadow-maint/shadow/blob/3e8c105f0703264e947d8c034b90419794955d49/src/newusers.c#L251

If I'm reading this correctly, this seems to say that newusers(8) treats any group string that starts with a digit as if it were a GID.

And indeed, it fails:

```sh
alx@devuan:~$ echo 'foo:p::1a::/tmp/nonexistent:/usr/bin/false' > x
alx@devuan:~$ sudo newusers ./x
newusers: invalid group ID '1a'
newusers: line 1: can't create group
```

But succeeds if it doesn't start in a digit:

```sh
alx@devuan:~$ echo 'foo:p::a1a::/tmp/nonexistent:/usr/bin/false' > x
alx@devuan:~$ sudo newusers ./x
alx@devuan:~$ tail -n1 /etc/group
a1a:x:1004:
alx@devuan:~$ tail -n1 /etc/passwd
foo:x:1004:1004::/tmp/nonexistent:/usr/bin/false
```

I think we should replace isdigit() by strisdigit().

alejandro-colomar:
Only consider a string to be a UID/GID if it is all digits.
    
Closes: <https://github.com/shadow-maint/shadow/issues/1474>

---

Cc: @uecker 

---

Revisions:

<details>
<summary>v1b</summary>

-  Add reproducer to commit message.

```
$ git rd 
1:  f26af1fee ! 1:  497cbf1fe src/newusers.c: A user/group name with a leading digit is valid
    @@ Commit message
     
         Only consider a string to be a UID/GID if it is all digits.
     
    +    Here's a reproducer of the bug:
    +
    +            $ echo 'foo:p::1a::/tmp/nonexistent:/usr/bin/false' > x
    +            $ sudo newusers ./x
    +            newusers: invalid group ID '1a'
    +            newusers: line 1: can't create group
    +
    +    Where the expected behavior would be the same as for a group name that
    +    doesn't start with a digit:
    +
    +            $ echo 'foo:p::a1a::/tmp/nonexistent:/usr/bin/false' > x
    +            $ sudo newusers ./x
    +            $ tail -n1 /etc/group
    +            a1a:x:1004:
    +            $ tail -n1 /etc/passwd
    +            foo:x:1004:1004::/tmp/nonexistent:/usr/bin/false
    +
         Closes: <https://github.com/shadow-maint/shadow/issues/1474>
     
```
</details>

<details>
<summary>v1c</summary>

-  Rebase

```
$ git rd 
1:  8d4aff7d = 1:  98677c4e src/newusers.c: A user/group name with a leading digit is valid
```
</details>

<details>
<summary>v1d</summary>

-  Rebase

```
$ git rd --creation-factor=99
1:  98677c4e39b2 ! 1:  8cbf06e0ebfb src/newusers.c: A user/group name with a leading digit is valid
    @@ Commit message
     
      ## src/newusers.c ##
    -@@
    - #include "shadow/gshadow/sgrp.h"
    - #include "shadowlog.h"
    - #include "sssd.h"
    -+#include "string/ctype/strisascii/strisdigit.h"
    - #include "string/sprintf/snprintf.h"
    - #include "string/strcmp/streq.h"
    - #include "string/strdup/strdup.h"
     @@ src/newusers.c: static int add_group (const char *name, const char *gid, gid_t *ngid, uid_t uid)
                return 0;
        }
      
    --  if (isdigit (gid[0])) {
    -+  if (!streq(gid, "") && strisdigit(gid)) {
    +-  if (isdigit_c(gid[0])) {
    ++  if (!streq(gid, "") && strisdigit_c(gid)) {
                /*
                 * The GID is a number, which means either this is a brand
                 * new group, or an existing group.
    @@ src/newusers.c: static int add_group (const char *name, const char *gid, gid_t *
        /*
         * Now I have all of the fields required to create the new group.
         */
    --  if (!streq(gid, "") && (!isdigit(gid[0]))) {
    -+  if (!streq(gid, "") && !strisdigit(gid)) {
    +-  if (!streq(gid, "") && (!isdigit_c(gid[0]))) {
    ++  if (!streq(gid, "") && !strisdigit_c(gid)) {
                grent.gr_name = xstrdup (gid);
        } else {
                grent.gr_name = xstrdup (name);
    @@ src/newusers.c: static int get_user_id (const char *uid, uid_t *nuid) {
         * The first guess for the UID is either the numerical UID that the
         * caller provided, or the next available UID.
         */
    --  if (isdigit (uid[0])) {
    -+  if (!streq(uid, "") && strisdigit(uid)) {
    +-  if (isdigit_c(uid[0])) {
    ++  if (!streq(uid, "") && strisdigit_c(uid)) {
                if ((get_uid(uid, nuid) == -1) || (*nuid == (uid_t)-1)) {
                        fprintf (stderr,
                                 _("%s: invalid user ID '%s'\n"),
```
</details>

### DIFF
--- a/src/newusers.c
+++ b/src/newusers.c
@@ -236,7 +236,7 @@ static int add_group (const char *name, const char *gid, gid_t *ngid, uid_t uid)
 		return 0;
 	}
 
-	if (isdigit_c(gid[0])) {
+	if (!streq(gid, "") && strisdigit_c(gid)) {
 		/*
 		 * The GID is a number, which means either this is a brand
 		 * new group, or an existing group.
@@ -280,7 +280,7 @@ static int add_group (const char *name, const char *gid, gid_t *ngid, uid_t uid)
 	/*
 	 * Now I have all of the fields required to create the new group.
 	 */
-	if (!streq(gid, "") && (!isdigit_c(gid[0]))) {
+	if (!streq(gid, "") && !strisdigit_c(gid)) {
 		grent.gr_name = xstrdup (gid);
 	} else {
 		grent.gr_name = xstrdup (name);
@@ -345,7 +345,7 @@ static int get_user_id (const char *uid, uid_t *nuid) {
 	 * The first guess for the UID is either the numerical UID that the
 	 * caller provided, or the next available UID.
 	 */
-	if (isdigit_c(uid[0])) {
+	if (!streq(uid, "") && strisdigit_c(uid)) {
 		if ((get_uid(uid, nuid) == -1) || (*nuid == (uid_t)-1)) {
 			fprintf (stderr,
 			         _("%s: invalid user ID '%s'\n"),


src/newusers.c
```diff
@@ -248,7 +249,7 @@ static int add_group (const char *name, const char *gid, gid_t *ngid, uid_t uid)
 		return 0;
 	}
 
-	if (isdigit (gid[0])) {
+	if (!streq(gid, "") && strisdigit(gid)) {
```
hallyn: strisdigit returns false if streq(s, "").  do you need the first !streq(gid, "") in this check?

src/newusers.c
```diff
@@ -292,7 +293,7 @@ static int add_group (const char *name, const char *gid, gid_t *ngid, uid_t uid)
 	/*
 	 * Now I have all of the fields required to create the new group.
 	 */
-	if (!streq(gid, "") && (!isdigit(gid[0]))) {
+	if (!streq(gid, "") && !strisdigit(gid)) {
```
hallyn: (this one i think does need it.  But i'm a little dizzy trying to get the cases straight :)

src/newusers.c
alejandro-colomar: I have plans to change that corner-case behavior of strisdigit().

-  See <https://github.com/shadow-maint/shadow/pull/1471>

More precisely, see the following two commits of that PR:

<https://github.com/shadow-maint/shadow/pull/1471/changes/7c37e884f8e0414ea739ebf1e829bfe29319cb5d>
<https://github.com/shadow-maint/shadow/pull/1471/changes/d311b0d09b9a69e3cd9c4cc67399ad9b099d5815>

hallyn (approved):

PR #1477, opened 2026-01-04
man/groupmems: Fix grammar

stoeckmann:
The groupmems utility is deprecated by now, but let's keep existing documentation properly maintained until it's gone.


alejandro-colomar (approved):
Thanks!

PR #1484, opened 2026-01-07
vipw: set SIGCHLD before fork

stoeckmann:
It could happen that, if SIGCHLD was set to SIG_IGN before calling vipw, the forked child is already gone before SIGCHLD is set to SIG_DFL after the fork.

Prevent this race condition and also properly set up SIGCHLD for child handling within the fork, even though system() should take care of that.

Follow up of https://github.com/shadow-maint/shadow/pull/440 (looks like older me is getting better at this ;) )


alejandro-colomar:
Since there have been no comments, I'll merge. I'll cherry-pick for the stable release 4.19.1. Thanks!

PR #1485, opened 2026-01-07
vipw: Use fmkomstemp for temporary file

stoeckmann:
Avoid fixed and thus predictable temporary file names. Especially avoid just opening already existing ones.

Proof of Concept:

1. Create a FIFO named /etc/shadow.edit as root, which should be readable by normal users
```
mkfifo /etc/shadow.edit
```
2. Wait for content of FIFO as regular user
```
cat /etc/shadow.edit
```
3. Run vipw as root in another terminal
```
vipw -s
```

You can see the content of /etc/shadow as regular user.

### DIFF
--- a/src/vipw.c
+++ b/src/vipw.c
@@ -43,6 +43,7 @@
 #endif				/* WITH_TCB */
 #include "shadowlog.h"
 #include "sssd.h"
+#include "fs/mkstemp/fmkomstemp.h"
 #include "string/sprintf/aprintf.h"
 #include "string/sprintf/snprintf.h"
 #include "string/strcmp/streq.h"
@@ -72,7 +73,7 @@ static bool tcb_mode = false;
 
 /* local function prototypes */
 static void usage (int status);
-static int create_backup_file (FILE *, const char *, struct stat *);
+static int create_backup_file (FILE *, char *, struct stat *);
 static void vipwexit (const char *msg, int syserr, int ret);
 static void vipwedit (const char *, int (*)(void), int (*)(bool));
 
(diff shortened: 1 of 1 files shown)


alejandro-colomar (approved):
Thanks!  LGTM.

```
```

@ikerexxe , @hallyn , Any comments?

alejandro-colomar:
BTW, I guess I should backport this to the stable release 4.19.1.

alejandro-colomar:
Since there have been no comments, I'll merge.  I'll cherry-pick for the stable release 4.19.1.  Thanks!

src/vipw.c
```diff
@@ -103,16 +104,13 @@ static void usage (int status)
...
 	struct utimbuf ub;
 	FILE *bkfp;
 	int c;
-	mode_t mask;
 
-	mask = umask (077);
-	bkfp = fopen (backup, "w");
-	(void) umask (mask);
+	bkfp = fmkomstemp(backup, 0, 0600);
 	if (NULL == bkfp) {
```
alejandro-colomar, line 114: I was trying to unify all calls to fmkomstemp(), and noticed that we might have introduced a bug here.

There are two calls to create_backup_file().

One is:
<https://github.com/shadow-maint/shadow/blob/f4777ac5421b13e63c5fe4b67e681ab91fa3cd29/src/vipw.c#L278>

That one is fine, as we correctly added the appropiate template `XXXXXX` to `fileedit`.

The other one is:
<https://github.com/shadow-maint/shadow/blob/f4777ac5421b13e63c5fe4b67e681ab91fa3cd29/src/vipw.c#L433>

But as far as I can see, `to_rename` , which AFAICS, doesn't have the `XXXXXX` needed for the template.

Am I right?

@stoeckmann

stoeckmann: Oh great, a TCB defined code part. Yes, this has to be adjusted. Hang on, I'll create a PR.

stoeckmann: I think that the idea of TCB prevents this, but ... The code creates a "user+" file in the TCB directory and might fail during rename, which could leave that temporary file in the directory.

But "user+" could be a valid username (with --allow-badnames at least). I have added the illegal character `,` which is still easy to use in filesystem to make this less bad ...

This is especially true for just appending the suffix XXXXXX since that would be a perfectly valid username (no matter the characters used to replace the pattern. So, a bit more magic than I would like there to happen.

stoeckmann: See https://github.com/shadow-maint/shadow/pull/1503

PR #1490, opened 2026-01-09
su: Simplify SIGALRM signal handling

stoeckmann:
The SIGALRM signal handling performs quite some actions and reads shared variables on heap.

To avoid any form of analysis if these specific actions are safe or not, fall back to a signal handler which just sets a flag, which is eventually read.

See https://github.com/shadow-maint/shadow/pull/423 for a good way to test this. Especially to avoid regressions.

CC @lrh2000 for review request due to awesome analysis back in 2021. :)

### DIFF
--- a/src/su.c
+++ b/src/su.c
@@ -172,7 +172,7 @@ kill_child(pid_t pid_child)
 {
 	kill(-pid_child, SIGKILL);
 	fputs(_(" ...killed.\n"), stderr);
-	_exit (255);
+	exit (255);
 }
 
 static void
@@ -402,7 +402,7 @@ static void prepare_pam_close_session (void)
 		if (sigprocmask (SIG_BLOCK, &ourset, NULL) != 0) {
 			fprintf (stderr, _("%s: signal masking malfunction\n"), Prog);
 			kill_child(pid_child);
-			/* Never reached (_exit called). */
+			/* Never reached (exit called). */
 		}
 
 		/* Send SIGKILL to the child if it doesn't
@@ -418,7 +418,7 @@ static void prepare_pam_close_session (void)
 			sigsuspend (&ourset);
 			if (timeout) {
 				kill_child(pid_child);
-				/* Never reached (_exit called). */
+				/* Never reached (exit called). */
 			}
 		}
 		pid_child = 0;


src/su.c
```diff
@@ -409,7 +417,7 @@ static void prepare_pam_close_session (void)
 
 		/* Send SIGKILL to the child if it doesn't
 		 * exit within 2 seconds (after SIGTERM) */
-		(void) signal (SIGALRM, kill_child);
+		(void) signal (SIGALRM, set_timeout);
```
alejandro-colomar, line 410: @@ commit message
>  su: Kill child outside of signal handler
>
> This simplifies the alarm handler to also just set a volatile
> sig_atomic_t, which makes the handler way easier to review.

I'm not sure I understand the word 'also' here.  What does it mean there?

src/su.c
```diff
@@ -175,12 +175,8 @@ set_timeout(int)
...
 kill_child(pid_t pid_child)
 {
-	if (0 != pid_child) {
-		(void) kill (-pid_child, SIGKILL);
-		fputs(_(" ...killed.\n"), stderr);
-	} else {
-		fputs(_(" ...waiting for child to terminate.\n"), stderr);
-	}
+	kill(-pid_child, SIGKILL);
+	fputs(_(" ...killed.\n"), stderr);
```
alejandro-colomar: @@ commit message
>  su: Remove dead code
> 
> The pid_child is never 0 when reaching kill_child.

Would you mind documenting why it is never reached?  That would help reviewing.

src/su.c
stoeckmann: "Also" refers to `catch_signals`, another signal handler which only sets a `volatile sig_atomic_t`. I'm not entirely happy about it, since it doesn't set a flag but writes the signal (an `int`) into it. I know that it fits on real world hardware, but not entirely sure if this is the 100 % correct C way.

Updated commit message.

src/su.c
stoeckmann: Added to the commit message. Since it's not a signal handler function anymore, the function calls are only happening within an if-block which explicitly checks that `pid_child` is not `0`.

Also adjusted commit message.

stoeckmann:
Just noticed that I messed up the alphabetical ordering of the functions which I've added (due to renaming during development). Fixed.

src/su.c
alejandro-colomar: Thanks!

stoeckmann:
> Thanks for the carefully small commits! It was very easy to review this code thanks to that.

And thanks for the remark. The combination of setuid and signals definitely screams for very small steps in changes. ;)

alejandro-colomar:
I'll leave the PR open for a day or two.  If someone else wants to have a look at it before I merge, please say so.

@stoeckmann , do you want to wait for @lrh2000 to have a look?  Or should I merge in a couple of days if no one comments?

stoeckmann:
Having more eyes on these changes won't hurt. I've CC'ed @lrh2000 due to the regression fix back then. No strong requirement from my side.

src/su.c
```diff
@@ -418,6 +426,10 @@ static void prepare_pam_close_session (void)
 
 		while (0 == waitpid (pid_child, &status, WNOHANG)) {
 			sigsuspend (&ourset);
+			if (timeout) {
+				kill_child(0);
```
lrh2000: *If* we want to follow the existing style, we may want to add a whitespace between the function name and the arguments:
```suggestion
				kill_child (0);
```

src/su.c
```diff
@@ -405,7 +404,7 @@ static void prepare_pam_close_session (void)
 		sigfillset (&ourset);
 		if (sigprocmask (SIG_BLOCK, &ourset, NULL) != 0) {
 			fprintf (stderr, _("%s: signal masking malfunction\n"), Prog);
-			kill_child (0);
+			kill_child(pid_child);
```
lrh2000, line 404: Same.

src/su.c
```diff
@@ -169,12 +169,8 @@ static bool iswheel (const char *username)
...
 kill_child(pid_t pid_child)
 {
-	if (0 != pid_child) {
-		(void) kill (-pid_child, SIGKILL);
-		fputs(_(" ...killed.\n"), stderr);
-	} else {
-		fputs(_(" ...waiting for child to terminate.\n"), stderr);
-	}
+	kill(-pid_child, SIGKILL);
+	fputs(_(" ...killed.\n"), stderr);
```
lrh2000, line 174: Same.

lrh2000 (approved):
Nice cleanups. All LGTM. Meanwhile, I really appreciate that you still remember my PR from five years (!) ago. Time flies...

By the way, I noticed that the existing code has white space between the function name and its arguments. However, I don't know if we should continue to follow that convention in new code. Of course, I'm not very involved with the codebase here, so feel free to ignore this point anyway.

So with or without addressing the nitpicks, feel free to add:
```
Reviewed-by: Ruihan Li <lrh2000@pku.edu.cn>
```

stoeckmann:
> Nice cleanups. All LGTM. Meanwhile, I really appreciate that you still remember my PR from five years (!) ago. Time flies...

Yes, time really flies. Didn't think it was that long ago. But my memories about instances when I screwed up signal handling adjustments are really strong. :D

> By the way, I noticed that the existing code has white space between the function name and its arguments. However, I don't know if we should continue to follow that convention in new code. Of course, I'm not very involved with the codebase here, so feel free to ignore this point anyway.

The preferred way is to have no whitespaces in between. It's adjusted "on the fly" when a line is modified, but there's no one big commit which adjusts everything at once.

> ```
> Reviewed-by: Ruihan Li <lrh2000@pku.edu.cn>
> ```

Nice, boosting my confidence that this one is correct. ;)

lrh2000:
> The preferred way is to have no whitespaces in between. It's adjusted "on the fly" when a line is modified, but there's no one big commit which adjusts everything at once.

Thanks for the clarification! Got it. Everything is perfect to me then.

src/su.c
alejandro-colomar: No, I hate those spaces, and prefer to remove them in new code.  Thanks anyway!  :)

alejandro-colomar:
Thanks!  I'll merge.

PR #1493, opened 2026-01-10
lib/motd.c: Use motd only for non-PAM builds

stoeckmann:
The motd functionality is only used in login if no PAM support was compiled in.

### DIFF
--- a/lib/motd.c
+++ b/lib/motd.c
@@ -9,6 +9,8 @@
 
 #include "config.h"
 
+#ifndef USE_PAM
+
 #ident "$Id$"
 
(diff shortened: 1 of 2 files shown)


alejandro-colomar:
I'd like to avoid unnecessary conditional compilation.  It clutters the source code without much benefit.  If one wants to get rid of the binary code, LTO could help.

I'd like to only wrap in #ifdef's stuff that would produce build errors otherwise.

Apart from cluttering the code, having the code compiled unconditionally also means we build-test more of our code always, independent of how you're building it.

PR #1496, opened 2026-01-11
lib/commonio.c: Drop lrename

stoeckmann:
The lrename function follows symlinks when renaming files. Since the source is a temporary file and the target is the database file itself, which is opened with O_NOFOLLOW, this function is only useful for an attacker who manages to win some form of race.

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -45,7 +45,6 @@
 
 
 /* local function prototypes */
-static int lrename (const char *, const char *);
 static int check_link_count (const char *file, bool log);
 static int do_lock_file (const char *file, const char *lock, bool log);
 static /*@null@*/ /*@dependent@*/FILE *fmkstemp_set_perms (
@@ -68,32 +67,6 @@ static /*@dependent@*/ /*@null@*/struct commonio_entry *next_entry_by_name (
 
 static int lock_count = 0;
 static bool nscd_need_reload = false;
-
-/*
- * Simple rename(P) alternative that attempts to rename to symlink
- * target.
- */
-int lrename (const char *old, const char *new)
-{
-	int res;
-	char *r = NULL;
-	struct stat sb;
-
-	if (lstat (new, &sb) == 0 && S_ISLNK (sb.st_mode)) {
-		r = realpath (new, NULL);
-		if (NULL == r) {
-			perror ("realpath in lrename()");
-		} else {
-			new = r;
-		}
-	}
-
-	res = rename (old, new);
-
-	free (r);
-
-	return res;
-}
 
 static int check_link_count (const char *file, bool log)
 {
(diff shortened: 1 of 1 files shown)


lib/commonio.c
```diff
@@ -977,7 +950,7 @@ commonio_close(struct commonio_db *db, MAYBE_UNUSED bool process_selinux)
 		goto fail;
 	}
 
-	if (lrename (buf, db->filename) != 0) {
+	if (rename (buf, db->filename) != 0) {
```
alejandro-colomar: The patch seems good to me, but I think it would be interesting to research why this was done in the first place.

I've git-blame(1)'d this, and it was introduced in commit

0fa908302660 (2007-10-07; "[svn-upgrade] Integrating new upstream version, shadow (4.0.16)")

which says:

```
+2006-05-07  Tomasz K<B3>oczko  <kloczek@pld.org.pl>
+
+       * NEWS, po/fr.po:
+       updated (by Christian Perrier <bubulle@kheops.frmug.org>).
+
+       * NEWS, lib/commonio.c:
+       fixed handle relative symlinks too in lib/commonio.c (merge patch from Fedora).
```

Cc: @kloczek

Also, since this says that the patch was merged from Fedora, maybe @ikerexxe knows something about it.  Maybe Fedora keeps a link to the bug discussion.

In any case, please add

```
Fixes: 0fa908302660 (2007-10-07; "[svn-upgrade] Integrating new upstream version, shadow (4.0.16)")
```

stoeckmann: Also, see https://github.com/shadow-maint/shadow/commit/391a3847157c49ace82d7b26e5a0e76736e98c43 which added the `O_NOFOLLOW` in 2010 (almost exactly 16 years ago). Long after your referenced adjustments.

I guess that this one would have broken systems by now which relied on symlinks.

alejandro-colomar: Maybe we should also add

```
Fixes: 391a3847157c (2010-03-04; "2010-01-30  Paweł Hajdan, Jr.  <phajdan.jr@gentoo.org>")
```

stoeckmann: Added

alejandro-colomar: Thanks!  The patch looks good to me.  Let's wait for a week or so in case we get any useful feedback.  If nobody says anything, we'll merge.

hallyn:
LGTM.

stoeckmann:
Rebased on master (to remove the new lrename call here as well).

lib/commonio.c
hallyn: > Thanks! The patch looks good to me. Let's wait for a week or so in case we get any useful feedback. If nobody says anything, we'll merge.

Honestly I don't think you'll get further useful feedback without merging it :)

alejandro-colomar: Quite possibly.  :)

alejandro-colomar:
As @hallyn  suggested, I'll merge now.

PR #1497, opened 2026-01-11
lib/commonio.c: Use unpredictable temporary names

stoeckmann:
Make sure that an attacker with sufficient privileges cannot simply create a file with expected temporary name to retrieve content of previous and/or future database.

Not as thoroughly tested as usual, but given that https://github.com/shadow-maint/shadow/pull/1485 is basically the same and planned to be added to 4.19.1, this should be investigated as well.

In fact, this one is slightly worse because it affects every database write, even the ones with `--prefix` which could point into directories which are not as protected as `/etc`.

Still no serious security issue in my point of view, but better be safe than sorry and protect against it.

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -872,7 +872,7 @@ int
 commonio_close(struct commonio_db *db, MAYBE_UNUSED bool process_selinux)
 {
 	bool         errors = false;
-	char         buf[1024];
+	char         tmpf[PATH_MAX];
 	struct stat  sb;
 
 	if (!db->isopen) {
@@ -937,7 +937,7 @@ commonio_close(struct commonio_db *db, MAYBE_UNUSED bool process_selinux)
 		sb.st_gid = db->st_gid;
 	}
 
-	if (stprintf_a(buf, "%s.cioXXXXXX", db->filename) == -1)
+	if (stprintf_a(tmpf, "%s.cioXXXXXX", db->filename) == -1)
 		goto fail;
 
 #ifdef WITH_SELINUX
@@ -947,7 +947,7 @@ commonio_close(struct commonio_db *db, MAYBE_UNUSED bool process_selinux)
 	}
 #endif
 
-	db->fp = fmkstemp_set_perms(buf, &sb);
+	db->fp = fmkstemp_set_perms(tmpf, &sb);
 	if (NULL == db->fp) {
 		goto fail;
 	}
@@ -971,11 +971,11 @@ commonio_close(struct commonio_db *db, MAYBE_UNUSED bool process_selinux)
 	db->fp = NULL;
 
 	if (errors) {
-		unlink (buf);
+		unlink(tmpf);
 		goto fail;
 	}
 
-	if (lrename(buf, db->filename) != 0) {
+	if (lrename(tmpf, db->filename) != 0) {
 		goto fail;
 	}
 


lib/commonio.c
```diff
@@ -266,24 +264,26 @@ static /*@null@*/ /*@dependent@*/FILE *fopen_set_perms (
...
+	/* fmkomstemp_set_perms is used for intermediate files */
 	(void) unlink (name);
 	return NULL;
 }
 
 
-static int create_backup (const char *backup, FILE * fp)
+static int create_backup (const char *name, FILE * fp)
 {
+	char buf[PATH_MAX], target[PATH_MAX];
```
stoeckmann: Please note that the db filename itself is limited to 1023 characters (+ NUL byte).
Since command line arguments for `--prefix` can be longer than that, silent truncation happens in quite some places within shadow, including cutting off prefixes so same files can be opened erroneously.

Fixing that is not part of this PR.

First step would be to check at least the return value of `commonio_set_name`, but definitely keep this in mind.

lib/commonio.c
```diff
@@ -240,17 +240,13 @@ static int do_lock_file (const char *file, const char *lock, bool log)
...
+	char *name,
 	const struct stat *sb)
 {
 	FILE *fp;
-	mode_t mask;
 
-	mask = umask (0777);
-	fp = fopen (name, mode);
-	(void) umask (mask);
+	fp = fmkomstemp(name, O_CLOEXEC, 0600);
```
alejandro-colomar: Since we're fchmod(2)ing the file a few lines below, should we create it with the final mode already?

Or do you think there's some reason to do it this way?

alejandro-colomar: BTW, the same question applies to the recently-merged PR.

lib/commonio.c
```diff
@@ -266,24 +262,26 @@ static /*@null@*/ /*@dependent@*/FILE *fopen_set_perms (
...
 	(void) fclose (fp);
-	/* fopen_set_perms is used for intermediate files */
+	/* fmkomstemp_set_perms is used for intermediate files */
 	(void) unlink (name);
 	return NULL;
 }
 
 
-static int create_backup (const char *backup, FILE * fp)
+static int create_backup (const char *name, FILE * fp)
```
alejandro-colomar, line 271: Since you're changing the name, would it make sense to call it `template`, such as in fmkomstemp() and mkostemp(3)?

lib/commonio.c
```diff
@@ -266,24 +262,26 @@ static /*@null@*/ /*@dependent@*/FILE *fopen_set_perms (
...
 {
+	char buf[PATH_MAX], target[PATH_MAX];
```
alejandro-colomar: Please use two spaces between `char` and `buf`.

lib/commonio.c
alejandro-colomar: Ahhh, sorry, this is not the template.

lib/commonio.c
stoeckmann: It's better to have it only read-write by the user first, THEN adjusting ownership and THEN mode:

If you set mode first and then ownership, you might grant a different group more access than wanted, e.g. "root shadow" with "rw-r-----". If you set the mode first, the current group gets more permissions than wanted. Granted, this is probably not important for root, who is most likely in group root.

Another possibility would be "rw----r--" for "root shadow". Setting mode first would give the group shadow, who wouldn't be owner, a short timeframe with read access.

Highly artificial examples, but that's why it's much easier to judge correctness by keeping it the same if there is no special reason to deviate from standard practice.

lib/commonio.c
```diff
@@ -299,22 +297,28 @@ static int create_backup (const char *backup, FILE * fp)
 	}
 	if ((c != EOF) || (ferror (fp) != 0) || (fflush (bkfp) != 0)) {
 		(void) fclose (bkfp);
-		/* FIXME: unlink the backup file? */
+		unlink(buf);
```
alejandro-colomar: This looks like it deserves a separate preceeding commit, right?

lib/commonio.c
stoeckmann: Adjusted. It as two spaces now (just repeating this to make sure that you really mean two spaces, since it looks a bit off to me)

lib/commonio.c
stoeckmann: Depends. I've thought about it, but it's not like we really created the backup file yet. But I'll split.

lib/commonio.c
alejandro-colomar: Yup, I took that practice from nginx:

<https://nginx.org/en/docs/dev/development_guide.html#code_style_variables>

lib/commonio.c
stoeckmann: Adjusted: Split into two

lib/commonio.c
alejandro-colomar: Using O_CLOEXEC is introduced in this patch (the current calls use `"w"`, and not `"we"`) without documenting it.  I think this should be introduced in a separate patch (probably, a preceeding one).

Also, such a patch should probably evaluate whether other calls to fmkomstemp() should also use it or not, and mention that and why in the commit message.

alejandro-colomar: Also, do we want to hardcode O_CLOEXEC here, or should we offer an open(2)-like argument like fmkomstemp()?

alejandro-colomar: I have in mind a couple of alternatives:

-  Call it `fmkestemp_set_perms()`, where `e` is for consistency with fopen(3)'s `"e"`.

   In this case, we'd hardcode O_CLOEXEC.

-  Call it `fmkostemp_set_perms()`, where `o` is for consistency with mkostemp(3) (meaning it takes an open(2)-like argument.

   In this case, we'd pass O_CLOEXEC as argument.

If we always want O_CLOEXEC, I guess the former is more reasonable.

lib/commonio.c
hallyn: We're no longer in the memory constrained days of yore (although people do some silly things with container memory limits).

Simply always allocating enough memory instead of doing e.g. char buf[PATH_MAX] might be the best approach.

I think it best to open an issue for this, and resolve this conversation.

Thanks for bringing it up.

lib/commonio.c
hallyn: > Using O_CLOEXEC is introduced in this patch (the current calls use `"w"`, and not `"we"`) without documenting it. I think this should be introduced in a separate patch (probably, a preceeding one).
> 
> Also, such a patch should probably evaluate whether other calls to fmkomstemp() should also use it or not, and mention that and why in the commit message.

A separate patch to call it out sounds good.

But deciding whether other calls should be changed should not be a part of this PR.  Feel free to open an issue for it.

lib/commonio.c
alejandro-colomar: I have doubts about that.  It requires many more malloc(3) calls that can fail, etc..  Also, it will require more lines of code.  I prefer being cautious about it.

I'm contrary to GNU Hurd's idea of no limits, because software always has limits somewhere; it's just pushing them elsewhere, and hiding the UB under the rug where you just don't see it so obviously.

lib/commonio.c
stoeckmann: My bad. I'm used to setting everything that's not supposed to land as open file descriptor in a child `O_CLOEXEC`, but these are short-living handles with no `exec` calls in between. Removed.

lib/commonio.c
stoeckmann: Created issue for more discussion: https://github.com/shadow-maint/shadow/issues/1498

lib/commonio.c
```diff
@@ -266,24 +262,26 @@ static /*@null@*/ /*@dependent@*/FILE *fopen_set_perms (
...
+	/* fmkomstemp_set_perms is used for intermediate files */
 	(void) unlink (name);
 	return NULL;
 }
 
 
-static int create_backup (const char *backup, FILE * fp)
+static int create_backup (const char *name, FILE * fp)
 {
+	char  buf[PATH_MAX], target[PATH_MAX];
```
alejandro-colomar: Can we use `tmpf` (temporary file) instead of `buf`?  That should make uses more readable.

That is, `unlink(tmpf)` seems more readable (it tells me we're unlinking the temporary file) than `unlink(buf)`.

lib/commonio.c
```diff
@@ -299,22 +297,28 @@ static int create_backup (const char *backup, FILE * fp)
...
-		unlink(backup);
+		unlink(buf);
+		return -1;
+	}
+
+	stprintf_a(target, "%s-", name);
+	if (lrename(buf, target) != 0) {
+		unlink(buf);
 		return -1;
 	}
```
alejandro-colomar, line 318: Why use a `-` when we were using `+` before?

lib/commonio.c
```diff
@@ -240,17 +240,13 @@ static int do_lock_file (const char *file, const char *lock, bool log)
...
+	char *name,
 	const struct stat *sb)
 {
 	FILE *fp;
-	mode_t mask;
 
-	mask = umask (0777);
-	fp = fopen (name, mode);
-	(void) umask (mask);
+	fp = fmkomstemp(name, 0, 0600);
```
alejandro-colomar, line 249: Since we don't get open(2)-like flags (nor mode_t), I think we should call this fmkstemp_set_perms().

lib/commonio.c
stoeckmann: Adjusted

lib/commonio.c
stoeckmann: This is create_backup, which used a `-` before as well. The `+` version has been replaced. This one does not come from create_backup but is a temporary file which is filled with new content.

https://github.com/shadow-maint/shadow/blob/d5638a5b2b7b9038fe3436d3ffb9b973ad479176/lib/commonio.c#L903 (backup file creation)
https://github.com/shadow-maint/shadow/blob/d5638a5b2b7b9038fe3436d3ffb9b973ad479176/lib/commonio.c#L942 ("new file" creation)

lib/commonio.c
stoeckmann: Adjusted

lib/commonio.c
stoeckmann: And I forgot to remove the now unnecessary `%s-` pattern, which is why I didn't see it in this PR diff when looking for it. D'oh! Thanks.

lib/commonio.c
stoeckmann: And changed the other `buf` into `tmpf` as well. Added a new commit for it.

lib/commonio.c
stoeckmann: To stay consistent, the other `buf[1024]` in `commonio_close` is `newf[PATH_MAX]` now, too. Since we append to the name, this should be enough for now. Pretty much "while at it."

lib/commonio.c
alejandro-colomar: If you want to do O_CLOEXEC, I think it would be a good idea.  I just didn't want it in the same commit.  If you open a PR in the future for that, it will be good, I think.  As you say, not using O_CLOEXEC should be rare.

lib/commonio.c
```diff
@@ -266,24 +262,26 @@ static /*@null@*/ /*@dependent@*/FILE *fopen_set_perms (
...
-static int create_backup (const char *backup, FILE * fp)
+static int create_backup (const char *name, FILE * fp)
 {
+	char  tmpf[PATH_MAX], target[PATH_MAX];
 	struct stat sb;
 	struct utimbuf ub;
 	FILE *bkfp;
 	int c;
 
+	stprintf_a(tmpf, "%s.cioXXXXXX", name);
```
hallyn, line 279: I do wish that this and the next stprintf_a also returned an error if

alejandro-colomar: Yeah, we could return -1 if it fails.  After all, this function already reports some errors, and this is clearly an error.

alejandro-colomar: I'll merge; we can do that in a separate PR.  As @stoeckmann said, this is no regression here.

lib/commonio.c
```diff
@@ -299,22 +297,28 @@ static int create_backup (const char *backup, FILE * fp)
...
+	if (lrename(tmpf, target) != 0) {
+		unlink(tmpf);
 		return -1;
 	}
 
 	ub.actime = sb.st_atime;
 	ub.modtime = sb.st_mtime;
-	(void) utime (backup, &ub);
+	(void) utime(tmpf, &ub);
 	return 0;
```
alejandro-colomar, line 322: @stoeckmann 

I just realized that this introduced a (minor) bug:

utime(2) here is trying to change the time of 'tmpf', but we just renamed it to target.
We should have changed the call to

```c
utime(target, &ub);
```

stoeckmann: Yeah, right ... Do you create a PR?

alejandro-colomar: Okay, I'll do.  Thanks for confirming!

stoeckmann: Guess I was a bit faster now and didn't want to push work to fix my bugs to someone else. See https://github.com/shadow-maint/shadow/pull/1502

alejandro-colomar: Thanks!  :)

Issue #1500, opened 2026-01-14
mode_t mask in backups


alejandro-colomar:
We have (at least) two places where we create backup files.

In one of them, we use a hardcoded mask of 0664:

<https://github.com/shadow-maint/shadow/blob/c745eea4a45e6c8eba551f4d82bc952b34740fc9/lib/commonio.c#L257>

which blames back to 45c6603cc86c (2007-10-07; "[svn-upgrade] Integrating new upstream version, shadow (19990709)").

In the other, we have no mask at all:

<https://github.com/shadow-maint/shadow/blob/c745eea4a45e6c8eba551f4d82bc952b34740fc9/src/vipw.c#L143>

which blames back to the same 45c6603cc86c (2007-10-07; "[svn-upgrade] Integrating new upstream version, shadow (19990709)").

Cc: @stoeckmann , @hallyn 

Should we change anything?

alejandro-colomar:
See also:

```diff
$ diff -U999 <(grepc create_backup lib/commonio.c) <(grepc create_backup_file src/vipw.c)
--- /dev/fd/63	2026-01-14 16:10:44.051293920 +0100
+++ /dev/fd/62	2026-01-14 16:10:44.051293920 +0100
@@ -1,44 +1,36 @@
-lib/commonio.c:static int create_backup (const char *, FILE *);
-lib/commonio.c:static int create_backup (const char *name, FILE * fp)
+src/vipw.c:static int create_backup_file (FILE *, char *, struct stat *);
+src/vipw.c:static int create_backup_file (FILE * fp, char *backup, struct stat *sb)
 {
-	char  tmpf[PATH_MAX], target[PATH_MAX];
-	struct stat sb;
 	struct utimbuf ub;
 	FILE *bkfp;
 
-	stprintf_a(tmpf, "%s.cioXXXXXX", name);
-	if (fstat (fileno (fp), &sb) != 0) {
-		return -1;
-	}
-
-	bkfp = fmkstemp_set_perms(tmpf, &sb);
+	bkfp = fmkomstemp(backup, 0, 0600);
 	if (NULL == bkfp) {
 		return -1;
 	}
 
 	if (fcopy(bkfp, fp) == -1) {
-		(void) fclose (bkfp);
-		unlink(tmpf);
+		fclose (bkfp);
+		unlink (backup);
 		return -1;
 	}
 	if (fsync (fileno (bkfp)) != 0) {
 		(void) fclose (bkfp);
[...]

PR #1501, opened 2026-01-14
vipw: Prefer file based functions over path functions in correct order

stoeckmann:
Avoid symlink dereferences and other race conditions. Also, set ownership and then modes to handle more edge cases.

Purely defensive measures, since vipw operates within /etc which should have proper permissions to avoid these issues to begin with.

Helps https://github.com/shadow-maint/shadow/issues/1500

### DIFF
--- a/src/vipw.c
+++ b/src/vipw.c
@@ -136,8 +136,8 @@ static int create_backup_file (FILE * fp, char *backup, struct stat *sb)
 	ub.actime = sb->st_atime;
 	ub.modtime = sb->st_mtime;
 	if (   (utime (backup, &ub) != 0)
-	    || (fchmod(fileno(bkfp), sb->st_mode) != 0)
-	    || (fchown(fileno(bkfp), sb->st_uid, sb->st_gid) != 0)) {
+	    || (fchown(fileno(bkfp), sb->st_uid, sb->st_gid) != 0)
+	    || (fchmod(fileno(bkfp), sb->st_mode) != 0)) {
 		fclose(bkfp);
 		unlink (backup);
 		return -1;


stoeckmann:
@alejandro-colomar, let me know if I should you add as "Reported-by" to these commits. I'm not sure if it fits your remarks in the issue, but without it, I would have skipped this detail (again). :)

src/vipw.c
```diff
@@ -132,16 +132,17 @@ static int create_backup_file (FILE * fp, char *backup, struct stat *sb)
...
 	if (   (utime (backup, &ub) != 0)
-	    || (chmod (backup, sb->st_mode) != 0)
-	    || (chown (backup, sb->st_uid, sb->st_gid) != 0)) {
+	    || (fchmod(fileno(bkfp), sb->st_mode) != 0)
+	    || (fchown(fileno(bkfp), sb->st_uid, sb->st_gid) != 0)) {
+		unlink (backup);
+		return -1;
+	}
+
+	if (fclose (bkfp) != 0) {
```
alejandro-colomar, line 146: Since you're moving the close(2) down, fchmod(2) failures should now fclose(bkfp) the file, right?

src/vipw.c
```diff
@@ -136,8 +136,8 @@ static int create_backup_file (FILE * fp, char *backup, struct stat *sb)
 	ub.actime = sb->st_atime;
 	ub.modtime = sb->st_mtime;
 	if (   (utime (backup, &ub) != 0)
-	    || (fchmod(fileno(bkfp), sb->st_mode) != 0)
-	    || (fchown(fileno(bkfp), sb->st_uid, sb->st_gid) != 0)) {
+	    || (fchown(fileno(bkfp), sb->st_uid, sb->st_gid) != 0)
+	    || (fchmod(fileno(bkfp), sb->st_mode) != 0)) {
```
alejandro-colomar: For patch 2:

```
```

alejandro-colomar:
> @alejandro-colomar, let me know if I should you add as "Reported-by" to these commits. I'm not sure if it fits your remarks in the issue, but without it, I would have skipped this detail (again). :)

Yup, feel free to add it.  Thanks!  :)

src/vipw.c
stoeckmann: Great catch! Fixed

stoeckmann:
> Yup, feel free to add it. Thanks! :)

Adjusted

PR #1502, opened 2026-01-14
lib/commonio.c: Call utime for correct target

stoeckmann:
Since tmpf has been already renamed to target at this point, call utime with target instead of tmpf.

Fixes: f8732b17dd1d (2026-01-11; "lib/commonio.c: Use unpredictable temporary names")

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -318,7 +318,7 @@ static int create_backup (const char *name, FILE * fp)
 
 	ub.actime = sb.st_atime;
 	ub.modtime = sb.st_mtime;
-	(void) utime(tmpf, &ub);
+	(void) utime(target, &ub);
 	return 0;
 }
 


lib/commonio.c
```diff
@@ -318,7 +318,7 @@ static int create_backup (const char *name, FILE * fp)
 
 	ub.actime = sb.st_atime;
 	ub.modtime = sb.st_mtime;
-	(void) utime(tmpf, &ub);
+	(void) utime(target, &ub);
```
alejandro-colomar, line 321: For the 'Fixes:' tag, could you please use the commit date?  That is a bit more useful, because commit dates are correlative, while author dates are a bit arbitrary.

To generate the tags, I use the following git alias:

In /etc/gitconfig:
```
[alias]
	ref = show --no-patch --abbrev=12 --date=short \
		--format=tformat:'%C(auto)%h%C(reset) %C(dim white)(%cd%x3B \"%C(reset)%C(auto)%s%C(reset)%C(dim white)\")%C(reset)'
```

and then:

```sh
$ git ref f8732b1
f8732b17dd1d (2026-01-14; "lib/commonio.c: Use unpredictable temporary names")
```

stoeckmann: Thanks, time to save that. Adjusted.

alejandro-colomar (approved):
Thanks!

```
```

PR #1503, opened 2026-01-14
vipw: Set correct pattern for fmkomstemp

stoeckmann:
The fmkomstemp call requires a suffix of XXXXXX for correct operation. Do so in TCB case as well.

Note: If something fails and the file resides in this directory, it could be interpreted as a username. Use the ',' character as an illegal character to prevent shadow tools from erroneously accessing this file and assuming that the user actually exists.

Fixes: a5b3d56e2902 (2026-01-09; "vipw: Use fmkomstemp for temporary file")

### DIFF
--- a/src/vipw.c
+++ b/src/vipw.c
@@ -426,7 +426,7 @@ vipwedit (const char *file, int (*file_lock) (void), int (*file_unlock) (bool))
 		if (stat (file, &st1) != 0) {
 			vipwexit (_("failed to stat edited file"), errno, 1);
 		}
-		to_rename = aprintf("%s+", file);
+		to_rename = aprintf("%s,XXXXXX", file);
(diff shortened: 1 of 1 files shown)


src/vipw.c
```diff
@@ -426,7 +426,7 @@ vipwedit (const char *file, int (*file_lock) (void), int (*file_unlock) (bool))
 		if (stat (file, &st1) != 0) {
 			vipwexit (_("failed to stat edited file"), errno, 1);
```
stoeckmann, line 427: I'd argue that this error message is incorrect. `file` should be the original file if I'm not mistaken. Not touching it in this PR though. And in general not touching it because I don't have a good TCB test system.

alejandro-colomar (approved):
Thanks!

```
```

Issue #1479, opened 2026-01-05
Unexpected `usermod -U` exit code


stoeckmann:
I would have expected a different result after performing these steps (and seeing the warning):

1. Create a new user
```
useradd testuser
```
2. Let's make sure that nothing unexpected shows up
```
grep testuser /etc/passwd /etc/shadow
```
```
/etc/passwd:testuser:x:1001:1001::/home/testuser:/usr/bin/bash
/etc/shadow:testuser:!:20458:0:99999:7:::
```
3. Unlock the account
```
usermod -U testuser
```
```
usermod: unlocking the user's password would result in a passwordless account.
You should set a password with usermod -p to unlock this user's password.
```
4. Check the return value
```
echo $?
```
```
0
```
5. Check the files again
```
grep testuser /etc/passwd /etc/shadow
```
```
/etc/passwd:testuser:x:1001:1001::/home/testuser:/usr/bin/bash
/etc/shadow:testuser:!:20458:0:99999:7:::
```

While I'm fine with a tool which makes sure that I do not accidentally create a passwordless acount (even though I asked for it), the exit code `0` seems off here. I would expect an error value, after all the operation was not successfully performed. Yet, I won't just open a PR for this, after all the code is like that since at least 2007. Quite a long time.

stoeckmann:
One possibility arises from what systemd does, as can be seen here:
https://github.com/systemd/systemd/blob/6a9f0641cd924a5bb41f63a2b9e18838b11db9b9/src/basic/user-util.h#L153
```
/* A locked *and* invalid password for "struct spwd"'s .sp_pwdp and "struct passwd"'s .pw_passwd field */
#define PASSWORD_LOCKED_AND_INVALID "!*"
```

If a user gets such an entry when a shadow entry is created, the operation would be successful, but a login with a password would still fail.

This would require an adjustment in how we create new entries. And it would bring us closer to what systemd does (if that's of interest).

Old entries before the shadow change get still this error message and old behavior, new entries can be unlocked without error message.

stoeckmann:
And last but not least comment: I'm not sure if `*` is really such a nice character for an invalid password. See: https://github.com/linux-pam/linux-pam/issues/866

If anyone somehow manages to go down the whole `crypt` path on a musl-system with the `*` error return value, it could suddenly allow user access.

alejandro-colomar:
Let's report an error code for now.  That's probably the spirit of the current code.  The `0` is probably a mistake that was never fixed.

[...]

PR #1504, opened 2026-01-14
is_valid_hash: accept two more cases

hallyn:
We currently allow ! followed by a valid hash but not ! followed by nothing, and * followed by nothing but not * followed by something.

Accept the other cases.

Cc: Alejandro Colomar <alx@kernel.org>
Cc: Chris Hofstaedtler <zeha@debian.org>

### DIFF
--- a/lib/chkhash.c
+++ b/lib/chkhash.c
@@ -40,9 +40,13 @@ match_regex(const char *pattern, const char *string)
 bool 
 is_valid_hash(const char *hash) 
 {
-	hash = strprefix(hash, "!") ?: hash;
+	// If the hash starts with '!', it means just lock the password.
+	if (strprefix(hash, "!"))
+		return true;
 
-	if (streq(hash, "*"))
+	// If the hash starts with '*', this is an intentional way to prevent
+	// a user from logging in with password.
+	if (strprefix(hash, "*"))
 		return true;
 
 	// Minimum hash length


lib/chkhash.c
```diff
@@ -40,9 +40,13 @@ match_regex(const char *pattern, const char *string)
...
 is_valid_hash(const char *hash) 
 {
-	hash = strprefix(hash, "!") ?: hash;
+	// If the hash starts with '!', it means just lock the account.
+	if (strprefix(hash, "!"))
+		return true;
 
-	if (streq(hash, "*"))
+	// If the hash starts with '*', this is an intentional way to prevent
+	// a user from loggingin with password.
```
stoeckmann: `loggingin` -> `logging in`

hallyn: Thanks, my laptops are kind of old and the keyboards are getting crufty.

lib/chkhash.c
```diff
@@ -40,9 +40,13 @@ match_regex(const char *pattern, const char *string)
 bool 
 is_valid_hash(const char *hash) 
 {
-	hash = strprefix(hash, "!") ?: hash;
+	// If the hash starts with '!', it means just lock the account.
+	if (strprefix(hash, "!"))
+		return true;
```
alejandro-colomar, line 45: Is this accurate?  shadow(5) says `!` means the password is locked, not the account.  One would still be able to log in via other methods, right?

Or is shadow(5) incorrect?

lib/chkhash.c
```diff
@@ -40,9 +40,13 @@ match_regex(const char *pattern, const char *string)
...
 
-	if (streq(hash, "*"))
+	// If the hash starts with '*', this is an intentional way to prevent
+	// a user from logging in with password.
+	if (strprefix(hash, "*"))
 		return true;
```
alejandro-colomar, line 50: According to shadow(5), `!` is the prefix to lock a password.  Please check the manual page to make sure both match.

lib/chkhash.c
stoeckmann: shadow(5) is correct (as is passwd's manual page). It's a locked password.

lib/chkhash.c
alejandro-colomar: As I understand it, `!` is the prefix to lock temporarily a password.

On the other hand, `*` is to permanently remove a password.  Thus, this should not be a prefix, but the entire hash.

Am I mistaken?

lib/chkhash.c
```diff
@@ -40,9 +40,13 @@ match_regex(const char *pattern, const char *string)
 bool 
 is_valid_hash(const char *hash) 
 {
-	hash = strprefix(hash, "!") ?: hash;
+	// If the hash starts with '!', it means just lock the account.
+	if (strprefix(hash, "!"))
+		return true;
 
-	if (streq(hash, "*"))
```
alejandro-colomar, line 45: Hmm, you're right that I forgot to allow an empty hash for a passwordless account.

```c
if (streq(hash, "")
   return true;
```

Should fix that.

lib/chkhash.c
alejandro-colomar: FWIW, I'd be happy accepting `*` as a prefix, if that's common practice.  That would make it essentially a synonym of `!`, right?

stoeckmann: > That would make it essentially a synonym of `!`, right?

Not fully. `usermod -U user` removes a `!` but not a `*`. So your remark about permanently removed password (or "locked" password) is the key difference. In essence, both mean that the hash cannot be matched with any entered password.

lib/chkhash.c
stoeckmann: "Ho hash a valid hash. is vacuously true, since the statement "no hash is an invalid hash" is true at the same time.

I think it would make sense to rename the function. It's used to figure out if the spwd field in /etc/shadow may be filled with given value. It's no statement about the hash, but the field content.

But this is derailing the PR and I'm bad at coming up with function names anyway. :)

So, despite the function name, an empty field is valid. Yes, just frowned upon by `usermod -U` for a new user, see https://github.com/shadow-maint/shadow/issues/1479

alejandro-colomar: Agree.

lib/chkhash.c
alejandro-colomar: The difference between `*hash` and `*` would be that `*` is a permanent removal of the password hash (you'll need to regenerate it), while `*hash` still allows manual restoration.  So, `*hash` would be a more inconvenient synonym of `!hash`, while `*` would mean "whatever your password hash was, we don't know it anymore, and can't recover it".

lib/chkhash.c
ikerexxe: Right, `!` is used for a locked password

lib/chkhash.c
ikerexxe: Yes, I'd update the comment and the man page to state the meaning of these things

lib/chkhash.c
hallyn: Just to make sure I understood right - you're just saying to update the comment?

(If so, done)

alejandro-colomar: Yes.

lib/chkhash.c
```diff
@@ -40,9 +40,13 @@ match_regex(const char *pattern, const char *string)
 bool 
 is_valid_hash(const char *hash) 
 {
-	hash = strprefix(hash, "!") ?: hash;
+	// If the hash starts with '!', it means just lock the password.
+	if (strprefix(hash, "!"))
+		return true;
```
alejandro-colomar, line 45: This change is accepting anything after a `!`, such as `!invalid_hash`.  We'd want to make sure that for a string `!hash` we still verify that `hash` is a valid hash for some algorithm.

I think what you want is to just add a check for an empty string after jumping over the `!`, as done in <https://github.com/shadow-maint/shadow/pull/1505>.

ikerexxe:
@hallyn can you take a look at the feedback? It should be quite straightforward, and we can merge this PR

alejandro-colomar:
> @hallyn can you take a look at the feedback? It should be quite straightforward, and we can merge this PR

Actually, I don't want to merge this PR; I'm not convinced that it's good.

alejandro-colomar:
> > @hallyn can you take a look at the feedback? It should be quite straightforward, and we can merge this PR
> 
> Actually, I don't want to merge this PR; I'm not convinced that it's good.

Instead, please have a look at <https://github.com/shadow-maint/shadow/pull/1505>, which should be ready to merge.

lib/chkhash.c
```diff
@@ -40,9 +40,13 @@ match_regex(const char *pattern, const char *string)
...
 
-	if (streq(hash, "*"))
+	// If the hash starts with '*', this is an intentional way to prevent
+	// a user from logging in with password.
+	if (strprefix(hash, "*"))
 		return true;
```
alejandro-colomar, line 50: @Zugschlus, @zeha , @hallyn 

Are you aware of uses of `*` followed by a hash?  As far as I know, `*` is intended to replace a hash, not to be a prefix.

ikerexxe:
Superseded by https://github.com/shadow-maint/shadow/pull/1505

PR #1505, opened 2026-01-15
lib/chkhash.c: is_valid_hash(): Allow passwordless account

alejandro-colomar:
Reported-by: @hallyn 
Reported-by: @Zugschlus
Cc: @stoeckmann 

Serge, this PR is smaller than yours, as I acknowledge this one was an obvious oversight from me, while the other we still need to talk about it.  Please merge this, and then we'll talk about allowing `*hash`.  Does it sound good?

---

Revisions:

<details>
<summary>v2</summary>

-  Add comments documenting each case.

```
$ git rd 
1:  12cc0aa0c ! 1:  bdf93b20a lib/chkhash.c: is_valid_hash(): Accept an empty hash
    @@ lib/chkhash.c: is_valid_hash(const char *hash)
      {
        hash = strprefix(hash, "!") ?: hash;
      
    ++  // Passwordless account; discouraged
     +  if (streq(hash, ""))
    -+          return true;  // Passwordless account; discouraged
    ++          return true;
     +
        if (streq(hash, "*"))
                return true;
2:  7e60f6d89 = 2:  051940ede lib/chkhash.c: is_valid_hash(): Update comment
-:  --------- > 3:  5bb2c2ecc lib/chkhash.c: is_valid_hash(): Comment meaning of !hash and *
```
</details>

<details>
<summary>v2b</summary>

-  Commit 1 was actually co-authored-by @hallyn , as I based my code on his.

```
$ git rd 
1:  bdf93b20a ! 1:  52f1342e3 lib/chkhash.c: is_valid_hash(): Accept an empty hash
    @@ Commit message
         Link: <https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1124835>
         Reported-by: Marc 'Zugschlus' Haber <mh+githubvisible@zugschlus.de>
         Reported-by: "Serge E. Hallyn" <serge@hallyn.com>
    +    Co-authored-by: "Serge E. Hallyn" <serge@hallyn.com>
         Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
      ## lib/chkhash.c ##
2:  051940ede = 2:  0bf0b7ccc lib/chkhash.c: is_valid_hash(): Update comment
3:  5bb2c2ecc = 3:  4186bdfe8 lib/chkhash.c: is_valid_hash(): Comment meaning of !hash and *
```
</details>

<details>
<summary>v2c</summary>

-  Rebase

```
$ git rd 
[...]

PR #1508, opened 2026-01-15
lib/chkhash.c: is_valid_hash(): Accept '!' as the hash

AdamWill:
Following ddc2549, also allow a bare '!' as the hash. anaconda, the installer used by Red Hat, Fedora and related distributions, has used this as the default hash for the root password (if the user does not explicitly set a root password during the install process) for a long time (at least since 2015). It has the same effect as a bare '*': it'll make the account inaccessible.

### DIFF
--- a/lib/chkhash.c
+++ b/lib/chkhash.c
@@ -40,6 +40,9 @@ match_regex(const char *pattern, const char *string)
 bool 
 is_valid_hash(const char *hash) 
 {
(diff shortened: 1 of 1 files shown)


alejandro-colomar:
This is a duplicate of <https://github.com/shadow-maint/shadow/pull/1505>.

Anyway, thanks for PR!

Cc: @hallyn

PR #1514, opened 2026-01-21
tests/system/framework/utils/tools.py: apply style fix

ikerexxe:
apply style fix for tuple unpacking

### DIFF
--- a/tests/system/framework/utils/tools.py
+++ b/tests/system/framework/utils/tools.py
@@ -61,7 +61,7 @@ class UnixObject(object):
             if len(o) != 2 or not isinstance(o[0], int) or not isinstance(o[1], str):
                 raise NotImplementedError(f"Unable to compare {type(o)} with {self.__class__}")
 
-            (id, name) = o
+            id, name = o
             return id == self.id and name == self.name
         elif isinstance(o, UnixObject):
             # Fallback to identity comparison


tests/system/framework/utils/tools.py
```diff
@@ -61,7 +61,7 @@ def __eq__(self, o: object) -> bool:
             if len(o) != 2 or not isinstance(o[0], int) or not isinstance(o[1], str):
                 raise NotImplementedError(f"Unable to compare {type(o)} with {self.__class__}")
 
-            (id, name) = o
+            id, name = o
```
alejandro-colomar, line 64: Doesn't the parenthesized version look more explicit?

Being a C programmer, the non-parenthesized version looks more like `id, (name = o)` to me.

ikerexxe: I'm not a big expert either but I have some experience with python

In python the **comma**, not the parentheses, creates a tuple or a 'target list' for unpacking. Python's operator precedence is designed so that `id, name` is evaluated together as the assignment target.

The reason for the change is that our project uses black. black’s rule is to remove all redundant parentheses to keep the code as minimal as possible. It flagged `(id, name) = o` as an error because the parentheses don't change the logic—they're just visual 'noise' in Python's eyes.

alejandro-colomar: Okay then; I guess.  :)

alejandro-colomar: BTW, it might be good to paste the shell where you run the 'black' command (or however that works) and its output.

ikerexxe: The message isn't very informative, to say the least. It's assumed that Black himself can fix the problem.

```
--- /home/runner/work/shadow/shadow/tests/system/framework/utils/tools.py	2026-01-22 01:48:42.422892+00:00
+++ /home/runner/work/shadow/shadow/tests/system/framework/utils/tools.py	2026-01-22 01:49:10.639515+00:00
@@ -59,11 +59,11 @@
             return o == self.id
         elif isinstance(o, tuple):
             if len(o) != 2 or not isinstance(o[0], int) or not isinstance(o[1], str):
                 raise NotImplementedError(f"Unable to compare {type(o)} with {self.__class__}")
 
-            (id, name) = o
+            id, name = o
             return id == self.id and name == self.name
         elif isinstance(o, UnixObject):
             # Fallback to identity comparison
             return NotImplemented
 
would reformat /home/runner/work/shadow/shadow/tests/system/framework/utils/tools.py

Oh no! 💥 💔 💥
1 file would be reformatted, 25 files would be left unchanged.

```

alejandro-colomar: > The message isn't very informative, to say the least. It's assumed that Black himself can fix the problem.
> 
> ```
> --- /home/runner/work/shadow/shadow/tests/system/framework/utils/tools.py	2026-01-22 01:48:42.422892+00:00
> +++ /home/runner/work/shadow/shadow/tests/system/framework/utils/tools.py	2026-01-22 01:49:10.639515+00:00
> @@ -59,11 +59,11 @@
>              return o == self.id
>          elif isinstance(o, tuple):
>              if len(o) != 2 or not isinstance(o[0], int) or not isinstance(o[1], str):
>                  raise NotImplementedError(f"Unable to compare {type(o)} with {self.__class__}")
>  
> -            (id, name) = o
> +            id, name = o
>              return id == self.id and name == self.name
>          elif isinstance(o, UnixObject):
>              # Fallback to identity comparison
>              return NotImplemented
>  
> would reformat /home/runner/work/shadow/shadow/tests/system/framework/utils/tools.py
> 
> Oh no! 💥 💔 💥
> 1 file would be reformatted, 25 files would be left unchanged.
> ```

That's still useful.  Can you also paste the command you used to print that?  That would help learn how to use it.

ikerexxe: It's `black --check --diff .` if you want to check only the patch or `black --check .` for the complete codebase. If you run `black .` it will check the complete database and fix any problem encountered.

https://github.com/shadow-maint/shadow/blob/master/.github/workflows/static-code-analysis.yml#L81 and following lines contain the exact commands to run the python linters

PR #1518, opened 2026-01-26
Remove suid account management tools

silverhadch:
Setuid account management has been disabled by default for over 7 years now and for way way longer on Distributions. So its time to permanently remove the option to not hold back the account managments tools with restrictions for being potential setuid binaries.

CC: @alejandro-colomar  @stoeckmann  @Conan-Kudo

### DIFF
--- a/etc/pam.d/Makefile.am
+++ b/etc/pam.d/Makefile.am
@@ -10,25 +10,13 @@ pamd_files = \
 	newusers \
 	passwd
 
-pamd_acct_tools_files = \
-	chgpasswd \
-	groupadd \
-	groupdel \
-	groupmod \
-	useradd \
-	userdel \
-	usermod
-
 if USE_PAM
 pamddir = $(sysconfdir)/pam.d
 pamd_DATA = $(pamd_files)
-if ACCT_TOOLS_SETUID
-pamd_DATA += $(pamd_acct_tools_files)
-endif
 endif
 
 if WITH_SU
 pamd_files += su
 endif
 
-EXTRA_DIST = $(pamd_files) $(pamd_acct_tools_files)
+EXTRA_DIST = $(pamd_files)
--- a/etc/pam.d/chgpasswd
+++ b/etc/pam.d/chgpasswd
@@ -1,4 +1,0 @@
-#%PAM-1.0
-auth		sufficient	pam_rootok.so
-account		required	pam_permit.so
-password	include		system-auth
--- a/etc/pam.d/groupadd
+++ b/etc/pam.d/groupadd
@@ -1,4 +1,0 @@
-#%PAM-1.0
-auth		sufficient	pam_rootok.so
-account		required	pam_permit.so
-password	include		system-auth
--- a/etc/pam.d/groupdel
+++ b/etc/pam.d/groupdel
@@ -1,4 +1,0 @@
-#%PAM-1.0
-auth		sufficient	pam_rootok.so
-account		required	pam_permit.so
-password	include		system-auth
--- a/etc/pam.d/groupmod
+++ b/etc/pam.d/groupmod
@@ -1,4 +1,0 @@
-#%PAM-1.0
-auth		sufficient	pam_rootok.so
-account		required	pam_permit.so
-password	include		system-auth
--- a/etc/pam.d/useradd
+++ b/etc/pam.d/useradd
@@ -1,4 +1,0 @@
-#%PAM-1.0
-auth		sufficient	pam_rootok.so
-account		required	pam_permit.so
-password	include		system-auth
--- a/etc/pam.d/userdel
+++ b/etc/pam.d/userdel
@@ -1,4 +1,0 @@
-#%PAM-1.0
-auth		sufficient	pam_rootok.so
-account		required	pam_permit.so
-password	include		system-auth
--- a/etc/pam.d/usermod
+++ b/etc/pam.d/usermod
@@ -1,4 +1,0 @@
-#%PAM-1.0
-auth		sufficient	pam_rootok.so
-account		required	pam_permit.so
-password	include		system-auth


Conan-Kudo:
It probably makes sense to rip the code out, but that can be a followup PR.

silverhadch:
@alejandro-colomar 
Should I remove the comments and also the output at the end of the configure script too? I kept them there for now.

configure.ac
```diff
@@ -674,7 +650,7 @@ AC_MSG_NOTICE([shadow ${PACKAGE_VERSION} has been configured with the following
 
 	auditing support:		$with_audit
 	PAM support:			$with_libpam
-	suid account management tools:	$enable_acct_tools_setuid
+	suid account management tools:	no (permanently disabled)
```
alejandro-colomar: Yeah, you could remove this entirely.

If anyone tries using this flag accidentally, the command should fail, so they'll realize.

configure.ac
```diff
@@ -579,24 +570,9 @@ else
...
-		AC_DEFINE([ACCT_TOOLS_SETUID],
-		          [1],
-		          [Define if account management tools should be installed setuid and authenticate the callers])
-	fi
-fi
-AM_CONDITIONAL([ACCT_TOOLS_SETUID], [test "x$enable_acct_tools_setuid" = "xyes"])
-
+dnl REMOVED: All conditional logic for account-tools-setuid - permanently disabled
+dnl No AC_DEFINE for ACCT_TOOLS_SETUID since it's disabled
+AM_CONDITIONAL([ACCT_TOOLS_SETUID], [false])
```
alejandro-colomar: I'd entirely remove this.

configure.ac
```diff
@@ -123,17 +123,8 @@ AC_ARG_ENABLE([man],
...
-	[case "${enableval}" in
-	 yes) enable_acct_tools_setuid="yes" ;;
-	  no) enable_acct_tools_setuid="no" ;;
-	   *) AC_MSG_ERROR([bad value ${enableval} for --enable-account-tools-setuid])
-	   ;;
-	 esac],
-	[enable_acct_tools_setuid="no"]
-)
+dnl REMOVED: AC_ARG_ENABLE for account-tools-setuid - permanently disabled
+enable_acct_tools_setuid="no"
```
alejandro-colomar: I'd entirely remove this.

configure.ac
alejandro-colomar: As a consequence, all code wrapped in `#ifdef ACCT_TOOLS_SETUID` should also be removed.

silverhadch: I am doing that right now.

alejandro-colomar:
> It probably makes sense to rip the code out, but that can be a followup PR.

I'd make it a follow-up commit in the same PR.

jubalh:
> Setuid account management has been disabled by default for over 7 years now and for way way longer on Distributions.

For which ones?
We have them still enabled in (open)SUSE.

silverhadch:
> > Setuid account management has been disabled by default for over 7 years now and for way way longer on Distributions.
> 
> For which ones? We have them still enabled in (open)SUSE.

Which exactly?

silverhadch:
> > Setuid account management has been disabled by default for over 7 years now and for way way longer on Distributions.
> 
> For which ones? We have them still enabled in (open)SUSE.

Is there any particular reason for OpenSUSE to have them enabled in the first place? What exactly wouldnt work if they werent setuid binaries?

silverhadch:
@alejandro-colomar 
I think this is right? Also there are now empty void functions that wont return anything as intended but now since setuid is gone they basically are a redundant wasted stack frame.

src/Makefile.am
```diff
@@ -86,11 +83,7 @@ LDADD          = $(INTLLIBS) \
 		 $(top_builddir)/lib/libshadow.la \
 		 $(LIBTCB)
 
-if ACCT_TOOLS_SETUID
-LIBPAM_SUID  = $(LIBPAM)
-else
 LIBPAM_SUID  =
-endif
```
alejandro-colomar, line 93: Should we wipe LIBPAM_SUID entirely?  Maybe in a third commit removing dead code resulting from the second one?

src/chgpasswd.c
```diff
@@ -304,42 +299,6 @@ static void check_flags (void)
  */
 static void check_perms (void)
 {
```
alejandro-colomar: Same here: a third commit removing dead code would be nice.

src/Makefile.am
silverhadch: Is there any nice way to find dead Code or do I again have to grep and remove manually?

alejandro-colomar: I don't know if there's anything "nice".  I use grep(1) and remove manually.  It actually feels great to open vim(1) and remove large amounts of code.  :D

silverhadch: I guess I could configure the compiler and linker to be extremly agressive to find some chunks of dead Code.

silverhadch: Never mind the output is over 5000 lines of the compiler screaming about from this is a GNU Extension to ISO C forbids blabla . The Pedantic Flag does seem to do its job well. Guess grep and nvim it is.

silverhadch:
@alejandro-colomar 
I think I got every last line of dead Code now.

alejandro-colomar:
> @alejandro-colomar I think I got every last line of dead Code now.

Thanks!

silverhadch:
@jubalh @ikerexxe

I dug around a bit in openSUSE land and I think I mostly understand what’s going on now.

From what I can tell, this option was already enabled when shadow was first packaged (around 13 years ago) and it’s basically just stuck around ever since. Later on, openSUSE ran into issues with tools like newuidmap / newgidmap, which then got explicitly handled.

What’s confusing to me is that the compile option affects these binaries:

chgpasswd
chpasswd
groupadd
groupdel
groupmod
newusers
useradd
userdel
usermod


But none of these appear in shadow.permissions.

That’s the part that worries me a bit: if they’re actually ending up setuid because of the build option, but aren’t managed through the permissions framework, then they’re effectively flying under the radar. If that’s the case, it feels like something that just never got cleaned up rather than an intentional decision.

So I’m trying to figure out which of these is true:

this is all legacy behavior that nobody revisited, or

the permissions are expected to come from somewhere else that I’m missing.

Does that sound right to you, or am I overlooking something obvious here?

jubalh:
Thank you for checking @silverhadch ! I didn't read the whole conversation here carefully yet. Unfortunately I'm a bit busy right now but I hope I can find some time in the next days. It might very well be that is just a leftover.
Regarding permissions we also have a separate permissions [package](https://build.opensuse.org/package/show/Base:System/permissions). See https://github.com/openSUSE/permissions/blob/master/profiles/permissions.secure for example.

silverhadch:
> Thank you for checking @silverhadch ! I didn't read the whole conversation here carefully yet. Unfortunately I'm a bit busy right now but I hope I can find some time in the next days. Regarding permissions we also have a separate permissions [package](https://build.opensuse.org/package/show/Base:System/permissions). See https://github.com/openSUSE/permissions/blob/master/profiles/permissions.secure for example.

I think I found a Bug in OpenSUSE. The binaries arent actually setuid at all despite being compiled with the flag. I am sending a Screenshot right now.

silverhadch:
@jubalh 
I added chsh as an exmaple for an setuid binary.
Screenshot_opensusetumbleweed_2026-01-27_10:47:56

silverhadch:
> Thank you for checking @silverhadch ! I didn't read the whole conversation here carefully yet. Unfortunately I'm a bit busy right now but I hope I can find some time in the next days. It might very well be that is just a leftover. Regarding permissions we also have a separate permissions [package](https://build.opensuse.org/package/show/Base:System/permissions). See https://github.com/openSUSE/permissions/blob/master/profiles/permissions.secure for example.

Even there, the binaries don’t appear. But after looking into it, the binaries also aren’t setuid at all. So somewhere in the build process, the flag we want to remove is already being ignored—and apparently has been for who knows how long. That makes me think openSUSE is using this flag by accident, simply because it’s been there from the very beginning, since it doesn’t seem to have any effect at all. So I think openSUSE should just remove the flag and see what happens, because it doesn’t appear to do anything anyway.

silverhadch:
@jubalh 
I could spin up a small OBS Project without the flag and install that and see if anything breaks since I doubt it.

silverhadch:
> Thank you for checking @silverhadch ! I didn't read the whole conversation here carefully yet. Unfortunately I'm a bit busy right now but I hope I can find some time in the next days. It might very well be that is just a leftover. Regarding permissions we also have a separate permissions [package](https://build.opensuse.org/package/show/Base:System/permissions). See https://github.com/openSUSE/permissions/blob/master/profiles/permissions.secure for example.

Okay, I now know with 100% certainty what’s going on. In the spec file, the binaries affected by the flag have their setuid bit manually stripped instead of the flag being removed. So openSUSE won’t actually be affected by this PR, but they *do* need to clean up their shadow spec file.

Line 147 enables setuid support, and lines 331–333, 336–338, and 341–342 manually strip it again.

[https://build.opensuse.org/projects/Base:System/packages/shadow/files/shadow.spec?expand=1](https://build.opensuse.org/projects/Base:System/packages/shadow/files/shadow.spec?expand=1)

@ikerexxe
You can remove the blocker.

silverhadch:
@jubalh 
I made a custom package and everything seemed to work just fine. Tho i didnt do 100% testing of everything.
https://build.opensuse.org/package/show/home:silverhadch:branches:Base:System/shadow

silverhadch:
@ikerexxe 
The test grow unhappy when I remove the unused PAM Files even if in Installation they never get used.

silverhadch:
@ikerexxe 
Nevermind, a cosmic ray wanted to troll me.

silverhadch:
@jubalh
I opened a Request. 
https://build.opensuse.org/requests/1329486

etc/pam.d/Makefile.am
```diff
@@ -8,9 +8,7 @@ pamd_files = \
 	groupmems \
 	login \
 	newusers \
-	passwd
-
-pamd_acct_tools_files = \
+	passwd \
 	chgpasswd \
 	groupadd \
```
alejandro-colomar: Maybe the contents of pamd_acct_tools_files should be removed in commit 2 without merging to pamd_files?

That would be effectively merging commits 2 and 3.

silverhadch:
@alejandro-colomar 
Now all commits only touch what is in their message.

hallyn:
I'm good with (make that, happy about) this PR, whenever @jubalh has time to verify.

jubalh:
I took a look now.
@silverhadch is right, it must have been a leftover. We use the configure flag and later strip the binaries of the setuid bit.
So removing seems fine!

Also thanks for the submit request @silverhadch, i'll comment on OBS since some things need to be changed.

I found that ALT Linux also still uses this configure flag btw: https://packages.altlinux.org/en/sisyphus/srpms/shadow/specfiles/

alejandro-colomar:
> I took a look now. @silverhadch is right, it must have been a leftover. We use the configure flag and later strip the binaries of the setuid bit. So removing seems fine!
> 
> Also thanks for the submit request @silverhadch, i'll comment on OBS since some things need to be changed.
> 
> I found that ALT Linux also still uses this configure flag btw: https://packages.altlinux.org/en/sisyphus/srpms/shadow/specfiles/

Cc: @ldv-alt, @sem-gh

silverhadch:
> I took a look now. @silverhadch is right, it must have been a leftover. We use the configure flag and later strip the binaries of the setuid bit. So removing seems fine!
> 
> Also thanks for the submit request @silverhadch, i'll comment on OBS since some things need to be changed.
> 
> I found that ALT Linux also still uses this configure flag btw: https://packages.altlinux.org/en/sisyphus/srpms/shadow/specfiles/

With Debian/Ubuntu, Fedora/RHEL, and Arch Linux being the big three that most distributions are based on, and all of them having this disabled for years already, this shouldn’t be much of an issue. OpenSUSE is something I’m currently working on. NixOS is allergic to setuid binaries anyway, and several others are as well. Realistically, about 99% of the Linux ecosystem will be fine. The remaining 1% will need to adjust their package, but that’s not overly complex either.

alejandro-colomar:
> > I took a look now. @silverhadch is right, it must have been a leftover. We use the configure flag and later strip the binaries of the setuid bit. So removing seems fine!
> > Also thanks for the submit request @silverhadch, i'll comment on OBS since some things need to be changed.
> > I found that ALT Linux also still uses this configure flag btw: https://packages.altlinux.org/en/sisyphus/srpms/shadow/specfiles/
> 
> With Debian/Ubuntu, Fedora/RHEL, and Arch Linux being the big three that most distributions are based on, and all of them having this disabled for years already, this shouldn’t be much of an issue. OpenSUSE is something I’m currently working on. NixOS is allergic to setuid binaries anyway, and several others are as well. Realistically, about 99% of the Linux ecosystem will be fine. The remaining 1% will need to adjust their package, but that’s not overly complex either.

I'd like to make sure that ALT is okay with this.  Otherwise, if we remove this flag, and they re-enable it manually in their packaging, we might introduce vulnerabilities in ALT if we start using PATH, for example.

Let's wait to hear them.

silverhadch:
> > > I took a look now. @silverhadch is right, it must have been a leftover. We use the configure flag and later strip the binaries of the setuid bit. So removing seems fine!
> > > Also thanks for the submit request @silverhadch, i'll comment on OBS since some things need to be changed.
> > > I found that ALT Linux also still uses this configure flag btw: https://packages.altlinux.org/en/sisyphus/srpms/shadow/specfiles/
> > 
> > With Debian/Ubuntu, Fedora/RHEL, and Arch Linux being the big three that most distributions are based on, and all of them having this disabled for years already, this shouldn’t be much of an issue. OpenSUSE is something I’m currently working on. NixOS is allergic to setuid binaries anyway, and several others are as well. Realistically, about 99% of the Linux ecosystem will be fine. The remaining 1% will need to adjust their package, but that’s not overly complex either.
> 
> I'd like to make sure that ALT is okay with this.  Otherwise, if we remove this flag, and they re-enable it manually in their packaging, we might introduce vulnerabilities in ALT if we start using PATH, for example.
> 
> Let's wait to hear them.

Yes, understandable. We can wait. Tho if they want to reenable setuid on these binaries they would need to revert this PR because the flag no longer exist.

silverhadch:
@jubalh 
New Request.
https://build.opensuse.org/requests/1329767

jubalh:
I get notified about new requests on OBS ;) No need to spam here the people for who it is not relevant :)

sem-gh:
This is also okay for ALT. In fact, these utilities are packaged in ALT without the SUID bit, and honestly, I don't even remember why I added --enable-account-tools-setuid at the time. I'll build our package without that option.
And thanks for your concern!

alejandro-colomar:
As @jubalh is okay with it, and @sem-gh is also okay with it, I'll merge.  Thanks!

alejandro-colomar:
Cc: @hallyn

Issue #1519, opened 2026-01-27
is_valid_hash rejects valid SHA-256, SHA-512 and MD5 hashes whose salt contains a `\` or a `n`


AdamWill:
In Fedora's openQA tests, we use a [kickstart](https://forge.fedoraproject.org/quality/createhdds/src/branch/main/uploads/root-user-crypted-net.ks) with two crypted passwords in it:
```
rootpw --iscrypted $6$ansiogjasd0io9u3$9E1vMbLbXW14grtguedFGVjvhyBz1T.KIA3MJl1SWnGbtTpiXIAjbazIQAUKRNkNIEmd3mI0NCkFIVBrN41fZ.
user --name=test --password=$6$ansioasgfgadsghd$O8O8zom5hx.V8ib1jV91xuvIgYqA2b99tzhibkk3URITdCrDtbRbwJjMK1kW4l0/9W0brraGC4NUBtDoGv4Kl. --iscrypted
```
The plaintext root password is '111111' and the user password is 'test'. it's been long enough since we created this kickstart I no longer remember for sure how we created those, but we *probably* did what's listed in the [kickstart docs](https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#rootpw):

`python -c 'import crypt; print(crypt.crypt("My Password", "$6$My Salt"))'`

and indeed:
```
[adamw@toolbx fedora-toolbox-43 anaconda (main %)]$ python -c 'import crypt; print(crypt.crypt("111111", "$6$ansiogjasd0io9u3"))'
$6$ansiogjasd0io9u3$9E1vMbLbXW14grtguedFGVjvhyBz1T.KIA3MJl1SWnGbtTpiXIAjbazIQAUKRNkNIEmd3mI0NCkFIVBrN41fZ.
```

Anyhow, those work, they're valid. With previous versions of shadow, user creation works fine and after install we can log in as those users with the plaintext passwords.

With shadow 4.19, the installer's user creation fails because `chpasswd` rejects the hashes as invalid.

AdamWill:
Hmm, I actually need to verify exactly what's going on here, let me poke into it a bit more...

...yeah, ok, it is what I thought:
```
[...]

PR #1520, opened 2026-01-28
lib/chkhash.c: fix escaping in SHA-256 / SHA-512 / MD5 regexes

AdamWill:
`\\n` inside square brackets doesn't include or exclude the newline character. It includes or excludes a literal slash and the literal character 'n'.

### DIFF
--- a/lib/chkhash.c
+++ b/lib/chkhash.c
@@ -62,15 +62,15 @@ is_valid_hash(const char *hash)
 		return true;
 
 	// SHA-512: $6$ + salt + $ + 86-char hash
-	if (match_regex("^\\$6\\$(rounds=[1-9][0-9]{3,8}\\$)?[^$:\\n]{1,16}\\$[./A-Za-z0-9]{86}$", hash))
+	if (match_regex("^\\$6\\$(rounds=[1-9][0-9]{3,8}\\$)?[^$:\n]{1,16}\\$[./A-Za-z0-9]{86}$", hash))
 		return true;
 
 	// SHA-256: $5$ + salt + $ + 43-char hash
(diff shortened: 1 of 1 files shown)


AdamWill:
The effect of this was that we rejected any SHA-512, SHA-256 or MD5 hash whose salt contained a `\` or a `n`. I verified this with a [test script](https://github.com/shadow-maint/shadow/issues/1519#issuecomment-3808151141) and also by verifying that a Fedora install from a [kickstart containing two affected hashes](https://fedorapeople.org/groups/qa/kickstarts/root-user-crypted-net.ks) fails without this patch but succeeds when using an installer image containing a shadow build with this patch applied (and I can successfully log in as either account on the installed system).

AdamWill:
I don't want to sound too harsh here, but...this check turned out to be broken in really quite a lot of ways. Are there any lessons to be learned here to avoid similar issues in future? One obvious thing is that there are no unit tests for the check, even though the project appears to have a pretty good 'house style' for such unit tests; it would seem to be fairly easy to adapt e.g. `test_chkname.c` into a `test_chkhash.c`.

Of course, you'd still have had to be aware of the various broken scenarios when deciding on the hashes to include in the unit tests, or find them via research, or possibly just got lucky with whatever salt you decided to use, for this issue. But not including any tests *at all* seems like an obvious miss.

AdamWill:
BTW, you can confirm the root cause of this by comparing the result of `match_regex("[\n]", "n")` (false) with the result of `match_regex("[\\n]", "n")` (true).

alejandro-colomar:
I will release 4.19.3 this weekend (most likely).

alejandro-colomar:
> LGTM! Thank you for the patch.
> 
> @alejandro-colomar is there any reason why you didn't merge this request? I'm fine with merging it now

Nothing specific.  I was waiting in case anyone had comments.  I'll merge now.

PR #1523, opened 2026-01-30
Clarify documentation regarding creation of homedir

### LINKED ISSUE #1522: Confusing phrasing in `useradd` manpage for `--home-dir` flag
It states

> The directory HOME_DIR is not created by default. However it will be created for non-system users if either the -m flag is specifed or CREATE_HOME in login.defs is set to true. However, it will never be created if the -M flag is specified.

However, `--system` states

> Note that useradd will not create a home directory for such a user, regardless of the default setting in /etc/login.defs (CREATE_HOME). You have to specify the -m options if you want a home directory for a system account to be created.

This is also what actually happens. A more clear phrasing would be to rearrange the "for non-system users":

> However it will be created if either the -m flag is specifed or (for non-system users) CREATE_HOME in login.defs is set to true.

septatrix:
The "for non-system users" only applies to the CREATE_HOME config and is does not matter when using the `-m` flag.

Fixes #1522.

```
podman run --rm -it fedora:44
[root@cae89bd5de3c /]# ls -l /home/
total 0
[root@cae89bd5de3c /]# grep CREATE_HOME /etc/login.defs
CREATE_HOME     yes
[root@cae89bd5de3c /]# useradd --system foo
[root@cae89bd5de3c /]# useradd --system --create-home bar
[root@cae89bd5de3c /]# ls -l /home/
total 0
drwx------. 1 bar bar 64 Jan 30 01:39 bar
```

### DIFF
--- a/man/useradd.8.xml
+++ b/man/useradd.8.xml
@@ -184,13 +184,30 @@
 	    user's login directory. The default is to append the
 	    <replaceable>LOGIN</replaceable> name to
 	    <replaceable>BASE_DIR</replaceable> and use that as the
-	    login directory name.  
+	    login directory name.
 	    The directory <replaceable>HOME_DIR</replaceable> is not created by
-	    default.  However it will be created for non-system users if either the
-	    <option>-m</option> flag is specified or
-	    <replaceable>CREATE_HOME</replaceable> in
-	    <filename>login.defs</filename> is set to true.  However, it will never
-	    be created if the <option>-M</option> flag is specified.
+	    default.
+		Whether it gets created
+		depends on the user type and command-line options.
+	  </para>
+	  <para>
+	    For regular users,
+		the home directory is created
+		if either the <option>-m</option> flag is specified
+		or <replaceable>CREATE_HOME</replaceable>
+		in <citerefentry><refentrytitle>login.defs</refentrytitle><manvolnum>5</manvolnum></citerefentry>.
+		is set to true.
+	  </para>
+	  <para>
+	    For system users,
+		the home directory is created
+		only if the <option>-m</option> flag is explicitly specified,
+	    regardless of the <replaceable>CREATE_HOME</replaceable> setting.
+	  </para>
+	  <para>
+	    The <option>-M</option> flag
+		prevents home directory creation in all cases,
+		overriding any other settings.
 	  </para>
 	</listitem>
       </varlistentry>
@@ -399,7 +416,7 @@
 	</term>
 	<listitem>
 	  <para>
-	    Create the user's home directory if it does not exist. 
+	    Create the user's home directory if it does not exist.
 	    The files and directories contained in the skeleton directory
 	    (which can be defined with the <option>-k</option> option)
 	    will be copied to the home directory.
@@ -489,7 +506,7 @@
             password himself.
 	  </para>
 	  <para>
-	    <emphasis role="bold">Note:</emphasis>Avoid this option on the command  
+	    <emphasis role="bold">Note:</emphasis>Avoid this option on the command
 	     line because the password (or encrypted password) will
 	    be visible by users listing the processes.
 	  </para>


septatrix:
I desired I can add a Signed-Off tag tomorrow, I just did this quickly via the GH web UI as I'm calling it a day right now.

ikerexxe (commented):
While this is an improvement the text is still quite confusing. I'd propose to make a complete rewrite of this section to clarify the beaviour by separating it by user type.

My proposal would be the following:
```
Home directory creation behavior:

The directory HOME_DIR is not created by default. Whether it gets created depends on the user type and command-line options:
- For regular users: the home directory will be created if either the -m flag is specified or if CREATE_HOME is set to yes in /etc/login.defs.
- For system users (created with -r): the home directory will only be created if the -m flag is explicitly specified, regardless of the CREATE_HOME setting.
- Override: the -M flag prevents home directory creation in all cases, overriding any other settings.
```

Feel free to reword or improve anything as it might contain errors.

man/useradd.8.xml
```diff
@@ -186,10 +186,9 @@
 	    <replaceable>BASE_DIR</replaceable> and use that as the
 	    login directory name.  
 	    The directory <replaceable>HOME_DIR</replaceable> is not created by
-	    default.  However it will be created for non-system users if either the
-	    <option>-m</option> flag is specified or
-	    <replaceable>CREATE_HOME</replaceable> in
-	    <filename>login.defs</filename> is set to true.  However, it will never
+	    default. It will only be created if the <option>-m</option> flag
```
alejandro-colomar: The right amount of spaces after period is two.  Please don't break that.

<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/patches/description#n41>
<https://web.archive.org/web/20171217060354/http://www.heracliteanriver.com/?p=324>

alejandro-colomar:
> While this is an improvement the text is still quite confusing. I'd propose to make a complete rewrite of this section to clarify the beaviour by separating it by user type.
> 
> My proposal would be the following:
> 
> ```
> Home directory creation behavior:
> 
> The directory HOME_DIR is not created by default. Whether it gets created depends on the user type and command-line options:

<https://github.com/shadow-maint/shadow/pull/1523#discussion_r2746789249>

I have some wording tweaks for the following suggested text:

> - For regular users: the home directory will be created if either the -m flag is specified or if CREATE_HOME is set to yes in /etc/login.defs.

`s/the home directory/it/`

`s/specified or if/specified, or/`

`s,/etc/login.defs,login.defs(5),`

> - For system users (created with -r): the home directory will only be created if the -m flag is explicitly specified, regardless of the CREATE_HOME setting.

`s/will only be created if/will be created only if/`

`s/(created with -r)/(-r)/`

> - Override: the -M flag prevents home directory creation in all cases, overriding any other settings.

`s/Override: the/The/`

> ```
> 
> Feel free to reword or improve anything as it might contain errors.

alejandro-colomar:
Ping.  Cc: @septatrix

man/useradd.8.xml
```diff
@@ -184,13 +184,26 @@
...
-	    be created if the <option>-M</option> flag is specified.
+	    default.  Whether it gets created depends on the user type and
+	    command-line options.
+	  </para>
+	  <para>
+	    For regular users, the home directory is created if either the
+	    <option>-m</option> flag is specified or <replaceable>CREATE_HOME</replaceable>
+	    in <citerefentry>
+	    <refentrytitle>login.defs</refentrytitle><manvolnum>5</manvolnum>
+	    </citerefentry> is set to true.
```
alejandro-colomar: Please use semantic newlines in new paragraphs.  See man-pages(7):

```
$ MANWIDTH=72 man man-pages | awk '/Use semantic newlines/,/^$/'
   Use semantic newlines
     In the source of a manual page, new sentences should be started on
     new lines, long sentences should be split  into  lines  at  clause
     breaks  (commas,  semicolons, colons, and so on), and long clauses
     should be split at phrase boundaries.  This convention,  sometimes
     known as "semantic newlines", makes it easier to see the effect of
     patches, which often operate at the level of individual sentences,
     clauses, or phrases.
```

septatrix: I assume you mean the long line with `CREATE_HOME` in it?

alejandro-colomar: I'd do this:

```
For regular users,
the home directory is created
if either the <option>-m</option> flag is specified
or <replaceable>CREATE_HOME</replaceable> is set to true in
<citerefentry><refentrytitle>login.defs</refentrytitle><manvolnum>5</manvolnum></citerefentry>.
```

man/useradd.8.xml
```diff
@@ -184,13 +184,26 @@
...
+	    For regular users, the home directory is created if either the
+	    <option>-m</option> flag is specified
+		or <replaceable>CREATE_HOME</replaceable> in <citerefentry>
+	    <refentrytitle>login.defs</refentrytitle><manvolnum>5</manvolnum>
+	    </citerefentry> is set to true.
+	  </para>
+	  <para>
+	    For system users, the home directory is created only if the
+	    <option>-m</option> flag is explicitly specified,
+	    regardless of the <replaceable>CREATE_HOME</replaceable> setting.
```
alejandro-colomar:
```
For system users,
the home directory is created
only if the <option>-m</option> flag is explicitly specified,
regardless of the <replaceable>CREATE_HOME</replaceable> setting.
```

man/useradd.8.xml
```diff
@@ -184,13 +184,26 @@
...
+	  </para>
+	  <para>
+	    The <option>-M</option> flag prevents home directory creation in all
+	    cases, overriding any other settings.
```
alejandro-colomar:
```
The <option>-M</option> flag
prevents home directory creation in all cases,
overriding any other settings.
```

man/useradd.8.xml
```diff
@@ -184,13 +184,26 @@
...
-	    login directory name.  
+	    login directory name.
 	    The directory <replaceable>HOME_DIR</replaceable> is not created by
-	    default.  However it will be created for non-system users if either the
-	    <option>-m</option> flag is specified or
-	    <replaceable>CREATE_HOME</replaceable> in
-	    <filename>login.defs</filename> is set to true.  However, it will never
-	    be created if the <option>-M</option> flag is specified.
+	    default.  Whether it gets created depends on the user type and
+	    command-line options.
```
alejandro-colomar:
```diff
 The directory <replaceable>HOME_DIR</replaceable> is not created by
-default.  However it will be created for non-system users if either the
-<option>-m</option> flag is specified or
-<replaceable>CREATE_HOME</replaceable> in
-<filename>login.defs</filename> is set to true.  However, it will never
-be created if the <option>-M</option> flag is specified.
+default.
+Whether it gets created
+depends on the user type and command-line options.
```

man/useradd.8.xml
septatrix: Aside: I noticed that the other references in this file just say `<filename>/etc/login.defs</filename>`. Should I also use that? It would be a bit shorter but lose the nice linking in some man-page readers. Other places like `-m` do not mention login.defs at all...

alejandro-colomar: I like the manual page reference; let's keep it.

man/useradd.8.xml
```diff
@@ -184,13 +184,30 @@
...
+		Whether it gets created
+		depends on the user type and command-line options.
+	  </para>
+	  <para>
+	    For regular users,
+		the home directory is created
+		if either the <option>-m</option> flag is specified
+		or <replaceable>CREATE_HOME</replaceable>
+		in <citerefentry><refentrytitle>login.defs</refentrytitle><manvolnum>5</manvolnum></citerefentry>.
+		is set to true.
```
alejandro-colomar, line 199: This end of sentence is confused (multiple periods).

alejandro-colomar: I'd move `is set to true` before `in`.

man/useradd.8.xml
```diff
@@ -184,13 +184,30 @@
 	    user's login directory. The default is to append the
```
alejandro-colomar, line 184: In the commit message:

`s/only applies/applies only/`
`s/is does/it does/`

PR #1524, opened 2026-02-01
updated Dutch translation

Frans-Spiesschaert:
updated po/nl.po based upon .pot file from 2025-12-30

### DIFF
--- a/po/nl.po
+++ b/po/nl.po
@@ -1,14 +1,14 @@
 # dutch po-file for shadow
 # Copyright (C) 2004 Free Software Foundation, Inc.
 # Bart Cornelis <cobaco@linux.be>, 2004, 2006.
-# Frans Spiesschaert <Frans.Spiesschaert@yucom.be>, 2014-2025.
+# Frans Spiesschaert <Frans.Spiesschaert@yucom.be>, 2014-2026.
 #
 msgid ""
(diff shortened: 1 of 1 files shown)


po/nl.po
```diff
@@ -1,14 +1,14 @@
 # dutch po-file for shadow
 # Copyright (C) 2004 Free Software Foundation, Inc.
 # Bart Cornelis <cobaco@linux.be>, 2004, 2006.
-# Frans Spiesschaert <Frans.Spiesschaert@yucom.be>, 2014-2025.
+# Frans Spiesschaert <Frans.Spiesschaert@yucom.be>, 2014-2026.
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: shadow 4.19.0\n"
+"Project-Id-Version: shadow 4.19.2\n"
```
alejandro-colomar: The 4.19.x versions are not released from the master branch.  Since this is applied to the master branch, the version should still be 4.19.0.

In fact, 4.19.2 doesn't have these changes, IIRC.

Frans-Spiesschaert: In that case, please change it to 4.19.0. Thank you.

alejandro-colomar: Pushed now.

PR #1528, opened 2026-02-05
Add privileged container and CI support for BTRFS system tests

silverhadch:
This PR introduces opt-in privileged container support to enable filesystem-level
system tests (e.g. BTRFS subvolume homes) while keeping all existing CI and local
testing unprivileged by default.

Key changes:
- Add privileged build containers and execution paths in the Ansible-based CI setup
- Wire privileged execution into the GitHub Actions workflow as a separate CI step
- Extend the Python system test framework with a dedicated privileged topology
- Explicitly exclude privileged tests from default test runs
- Add a privileged useradd system test covering BTRFS subvolume home directories

Safety:
Privileged tests may mount filesystems, create loop devices, and modify kernel-visible
state. They must never be run on a host system. Execution is restricted to disposable
VMs or explicitly created privileged containers.

This keeps destructive tests clearly isolated while allowing coverage for features
that cannot be tested in unprivileged environments.

CC: @Conan-Kudo @ikerexxe 

@ikerexxe wanted a Test Case for this Functionality, we would be relying on in Fedora and KDE Linux.
https://github.com/shadow-maint/shadow/pull/1418#pullrequestreview-3576981595

### DIFF
--- a/.github/workflows/runner.yml
+++ b/.github/workflows/runner.yml
@@ -16,6 +16,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v3
+
     - name: debug
       run: |
         id
@@ -27,17 +28,22 @@ jobs:
         cat /proc/self/uid_map
         cat /proc/self/status
         systemd-detect-virt
+
     - name: Install dependencies
       id: dependencies
       uses: ./.github/actions/install-dependencies
+
     - name: configure
       run: |
         autoreconf -v -f --install
         ./autogen.sh --without-selinux --disable-man --with-yescrypt
+
     - run: make
     - run: make install DESTDIR=${HOME}/rootfs
     - run: sudo make install
+
     - name: run tests in shell with tty
+      # Run tests inside a pseudo-TTY because some parts expect one.
       shell: 'script -q -e -c "bash {0}"'
       run: |
         set -e
@@ -71,10 +77,12 @@ jobs:
 
   container-build:
     runs-on: ubuntu-latest
+    name: "container-build (${{ matrix.os }}${{ matrix.privileged && ', privileged' || '' }})"
     strategy:
       fail-fast: false
       matrix:
         os: [alpine, debian, fedora, opensuse]
+        privileged: [false, true]
 
     steps:
     - uses: actions/setup-python@v5
@@ -89,17 +97,21 @@ jobs:
         sudo apt-get update
         sudo apt-get -y install ansible
 
-    - name: Build container
+    - name: "Build container (${{ matrix.privileged && 'privileged' || 'unprivileged'}})"
       run: |
         pushd share/ansible/
-        ansible-playbook playbook.yml -i inventory.ini -e 'distribution=${{ matrix.os }}'
+        ${{ matrix.privileged && 'sudo' || '' }} ansible-playbook playbook.yml \
+          -i inventory.ini \
+          -e "distribution=${{ matrix.os }}" \
+          -e "privileged_mode=${{ matrix.privileged }}"
         popd
 
     - name: Store artifacts
+      # Always upload logs to allow debugging even on failure.
       if: always()
       uses: actions/upload-artifact@v4
       with:
-        name: ${{ matrix.os }}-build
+        name: "${{ matrix.os }}-${{ matrix.privileged && 'privileged' || 'unprivileged'}}-build"
         path: |
           ./share/ansible/build-out/config.log
           ./share/ansible/build-out/config.h


silverhadch:
The Python Linter is complaining and I want to polish the test a little with comments like in the other tests. A TODO for tomorrow.

tests/system/mhc.yaml
```diff
@@ -9,4 +10,5 @@ domains:
       type: podman
       container: builder
     artifacts:
-    - /var/log/*
\ No newline at end of file
+    - /var/log/*
+
```
alejandro-colomar: What's this exactly?

silverhadch: A added newline since the diff looked off without it as you can see at the removed part.

alejandro-colomar: But you added two instead of one, it seems.

silverhadch: No, its just one newline. Or do you mean at the end of the orginal line? Kate saves files this way as its a good practice as already explained with keeping diffs cleaner.

alejandro-colomar: Yes, the old file ended with `"*"`.  The new file ends with `"*\n\n"`.  It should end with `"*\n"`.

That is, it should terminate in a newline character, but it shouldn't have a blank line.

alejandro-colomar: You can check the file with hd(1).

doc/contributions/tests.md
```diff
@@ -69,6 +74,33 @@ The Python framework provides several advantages over Bash tests:
...
+- **Privileged tests (opt-in)**  
+  These tests require elevated privileges and may perform operations such as:
+  - Mounting filesystems
+  - Creating loop devices
+  - Working with BTRFS subvolumes
+  - Modifying kernel-visible system state
+
+  **Privileged tests are potentially destructive and must never be run on a host
+  system.** Writing new privileged tests is strongly discouraged unless absolutely
+  necessary.
```
ikerexxe: I'd rephrase this. It seems like you don't want new tests there, when in reality we should just prioritize one over the other

ikerexxe (commented):
Thanks for putting this together. I'm really liking the direction of this proposal. I’ve been looking through it and wanted to touch on a few architectural points before we get into the the code review itself.
Regarding the environment, since we’re currently running with full privileges, I’m wondering if we could tighten the security surface by only granting the specific Linux capabilities (like `CAP_SETUID` or `CAP_CHOWN`) actually required for the test? It would be great to avoid full root access if we can help it.
I also think it would be worth decoupling the privileged and unprivileged container runs right from the start; by separating them, we could run them in parallel to keep the CI fast and prevent the unprivileged results from being overwritten by the second run.
Lastly, and most importantly, while the logic here is solid, it adds a fair amount of complexity to the test suite. I’m curious if you’ve identified other areas in the current shadow project that would actually require this kind of infrastructure? I fear this will increase the maintenance burden and having other test cases running in such an environment would justify for this complexity increase.

doc/contributions/tests.md
silverhadch: Thanks for the review. Will apply the suggestions tomorrow.

silverhadch: One clear area this already covers is support for subuids and subgids via newuidmap and newgidmap, which allows testing real parts of that functionality. Some minor parts of the subuid and subgid logic don’t work in a rootful container yet, since they rely on the host’s /etc/subuid and /etc/subgid files, but being able to test parts of this is still far better than not testing it at all.

There’s also filesystem and mount support, even if it’s not entirely clear how much of that logic is needed directly in shadow.

There may be some SELinux or service-related aspects involved as well.

So there is meaningful coverage here. The subuid and subgid handling alone touches a fairly large part of the codebase, even if not everything is mapped out in detail.

On top of that, more tests can be written that require a working /proc, /dev, and kernel interfaces, which is a good justification by itself—especially since these features are already being used heavily by Linux distributions.

For me, the privileged Btrfs subvolume home test alone is worth the maintenance burden of only around 140 lines of more code: KDE Linux and Fedora will rely on this functionality, with KDE Linux depending on it at the RootFS level, so catching regressions here directly protects major and hopefully soon major :) distributions.

And it’s not like the Ansible-related code is changing all the time. Looking at the git history, minor updates seem to happen every three to four months on average.

ikerexxe: I’ve spent some time weighing the long-term maintenance overhead against the value this brings to the ecosystem, and you’ve convinced me. As a project, we need to ensure that `shadow` remains robust for the modern filesystem layouts and user-management workflows that major distributions are actually shipping.

Furthermore, the fact that this infrastructure finally gives us a "real-world" testing ground for subuid/subgid logic, which has historically been difficult to verify in standard unprivileged environments, is a major technical win. I agree that the benefit to the project’s stability outweighs the "complexity tax."

silverhadch:
@ikerexxe 
Split the playbooks from each other.

.github/workflows/runner.yml
```diff
@@ -92,7 +92,8 @@ jobs:
     - name: Build container
       run: |
         pushd share/ansible/
-        ansible-playbook playbook.yml -i inventory.ini -e 'distribution=${{ matrix.os }}'
+        ansible-playbook playbook-unprivileged.yml -i inventory.ini -e 'distribution=${{ matrix.os }}'
+        sudo ansible-playbook playbook-privileged.yml -i inventory.ini -e 'distribution=${{ matrix.os }}'
         popd
```
ikerexxe, line 107: This is still running the tests in CI sequentially, not in parallel. Moreover, the test results are being overwritten due to this

tests/system/tests/privileged/test_privileged_useradd.py
```diff
@@ -0,0 +1,70 @@
...
+            [ -b /dev/loop$i ] || mknod /dev/loop$i b 7 $i
+        done
+
+        mkdir -p /var/lib/shadow-ci
+        rm -f {image_path}
+        truncate -s 128M {image_path}
+        mkfs.btrfs -f {image_path}
+
+        mount -t btrfs -o loop {image_path} /home
+        """)
```
ikerexxe: This should be wrapped inside the framework. The expectation is to call a framework method that is able to run this in the background hiding all the complexities of a script to the test developer

tests/system/tests/privileged/test_privileged_useradd.py
```diff
@@ -0,0 +1,70 @@
...
+        result = shadow.host.conn.run("btrfs subvolume list /home").stdout
+
+        assert "tuser" in result, "/home/tuser should be a Btrfs subvolume"
+
+    finally:
+        shadow.host.conn.run(f"""
+            set +e
+            mountpoint -q /home && umount /home
+            rm -f {image_path}
+            """)
```
ikerexxe: Same here

tests/system/tests/privileged/test_privileged_useradd.py
ikerexxe: In reality, this is more for [pytest-mh](https://github.com/next-actions/pytest-mh), probably [LinuxFileSystem](https://github.com/next-actions/pytest-mh/blob/48c68f8b9196380a7333996105c0b70f417f1c37/pytest_mh/utils/fs.py) or a new class

silverhadch: Well mounting a BTRFS Subvolume as a Home Directory is quite specific for us. So I created a BTRFS class for now and will consult upstream about it if it makes sense to integrate something like this in LinuxFileSystem or anything else. But the logic is now seperated from the test.

silverhadch: But I also refactored the BTRFS Class now so it can be upstreamed later when I consult upstream for this functionality. Right now the Class is pretty basic. It allows you to check if a certain subvolume exists under a certain path, returns a list of subvolumes under a certain path and can create a loop based BTRFS Filesystem of a specified size and mountpoint and then cleans it up.

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,83 @@
...
+            if [ -f {image_path}.loop ]; then
+                LOOP_DEV=$(cat {image_path}.loop)
+                losetup -d "$LOOP_DEV" 2>/dev/null || true
+                rm -f {image_path}.loop
+            fi
+
+            rm -f {image_path}
+        """
+
+        self.host.logger.info("Setting up loop-backed Btrfs {mountpoint}")
```
ikerexxe: Missing `f` prefix

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,83 @@
...
+        self.host.conn.run(setup_script)
+
+        try:
+            yield image_path
+        finally:
+            self.host.logger.info("Cleaning up loop-backed Btrfs {mountpoint}")
```
ikerexxe: Missing `f` prefix

tests/system/tests/privileged/test_privileged_useradd.py
```diff
@@ -0,0 +1,32 @@
...
+        1. Create a user with --btrfs-subvolume-home
+        2. Verify the home directory exists
+        3. Verify the home directory is a Btrfs subvolume
+    :expectedresults:
+        1. The user home directory exists at /home/tuser
+        2. /home/tuser is listed as a Btrfs subvolume
+    :customerscenario: False
+    """
+    with shadow.host.btrfs.loopback_mount("/home"):
+
```
ikerexxe: No need to add an empty line here

tests/system/tests/privileged/test_privileged_useradd.py
```diff
@@ -0,0 +1,32 @@
...
+        shadow.useradd("tuser --btrfs-subvolume-home")
+
+        assert shadow.fs.exists("/home/tuser"), "Home directory should exist"
+
```
ikerexxe, line 34: Same here

share/ansible/roles/build_container_privileged/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Example Playbook
+----------------
+
+Usage example:
+
+    - hosts: localhost
+      become: true
+      roles:
+        - role: build_container
```
ikerexxe: s/build_container/build_container_privileged

doc/contributions/tests.md
```diff
@@ -69,6 +74,44 @@ The Python framework provides several advantages over Bash tests:
...
+
+### Privileged tests (opt-in)
+
+These tests require elevated capabilities and may perform operations
+such as:
+
+-   Mounting filesystems
+-   Creating loop devices
+-   Working with BTRFS subvolumes
+-   Modifying kernel-visible system state
```
ikerexxe, line 95: This can be enhanced by giving concrete examples:
```
- Accessing and modifying any file on the host system through bind mounts
- Interfering with host services by consuming loop devices
- Filling host disk space by creating large filesystem images
- Leaving orphaned mount points that prevent clean system shutdown
```

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,83 @@
+from __future__ import annotations
```
ikerexxe, line 1: @pbrezina do you think this functionality fits in `pytest-mh` or should we keep it in shadow?

tests/system/framework/hosts/base.py
```diff
@@ -43,6 +44,7 @@ def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
 
         self.fs: LinuxFileSystem = LinuxFileSystem(self)
+        self.btrfs = Btrfs(self)
```
ikerexxe: Add type hint: `Btrfs`

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,83 @@
...
+
+import uuid
+from contextlib import contextmanager
+
+
+class Btrfs:
```
ikerexxe: This class should inherit from `MultihostReentrantUtility[MultihostHost]` just like `LinuxFileSystem` and other classes do. That should help with the clean up after the test finishes execution. I'd recommend you to look at these classes to understand their lifecycle and try to mimic their behaviour.

In the end we are looking to restore the clean environment after each test and each class is responsible for managing the "mess" they themselves made.

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self, host):
+        self.host = host
+
+    def subvolumes(self, path: str) -> list[str]:
+        """
+        Returns a list of Btrfs Subvolumes under the specified path.
+        """
+        result = self.host.conn.run(f"btrfs subvolume list {path}").stdout
```
ikerexxe: You should validate the command result (result.rc)

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,83 @@
...
+                LOOP_DEV=$(cat {image_path}.loop)
+                losetup -d "$LOOP_DEV" 2>/dev/null || true
+                rm -f {image_path}.loop
+            fi
+
+            rm -f {image_path}
+        """
+
+        self.host.logger.info(f"Setting up loop-backed Btrfs {mountpoint}")
+        self.host.conn.run(setup_script)
```
ikerexxe: What happens if the script fails?

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,83 @@
...
+        names = []
+        for line in result.strip().splitlines():
+            if " path " in line:
+                names.append(line.split(" path ", 1)[1].strip())
+
+        return names
+
+    def subvolume_exists(self, path: str, name: str) -> bool:
+        """
+        Returns true if the Btrfs Subvolume exists under the specified path.
```
ikerexxe: The under specified path is quite ambiguous. Does it mean only direct children? Or any descendant?

In addition, `names.append(line.split(" path ", 1)[1].strip())` strips the names in a way that might cause some problems. As an example btrfs subvolume list /home returns paths like `home/alice`, but the method searches for just `alice`.

tests/system/framework/hosts/btrfs.py
silverhadch: Now it gets cleaned up properly at all times.

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,191 @@
...
+            yield image_path
+
+        finally:
+            # Always attempt cleanup
+            self.logger.info(f"Cleaning up loop‑backed Btrfs at {mountpoint}")
+            self.host.conn.run(cleanup_script)  # ignore errors
+
+            # Remove from tracking list if it was added
+            if resource in self._resources:
+                self._resources.remove(resource)
```
ikerexxe: This block should be part of the teardown phase

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,191 @@
...
+                LOOP_DEV=$(cat {image_path}.loop)
+                losetup -d "$LOOP_DEV" 2>/dev/null || true
+                rm -f {image_path}.loop
+            fi
+
+            # Remove the image file
+            rm -f {image_path}
+        """
+
+        try:
```
ikerexxe, line 72: No need for the try either

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,191 @@
+from __future__ import annotations
+
+import uuid
+from contextlib import contextmanager
+from typing import Dict, List, Optional
+
+from pytest_mh import MultihostHost, MultihostReentrantUtility
+
+
+class Btrfs(MultihostReentrantUtility[MultihostHost]):
```
ikerexxe: Sorry but I mixed things last time. This should inherit from `MultihostUtility`, not `MultihostReentrantUtility`. The class only needs basic setup-teardown

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,191 @@
...
+        # loop_dev may be None initially, then becomes a string after setup.
+        self._resources: List[Dict[str, Optional[str]]] = []
+
+    def __enter__(self) -> Btrfs:
+        """Enter the runtime context for this utility."""
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        """Exit the runtime context and clean up any leftover resources."""
+        self.teardown()
```
ikerexxe: You can remove these two when inheriting from `MultihostUtility`

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,191 @@
...
+        Args:
+            path: Mount point of the Btrfs filesystem.
+            name: Subvolume name (relative path) to look for.
+
+        Returns:
+            True if the subvolume exists, False otherwise.
+        """
+        return name in self.subvolumes(path)
+
+    @contextmanager
```
ikerexxe, line 36: No need for `@contextmanager`

tests/system/framework/hosts/btrfs.py
```diff
@@ -0,0 +1,191 @@
...
+            cleanup_script = f"""
+                mountpoint -q {mountpoint} && umount {mountpoint}
+                if [ -f {image_path}.loop ]; then
+                    LOOP_DEV=$(cat {image_path}.loop)
+                    losetup -d "$LOOP_DEV" 2>/dev/null || true
+                    rm -f {image_path}.loop
+                fi
+                rm -f {image_path}
+            """
+            self.host.conn.run(cleanup_script)
```
ikerexxe: If this fails, there will be nothing we can do, but at least we should record the failures

ikerexxe (commented):
I have added a few more points for improvement

tests/system/framework/hosts/btrfs.py
silverhadch: Well it is cleaner for the test using them as they are tied to the lifecycle if I am not mistaken. so once you go out of ``with ...btrfs...(/opt):`` the mountpoint will become normal again.

tests/system/framework/hosts/btrfs.py
silverhadch: I added them to teardown but it seems to race against other teardown functions. Unless I call teardown in the finally some other teardown function wants to cleanup the BTRFS Loop Home and fail with device busy. So there is a race condition somewhere I dont understand.

ikerexxe: You can run system tests with additional debugging (`-vvv`) to understand which other teardown function is removing this.

Another option would be to clean this conditionally.

tests/system/framework/hosts/btrfs.py
pbrezina: It looks like a nice feature to integrate with pytest-mh, but let's do it for shadow only first so we do not delay this PR. It can be proposed to pytest-mh later.

tests/system/framework/hosts/btrfs.py
silverhadch: I made it so the resources get cleaned after being used and teardown is a safety net incase anything fell through somehow.

share/ansible/roles/build_container/tasks/main.yml
```diff
@@ -1,20 +0,0 @@
----
-# tasks file for build_container
-- name: Pull container image
-  containers.podman.podman_image:
-    name: '{{ image[distribution] }}'
-
-- name: Create and start container
```
ikerexxe, line 7: This can be `"Create and start {{ 'privileged' if privileged | default(false) else 'unprivileged' }} container"`

share/ansible/roles/build_container/tasks/main.yml
```diff
@@ -1,20 +0,0 @@
...
-  containers.podman.podman_container:
-    name: builder
```
ikerexxe, line 9: And this can be `"{{ container_name | default('builder') }}"`

share/ansible/roles/build_container/tasks/main.yml
```diff
@@ -1,20 +0,0 @@
...
-    state: started
-    image: '{{ image[distribution] }}'
-    command: "sleep 1d"
```
ikerexxe, line 12: Add an additional line after this one with `privileged: "{{ privileged | default(false) }}"`

share/ansible/roles/build_container/tasks/main.yml
```diff
@@ -1,20 +0,0 @@
...
-
-- name: Create repo
-  ansible.builtin.shell:
-    podman exec builder mkdir -p /usr/local/src
```
ikerexxe, line 16: Substitute by `podman exec {{ container_name | default('builder') }} mkdir -p /usr/local/src`

share/ansible/roles/build_container/tasks/main.yml
```diff
@@ -1,20 +0,0 @@
...
-
-- name: Copy repo
-  ansible.builtin.shell:
-    podman cp ../../ builder:/usr/local/src/shadow
```
ikerexxe, line 20: Substitute by `podman cp ../../ {{ container_name | default('builder')}}:/usr/local/src/shadow`

share/ansible/playbook-unprivileged.yml
```diff
@@ -1,14 +1,14 @@
-- name: Start build container
+---
+- name: Start unprivileged build container
```
ikerexxe: This can be `"Start {{ 'privileged' if privileged_mode | default(false) else 'unprivileged' }} build container"`

share/ansible/playbook-unprivileged.yml
```diff
@@ -1,14 +1,14 @@
...
   hosts: localhost
   vars:
     image:
```
ikerexxe: Add below in the same level as this line: `container_name: "{{ 'builder-privileged' if privileged_mode | default(false) else 'builder' }}"`

share/ansible/playbook-unprivileged.yml
```diff
@@ -1,14 +1,14 @@
...
       fedora: registry.fedoraproject.org/fedora:latest
       alpine: docker.io/library/alpine:latest
       debian: docker.io/library/debian:latest
       opensuse: docker.io/opensuse/tumbleweed:latest
-
   roles:
-    - role: build_container
+    - role: build_container_unprivileged
```
ikerexxe: This should be:
```
roles:
      - role: build_container
        vars:
          privileged: "{{ privileged_mode | default(false) }}"
          name: "{{ container_name }}"
```

share/ansible/playbook-unprivileged.yml
```diff
@@ -1,14 +1,14 @@
...
 
 - name: CI run
   hosts: builder
```
ikerexxe: s/builder/"{{ container_name }}"

share/ansible/playbook-unprivileged.yml
```diff
@@ -17,7 +17,11 @@
   roles:
     - role: ci_run
 
-- name: Run system tests
+- name: Run unprivileged system tests
```
ikerexxe: Replace by `"Run {{ 'privileged' if privileged_mode | default(false) else 'unprivileged' }} system tests"`

share/ansible/playbook-unprivileged.yml
```diff
@@ -17,7 +17,11 @@
...
   hosts: localhost
   roles:
     - role: run_system_tests
+      vars:
+        mhc: "mhc.yaml"
```
ikerexxe: Replace with `"{{ 'mhc-privileged.yaml' if privileged_mode | default(false) else 'mhc.yaml' }}"`

share/ansible/playbook-unprivileged.yml
```diff
@@ -17,7 +17,11 @@
...
+        pytest_extra_args: '-m "not privileged"'
```
ikerexxe: Replace with `"{{ '-m \"privileged\"' if privileged_mode | default(false) else '-m \"not privileged\"' }}"`

.github/workflows/runner.yml
```diff
@@ -69,7 +75,54 @@ jobs:
...
+          ./share/ansible/build-out/config.h
+          ./share/ansible/build-out/build.log
+          ./share/ansible/build-out/test-suite.log
+          ./tests/system/pytest.log
+          ./tests/system/artifacts
+        if-no-files-found: ignore
+
+  # Run container builds requiring elevated privileges.
+  # Separated from unprivileged runs to make CI faster to debug and structurally clearer.
+  container-build-privileged:
```
ikerexxe: `container-build:` should be enough here

.github/workflows/runner.yml
ikerexxe: Add another line after this one with `privileged: [false, true]`

.github/workflows/runner.yml
```diff
@@ -89,17 +142,20 @@ jobs:
         sudo apt-get update
         sudo apt-get -y install ansible
 
-    - name: Build container
+    - name: Build container (privileged)
```
ikerexxe: Replace with `"Build container (${{ matrix.privileged && 'privileged' || 'unprivileged'}})"`

.github/workflows/runner.yml
```diff
@@ -89,17 +142,20 @@ jobs:
...
       run: |
         pushd share/ansible/
-        ansible-playbook playbook.yml -i inventory.ini -e 'distribution=${{ matrix.os }}'
+        sudo ansible-playbook playbook-privileged.yml \
+          -i inventory.ini \
+          -e "distribution=${{ matrix.os }}"
```
ikerexxe: This should be:
```
${{ matrix.privileged && 'sudo' || '' }} ansible-playbook playbook.yml \
            -i inventory.ini \
            -e "distribution=${{ matrix.os }}" \
            -e "privileged_mode=${{ matrix.privileged }}"
```

.github/workflows/runner.yml
```diff
@@ -89,17 +142,20 @@ jobs:
...
         popd
 
     - name: Store artifacts
+      # Always upload logs to allow debugging even on failure.
       if: always()
       uses: actions/upload-artifact@v4
       with:
-        name: ${{ matrix.os }}-build
+        name: ${{ matrix.os }}-privileged-build
```
ikerexxe: Replace with `"${{ matrix.os }}-${{ matrix.privileged && 'privileged' || 'unprivileged'}}-build"`

silverhadch:
@ikerexxe 
Has been reduced to net 500~ lines to 330~ ones.
And almost half of it is documentation.

Issue #1530, opened 2026-02-11
[Regression] Build error with gcc 10: error: parameter name omitted


ChenQi1989:
With 9e711e2240c72017dacf088d2931fb26dc6e8fed and 40be785f8a6939b1bfc48faa327ca81a2499efd1, the current 4.19.3 has build error with gcc 10.
For example, on Debian 11 (bullseye), the gcc version is 10.2.1, and it fails to compile shadow.

Some error messages are as below:
../../sources/shadow-4.19.3/lib/copydir.c:103:11: error: parameter name omitted
  103 | error_acl(struct error_context *, const char *fmt, ...)
      |           ^~~~~~~~~~~~~~~~~~~~~~
.........
.........
../../sources/shadow-4.19.3/lib/sssd.h:14:18: error: parameter name omitted
   14 | sssd_flush_cache(int)
      |                  ^~~

PR #1531, opened 2026-02-11
Gcc 10 error

ChenQi1989:
This PR fixes the error for gcc10:
error: parameter name omitted

### DIFF
--- a/lib/copydir.c
+++ b/lib/copydir.c
@@ -91,7 +91,7 @@ static int fchown_if_needed (int fdst, const struct stat *statp,
  */
 format_attr(printf, 2, 3)
 static void
-error_acl(struct error_context *, const char *fmt, ...)
+error_acl(MAYBE_UNUSED struct error_context *_1, const char *fmt, ...)
 {
 	va_list ap;
 
--- a/lib/loginprompt.c
+++ b/lib/loginprompt.c
@@ -27,7 +27,7 @@
 
 
 static void
-login_exit(int)
+login_exit(MAYBE_UNUSED int _1)
 {
 	_exit (EXIT_FAILURE);
 }
--- a/src/gpasswd.c
+++ b/src/gpasswd.c
@@ -109,14 +109,14 @@ static void update_group (struct group *gr);
 static void change_passwd (struct group *gr);
 #endif
 static void log_gpasswd_failure (const char *suffix);
-static void log_gpasswd_failure_system (/*@null@*/void *);
-static void log_gpasswd_failure_group (/*@null@*/void *);
+static void log_gpasswd_failure_system (/*@null@*/MAYBE_UNUSED void *_1);
+static void log_gpasswd_failure_group (/*@null@*/MAYBE_UNUSED void *_1);
 #ifdef SHADOWGRP
-static void log_gpasswd_failure_gshadow (/*@null@*/void *);
+static void log_gpasswd_failure_gshadow (/*@null@*/MAYBE_UNUSED void *_1);
 #endif
 static void log_gpasswd_success (const char *suffix);
-static void log_gpasswd_success_system (/*@null@*/void *);
-static void log_gpasswd_success_group(/*@null@*/void *);
+static void log_gpasswd_success_system (/*@null@*/MAYBE_UNUSED void *_1);
+static void log_gpasswd_success_group (/*@null@*/MAYBE_UNUSED void *_1);
 
 /*
  * usage - display usage message
@@ -467,13 +467,13 @@ static void log_gpasswd_failure (const char *suffix)
 }
 
 static void
-log_gpasswd_failure_system(void *)
+log_gpasswd_failure_system(MAYBE_UNUSED void *_1)
 {
 	log_gpasswd_failure ("");
 }
 
 static void
-log_gpasswd_failure_group(void *)
+log_gpasswd_failure_group(MAYBE_UNUSED void *_1)
 {
 	char  buf[1024];
 
@@ -483,7 +483,7 @@ log_gpasswd_failure_group(void *)
 
 #ifdef SHADOWGRP
 static void
-log_gpasswd_failure_gshadow(void *)
+log_gpasswd_failure_gshadow(MAYBE_UNUSED void *_1)
 {
 	char  buf[1024];
 
@@ -574,13 +574,13 @@ static void log_gpasswd_success (const char *suffix)
 }
 
 static void
-log_gpasswd_success_system(void *)
+log_gpasswd_success_system(MAYBE_UNUSED void *_1)
 {
 	log_gpasswd_success ("");
 }
 
 static void
-log_gpasswd_success_group(void *)
+log_gpasswd_success_group(MAYBE_UNUSED void *_1)
 {
 	char  buf[1024];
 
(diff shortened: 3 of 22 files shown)


lib/copydir.c
```diff
@@ -90,8 +90,7 @@ static int fchown_if_needed (int fdst, const struct stat *statp,
  * error_acl - format the error messages for the ACL and EQ libraries.
  */
 format_attr(printf, 2, 3)
-static void
-error_acl(struct error_context *, const char *fmt, ...)
+static void error_acl (MAYBE_UNUSED struct error_context *ctx, const char *fmt, ...)
```
alejandro-colomar: Would it be possible for you to use a more recent version of GCC?  I've been able to test GCC 11 and it works.

I would like to understand the use case of a bleeding-edge shadow-utils with an old compiler.

I would really like to avoid having that source code.

lib/loginprompt.c
```diff
@@ -26,8 +26,7 @@
 #include "string/strtok/stpsep.h"
 
 
-static void
-login_exit(int)
+static void login_exit (MAYBE_UNUSED int sig)
```
alejandro-colomar: If we really need to fix the regression, I'd like to not revert the commits.  Instead, please call the parameters `_` in a new commit.  When there is more than one, please call them `_`, `__`, `___`, etc.

lib/copydir.c
ChenQi1989: I'm working on Yocto. For its master branch, the minimum host gcc version required is 10.1. See https://git.openembedded.org/openembedded-core/tree/meta/classes-global/sanity.bbclass#n520
So we need to support gcc 10 at the moment. The host that uses gcc 10 is Debian 11 (bullseye). 
We're using host gcc to build shadow-native which runs on the build host.

lib/loginprompt.c
ChenQi1989: Got it. I'll squash these into one patch and use parameters like '_'.

lib/copydir.c
alejandro-colomar: Okay, let's fix the regression.

lib/loginprompt.c
alejandro-colomar: Sorry, after some thought, `_` will collide with the `_()` function from GNU gettext.  Also, `__` is a reserved identifier.

So, we'll need `_1`, `_2`, ...

lib/copydir.c
alejandro-colomar: Oh, Debian Bullseye LTS will die later this year.  This is nice!  We'll be able to revert this fix rather soon.  :)

lib/loginprompt.c
alejandro-colomar: Ping, @ChenQi1989 .

ChenQi1989: @alejandro-colomar I just came back from a vacation. I will work on this ASAP. Thanks for your above _1 suggestion.

ChenQi1989:
@alejandro-colomar I've squashed the codes with extra fixes for the tests/unit directory.
The build test was done in both ubuntu:24.04 (gcc 13) and debian:bullseye (gcc 10) containers with the following commands:

```
$ apt update && apt install --fix-missing gawk wget git diffstat unzip texinfo \
            gcc build-essential chrpath socat cpio python3 autopoint \
            autoconf automake libtool pkg-config  libltdl-dev \
            xsltproc libbsd-dev gettext libcmocka-dev
$ ./autogen.sh
$ ./configure --disable-logind --without-libpam
$ make -j 10
$ make -C tests/unit/ check-TESTS
```

lib/copydir.c
```diff
@@ -90,8 +90,7 @@ static int fchown_if_needed (int fdst, const struct stat *statp,
  * error_acl - format the error messages for the ACL and EQ libraries.
  */
 format_attr(printf, 2, 3)
-static void
-error_acl(struct error_context *, const char *fmt, ...)
+static void error_acl (MAYBE_UNUSED struct error_context *_1, const char *fmt, ...)
```
alejandro-colomar: Please keep the return type in a separate line (and don't add white space).

tests/unit/test_logind.c
```diff
@@ -14,26 +14,27 @@
...
+__wrap_sd_uid_get_sessions(MAYBE_UNUSED uid_t _1, MAYBE_UNUSED int _2, MAYBE_UNUSED char ***_3)
 {
     return mock();
 }
 
 /***********************
  * TEST
  **********************/
-static void test_active_sessions_count_return_ok(void **)
+static void test_active_sessions_count_return_ok(MAYBE_UNUSED MAYBE_UNUSED void ** _1_1)
```
alejandro-colomar: Why _1_1 instead of _1?

alejandro-colomar: Oh, and MAYBE_UNUSED is also dup'd.  It seems some script went wrong.  :)

lib/cleanup_user.c
```diff
@@ -9,6 +9,7 @@
 #include <assert.h>
 #include <stdio.h>
```
alejandro-colomar, line 11: (in the commit message)

> Fixes: https://github.com/shadow-maint/shadow/issues/1530

This should be `Closes`, not `Fixes`.

tests/unit/test_logind.c
alejandro-colomar: BTW, if you use a script, please document it in the commit message (script contents and/or shell session).

ChenQi1989: I was manually changing things. Sorry for the typo. I think I was hitting Ctrl-y twice in emacs. I'll fix all the above issues and update this PR.

ChenQi1989:
@alejandro-colomar I've updated this PR to address the above concerns. The testing method is the same as above.

lib/copydir.c
```diff
@@ -91,7 +91,7 @@ static int fchown_if_needed (int fdst, const struct stat *statp,
  */
 format_attr(printf, 2, 3)
 static void
-error_acl(struct error_context *, const char *fmt, ...)
+error_acl (MAYBE_UNUSED struct error_context *_1, const char *fmt, ...)
```
alejandro-colomar: Please don't introduce white space.

src/gpasswd.c
```diff
@@ -467,13 +467,13 @@ static void log_gpasswd_failure (const char *suffix)
 }
 
 static void
-log_gpasswd_failure_system(void *)
+log_gpasswd_failure_system (MAYBE_UNUSED void *_1)
```
alejandro-colomar: Please don't introduce white space.

ChenQi1989: Got it. I'll remove the whitespace and update this PR.

ChenQi1989:
PR updated with unneeded whitespaces removed. Testing was done in the same method as above.

lib/cleanup_user.c
```diff
@@ -9,6 +9,7 @@
...
 
+#include "attr.h"
```
alejandro-colomar: This file doesn't seem to need this include.  There are no other changes, right?

ChenQi1989: You're right! I guess it gets in when I squashed the patches. I've double checked and there are no other place like this. I'll update this PR to fix it.

ChenQi1989:
PR updated to remove the unneeded "#include". Testing was done in the same method as above.

alejandro-colomar:
I'll backport this into a new stable release this week.

PR #1533, opened 2026-02-11
src/passwd.c: add audit messages for passwd

ikerexxe:
Add comprehensive audit messages for password operations, including unlock, delete, expire and aging operations.

### DIFF
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -34,6 +34,7 @@
 #include "sssd.h"
 #include "string/memset/memzero.h"
 #include "string/sprintf/aprintf.h"
+#include "string/sprintf/snprintf.h"
 #include "string/strcmp/streq.h"
 #include "string/strcmp/strprefix.h"
 #include "string/strcpy/strtcpy.h"
@@ -745,6 +746,17 @@ static void update_shadow(bool process_selinux)
 	if (eflg) {
 		audit_logger(AUDIT_USER_MGMT,
 		             "expired-password",
+		             NULL, pw->pw_uid,
+		             ret ? SHADOW_AUDIT_SUCCESS : SHADOW_AUDIT_FAILURE);
+	}
+	/* Audit aging parameter changes if any were modified */
+	if (xflg || nflg || wflg || iflg) {
+		char aging_msg[256];
+		stprintf_a(aging_msg,
+		           "changed-password-aging min=%ld max=%ld warn=%ld inact=%ld",
+		           nsp->sp_min, nsp->sp_max, nsp->sp_warn, nsp->sp_inact);
+		audit_logger(AUDIT_USER_MGMT,
+		             aging_msg,
 		             NULL, pw->pw_uid,
 		             ret ? SHADOW_AUDIT_SUCCESS : SHADOW_AUDIT_FAILURE);
 	}


src/passwd.c
```diff
@@ -642,17 +664,26 @@ static void update_noshadow(bool process_selinux)
...
 	struct spwd *nsp;
+	int ret;
+
+	pw = pw_locate(name);
+	if (NULL == pw) {
+		(void) fprintf (stderr,
+		                _("%s: user '%s' does not exist in %s\n"),
+		                Prog, name, pw_dbname ());
+		fail_exit (E_NOPERM, process_selinux);
+	}
```
alejandro-colomar, line 679: Could you please clarify in the commit message why this is necessary?  It seems different from the rest of the commit.

ikerexxe: Done. Can you check whether the explanation makes sense to you?

src/passwd.c
```diff
@@ -642,17 +664,26 @@ static void update_noshadow(bool process_selinux)
...
 {
-	const struct passwd pw = { .pw_passwd = SHADOW_PASSWD_STRING };
+	const struct passwd *pw;
 	const struct spwd *sp;
 	struct spwd *nsp;
+	int ret;
+
+	pw = pw_locate(name);
+	if (NULL == pw) {
+		(void) fprintf (stderr,
```
alejandro-colomar: Please don't cast to void, and don't separate the function designator from the `(`.

src/passwd.c
alejandro-colomar: Thanks!

src/passwd.c
```diff
@@ -691,7 +722,45 @@ static void update_shadow(bool process_selinux)
...
+	if (xflg || nflg || wflg || iflg) {
+		char aging_msg[256];
+		snprintf(aging_msg, sizeof(aging_msg),
+		         "changed-password-aging min=%ld max=%ld warn=%ld inact=%ld",
+		         nsp->sp_min, nsp->sp_max, nsp->sp_warn, nsp->sp_inact);
+		audit_logger(AUDIT_USER_MGMT,
+		             aging_msg,
+		             NULL, pw->pw_uid,
+		             ret ? SHADOW_AUDIT_SUCCESS : SHADOW_AUDIT_FAILURE);
+	}
```
alejandro-colomar, line 762: Can we please omit this?  I think it's better to not add any more support for aging information.  The existing support is deprecated, so let it die.

src/passwd.c
ikerexxe: I think I copied this from somewhere else and that's how it ended up like that. It should be fixed now

src/passwd.c
ikerexxe: Unfortunately no. As you have already seen, this feature is still in use by our users, and some of them are required to audit this type of event, so I have no choice but to add it

alejandro-colomar: Could you please keep that code as a downstream extension for now?  Let's discuss it again in the future, in a separate patch.

src/passwd.c
```diff
@@ -691,7 +722,45 @@ static void update_shadow(bool process_selinux)
...
+	if (eflg) {
+		audit_logger(AUDIT_USER_MGMT,
+		             "expired-password",
+		             NULL, pw->pw_uid,
+		             ret ? SHADOW_AUDIT_SUCCESS : SHADOW_AUDIT_FAILURE);
+	}
+	/* Audit aging parameter changes if any were modified */
+	if (xflg || nflg || wflg || iflg) {
+		char aging_msg[256];
+		snprintf(aging_msg, sizeof(aging_msg),
```
alejandro-colomar: snprintf(3) is not allowed.  See the string README.

src/passwd.c
ikerexxe: I’m afraid I have to insist on this. While I understand the intent to move away from password aging, the reality for our users hasn’t caught up to that vision yet.

The decision to deprecate aging is still quite controversial. Many of our users have no choice in the matter; they are legally or contractually obligated to expire passwords and, crucially, to **audit those changes**. As long as the aging functionality remains in the codebase, it must be properly supported by the audit trail to keep those users compliant.

On a more practical note, I am very close to achieving a "zero-downstream-patch" goal for fedora. Adding this as a downstream extension would be a significant step backward in maintainability for us. Since the feature is still present in the upstream source, let’s ensure it’s functionally complete and auditable until the day it’s actually removed.

alejandro-colomar: Okay.  Would you mind at least splitting it to a separate commit?  That will make it easier to revert it once aging features are removed.

src/passwd.c
alejandro-colomar: You should use stprintf_a().

alejandro-colomar: Also, what should happen if the string is truncated?

src/passwd.c
```diff
@@ -755,6 +824,10 @@ main(int argc, char **argv)
 		do_update_age = true;
 	}
 
+#ifdef WITH_AUDIT
+	audit_help_open ();
```
alejandro-colomar: Please remove the space.

src/passwd.c
```diff
@@ -1002,6 +1081,12 @@ main(int argc, char **argv)
 	/* only do this check when getuid()==0 because it's a pre-condition for
 	   changing a password without entering the old one */
 	if (amroot && (check_selinux_permit (Prog) != 0)) {
+#ifdef WITH_AUDIT
+		audit_logger(AUDIT_USER_CHAUTHTOK,
+		             "attempted-to-change-password",
+		             NULL, pw->pw_uid,
+		             SHADOW_AUDIT_FAILURE);
+#endif /* WITH_AUDIT */
```
alejandro-colomar, line 1090: I think we should define audit_logger() to empty if compiled `!WITH_AUDIT`.  That would remove a lot a preprocessor conditionals.

But I'll do that in a separate PR.

src/passwd.c
ikerexxe: Yes, that's done

src/passwd.c
ikerexxe: Done

src/passwd.c
ikerexxe: It's not possible to truncate that string, at least nowadays. There are around 50 characters in the audit message + another 20 characters per `long int` in a 64 bit system, that makes it 50 + 20 * 4 = 130. The maximum buffer size is 256. We still have plenty of room.

In addition, aging parameters are supposed to be low, not big numbers that can never realistically be hit. For those purposes you leave the field empty.

src/passwd.c
ikerexxe: How does the kernel manage these things? I don't remember seeing compilation macros there

src/passwd.c
alejandro-colomar: I would still add a fatal_exit() call if it fails.  Otherwise, it may eventually start truncating once something changes.

alejandro-colomar: Although, since we're going to wipe this code eventually, I think it's fine to keep it like that.  I don't expect we'll change it much, and if it ever breaks, we'll have more reasons to remove it.

So, let's not add error handling.

src/passwd.c
alejandro-colomar: It depends on the case.  They tend to avoid or minimize #ifdef in general.

alejandro-colomar:
LGTM.  I didn't check it thoroughly, as I'm starting to have fever, and my brain isn't ready for that, but feel free to commit.  We could fix it later if it has any issues.  It looks reasonably good.

ikerexxe:
> LGTM. I didn't check it thoroughly, as I'm starting to have fever, and my brain isn't ready for that, but feel free to commit. We could fix it later if it has any issues. It looks reasonably good.

Take care Alejandro! There are a lot of viruses going around these days, so I hope it's nothing serious

PR #1537, opened 2026-02-16
Clarify allowed user and group names

### LINKED ISSUE #1536: Clarify allowed user/group names man page wording
The `useradd(8)` man page states:

Usernames may contain only lower and upper case letters, digits, underscores, or dashes. [...] Fully numeric usernames and usernames . or .. are also disallowed. It is not recommended to use usernames beginning with . character [...]

It looks like dots should be mentioned as allowed characters for user and group names?

myllynen:
Resolves #1536

### DIFF
--- a/man/groupadd.8.xml
+++ b/man/groupadd.8.xml
@@ -65,7 +65,7 @@
     </para>
      <para>
        Groupnames may contain only lower and upper case letters, digits,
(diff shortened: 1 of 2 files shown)


alejandro-colomar (approved):
Thanks!  LGTM.

PR #1538, opened 2026-02-16
configure.ac: fix detection of secure_getenv

floppym:
lib/defines.h was looking for HAVE_SECURE_GETENV instead of of HAS_SECURE_GETENV as defined in configure. This resulted in shadow_getenv always being defined to getenv.

AC_CHECK_FUNC is linker test; it does not check for declarations. Replace this with AC_CHECK_DECLS/HAVE_DECL_SECURE_GETENV.

### DIFF
--- a/lib/defines.h
+++ b/lib/defines.h
@@ -174,7 +174,7 @@
 /* Maximum length of passwd entry */
 #define PASSWD_ENTRY_MAX_LENGTH 32768
 
-#ifdef HAVE_SECURE_GETENV
+#if HAVE_DECL_SECURE_GETENV
 #  define shadow_getenv(name) secure_getenv(name)
 # else
 #  define shadow_getenv(name) getenv(name)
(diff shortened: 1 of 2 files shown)


lib/defines.h
```diff
@@ -174,7 +174,7 @@
 /* Maximum length of passwd entry */
 #define PASSWD_ENTRY_MAX_LENGTH 32768
 
-#ifdef HAVE_SECURE_GETENV
+#if HAVE_DECL_SECURE_GETENV
```
alejandro-colomar, line 177: Should we backport this to stable?

alejandro-colomar, line 177: Please add a Fixes tag in the commit message.

Here's some guidelines for that:

<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/git#n46>

floppym: It has potential security implications I suppose, so a stable backport seems reasonable.

floppym: Done!

floppym: On the other hand, this would be a change in behavior, so it might be better to wait for a mainline release.

alejandro-colomar: Okay; sounds reasonable.

alejandro-colomar: Anyway, we'll be releasing in a month.  :)

PR #1542, opened 2026-02-17
Clean up unused code in configure

floppym:
See commit messages for details.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -581,7 +581,7 @@ if test "X$with_skey" = "Xyes"; then
 	]])],[AC_DEFINE([SKEY_BSD_STYLE], [1], [Define to support newer BSD S/Key API])],[])
 fi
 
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -3,7 +3,6 @@ AC_PREREQ([2.69])
 m4_define([libsubid_abi_major], [5])
 m4_define([libsubid_abi_minor], [0])
 m4_define([libsubid_abi_micro], [0])
-m4_define([libsubid_abi], [libsubid_abi_major.libsubid_abi_minor.libsubid_abi_micro])
```
alejandro-colomar, line 6: Would this break any users of libsubid?  I don't know how that library works or who uses it.

floppym: I don't think so: libsubid/subid.h.in references the 3 individual version components, but does not reference the combined LIBSUBID_ABI.

alejandro-colomar: Okay, let's merge.

PR #1545, opened 2026-02-18
src/useradd.c: Refactor unreadable conditionals.

alejandro-colomar:
Co-authored-by: @hallyn 
Cc: @acdn-ndostert

---

Revisions:

<details>
<summary>v2</summary>

-  Signed-off-by: @hallyn 
-  tfix [@hallyn ]

```
$ git rd 
1:  1427d2df087b ! 1:  4fa43ef9f01e src/useradd.c: Factor out logic to helper functions
    @@ Commit message
         src/useradd.c: Factor out logic to helper functions
     
         Co-authored-by: Serge Hallyn <serge@hallyn.com>
    +    Signed-off-by: Serge Hallyn <serge@hallyn.com>
         Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
      ## src/useradd.c ##
    @@ src/useradd.c: int main (int argc, char **argv)
     -      (!rflg || Fflg) &&
     -      (!user_id || (user_id <= uid_max && user_id >= uid_min));
     +  is_sub_uid = should_assign_subuid();
    -+  is_sub_gid = should_assign_subuid();
    ++  is_sub_gid = should_assign_subgid();
      #endif                            /* ENABLE_SUBIDS */
      
        if (run_parts ("/etc/shadow-maint/useradd-pre.d", user_name,
2:  9d6b44cf0029 ! 2:  1a1cabf8a463 src/useradd.c: Refactor long expression into many simple conditionals
    @@ Commit message
         src/useradd.c: Refactor long expression into many simple conditionals
     
         Co-authored-by: Serge Hallyn <serge@hallyn.com>
    +    Signed-off-by: Serge Hallyn <serge@hallyn.com>
         Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
      ## src/useradd.c ##
```
</details>

PR #1511, opened 2026-01-17
useradd: Correctly set subuid/subgid when using -F

acdn-ndostert:
Fixes https://github.com/shadow-maint/shadow/issues/1255

I was not able to run the tests as described in doc/contributions/tests.md at the moment but I was able to build and test that `useradd -F` work as intended.

PR #1548, opened 2026-02-23
useradd: Correctly set subuid/subgid when using -F

acdn-ndostert:
The -F flag should bypass the -r flag and UID checks.

Closes: <shadow-maint#1255>

This is a followup of [#1511](https://github.com/shadow-maint/shadow/pull/1511) and https://github.com/shadow-maint/shadow/pull/1545

In the current implementation the case `useradd -r -F alice` was not working anymore, this is related to the change from `!user_id` to `if (user_id == 0)` while the user_id is not set at this moment.
As a side effect this PR restore the old behavior and `useradd -r -F alice` again add entries to `/etc/subuid` and `/etc/subgid`.

Cc: @alejandro-colomar 
Cc: @hallyn

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2463,7 +2463,9 @@ should_assign_subuid(void)
 		return false;
 	if (!sub_uid_file_present())
 		return false;
-	if (rflg && !Fflg)
+	if (Fflg)
+		return true;
+	if (rflg)
 		return false;
 	if (user_id == 0)
 		return false;
@@ -2491,7 +2493,9 @@ should_assign_subgid(void)
 		return false;
 	if (!sub_gid_file_present())
 		return false;
-	if (rflg && !Fflg)
+	if (Fflg)
+		return true;
+	if (rflg)
 		return false;
 	if (user_id == 0)
 		return false;


alejandro-colomar:
If this fixes something, the commit message should include a Fixes tag.

See
and
<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/git#n46>.

src/useradd.c
```diff
@@ -2463,12 +2463,14 @@ should_assign_subuid(void)
...
-	if (user_id > uid_max || user_id < uid_min)
-		return false;
+	if (!Fflg) {
+		if (rflg)
+			return false;
+		if (user_id == 0)
+			return false;
+		if (user_id > uid_max || user_id < uid_min)
+			return false;
+	}
```
alejandro-colomar: Please don't indent stuff.

You could just add

```c
if (Fflg)
	return true;
```

acdn-ndostert: I'm trusting you on this :)
I added you as a co-author if thats how it work ?

alejandro-colomar: If you want, that's fine.  :)

Although that's just a small style change, so you don't need to add it.

Also, I don't understand this code very much.  (I don't know much about subids, to be honest..)

hallyn (approved):
Thanks! Lgtm

PR #1552, opened 2026-02-25
doc/contributions/coding_style.md: Refer to the man-pages style guide

alejandro-colomar:
This adds a link to the style I've been using in new code.

Reported-by: @richardweinberger
Reported-by: @Karlson2k

### DIFF
--- a/doc/contributions/coding_style.md
+++ b/doc/contributions/coding_style.md
@@ -6,7 +6,10 @@ Each language follows its own established conventions and style guidelines.
 
 ## C code
 
-* For a general guidance refer to the
+* For a general guidance refer to this
+[style guide](https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c).
+
+* For stuff not covered there, follow the
 [Linux kernel coding style](https://www.kernel.org/doc/html/latest/process/coding-style.html)
 
 * Patches that change the existing coding style are not welcome, as they make


alejandro-colomar:
Here's the full text of the man-pages style guide, as of now:

<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c>
<details>
<summary>style guide</summary>

```
$ cat CONTRIBUTING.d/style/c | expand;
Name
       style/c - C coding style

Description
    Indentation
        Use tabs for indentation.

                if (foo)
                        bar();

        Indent preprocessor directives after the hash by 1 space.

                #ifndef  FOO
                # define FOO
                #endif

        'case' is not indented within a 'switch'.

                switch (x) {
                case 1:
                        foo();
                        break;
                default:
                        break;
                }

    Alignment
        However, alignment is done with spaces.

        When breaking a function prototype, start the continuation line
        with 4 spaces.

                void foo(const char *restrict x,
                    const char *restrict y);

        When breaking a function call, align at the opening parenthesis.

                foobar(qwertyuiopasdfghjklzxcvbnm, qwertyuiop,
                       asdfghjkl);

    Line length
        Lines should not be longer than 80 columns.  Except that if they
        contain string literals, they can be longer; don't break
        user-visible string literals.

    Braces
        its own.

                if (foo)
                        bar();

                {
                        baz();
                }

    Spaces
        Treat sizeof() and similar operators as functions, not keywords.
        Use a space after keywords, but not after functions.

        Use a space to separate binary and ternary operators (except
        `.` and `->`), but not to separate unary operators.

        Use a space between a cast and the expression it converts.

    Naming
        Use short names.  Long names should be an exception, and
        indicate that something probably isn't well designed.

    Functions
        Functions should be short and sweet.

        All functions should have prototypes.

    Macros
        Don't be worried about using macros.  They can and do improve
        safety, if used judiciously.

    Error handling
        goto is good for error handling.  It's certainly better than the
        alternatives most of the time.

        Check for explicit error codes (connect(sfd, &sa, len) == -1)
        instead of vague comparisons (connect(sfd, &sa, len) < 0).

    Includes
        Follow include-what-you-use guidelines.

    Comments
        Comments lie; don't write comments.  If you need to comment
        code, do it in the commit message.  If that's not enough, maybe
        the code isn't good.

        In most cases, a function with an appropriate name is better
        than a comment.  A function is also better than a named loop.

    Variables
        Variable should be declared at the top of the block in which
        they are used.  That is, use C89 declarations.  The exception is
        loop variables; we use C99 for-loop variable declarations.

        The '*' goes with the variable name, not with the type name.
        Except if the pointer is qualified, in which case the '*' goes
        with the type name.

                int     i;
                char    c;
                char    *p;
                size_t  size;

    Dialect
        We use the latest GNU C dialect.  Feel free to use new language
        features, unless they are evil.

See also
        For anything not explicitly covered above, you can check the
        following coding styles, roughly in order of appearance:

        <https://include-what-you-use.org/>
        <https://doc.cat-v.org/bell_labs/pikestyle>
        <https://www.kernel.org/doc/html/latest/process/coding-style.html>
        <https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/CodingGuidelines>
        <https://mbsd.evolvis.org/htman/i386/man9/style.htm>
        <https://nginx.org/en/docs/dev/development_guide.html#code_style>
        <https://google.github.io/styleguide/>
        <https://www.gnu.org/prep/standards/standards.html>
        <https://www.cis.upenn.edu/~lee/06cse480/data/cstyle.ms.pdf>
```
</details>

hallyn (approved):

Issue #1018, opened 2024-06-06
libsubid: Memory Management in Different Heap States


daskol:
As in release shadow 4.15.1, libsubid assumes that the same allocator (presumably glibc) is used across caller and callee code. The issue is getting worse in a situation when libsubid forward subordinate management to a plugin according to nsswitch.conf. In this case, memory management gets more complicated: one must be sure that a plugin and a caller use the same allocator which shares the same heap state. The issue is that a caller can use glibc malloc while a plugin uses jemalloc/tcmalloc/etc or none of them at all if  a plugin is not written in C/C++.

I propose two ways to tackle the issue. In the first solution, libsubid public API should be extended with subid_free routine to release memory allocated in a plugin. The advantage of this solution is its simplicity and it requires only small changes (writing new code, mostly).

```c
struct subid_range *ranges;
int count = subid_get_uid_ranges("username", &ranges);
// do something with ranges.
subid_free(ranges);
```

The second approach is wide-spread as well. It requires API call to estimate buffer size first and allocating a buffer on a caller side.

```cpp
int count = subid_get_uid_ranges_count("username");
if (count < 0) {
    throw std::runtime_exception(...);
}
auto ranges = std::make_unique<subid_range[]>(count);
subid_get_uid_ranges("username", ranges.get());
// do something with ranges.
```

What do you think?

daskol:
@hallyn @alejandro-colomar Any comments?

alejandro-colomar:
> I propose two ways to tackle the issue. In the first solution, libsubid public API should be extended with subid_free routine to release memory allocated in a plugin. The advantage of this solution is its simplicity and it requires only small changes (writing new code, mostly).
> 
> ```c
> struct subid_range *ranges;
> int count = subid_get_uid_ranges("username", &ranges);
> // do something with ranges.
> subid_free(ranges);
> ```
> 
> The second approach is wide-spread as well. It requires API call to estimate buffer size first and allocating a buffer on a caller side.
> 
> ```c++
> int count = subid_get_uid_ranges_count("username");
> if (count < 0) {
>     throw std::runtime_exception(...);
> }
> auto ranges = std::make_unique<subid_range[]>(count);
> subid_get_uid_ranges("username", ranges.get());
> // do something with ranges.
> ```
> 
> What do you think?

Both sound good to me.

daskol:
Great. I will prepare pull request then.

alejandro-colomar:
[...]

PR #1562, opened 2026-03-02
subid: Support multiple subid NSS modules

YiPrograms:
Currently, only a single provider is supported for subid delegation. If subid is managed by `sss`, local users are unable to have subids. This introduces significant limitations, particularly for the local root user, which should not be managed by LDAP. Because of this, the root user cannot allocate subids for rootless containers via `newuidmap`. This impacts LXC, which fails to start unprivileged containers if subid is managed by `sss`.
This limitation is also discussed in a Red Hat page regarding podman with local users: https://access.redhat.com/solutions/6961540

This PR addresses this limitation by introducing support for configuring and querying multiple subid delegation modules sequentially, allowing both local files and network providers.

The implementation follows the following strict resolution processes:
- The NSS initialization in `lib/nss.c` has been updated to tokenize the nsswitch.conf configuration line. It dynamically loads and appends resolved modules sequentially into a linked list structure called `subid_nss_db` to track multiple subid backends.
- When querying subid ranges via functions like `list_owner_ranges()`, `have_sub_uids()`, and `have_sub_gids()`, it consults the NSS providers in the exact order specified. The search stops immediately as soon as a module indicates the user exists in their db by returning `SUBID_STATUS_SUCCESS` (rather than `SUBID_STATUS_UNKNOWN_USER`). This short-circuiting occurs even if the successful module returns no subid ranges for that user.
- If a provider error occurs at any point during the query sequence, the query returns an error immediately. This prevents getting an entirely different subid ranges from subsequent fallback providers just because a primary provider experienced downtime.
- The `find_subid_owners()` function queries all NSS providers, returning an aggregated list of all users who may have the target subid.
- Local file mutations, such as `sub_uid_add()`, is only allowed if there is no NSS delegation, or if `files` is listed as the very first configured provider in the sequence.

`list_owner_ranges()` and `find_subid_owners()` now copy the data returned from an NSS module into locally allocated memory. The module's allocated memory is immediately freed using its specific free handler. As such, `free_subid_pointer()` no longer needs to manage module-specific freeing and can safely call free() on all publicly returned pointers. (Related issue #1018)

The test suites in `tests/libsubid/04_nss` has been updated to validate the new multi-provider resolution logic.
Manpages in `man/subuid.5` and `man/subgid.5` has also been updated to document the new support for multiple subid delegation sources.

Although this PR introduces a substantial amount of code changes to support multi-module scenarios, it is designed with strict backward compatibility in mind. If an environment only has a single provider configured, the behavior remains completely unchanged.

This is my first contribution to this project. Please let me know if there are any coding style issues, commit format errors, or other adjustments you'd like me to make. I am more than happy to update them.

Thank you for your time and for reviewing this PR!


lib/nss.c
```diff
@@ -86,66 +155,62 @@ nss_init(const char *nsswitch_path) {
...
+		// Use NULL to indicate the built-in "files" database
+		subid_nss_db_head = NULL;
+		goto done;
 	}
-	subid_nss->free = dlsym(h, "shadow_subid_free");
-	if (!subid_nss->free) {
-		fprintf(log_get_logfd(), "%s did not provide @subid_free@\n", libname);
-		goto close_lib;
+
+	token = strtok_r(p, delimiters, &saveptr);
```
alejandro-colomar: strtok(3) is not allowed.  Please use stpsep() or strsep(3).

See <https://github.com/shadow-maint/shadow/pull/1555>.

lib/nss.c
```diff
@@ -86,66 +155,62 @@ nss_init(const char *nsswitch_path) {
...
+		if (streq(token, "files")) {
+			// Use NULL to indicate the built-in "files" database
+			ops = NULL;
+		} else {
+			if (strlen(token) > 50) {
+				fprintf(log_get_logfd(), "Subid NSS module name too long: %s\n", token);
+				goto skip;
+			}
+
+			snprintf(libname, sizeof(libname), "libsubid_%s.so", token);
```
alejandro-colomar: snprintf(3) is not allowed.  Please use stprintf() or something else.

See lib/string/README.

lib/prototypes.h
```diff
@@ -228,6 +228,11 @@ extern /*@null@*//*@only@*/struct passwd *get_my_pwent (void);
 extern void nss_init(const char *nsswitch_path);
 extern bool nss_is_initialized(void);
 
+struct subid_nss_db {
+	struct subid_nss_ops* ops;
+	struct subid_nss_db* next;
```
alejandro-colomar: Please put the `*` together with the member, not the type.

lib/nss.c
```diff
@@ -86,66 +155,62 @@ nss_init(const char *nsswitch_path) {
...
-		goto close_lib;
+		// Use NULL to indicate the built-in "files" database
+		subid_nss_db_head = NULL;
+		goto done;
 	}
-	subid_nss->free = dlsym(h, "shadow_subid_free");
-	if (!subid_nss->free) {
-		fprintf(log_get_logfd(), "%s did not provide @subid_free@\n", libname);
-		goto close_lib;
+
```
alejandro-colomar, line 176: This patch is huge, which makes it difficult to review.  Could you please split it into several smaller ones?  I expect some parts could be split to refactoring patches before the main one.

YiPrograms: Thanks for the review! I will work on moving some of the preparatory changes to several smaller patches.

lib/nss.c
```diff
@@ -48,6 +48,53 @@ static void nss_exit(void) {
 	}
 }
 
+static struct subid_nss_ops* open_and_check_nss_module(const char *libname) {
```
alejandro-colomar: Please put the return type in a separate line, and the brace in another separate line:

```c
static struct foo *
func(...)
{
```

lib/nss.c
```diff
@@ -48,6 +48,53 @@ static void nss_exit(void) {
...
+	void *h;
+	struct subid_nss_ops *nss_ops;
```
alejandro-colomar: Please align variables as

```c
void                  *h
struct subid_nss_ops  *nss_ops;
```

(Note two spaces at the minimum distance point.)

lib/nss.c
```diff
@@ -108,42 +155,13 @@ nss_init(const char *nsswitch_path) {
 		goto null_subid;
 	}
 	stprintf_a(libname, "libsubid_%s.so", p);
-	h = dlopen(libname, RTLD_LAZY);
-	if (!h) {
-		fprintf(log_get_logfd(), "Error opening %s: %s\n", libname, dlerror());
-		fprintf(log_get_logfd(), "Using files\n");
```
alejandro-colomar, line 114: What happened to this error line?  It's not mentioned in the commit message.  If you want to remove diagnostic lines (which I'm okay with), that should happen on a separate commit, with justification.

lib/nss.c
```diff
@@ -178,3 +179,9 @@ struct subid_nss_ops *get_subid_nss_handle() {
 	nss_init(NULL);
 	return subid_nss;
 }
+
+struct subid_nss_db *get_subid_nss_db() {
```
alejandro-colomar: Please use `(void)` instead of `()`.

lib/nss.c
```diff
@@ -134,37 +138,61 @@ nss_init(const char *nsswitch_path) {
...
 	}
+
 	if (p == NULL) {
-		goto null_subid;
+		// Use NULL to indicate the built-in "files" database
+		subid_nss_db_head = NULL;
+		goto done;
+	}
+
+	while ((token = strsep(&p, delimiters)) != NULL) {
```
alejandro-colomar: Please put the NULL before the assignment.  That way, it's more robust against accidents, such as `=` instead of `!=`

```c
while (NULL != (token = strsep(...))) {
```

lib/nss.c
alejandro-colomar: Here's the coding style:

<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c>

lib/nss.c
alejandro-colomar: Why should strsep(3) separate on `\n`?  This doesn't seem right.

alejandro-colomar: Oh, actually, the old code already used that.  Still weird, though.

lib/nss.c
```diff
@@ -99,10 +99,14 @@ static struct subid_nss_ops* open_and_check_nss_module(const char *libname) {
 // nsswitch_path is an argument only to support testing.
 void
 nss_init(const char *nsswitch_path) {
+	struct subid_nss_db **tail = &subid_nss_db_head;
+	struct subid_nss_ops *ops;
+	struct subid_nss_db *new_db;
+	const char *delimiters = " \t\n";
```
alejandro-colomar: Is the format of subid: lines with multiple databases in /etc/nsswitch.conf defined anywhere?

Or are we inventing this format?

lib/nss.c
```diff
@@ -134,37 +138,61 @@ nss_init(const char *nsswitch_path) {
...
+	while ((token = strsep(&p, delimiters)) != NULL) {
+		if (*token == '\0') {
+      continue;
+    }
+
+		if (streq(token, "files")) {
+			// Use NULL to indicate the built-in "files" database
+			ops = NULL;
+		} else {
+			if (stprintf_a(libname, "libsubid_%s.so", token) < 0) {
```
alejandro-colomar: The error code should be explicitly checked `== -1`.

lib/nss.c
```diff
@@ -134,37 +138,61 @@ nss_init(const char *nsswitch_path) {
...
+        fprintf(log_get_logfd(), "Subid NSS module name too long: %s\n", token);
```
alejandro-colomar: This line is using spaces instead of tabs.

lib/nss.c
YiPrograms: The `getline()` above gets only the `subid: ` line, but includes the trailing \n.
I believe we are passing \n as a delimiter to prevent strsep() from returning a token with the trialing newline.

lib/subordinateio.c
```diff
@@ -626,17 +626,53 @@ bool local_sub_uid_assigned(const char *owner)
 
 bool have_sub_uids(const char *owner, uid_t start, unsigned long count)
```
YiPrograms, line 627: It seems that `lib/subordinateio.c` follows a different coding style as the style guide.

Should I make a patch that reformats the whole file and rebase other patches on top of it? Or should I fix the formatting in a follow-up patch or PR?

alejandro-colomar: > It seems that `lib/subordinateio.c` follows a different coding style as the style guide.

Yeah, some old code is really ugly.

> 
> Should I make a patch that reformats the whole file and rebase other patches on top of it? Or should I fix the formatting in a follow-up patch or PR?

As you wish.

YiPrograms: Ah, I just noticed that the coding style guide states: "Patches that change the existing coding style are not welcome."
I think it's safest if I just ensure the lines and functions I'm modifying adhere to the style guide, and leave the rest of the file untouched.
Let me know if you'd prefer otherwise!

YiPrograms: By the way, should we consider introducing clang-format to the project? It would make formatting automatic and consistent for everyone.
To avoid messing with the existing codebase, we could use `git clang-format` to ensure only touched lines are formatted on new contributions.

alejandro-colomar: > Ah, I just noticed that the coding style guide states: "Patches that change the existing coding style are not welcome." I think it's safest if I just ensure the lines and functions I'm modifying adhere to the style guide, and leave the rest of the file untouched. Let me know if you'd prefer otherwise!

They're not welcome by themselves, but if you send them together with other changes, they could be more interesting.

But not changing the existing style is also a valid option.  Do either of them, as you wish.

What we do not welcome is someone opening many PRs just to fix style.

alejandro-colomar: > By the way, should we consider introducing clang-format to the project? It would make formatting automatic and consistent for everyone. To avoid messing with the existing codebase, we could use `git clang-format` to ensure only touched lines are formatted on new contributions.

Maybe eventually.

YiPrograms:
Blocked by #1563, #1573 and #1586.
I will rebase once those patches are landed.

lib/nss.c
alejandro-colomar: We should then remove the `\n` before strsep(3) (with stpsep()), so that the delimiters are `" \t"` as usual.

alejandro-colomar: We've merged the related fixes to the handling of lines (#1563 ).  Please rebase.

alejandro-colomar:
> Blocked by #1563, #1573 and #1586. I will rebase once those patches are landed.

Those three have been merged.  This is thus unblocked.

lib/nss.c
```diff
@@ -108,42 +157,13 @@ nss_init(const char *nsswitch_path) {
 		goto null_subid;
 	}
 	stprintf_a(libname, "libsubid_%s.so", p);
-	h = dlopen(libname, RTLD_LAZY);
-	if (!h) {
```
alejandro-colomar, line 112: `h` is now unused in this function.  You should remove the declaration.

lib/prototypes.h
```diff
@@ -228,6 +228,11 @@ extern /*@null@*//*@only@*/struct passwd *get_my_pwent (void);
 extern void nss_init(const char *nsswitch_path);
 extern bool nss_is_initialized(void);
 
+struct subid_nss_db {
+	struct subid_nss_ops *ops;
+	struct subid_nss_db *next;
+};
+
```
alejandro-colomar, line 235: Do we really need a linked list?  Or is an array (maybe `NULL`-terminated) simpler and enough?

lib/nss.c
```diff
@@ -25,18 +25,27 @@
...
+//    subid: sss files
+// then databases will be consulted in the specified order. The search
+// stops as soon as the user is found in a database, even if no subids
+// are defined there. For example, if 'sss' knows the user but provides
+// no subids, 'files' will not be consulted.
+//
+// While multiple databases are now supported, the subids are a pretty
+// limited resource. Mixing local files with network allocations
+// (like sssd) requires careful management. Misconfigurations would
+// lead to overlapping ID mappings. Use with caution.
```
alejandro-colomar, line 40: Why did you change from two spaces of inter-sentence space to just one???

lib/subordinateio.c
```diff
@@ -1400,12 +1400,7 @@ bool release_subid_range(struct subordinate_range *range, enum subid_type id_typ
...
 void free_subid_pointer(void *ptr)
 {
-	struct subid_nss_ops *h = get_subid_nss_handle();
-	if (h) {
-		h->free(ptr);
-	} else {
-		free(ptr);
-	}
+	free(ptr);
 }
```
alejandro-colomar, line 1404: How is this related at all to supporting multiple NSS modules?

If you want to refactor, which is fine (and even welcome, if it simplifies), you should do that in a separate PR.  Otherwise, it's impossible to review this.

tests/libsubid/04_nss/passwd
```diff
@@ -0,0 +1,24 @@
...
+list:x:38:38:Mailing List Manager:/var/list:/bin/sh
+irc:x:39:39:ircd:/var/run/ircd:/bin/sh
+gnats:x:41:41:Gnats Bug-Reporting System (admin):/var/lib/gnats:/bin/sh
+nobody:x:65534:65534:nobody:/nonexistent:/bin/sh
+Debian-exim:x:102:102::/var/spool/exim4:/bin/false
+ubuntu:x:1000:1000:Ubuntu:/home/ubuntu:/bin/bash
+user1:x:1001:1001::/home/user1:/bin/bash
+user2:x:1002:1002::/home/user2:/bin/bash
+error:x:1003:1003::/home/error:/bin/bash
+conn:x:1004:1004::/home/conn:/bin/bash
```
Skyb0rg007, line 24: This file doesn't have a trailing newline, which seems wrong

Skyb0rg007 (commented):
This PR would be very useful to me, as it would let me configure immutable /etc/sub{u,g}id files for system accounts and use a custom module for mutable (ex. systemd-homed) accounts.

alejandro-colomar:
> This PR would be very useful to me, as it would let me configure immutable /etc/sub{u,g}id files for system accounts and use a custom module for mutable (ex. systemd-homed) accounts.

The author hasn't worked on this PR for a long time, so if you want to, I guess you can take over (you may need to open a new PR).

YiPrograms:
Sorry, I’m quite busy at the moment and don’t have much time to work on this PR recently.

It may be worth getting #1679 finished first, since `subordinateio.c` currently has quite a bit of old and inconsistent style, which makes changes in this area harder to read, review, and maintain.

Once #1679 is finished, I think it would make sense to rebase/rework this PR on top of that cleanup. I’d be happy for someone else to pick this up and open a new PR if needed, or I may also be able to come back and rework it myself then.

lib/subordinateio.c
YiPrograms: This logic has landed in #1740, so we can drop this when reworking the PR.

PR #1565, opened 2026-03-07
lib/commonio.c: Use block I/O for backup copies

cuiweixie:
Replace byte-by-byte backup copying in create_backup() with fread/fwrite buffering. This removes the long-standing TODO and keeps explicit cleanup paths for seek/read/write failures before flushing and syncing.

### DIFF
--- a/lib/commonio.c
+++ b/lib/commonio.c
@@ -247,7 +247,6 @@ static int create_backup (const char *name, FILE * fp)
 	struct stat sb;
 	struct utimbuf ub;
 	FILE *bkfp;
-	int c;
 
 	stprintf_a(tmpf, "%s.cioXXXXXX", name);
 	if (fstat (fileno (fp), &sb) != 0) {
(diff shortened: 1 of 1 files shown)


lib/commonio.c
```diff
@@ -259,16 +258,29 @@ static int create_backup (const char *name, FILE * fp)
...
 			}
 		}
+		if (ferror (fp) != 0) {
+			(void) fclose (bkfp);
+			unlink(tmpf);
+			return -1;
+		}
 	}
-	if ((c != EOF) || (ferror (fp) != 0) || (fflush (bkfp) != 0)) {
+	if (fflush (bkfp) != 0) {
```
alejandro-colomar, line 283: Did you notice any performance issues or issues of other kind?  Otherwise, I see this as a readability pessimization, which has a high chance of having bugs in there.

alejandro-colomar: You may be interested in this instead:
-  <https://github.com/shadow-maint/shadow/pull/1499>

PR #1569, opened 2026-03-09
Usermod lock what you need

jcpunk:
This PR has a few related bits:
* Only lock "`/etc/shadow`" if we're actually changing something relevant to it
* The unlock steps where not the reverse of the lock steps. The order probably doesn't matter but if the order of the locking is important, then unlock should rewind the the right sequence.
* The unlock steps at the successful branch are now conditional based on if things were actually locked.

Closes: https://github.com/shadow-maint/shadow/issues/1540

I have no idea how to test this...

### DIFF
--- a/src/usermod.c
+++ b/src/usermod.c
@@ -660,13 +660,22 @@ NORETURN
 static void
 fail_exit (int code, bool process_selinux)
 {
-	if (gr_locked) {
-		if (gr_unlock (process_selinux) == 0) {
-			fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, gr_dbname ());
-			SYSLOG(LOG_ERR, "failed to unlock %s", gr_dbname());
+#ifdef ENABLE_SUBIDS
+	if (sub_gid_locked) {
+		if (sub_gid_unlock (process_selinux) == 0) {
+			fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, sub_gid_dbname ());
+			SYSLOG(LOG_ERR, "failed to unlock %s", sub_gid_dbname());
 			/* continue */
 		}
 	}
+	if (sub_uid_locked) {
+		if (sub_uid_unlock (process_selinux) == 0) {
+			fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, sub_uid_dbname ());
+			SYSLOG(LOG_ERR, "failed to unlock %s", sub_uid_dbname());
+			/* continue */
+		}
+	}
+#endif				/* ENABLE_SUBIDS */
 #ifdef	SHADOWGRP
 	if (sgr_locked) {
 		if (sgr_unlock (process_selinux) == 0) {
@@ -676,6 +685,13 @@ fail_exit (int code, bool process_selinux)
 		}
 	}
 #endif
+	if (gr_locked) {
+		if (gr_unlock (process_selinux) == 0) {
+			fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, gr_dbname ());
+			SYSLOG(LOG_ERR, "failed to unlock %s", gr_dbname());
+			/* continue */
+		}
+	}
 	if (spw_locked) {
 		if (spw_unlock (process_selinux) == 0) {
 			fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, spw_dbname ());
@@ -690,22 +706,6 @@ fail_exit (int code, bool process_selinux)
 			/* continue */
 		}
 	}
-#ifdef ENABLE_SUBIDS
-	if (sub_uid_locked) {
-		if (sub_uid_unlock (process_selinux) == 0) {
-			fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, sub_uid_dbname ());
-			SYSLOG(LOG_ERR, "failed to unlock %s", sub_uid_dbname());
-			/* continue */
-		}
-	}
-	if (sub_gid_locked) {
-		if (sub_gid_unlock (process_selinux) == 0) {
-			fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, sub_gid_dbname ());
-			SYSLOG(LOG_ERR, "failed to unlock %s", sub_gid_dbname());
-			/* continue */
-		}
-	}
-#endif				/* ENABLE_SUBIDS */
 
 #ifdef WITH_AUDIT
 	audit_logger (AUDIT_USER_MGMT,


src/usermod.c
```diff
@@ -660,13 +660,22 @@ NORETURN
 static void
 fail_exit (int code, bool process_selinux)
 {
```
alejandro-colomar, line 662: Why does the patch only change usermod(8)?  Don't any other programs behave similarly?

src/usermod.c
```diff
@@ -660,13 +660,22 @@ NORETURN
...
-	if (gr_locked) {
-		if (gr_unlock (process_selinux) == 0) {
-			fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, gr_dbname ());
-			SYSLOG(LOG_ERR, "failed to unlock %s", gr_dbname());
+#ifdef ENABLE_SUBIDS
+	if (sub_gid_locked) {
+		if (sub_gid_unlock (process_selinux) == 0) {
+			fprintf (stderr, _("%s: failed to unlock %s\n"), Prog, sub_gid_dbname ());
+			SYSLOG(LOG_ERR, "failed to unlock %s", sub_gid_dbname());
```
alejandro-colomar, line 667: I'd expect the reorder to happen in a different patch from the one that does other changes (such as adding guards).

alejandro-colomar:
Cc: @stoeckmann

src/usermod.c
jcpunk: I'll confess, this was the only tool I use with any frequency - and at that only for subids.  I tend to let ansible/puppet do most of my user management.

src/usermod.c
jcpunk: Since I'm now conditionally unlocking only if things were locked, the reorder felt relevant - or is that the part you'd like moved out?

Would you prefer a totally separate PR, or would moving the reorder to another commit off this same PR be sufficient?

src/usermod.c
alejandro-colomar: It would be good to have a look at any other programs that do locking, and check if they do the same thing.  We don't want tools having inconsistent locking rules.

src/usermod.c
alejandro-colomar: A different commit in this PR would be sufficient.

The thing is, I can review the commit that does the move easily with git show --color-moved, and then review the commit that does actual changes more carefully.

jcpunk:
I don't have a great handle on how all the utilities handle locking, and my `C` is "not great".  Refactoring locking across the project is not something I think I'd be qualified to do well.

alejandro-colomar:
On Tue, Mar 10, 2026 at 06:19:16AM -0700, Pat Riehecky wrote:
> jcpunk left a comment (shadow-maint/shadow#1569)
> 
> I don't have a great handle on how all the utilities handle locking,
> and my `C` is "not great".  Refactoring locking across the project is
> not something I think I'd be qualified to do well.

Okay.  Then, let's do this one well, and I'll take the work of spreading
the same changes accross the project.  Thanks!

-- 
<https://www.alejandro-colomar.es>
Use port 80 (that is, <...:80/>).

src/usermod.c
jcpunk: I think I've got this chopped into 4 small, hopefully rational looking, commits.

src/usermod.c
```diff
@@ -1583,44 +1606,15 @@ static void close_files(const struct option_flags *flags)
...
 	 */
-	endpwent ();
-	endspent ();
-	endgrent ();
 #ifdef	SHADOWGRP
 	endsgent ();
 #endif
+	endgrent ();
+	endspent ();
+	endpwent ();
```
jcpunk, line 1617: I don't have a good grip on what these do.  Should they be under the close/unlock?

stoeckmann:
Since I've been CC'ed and looked at this topic before (noticing the passwd locking issue):

> Looking at the repetition across the project, do you think there’s a need for a centralized functions for file handling in the library?

100 % agree. There are mostly two different idioms in place:

- usermod etc. which have a fail_exit function taking an exit code and process_selinux boolean
- groupadd/groupdel which register exit handlers for this

> ensure we're always handling these files consistently.

All of them should lock files in same order to avoid deadlocks. Unlocking can be done in reverse. It would be easiest to create some kind of lock function which takes a bitmask of files you want to lock, which iterates over these bits to always do the correct order. And in reverse for unlock.

> It might be out of scope for this specific PR, but I'd love to hear your thoughts on the pattern.

Agreed as well. What's also outside of this PR but should be noted: [systemd only locks /etc/.pwd.lock no matter if it's processing users or groups](https://github.com/systemd/systemd/blob/43116c56621c7317852c01e67dbcaa78b5e7ff70/src/basic/user-util.c#L688-L713).

So it can easily happen that shadow and systemd lock different files and trigger races.

jcpunk:
> So it can easily happen that shadow and systemd lock different files and trigger races.

Oh my, does that mean more files should be locked for safety rather than less?

stoeckmann:
> > So it can easily happen that shadow and systemd lock different files and trigger races.
> 
> Oh my, does that mean more files should be locked for safety rather than less?

When it comes to reviewing all files, including usermod, I'm happy to assist/help. But for this very PR I would say that it's out of scope.

Also, the situation is worse with libraries like musl-libc which have a no-op lckpwdf implementation. On such systems, shadow does not lock because it trusts that a locking is performed if lckpwdf exists. Which is definitely sane to believe, but ... outside of this PR as well. :)

src/usermod.c
```diff
@@ -1646,18 +1646,21 @@ static void open_files (bool process_selinux)
...
-		fail_exit (E_PW_UPDATE, process_selinux);
-	}
-	spw_locked = true;
-	if (is_shadow_pwd && (spw_open (O_CREAT | O_RDWR) == 0)) {
-		fprintf (stderr,
-		         _("%s: cannot open %s\n"),
-		         Prog, spw_dbname ());
-		fail_exit (E_PW_UPDATE, process_selinux);
+	if (is_shadow_pwd
+	    && (lflg || uflg || pflg || eflg || fflg || Lflg || Uflg)) {
```
alejandro-colomar: It looks like it fits in a single line in 80 columns.  I'd avoid breaking the line.

src/usermod.c
```diff
@@ -1646,18 +1646,21 @@ static void open_files (bool process_selinux)
...
+				Prog, spw_dbname ());
+			fail_exit (E_PW_UPDATE, process_selinux);
+		}
+		if (is_shadow_pwd && (spw_open (O_CREAT | O_RDWR) == 0)) {
+			fprintf (stderr,
+				_("%s: cannot open %s\n"),
+				Prog, spw_dbname ());
+			fail_exit (E_PW_UPDATE, process_selinux);
+		}
+		spw_locked = true;
```
alejandro-colomar: Why did you move this assignment?

src/usermod.c
```diff
@@ -1513,15 +1513,15 @@ static void close_files(const struct option_flags *flags)
 		SYSLOG(LOG_ERR, "failure while writing changes to %s", pw_dbname());
 		fail_exit (E_PW_UPDATE, process_selinux);
 	}
-	if (is_shadow_pwd && (spw_close (process_selinux) == 0)) {
+	if (spw_locked && (spw_close (process_selinux) == 0)) {
```
alejandro-colomar: For the (2nd) commit, which is 7dfe592a ("usermod: only unlock on close_files if we actually locked them "):

```
```

It looks very reasonable to me.  I would appreciate if someone else would review it too, but I think it's good.

src/usermod.c
alejandro-colomar: I think it made more sense right after `spw_lock()`, no?

src/usermod.c
alejandro-colomar: This is very weird.  usermod(8) doesn't call setpwent(3), so it shouldn't be calling endpwent(3) at all.  I'll try to research more into this tomorrow.

alejandro-colomar: At first glance, it looks like a bug introduced in 45c6603c (2007-10-07; "[svn-upgrade] Integrating new upstream version, shadow (19990709)"), which is when usermod(8) was added.

I'm surprised such a bug has not caused serious damage so far, TBH.

alejandro-colomar: We need to check this more thoroughly.

src/usermod.c
```diff
@@ -1563,8 +1573,14 @@ static void close_files(const struct option_flags *flags)
 		}
 		gr_locked = false;
 	}
-
-	if (is_shadow_pwd) {
+	if (spw_locked) {
+		if (spw_close (process_selinux) == 0) {
+			fprintf (stderr,
+		         	_("%s: failure while writing changes to %s\n"),
+		         	Prog, spw_dbname ());
```
alejandro-colomar: These two lines change white space accidentally (a tab was replaced by spaces).

src/usermod.c
jcpunk: Fixed

src/usermod.c
jcpunk: I'll put it back in the original location.

src/usermod.c
jcpunk: As part of fixing the items you've identified in the first commit, the hash changed.

I don't want to attach you're approval to a hash you didn't specify.  That feels rude.

The second commit has no changes, but, since the hash changed, I want to be sure I'm behaving and don't do something unfair.  https://github.com/shadow-maint/shadow/pull/1569/changes/21208ade9316cf06d7e64926be3cf3995373e51e should be exactly the same as before.

src/usermod.c
jcpunk: This should be fixed now.

src/usermod.c
alejandro-colomar: As long as the patch is identical, please add it.  If the patch changes (which can be verified with git-range-diff(1)), of course consider dropping it.

jcpunk: Reviewed-by added to "`usermod: only unlock on close_files if we actually locked them `"

jcpunk:
As a reminder, I have zero ideas how to test this PR

alejandro-colomar:
> As a reminder, I have zero ideas how to test this PR

Thanks for the reminder!

For now, I'm focusing on what I'm able to understand without testing.  It seems simpler than it would look at first.

src/usermod.c
```diff
@@ -1646,18 +1646,20 @@ static void open_files (bool process_selinux)
...
+				Prog, spw_dbname ());
+			fail_exit (E_PW_UPDATE, process_selinux);
+		}
+		spw_locked = true;
+		if (is_shadow_pwd && (spw_open (O_CREAT | O_RDWR) == 0)) {
+			fprintf (stderr,
+				_("%s: cannot open %s\n"),
+				Prog, spw_dbname ());
+			fail_exit (E_PW_UPDATE, process_selinux);
+		}
```
alejandro-colomar, line 1653: The first patch seems reasonable to me.  Having locked /etc/passwd, I feel safe, and since we don't use /etc/shadow, we shouldn't need locking it.

I'll have another look later, to make sure there's no flag we forgot about that might modify /etc/shadow, but certainly LGTM.

src/usermod.c
stoeckmann: From getpwent(3):
```
     The getpwent() function returns a pointer to a structure containing the broken-out fields of a record from the password database (e.g., the local password
     file /etc/passwd, NIS, and LDAP).  The first time getpwent() is called, it returns the first entry; thereafter, it returns successive entries.

     The setpwent() function rewinds to the beginning of the password database.

    The endpwent() function is used to close the password database after all processing has been performed.
```

And in case of (current) glibc, endpwent does nothing if the password database was not opened (see [here](https://sourceware.org/git/?p=glibc.git;a=blob;f=nss/getXXent_r.c;h=9925c3a50b923c8728011d4c71b920bae01265c2;hb=HEAD#l138)).

But I agree, there is no getpwent (and also no prefix_getpwent call which could use getpwent) reachable within usermod.

alejandro-colomar: I'll prepare a separate PR after this one is done.  Thanks!

src/usermod.c
```diff
@@ -1646,18 +1646,20 @@ static void open_files (bool process_selinux)
...
-		         Prog, spw_dbname ());
-		fail_exit (E_PW_UPDATE, process_selinux);
-	}
-	spw_locked = true;
-	if (is_shadow_pwd && (spw_open (O_CREAT | O_RDWR) == 0)) {
-		fprintf (stderr,
-		         _("%s: cannot open %s\n"),
-		         Prog, spw_dbname ());
-		fail_exit (E_PW_UPDATE, process_selinux);
+	if (is_shadow_pwd && (lflg || uflg || pflg || eflg || fflg || Lflg || Uflg)) {
```
stoeckmann: Could you explain where `uflg` becomes relevant for shadow? I can't find a branch for it, which is not a big surprise for me, given that the shadow file does not contain UIDs.

jcpunk: That is human error, I grabbed common flags to scatter about and missed removal of that one.

I'll get it dropped momentarily.

src/usermod.c
stoeckmann: Can't find a missing flag for now either. But given all these global variables, I'm not sure enough to say for certain that no weird path exists.

src/usermod.c
```diff
@@ -1646,18 +1646,20 @@ static void open_files (bool process_selinux)
...
-		fail_exit (E_PW_UPDATE, process_selinux);
+	if (is_shadow_pwd && (lflg || pflg || eflg || fflg || Lflg || Uflg)) {
+		if (spw_lock () == 0) {
+			fprintf (stderr,
+				_("%s: cannot lock %s; try again later.\n"),
+				Prog, spw_dbname ());
+			fail_exit (E_PW_UPDATE, process_selinux);
+		}
+		spw_locked = true;
+		if (is_shadow_pwd && (spw_open (O_CREAT | O_RDWR) == 0)) {
```
alejandro-colomar, line 1648: If we keep `is_shadow_pwd` in the second test, maybe it would be more consistent to also keep it in the first one?

That is, write it as this?:

```c
	if (lflg || pflg || eflg || fflg || Lflg || Uflg) {
		if (is_shadow_pwd && spw_lock () == 0) {
```

alejandro-colomar: Hmm, no, that would be wrong.  We'd need to wrap `spw_locked = true;` too.

Maybe we should later remove dead code in a separate patch.

alejandro-colomar, line 1648: Should we remove `is_shadow_pwd && ` here?

alejandro-colomar: I'll do this while extending this to the rest of the project.  No need to do it here.

alejandro-colomar:
Thanks!  I'll merge now.  It's good already

> Looking at the repetition across the project, do you think there’s a need for a centralized functions for file handling in the library? The open/close sequence can be managed in one place with arguments for specific files would make the code a bit more robust and ensure we're always handling these files consistently. It might be out of scope for this specific PR, but I'd love to hear your thoughts on the pattern.

I'll start working on a PR to move this to library code, and use it everywhere.

Issue #1554, opened 2026-02-26
`usermod` flag to write `subuid` and `subgid` entries by UID rather than username


jcpunk:
From the subuid manpage:
>  When large number of entries (10000-100000 or more) are defined in /etc/subuid, parsing performance penalty will become noticeable. In this case it is recommended to use UIDs instead of login names. Benchmarks have shown speed-ups up to 20x.

It would be helpful if there was something I could set in `/etc/login.defs` or with `usermod` to write out subids with their UID instead of the username.

I suspect this would be of value to `useradd` as well.

alejandro-colomar:
> From the subuid manpage:

Please use proper manual page references: subuid(5).  Otherwise, it might be unclear to which page you refer to.

> 
> > When large number of entries (10000-100000 or more) are defined in /etc/subuid, parsing performance penalty will become noticeable. In this case it is recommended to use UIDs instead of login names. Benchmarks have shown speed-ups up to 20x.
> 
> It would be helpful if there was something I could set in `/etc/login.defs` or with `usermod` to write out subids with their UID instead of the username.
> 
> I suspect this would be of value to `useradd` as well.

Cc: @richardweinberger , @ikerexxe , @hallyn

hallyn:
No objection.  login.defs probably makes the most sense, since having just one of the entries use UID isn't going to help if it's performance you care about.

PR #1570, opened 2026-03-09
subid: permit storage by id number rather than username

jcpunk:
This adds two new options to /etc/login.defs:
* SUB_UID_STORE_BY_UID
* SUB_GID_STORE_BY_UID

They default to 'no' but when set 'yes' the
subuid/subgid entries will be written by uid

Folks may end up with a mixed database, so userdel and usermod have been updated to scan for both by uid and by username when removing items.

Closes: https://github.com/shadow-maint/shadow/issues/1554

I will need help with the testing...


alejandro-colomar:
> Fixes: #1554

You should use `Closes:` for referring to issue tickets.  `Fixes:` is for referring to commits which introduced a bug which is fixed now.

src/newusers.c
```diff
@@ -1173,46 +1174,66 @@ int main (int argc, char **argv)
...
-					_("%s: failed to prepare new %s entry\n"),
-					Prog, sub_uid_dbname ());
-				fail_exit (EXIT_FAILURE, process_selinux);
+		{
+			char *sub_uid_owner = getdef_bool ("SUB_UID_STORE_BY_UID")
+			                      ? xaprintf ("%u", (unsigned int) newpw.pw_uid)
+			                      : xstrdup (fields[0]);
+			char *sub_gid_owner = getdef_bool ("SUB_GID_STORE_BY_UID")
+			                      ? xaprintf ("%u", (unsigned int) newpw.pw_uid)
+			                      : xstrdup (fields[0]);
```
alejandro-colomar, line 1183: Please declare variables separately from their use.

Also, please don't use white space between function names and `(` in new code.

See also: <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c>

src/newusers.c
```diff
@@ -1173,46 +1174,66 @@ int main (int argc, char **argv)
...
-			}
-			if (sub_uid_add(fields[0], sub_uid_start, sub_uid_count) == 0)
-			{
-				fprintf (stderr,
-					_("%s: failed to prepare new %s entry\n"),
-					Prog, sub_uid_dbname ());
-				fail_exit (EXIT_FAILURE, process_selinux);
+		{
+			char *sub_uid_owner = getdef_bool ("SUB_UID_STORE_BY_UID")
+			                      ? xaprintf ("%u", (unsigned int) newpw.pw_uid)
```
alejandro-colomar, line 1179: I'd prefer using `(u)intmax_t` for printing typedefs.

src/newusers.c
```diff
@@ -1173,46 +1174,66 @@ int main (int argc, char **argv)
...
-					Prog);
-				fail_exit (EXIT_FAILURE, process_selinux);
-			}
-			if (sub_uid_add(fields[0], sub_uid_start, sub_uid_count) == 0)
-			{
-				fprintf (stderr,
-					_("%s: failed to prepare new %s entry\n"),
-					Prog, sub_uid_dbname ());
-				fail_exit (EXIT_FAILURE, process_selinux);
+		{
```
alejandro-colomar, line 1177: Can we abstract this into a separate function?  The loop is growing too much, and dangling blocks are an indicator of that.

jcpunk: Can do!

Is this something you'd like to see made generic enough it could be linked in by newusers.c, useradd.c, userdel.c, and usermod.c (under lib?) or would the duplicate code be fine once it is broken out into a dedicated function?

alejandro-colomar: If several programs do the same exact thing, you're welcome to move it to somewhere in `lib/` and call it everywhere.  That would be a separate PR, I guess, and I'd welcome that very much!  De-duplicating code reduces the maintenance work.

If you do so, please separate the PR into many small commits that can be easily reviewed.

jcpunk: Opened https://github.com/shadow-maint/shadow/pull/1573 to try a more centralized approach.

PR #1381, opened 2025-11-03
Preserve errno; and add fprinte() and SYSLOGE(), which append strerrno()

alejandro-colomar:
These are similar to libc's err(3), but more flexible.

This PR does fix uncountable bugs related to clobbering of errno.  @stoeckmann shared one reproducer in

-  <https://github.com/shadow-maint/shadow/pull/1446>

---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase

```
$ git range-diff syslog_c..gh/e se2..e
1:  f9a2c5170 = 1:  c9324c952 lib/io/fprintf/: [v]fprinte(): Add function
2:  22d737344 = 2:  e1cb59b9e lib/: Use [v]fprinte() instead of its pattern
3:  a79ac3072 = 3:  14f2b7c9f lib/io/syslog.h: SYSLOGE(): Add macro
4:  9f5390550 = 4:  0bf285d61 lib/, src/: Use SYSLOGE() instead of its pattern
```
</details>

<details>
<summary>v2</summary>

-  Use `"%m"` with syslog(3).

```
$ git range-diff --creation-factor=99 se2 gh/e e 
1:  c9324c952 = 1:  c9324c952 lib/io/fprintf/: [v]fprinte(): Add function
2:  e1cb59b9e = 2:  e1cb59b9e lib/: Use [v]fprinte() instead of its pattern
3:  14f2b7c9f ! 3:  db29cee50 lib/io/syslog.h: SYSLOGE(): Add macro
    @@ Metadata
      ## Commit message ##
         lib/io/syslog.h: SYSLOGE(): Add macro
     
    +    syslog(3) supports "%m" (guaranteed by POSIX); we don't need to call
    +    strerrno() here.
    +
         Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
      ## lib/io/syslog.h ##
    -@@
    - #include <string.h>
    - #include <syslog.h>
    - 
    -+#include "string/strerrno.h"
    -+
    - 
    - #ifndef LOG_WARN
    - #define LOG_WARN LOG_WARNING
     @@
      #define OPENLOG(progname) openlog(progname, SYSLOG_OPTIONS, SYSLOG_FACILITY)
      
    @@ lib/io/syslog.h
     +// system log errno
     +#define SYSLOGE(prio, fmt, ...)  do                                   \
     +{                                                                     \
    -+  SYSLOG(prio, "" fmt ": %s" __VA_OPT__(,) __VA_ARGS__, strerrno()); \
[...]

PR #1571, opened 2026-03-09
Deterministic subid calculation

jcpunk:
This patch solves a long standing problem at my site.

We have two main environments: LDAP and "not LDAP".

Our LDAP server doesn't have a schema for subid, so that is local to each system.  Users are added to LDAP by a different group and we aren't usually told when a new user appears.

Our "not LDAP" systems are expected to have their UID match whatever is in LDAP, but beyond that we are free to customize.  This is typically a lab environment where folks have groups specific to their access within that lab cluster.  So here we explicitly want to mangage the subids.

More often then I'd like, users want us to setup an sync job where their home area from the LDAP is synced down to a sub area on the test cluster.  So that makes keeping the subids in sync a bit more of an adventure.

Then it gets worse. My site has currently around 63,000 users.  Which means if I give every user 65,536 subids to play with, in a few years I'll be out of space since uid_t is u_int_32.  Thankfully, my test labs don't the full user list.  They generally have about 100 or so users.

Before the merge of https://github.com/shadow-maint/shadow/commit/1ed06fee4c8d10d785ac4569cb6cb81cba7e6e0c I figured this was a 100% my problem thing since usermod didn't have a "recommend" mode.  But now it does.

This patch makes it automatic for me to keep all my "16bit" user subids in perfect sync by just setting up a job to run `usermod -S` as needed.

I'm going to need a subid solution for user 75537 (since `UID_MIN==1000`) eventually. Since the only place where I'd consider enabling wrap mode is in my heavily curated lab computers, seeding an imperfect solution.  I can't shake the feeling that I'm going to need this one day.  When LDAP "rolls over" I think we may just drop user namespaces on those hosts...

I'll certainly need help with tests.

### DIFF
--- a/lib/find_new_sub_gids.c
+++ b/lib/find_new_sub_gids.c
@@ -32,10 +32,22 @@
  *   start_id       = SUB_GID_MIN + logical_offset
  *   end_id         = start_id + SUB_GID_COUNT - 1
  *
- * DETERMINISTIC MODE:
+ * DETERMINISTIC-SAFE MODE (default):
  *   All arithmetic overflow is a hard error.  The assigned range must fit
  *   entirely within [SUB_GID_MIN, SUB_GID_MAX].  Allocation is monotonic
  *   and guaranteed non-overlapping.
+ *
+ * UNSAFE_SUB_GID_DETERMINISTIC_WRAP MODE:
+ *   Activated with UNSAFE_SUB_GID_DETERMINISTIC_WRAP yes
+ *
+ *   WARNING: SECURITY RISK!
+ *   WARNING: MAY CAUSE RANGE OVERLAPS!
+ *   WARNING: MAY CAUSE CONTAINER ESCAPES!
+ *
+ *   The subordinate GID space is treated as a ring.  Arithmetic overflow
+ *   is normalised via modulo over [SUB_GID_MIN, SUB_GID_MAX].
+ *   This means ranges MAY overlap for large UID populations!
+ *   Intended only for development, testing, or constrained lab environments.
  *
  * Return 0 on success, -1 if no GIDs are available.
  */
@@ -44,7 +56,9 @@ find_new_sub_gids_deterministic(uid_t uid,
 				id_t *range_start,
 				unsigned long *range_count)
 {
+	bool           allow_wrap;
 	unsigned long  count;
+	unsigned long  slot;
 	unsigned long  slots;
 	unsigned long  space;
 	unsigned long  uid_min;
@@ -56,6 +70,7 @@ find_new_sub_gids_deterministic(uid_t uid,
 	sub_gid_min = getdef_ulong ("SUB_GID_MIN", 65536UL);
 	sub_gid_max = getdef_ulong ("SUB_GID_MAX", 4294967295UL);
 	count = getdef_ulong ("SUB_GID_COUNT", 65536UL);
+	allow_wrap = getdef_bool ("UNSAFE_SUB_GID_DETERMINISTIC_WRAP");
 
 	if (uid < uid_min) {
 		fprintf(log_get_logfd(),
@@ -96,17 +111,22 @@ find_new_sub_gids_deterministic(uid_t uid,
 
 	uid_offset = uid - uid_min;
 	slots = space / count;
+	slot = uid_offset;
 
 	if (uid_offset >= slots) {
-		fprintf(log_get_logfd(),
-		         _("%s: Deterministic subordinate GID range"
-		           " for UID %ju exceeds SUB_GID_MAX (%lu)\n"),
-		         log_get_progname(),
-		         (uintmax_t)uid, sub_gid_max);
-		return -1;
+		if (allow_wrap) {
+			slot = uid_offset % slots;
+		} else {
+			fprintf(log_get_logfd(),
+	         		_("%s: Deterministic subordinate GID range"
+				  " for UID %ju exceeds SUB_GID_MAX (%lu)\n"),
+	         		log_get_progname(),
+	         		(uintmax_t)uid, sub_gid_max);
+			return -1;
+		}
 	}
 
-	*range_start = sub_gid_min + uid_offset * count;
+	*range_start = sub_gid_min + slot * count;
 	*range_count = count;
 	return 0;
 }
--- a/lib/find_new_sub_uids.c
+++ b/lib/find_new_sub_uids.c
@@ -32,10 +32,22 @@
  *   start_id       = SUB_UID_MIN + logical_offset
  *   end_id         = start_id + SUB_UID_COUNT - 1
  *
- * DETERMINISTIC MODE:
+ * DETERMINISTIC-SAFE MODE (default):
  *   All arithmetic overflow is a hard error.  The assigned range must fit
  *   entirely within [SUB_UID_MIN, SUB_UID_MAX].  Allocation is monotonic
  *   and guaranteed non-overlapping.
+ *
+ * UNSAFE_SUB_UID_DETERMINISTIC_WRAP MODE:
+ *   Activated with UNSAFE_SUB_UID_DETERMINISTIC_WRAP yes
+ *
+ *   WARNING: SECURITY RISK!
+ *   WARNING: MAY CAUSE RANGE OVERLAPS!
+ *   WARNING: MAY CAUSE CONTAINER ESCAPES!
+ *
+ *   The subordinate UID space is treated as a ring.  Arithmetic overflow
+ *   is normalised via modulo over [SUB_UID_MIN, SUB_UID_MAX].
+ *   This means ranges MAY overlap for large UID populations!
+ *   Intended only for development, testing, or constrained lab environments.
  *
  * Return 0 on success, -1 if no UIDs are available.
  */
@@ -44,7 +56,9 @@ find_new_sub_uids_deterministic(uid_t uid,
 				id_t *range_start,
 				unsigned long *range_count)
 {
+	bool           allow_wrap;
 	unsigned long  count;
+	unsigned long  slot;
 	unsigned long  slots;
 	unsigned long  space;
 	unsigned long  uid_min;
@@ -56,6 +70,7 @@ find_new_sub_uids_deterministic(uid_t uid,
 	sub_uid_min = getdef_ulong ("SUB_UID_MIN", 65536UL);
 	sub_uid_max = getdef_ulong ("SUB_UID_MAX", 4294967295UL);
 	count = getdef_ulong ("SUB_UID_COUNT", 65536UL);
+	allow_wrap = getdef_bool ("UNSAFE_SUB_UID_DETERMINISTIC_WRAP");
 
 	if (uid < uid_min) {
 		fprintf(log_get_logfd(),
@@ -96,17 +111,22 @@ find_new_sub_uids_deterministic(uid_t uid,
 
 	uid_offset = uid - uid_min;
 	slots = space / count;
+	slot = uid_offset;
 
-	if (uid_offset > slots) {
-		fprintf(log_get_logfd(),
-		         _("%s: Deterministic subordinate UID range"
-		           " for UID %ju exceeds SUB_UID_MAX (%lu)\n"),
-		         log_get_progname(),
-		         (uintmax_t)uid, sub_uid_max);
-		return -1;
+	if (uid_offset >= slots) {
+		if (allow_wrap) {
+			slot = uid_offset % slots;
+		} else {
+			fprintf(log_get_logfd(),
+				_("%s: Deterministic subordinate UID range"
+				  " for UID %ju exceeds SUB_UID_MAX (%lu)\n"),
+				log_get_progname(),
+				(uintmax_t)uid, sub_uid_max);
+			return -1;
+		}
 	}
 
-	*range_start = sub_uid_min + uid_offset * count;
+	*range_start = sub_uid_min + slot * count;
 	*range_count = count;
 	return 0;
 }


lib/find_new_sub_gids.c
```diff
@@ -46,6 +51,87 @@ int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
 		return -1;
 	}
 
+	if (getdef_bool ("SUB_GID_DETERMINISTIC")) {
```
alejandro-colomar: Should this block go into a separate function?

lib/find_new_sub_gids.c
```diff
@@ -46,6 +51,87 @@ int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
...
+		unsigned long uid_min;
+		unsigned long uid_offset;
+		unsigned long space;
+		bool allow_wrap;
+
+		uid_min = getdef_ulong ("UID_MIN", 1000UL);
+		allow_wrap = getdef_bool ("UNSAFE_SUB_GID_DETERMINISTIC_WRAP");
+
+		if ((unsigned long)uid < uid_min) {
+			(void) fprintf (log_get_logfd(),
```
alejandro-colomar: Please don't cast to void in new code.

lib/find_new_sub_gids.c
```diff
@@ -46,6 +51,87 @@ int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
...
+					_("%s: UID %lu is less than UID_MIN %lu,"
+					  " cannot calculate deterministic subordinate GIDs\n"),
+				log_get_progname(),
+				(unsigned long)uid, uid_min);
+			return -1;
+		}
+
+		uid_offset = (unsigned long)uid - uid_min;
```
alejandro-colomar: Why cast?

lib/find_new_sub_gids.c
```diff
@@ -46,6 +51,87 @@ int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
...
+
+		uid_min = getdef_ulong ("UID_MIN", 1000UL);
+		allow_wrap = getdef_bool ("UNSAFE_SUB_GID_DETERMINISTIC_WRAP");
+
+		if ((unsigned long)uid < uid_min) {
+			(void) fprintf (log_get_logfd(),
+					_("%s: UID %lu is less than UID_MIN %lu,"
+					  " cannot calculate deterministic subordinate GIDs\n"),
+				log_get_progname(),
+				(unsigned long)uid, uid_min);
```
alejandro-colomar: I'd prefer `uintmax_t`.

lib/find_new_sub_gids.c
jcpunk: I have massively increased the size of this function, but I think it still has a clear purpose.  If I broke it out to another function, should I do the same for the existing code something like:
```C
 int find_new_sub_gids (uid_t uid, id_t *range_start, unsigned long *range_count) {
	if (getdef_bool("SUB_GID_DETERMINISTIC")) {
        find_new_sub_gids_deterministic(uid, range_start, range_count)
    } else {
        find_new_sub_gids_linear(uid, range_start, range_count)
    }
}
```
Would that provide more clarity or just make a branch for branch sake?

lib/find_new_sub_gids.c
jcpunk: Human error, I had it in my head that I was at risk for a rollover.  It is clearly not a risk in that regard...

lib/find_new_sub_gids.c
alejandro-colomar: Sounds very reasonable!

jcpunk:
I think I've reworked this into a more linear patch set. Hopefully I've gotten the overflow checks clearly documented.

lib/find_new_sub_gids.c
```diff
@@ -9,56 +9,74 @@
...
+	unsigned long min, max, count;
 	id_t start;
 
 	assert (range_start != NULL);
 	assert (range_count != NULL);
 
 	min = getdef_ulong ("SUB_GID_MIN", 100000UL);
 	max = getdef_ulong ("SUB_GID_MAX", 600100000UL);
-	count = getdef_ulong ("SUB_GID_COUNT", 65536);
+	count = getdef_ulong ("SUB_GID_COUNT", 65536UL);
```
alejandro-colomar, line 72: The `UL` suffix seems completely unnecessary.

alejandro-colomar: Oh, I see it's for consistency with the lines above.  Hmmm, okay.

lib/find_new_sub_gids.c
```diff
@@ -9,56 +9,74 @@
...
-	start = sub_gid_find_free_range(min, max, count);
+	start = sub_gid_find_free_range (min, max, count);
 	if (start == (id_t)-1) {
-		fprintf (log_get_logfd(),
+		fprintf (log_get_logfd (),
 		         _("%s: Can't get unique subordinate GID range\n"),
-		         log_get_progname());
-		SYSLOG(LOG_WARN, "no more available subordinate GIDs on the system");
+		         log_get_progname ());
+		SYSLOG ((LOG_WARN, "no more available subordinate GIDs on the system"));
```
alejandro-colomar: This change is wrong.

lib/find_new_sub_gids.c
```diff
@@ -9,56 +9,74 @@
...
+	if (count == 0 || min > max || count > (max - min + 1)) {
+		fprintf (log_get_logfd (),
+		         _("%s: Invalid configuration: SUB_GID_MIN (%lu),"
+		           " SUB_GID_MAX (%lu), SUB_GID_COUNT (%lu)\n"),
+		         log_get_progname (), min, max, count);
 		return -1;
 	}
 
-	start = sub_gid_find_free_range(min, max, count);
+	start = sub_gid_find_free_range (min, max, count);
```
alejandro-colomar: Why?

lib/find_new_sub_uids.c
```diff
@@ -9,56 +9,74 @@
...
 	max = getdef_ulong ("SUB_UID_MAX", 600100000UL);
-	count = getdef_ulong ("SUB_UID_COUNT", 65536);
+	count = getdef_ulong ("SUB_UID_COUNT", 65536UL);
 
-	if (min > max || count >= max || (min + count - 1) > max) {
-		(void) fprintf (log_get_logfd(),
-				_("%s: Invalid configuration: SUB_UID_MIN (%lu),"
-				  " SUB_UID_MAX (%lu), SUB_UID_COUNT (%lu)\n"),
-			log_get_progname(), min, max, count);
+	if (count == 0 || min > max || count > (max - min + 1)) {
```
alejandro-colomar: Why does the expression change?

lib/find_new_sub_uids.c
```diff
@@ -9,56 +9,74 @@
...
 	*range_count = count;
 	return 0;
 }
+
+/*
+ * find_new_sub_uids - Find a new unused range of subordinate UIDs.
+ *
+ * Return 0 on success, -1 if no unused UIDs are available.
+ */
+int find_new_sub_uids (uid_t uid, id_t *range_start, unsigned long *range_count)
```
alejandro-colomar: uid is unused in this commit.

alejandro-colomar: Ahh, sorry; my bad.  I was looking at another window.  Ignore this.

alejandro-colomar: Ahh, no, my comment was correct.  I'll re-open.

lib/find_new_sub_uids.c
```diff
@@ -19,6 +19,175 @@
...
+	assert (range_start != NULL);
+	assert (range_count != NULL);
+
+	uid_min = getdef_ulong ("UID_MIN", 1000UL);
+	sub_uid_min = getdef_ulong ("SUB_UID_MIN", 65536UL);
+	sub_uid_max = getdef_ulong ("SUB_UID_MAX", 4294967295UL);
+	count = getdef_ulong ("SUB_UID_COUNT", 65536UL);
+	allow_wrap = getdef_bool  ("UNSAFE_SUB_UID_DETERMINISTIC_WRAP");
+
+	if ((unsigned long)uid < uid_min) {
```
alejandro-colomar: I'd remove the cast.

uid_t isn't going to be a signed integer wider than ulong.

The cast is a source of problems.

lib/find_new_sub_uids.c
```diff
@@ -19,6 +19,175 @@
...
+		fprintf (log_get_logfd (),
+		         _("%s: UID %lu is less than UID_MIN %lu,"
+		           " cannot calculate deterministic subordinate UIDs\n"),
+		         log_get_progname (),
+		         (unsigned long)uid, uid_min);
```
alejandro-colomar: Please use uintmax_t.

lib/find_new_sub_uids.c
```diff
@@ -19,6 +19,175 @@
...
+		return -1;
+	}
+
+	/*
+	 * Validate configuration before using min/max in any arithmetic.
+	 * If sub_uid_min > sub_uid_max, the space calculation below would
+	 * unsigned-wrap and silently produce a nonsense value.
+	 */
+	if (sub_uid_min > sub_uid_max || count == 0) {
```
alejandro-colomar: The comment is redundant with the code, IMO.  Please move it to the commit message.

lib/find_new_sub_uids.c
```diff
@@ -19,6 +19,175 @@
...
+	}
+
+	uid_offset = (unsigned long)uid - uid_min;
+	space = sub_uid_max - sub_uid_min + 1;
+
+	/*
+	 * A range larger than the entire configured space can never be placed
+	 * without overlap, regardless of mode.
+	 */
+	if (count > space) {
```
alejandro-colomar: The comment is superfluous.  Please remove.

lib/find_new_sub_uids.c
```diff
@@ -19,6 +19,175 @@
...
+	 * 32-bit platforms where unsigned long is 32 bits.  The modulo
+	 * folds the result back into [0, space) before adding min.
+	 */
+	uintmax_t logical_offset = (uintmax_t)uid_offset * (uintmax_t)count;
+
+	*range_start = (id_t)(sub_uid_min + (unsigned long)(logical_offset % space));
+	*range_count = count;
+	return 0;
+}
+
```
alejandro-colomar: Why would we want an unsafe mode?

lib/find_new_sub_uids.c
```diff
@@ -19,6 +19,175 @@
...
+		if (end_id > sub_uid_max) {
+			fprintf (log_get_logfd (),
+			         _("%s: Deterministic subordinate UID range"
+			           " for UID %lu exceeds SUB_UID_MAX (%lu)\n"),
+			         log_get_progname (),
+			         (unsigned long)uid, sub_uid_max);
+			return -1;
+		}
+
+		*range_start = (id_t)start_id;
```
alejandro-colomar: Why cast?

lib/find_new_sub_uids.c
```diff
@@ -19,6 +19,175 @@
...
+		}
+
+		if (end_id > sub_uid_max) {
+			fprintf (log_get_logfd (),
+			         _("%s: Deterministic subordinate UID range"
+			           " for UID %lu exceeds SUB_UID_MAX (%lu)\n"),
+			         log_get_progname (),
+			         (unsigned long)uid, sub_uid_max);
+			return -1;
+		}
```
alejandro-colomar: Should we have a goto to a single place where we log an error message?

alejandro-colomar: Something like this:

```c

		if (__builtin_mul_overflow(uid_offset, count, &product))
			goto ovf;
		if (__builtin_add_overflow(sub_uid_min, product, &start_id))
			goto ovf;
		if (__builtin_add_overflow(start_id, count - 1, &end_id))
			goto ovf;

		if (end_id > sub_uid_max)
			goto something-else;

...
ovf:
		fprintf (log_get_logfd (), ...);
		return -1;
something-else:
...
```

lib/find_new_sub_uids.c
```diff
@@ -19,6 +19,175 @@
...
+		 *      Overflows when the offset pushes start past UINTMAX_MAX.
+		 *
+		 *   3. start_id + (count-1) -> end_id
+		 *      Overflows when start_id is near UINTMAX_MAX even after
+		 *      guards 1 and 2 pass.  Required on platforms where
+		 *      uintmax_t is wider than the sub-UID value space.
+		 *
+		 * Omitting any one leaves a range-escape vector on some
+		 * possible arch/config combinations.
+		 */
```
alejandro-colomar: This is redundant with the code.  This would be interesting in the commit message, but not in the code.

lib/find_new_sub_gids.c
jcpunk: When I started on this patch, I didn't realize the space between the function and the `(` was the "old style" rather than the "current style" and was trying to use the correct style.

I'll clean that up.

alejandro-colomar: No problem.

lib/find_new_sub_uids.c
jcpunk: A zero count can't result in a valid range.  I'll break this into a separate commit with a longer explanation and see if that holds water.

jcpunk: Digging in deeper, `find_free_range` from `subordinateio.c` already has the `count == 0` check, so the expression here would be redundant.

lib/find_new_sub_uids.c
jcpunk: I've never really been comfortable with `goto`.  For a print and return is there really that much savings?

lib/find_new_sub_gids.c
jcpunk: Fixed

lib/find_new_sub_uids.c
jcpunk: I've moved the function arguments change to commit 3 so the earlier commits don't have unused parameters.

lib/find_new_sub_uids.c
jcpunk: Dropped

lib/find_new_sub_uids.c
jcpunk: Removed

lib/find_new_sub_uids.c
jcpunk: The initial definitions were not guaranteed to be the same size. I've changed them to be an identical width integer.

After doing all those overflow checks, I wanted to be sure I didn't then foolishly overflow

lib/find_new_sub_uids.c
jcpunk: Moved to commit message

lib/find_new_sub_uids.c
jcpunk: Done

lib/find_new_sub_uids.c
jcpunk: Dropped.

lib/find_new_sub_uids.c
jcpunk: At my site, once a UNIX UID is assigned, that is your ID forever.  It is never reused or reassigned.  It is expected to be identical across all systems.

In the next 5 years I expect to have more assigned UIDs than will allow me to give each user 65535 subids.

Eventually I'm going to hit a point where my newest users can't get subids because a bunch of subids are assigned to folks who've left the laboratory.

In my systems with the full user list, I'm going to probably just disable user namespaces since there is no safe way for me to give everyone their own big subid space.  I don't really like that plan, but I don't have a better one.

My systems with a limited user list are more interesting... On hosts dedicated to experimental development, I'm willing to take more risks. These hosts typically have less than 100 users.

The great thing about strong determinism in subids is that, so long as all the ids fit between SUBID_MIN and SUBID_MAX, I have perfect consistency across the whole site.  Once I run out of space things get rough.

On these hosts as the sysadmin responsible for safety and security of the system, I can see that my UIDs are not contiguous. At times I've got a 1,000+ "unallocated" UIDs on a given host because there are no users in that range who have access to the system.  I'm willing to "risk" some overlap when I can see that I've zero interactive UIDs under 2000.

"Unsafe mode" is my best compromise.  I get the determinism I want so subids are predicable, and, when I run out of space, I at least have an option I can consider for how to approach the problem.

With a UNIX UID being uint_32, and the traditional subid allocation being "each user gets their own uint_16", I'm going to hit that wall and I don't know what else to do.

If I'm reading the defaults correctly, the defaults from `/etc/login.defs` only have `9148` `65536` ranges.  After which some sort of plan is required.

lib/find_new_sub_uids.c
alejandro-colomar: Integer conversions do never overflow (AFAIR); they wrap around, no?

lib/find_new_sub_uids.c
alejandro-colomar: We turn dozens or so lines of code into 8, so readability-wise I'd say yes.

alejandro-colomar: But feel free to not do it.  I can clean up afterwards.

lib/find_new_sub_gids.c
```diff
@@ -58,7 +61,22 @@ int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
...
+ */
+int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
+{
+	if (!range_start || !range_count) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	return find_new_sub_gids_linear (range_start, range_count);
+}
```
alejandro-colomar: EINVAL here seems an addition, and seems to be unnecessary, right?  I'm not fundamentally against it, but in this commit, it seems superfluous.

While fixing that, you may want to follow the coding style for new code:

```c
int
find_new_sub_gids(id_t *range_start, unsigned long *range_count)
{
	if (range_start == NULL || range_count == NULL)
		return -1;

	return find_new_sub_gids_linear(range_start, range_count);
}
```

lib/find_new_sub_gids.c
```diff
@@ -36,7 +39,7 @@ int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
 
 	min = getdef_ulong ("SUB_GID_MIN", 100000UL);
 	max = getdef_ulong ("SUB_GID_MAX", 600100000UL);
-	count = getdef_ulong ("SUB_GID_COUNT", 65536);
+	count = getdef_ulong ("SUB_GID_COUNT", 65536UL);
```
alejandro-colomar: I'd be okay with this, but since the commit is otherwise very clean, I think it would be better not having this at all.

lib/prototypes.h
```diff
@@ -133,10 +133,10 @@ extern int find_new_uid (bool sys_user,
 
 #ifdef ENABLE_SUBIDS
 /* find_new_sub_gids.c */
-extern int find_new_sub_gids (id_t *range_start, unsigned long *range_count);
+extern int find_new_sub_gids (uid_t uid, id_t *range_start, unsigned long *range_count);
 
 /* find_new_sub_uids.c */
-extern int find_new_sub_uids (id_t *range_start, unsigned long *range_count);
+extern int find_new_sub_uids (uid_t uid, id_t *range_start, unsigned long *range_count);
```
alejandro-colomar: This doesn't seem to belong in the second commit, I think.  I guess this was an accident trying to move it to the third commit?

lib/find_new_sub_gids.c
```diff
@@ -67,13 +210,16 @@ static int find_new_sub_gids_linear (id_t *range_start, unsigned long *range_cou
...
 {
 	if (!range_start || !range_count) {
 		errno = EINVAL;
 		return -1;
 	}
 
+	if (getdef_bool ("SUB_GID_DETERMINISTIC"))
+		return find_new_sub_gids_deterministic (uid, range_start, range_count);
+
 	return find_new_sub_gids_linear (range_start, range_count);
```
alejandro-colomar: Should we have the "unsafe" variant be a separate feature independent of `SUB_GID_DETERMINISTIC`?

alejandro-colomar: Sorry, I changed my mind.

lib/find_new_sub_gids.c
```diff
@@ -19,6 +19,149 @@
...
+		           " SUB_GID_COUNT=%lu)\n"),
+		         log_get_progname(),
+		         sub_gid_min, sub_gid_max, count);
+		return -1;
+	}
+
+	if (!allow_wrap) {
+		/*
+		 * DETERMINISTIC-SAFE MODE
+		 */
```
alejandro-colomar: I would revert this conditional:

```c
if (allow_wrap) {
    ... unsafe code
}

... safe code
```

That makes it easier to see that the unsafe code is constrained to that condition.

lib/find_new_sub_uids.c
jcpunk: I have honestly no idea.  My `C` mostly consists of what I remember from CS3 back in 2002 plus whatever doc I find and code I can pattern after.

I spend nearly all my time in dynamically typed languages...

lib/find_new_sub_gids.c
jcpunk: I think I've sorted this

lib/find_new_sub_gids.c
jcpunk: I've removed my change.

lib/prototypes.h
jcpunk: I think I fixed this now.

lib/find_new_sub_gids.c
jcpunk: I've moved the unsafe blocks as you've described here.

lib/find_new_sub_uids.c
alejandro-colomar: Yes, in C there's no overflow when converting from one integer type to another integer type (especially, after C23 mandated 2's complement; prior to that, there might be some corner case, in the very rare case where an implementation wouldn't use 2's complement).

You can only get integer overflow when doing arithmetic.

lib/find_new_sub_gids.c
```diff
@@ -58,7 +62,21 @@ int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
...
+ *
+ * Return 0 on success, -1 if no unused GIDs are available.
+ */
+int
+find_new_sub_gids (id_t *range_start, unsigned long *range_count)
+{
+	if (!range_start || !range_count)
+		return -1;
+
+	return find_new_sub_gids_linear (range_start, range_count);
```
alejandro-colomar: Please don't use spaces between a function name and the opening `(`.

<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c#n62>

lib/find_new_sub_gids.c
```diff
@@ -9,23 +9,27 @@
...
- * If successful, find_new_sub_gids provides a range of unused
- * user IDs in the [SUB_GID_MIN:SUB_GID_MAX] range.
+ * Loads SUB_GID_COUNT from login.defs and writes the allocated count back
+ * to *range_count on success.
  *
  * Return 0 on success, -1 if no unused GIDs are available.
  */
-int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
+static int
+find_new_sub_gids_linear (id_t *range_start, unsigned long *range_count)
```
alejandro-colomar: wsfix

lib/find_new_sub_uids.c
```diff
@@ -58,7 +62,21 @@ int find_new_sub_uids (id_t *range_start, unsigned long *range_count)
...
+ * Return 0 on success, -1 if no unused UIDs are available.
+ */
+int
+find_new_sub_uids (id_t *range_start, unsigned long *range_count)
+{
+	if (!range_start || !range_count)
+		return -1;
+
+	return find_new_sub_uids_linear (range_start, range_count);
+}
```
alejandro-colomar, line 175: wsfix

lib/find_new_sub_uids.c
```diff
@@ -9,23 +9,27 @@
...
- * If successful, find_new_sub_uids provides a range of unused
- * user IDs in the [SUB_UID_MIN:SUB_UID_MAX] range.
+ * Loads SUB_UID_COUNT from login.defs and writes the allocated count back
+ * to *range_count on success.
  *
  * Return 0 on success, -1 if no unused UIDs are available.
  */
-int find_new_sub_uids (id_t *range_start, unsigned long *range_count)
+static int
+find_new_sub_uids_linear (id_t *range_start, unsigned long *range_count)
```
alejandro-colomar: wsfix

lib/find_new_sub_gids.c
alejandro-colomar: Do we really need the NULL checks?  Why are they being added here?

jcpunk: Fixed

lib/find_new_sub_gids.c
jcpunk: Fixed

lib/find_new_sub_uids.c
jcpunk: Fixed

lib/find_new_sub_gids.c
jcpunk: Looking over the code, I wasn't sure I could prove that range_start and range_count were always non-NULL.

I can drop it if you think it is unnecessary, but it felt like a safe defensive check.

alejandro-colomar: We only have two calls, and they all pass an address, which is necessarily non-null.

```c
$ grep -rn 'find_new_sub_gids *([^)]'
src/newusers.c:1204:			if (find_new_sub_gids(&sub_gid_start, &sub_gid_count) != 0) {
src/useradd.c:2664:		if (find_new_sub_gids(&sub_gid_start, &subgid_count) < 0) {
lib/find_new_sub_gids.c:28:int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
lib/prototypes.h:136:extern int find_new_sub_gids (id_t *range_start, unsigned long *range_count);
```

alejandro-colomar: Oh, we also use it as a callback once:

```c
src/usermod.c:2261:		if (find_range (&add_sub_gids, find_new_sub_gids) == 0) {
```

But even there, it's passed addresses:

```
$ grepc find_range src/usermod.c | grep -e '^\S' -e find_fn
src/usermod.c:static int
find_range(struct id_range_list_entry **head,
           int (*find_fn)(id_t *range_start, unsigned long *range_count))
{
	if (find_fn(&range.first, &count) != 0)
}
```

alejandro-colomar: > I can drop it if you think it is unnecessary,

Yes, please.

> but it felt like a safe defensive check.

I don't like such defensive checks, as they make the code harder to understand: they make it as if a null pointer were a valid/possible input, which makes maintenance more difficult.

It prevents problems in the short term, but introduces more subtle problems in the long term.

jcpunk: Dropped.

lib/find_new_sub_gids.c
```diff
@@ -9,23 +9,168 @@
...
+	if (sub_gid_min > sub_gid_max || count == 0) {
+		fprintf(log_get_logfd(),
+		         _("%s: Invalid configuration: SUB_GID_MIN (%lu),"
+		           " SUB_GID_MAX (%lu), SUB_GID_COUNT (%lu)\n"),
+		         log_get_progname(),
+		         sub_gid_min, sub_gid_max, count);
+		return -1;
+	}
+
+	uid_offset = (unsigned long)uid - uid_min;
```
alejandro-colomar: This cast is unnecessary.  Usual arithmetic conversions turn everything into u_long anyway.

<https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf#subsubsection.6.3.1.8>

> if both operands have signed integer types or both have unsigned integer types, the operand with the type of lesser integer conversion rank is converted to the type of the operand with greater rank.

etc/login.defs
```diff
@@ -240,6 +240,37 @@ SUB_UID_MIN		   100000
...
+#
+# WARNING: SECURITY RISK - MAY CAUSE RANGE OVERLAPS AND PRIVILEGE ESCALATION!
+#
+# When enabled (WRAP MODE), the subordinate ID space is treated as a ring
+# buffer. Uses modulo arithmetic to handle overflow. May cause range
+# overlaps between users.
+#
+# Default: no
+#
+#UNSAFE_SUB_UID_DETERMINISTIC_WRAP	no
```
alejandro-colomar, line 276: Could you please add the unsafe mode in a separate commit?

lib/find_new_sub_gids.c
```diff
@@ -19,6 +19,119 @@
...
+	space = sub_gid_max - sub_gid_min + 1;
+
+	if (count > space) {
+		fprintf(log_get_logfd(),
+		         _("%s: Not enough space for any subordinate GIDs"
+		           " (SUB_GID_MIN=%lu, SUB_GID_MAX=%lu,"
+		           " SUB_GID_COUNT=%lu)\n"),
```
alejandro-colomar, line 106: I would simplify this to just the first line.  One can check the configuration to verify the values.

This would make the source code cleaner.

lib/find_new_sub_gids.c
```diff
@@ -19,6 +19,119 @@
...
+		         (uintmax_t)uid, uid_min);
+		return -1;
+	}
+
+	if (sub_gid_min > sub_gid_max || count == 0) {
+		fprintf(log_get_logfd(),
+		         _("%s: Invalid configuration: SUB_GID_MIN (%lu),"
+		           " SUB_GID_MAX (%lu), SUB_GID_COUNT (%lu)\n"),
+		         log_get_progname(),
+		         sub_gid_min, sub_gid_max, count);
```
alejandro-colomar, line 89: I'd simplify to `"%s: Invalid SUB_GID_* configuration.\n"`

lib/find_new_sub_gids.c
```diff
@@ -19,6 +19,119 @@
...
+	sub_gid_min = getdef_ulong ("SUB_GID_MIN", 65536UL);
+	sub_gid_max = getdef_ulong ("SUB_GID_MAX", 4294967295UL);
+	count = getdef_ulong ("SUB_GID_COUNT", 65536UL);
+
+	if (uid < uid_min) {
+		fprintf(log_get_logfd(),
+		         _("%s: UID %ju is less than UID_MIN %lu,"
+		           " cannot calculate deterministic subordinate GIDs\n"),
+		         log_get_progname(),
+		         (uintmax_t)uid, uid_min);
```
alejandro-colomar, line 80: I'd simplify to `"%s: UID < UID_MIN.\n"`.  The user could verify the values easily.

lib/find_new_sub_gids.c
```diff
@@ -19,6 +19,119 @@
...
+
+	id_t  end_id;
+	id_t  product;
+	id_t  start_id;
+
+	if (__builtin_mul_overflow(uid_offset, count, &product)) {
+		fprintf(log_get_logfd(),
+		         _("%s: Overflow calculating deterministic"
+		           " subordinate GID range for UID %ju\n"),
+		         log_get_progname(), (uintmax_t)uid);
```
alejandro-colomar: How about something like this? `"%s: %s: %s\n", log_get_progname(), __func__, strerror(EOVERFLOW));`

lib/find_new_sub_gids.c
alejandro-colomar: Maybe, I would even move the fprintf() calls to outside of this function, and just set errno here.

I'd set `errno = EDOM;` for this one, since the input is outside the valid domain of the function.

alejandro-colomar: Indeed.  Please pick the commits from <https://github.com/shadow-maint/shadow/pull/1586> for this PR.

etc/login.defs
jcpunk: I've put the implementation in its own commit with detailed reasoning.  Should I move the documentation out too?

alejandro-colomar: Yeah, let's separate them so that the commits are fully independent.  I'd like to review separately the documentation for the safe mode and the unsafe mode.

jcpunk: Broke out further into more parts.  The detailed justification remains on the commit that adds the code rather than the documentation.

lib/find_new_sub_gids.c
```diff
@@ -58,7 +62,21 @@ int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
...
+ * find_new_sub_gids - Find a new unused range of subordinate GIDs.
+ *
+ * Return 0 on success, -1 if no unused GIDs are available.
+ */
+int
+find_new_sub_gids(id_t *range_start, unsigned long *range_count)
+{
+	if (!range_start || !range_count)
+		return -1;
+
```
alejandro-colomar: This should still be removed.

hallyn:
FWIW I like this feature, thank you.

etc/login.defs
```diff
@@ -253,6 +266,19 @@ SUB_GID_MIN		   100000
...
+# When set to "yes", subordinate GID ranges are calculated using a
+# deterministic formula instead of searching for the next free range:
+#
+# WARNING: Do not mix deterministic and linear (default) allocation
+# on the same system or across systems sharing /etc/subgid. Mixing
+# methods will cause subordinate ID range conflicts and overlaps.
+#
+# Default: no
+#
+#SUB_GID_DETERMINISTIC	no
```
hallyn, line 306: I'm not sure it makes sense to have SUB_UID be linear and SUB_GID be deterministic.  Does it?  Is there a use case for that?

lib/find_new_sub_gids.c
jcpunk: As I work on adding in those commits, I've started getting:
```
find_new_sub_gids.c: In function 'find_new_sub_gids_linear':
find_new_sub_gids.c:185:19: warning: comparison of integer expressions of different signedness: 'id_t' {aka 'unsigned int'} and 'int' [-Wsign-compare]
  185 |         if (start == -1)
      |                   ^~
At top level:
cc1: note: unrecognized command-line option '-Wno-unknown-warning-option' may have been intended to silence earlier diagnostics
cc1: note: unrecognized command-line option '-Wno-unknown-attributes' may have been intended to silence earlier diagnostics
  CC       libshadow_la-find_new_sub_uids.lo
find_new_sub_uids.c: In function 'find_new_sub_uids_linear':
find_new_sub_uids.c:185:19: warning: comparison of integer expressions of different signedness: 'id_t' {aka 'unsigned int'} and 'int' [-Wsign-compare]
  185 |         if (start == -1)
```
Should I move `start` to type `int` instead of `id_t`?

lib/find_new_sub_gids.c
jcpunk: Fixed

lib/find_new_sub_gids.c
alejandro-colomar: No; this is a false positive from GCC.  Just ignore it.

Using `int` would be incorrect.

lib/find_new_sub_gids.c
```diff
@@ -20,6 +20,119 @@
...
+	sub_gid_min = getdef_ulong ("SUB_GID_MIN", 65536UL);
+	sub_gid_max = getdef_ulong ("SUB_GID_MAX", 4294967295UL);
+	count = getdef_ulong ("SUB_GID_COUNT", 65536UL);
+
+	if (uid < uid_min) {
+		fprintf(log_get_logfd(),
+		         _("%s: UID %ju is less than UID_MIN %lu,"
+		           " cannot calculate deterministic subordinate GIDs\n"),
+		         log_get_progname(),
+		         (uintmax_t)uid, uid_min);
```
alejandro-colomar, line 80: Please set errno without logging anything.  The caller already logs errors.

This should be EINVAL, I think.

lib/find_new_sub_gids.c
```diff
@@ -20,6 +20,119 @@
...
+	}
+
+	if (sub_gid_min > sub_gid_max || count == 0) {
+		fprintf(log_get_logfd(),
+		         _("%s: Invalid configuration: SUB_GID_MIN (%lu),"
+		           " SUB_GID_MAX (%lu), SUB_GID_COUNT (%lu)\n"),
+		         log_get_progname(),
+		         sub_gid_min, sub_gid_max, count);
+		return -1;
+	}
```
alejandro-colomar, line 91: ERANGE

lib/find_new_sub_gids.c
```diff
@@ -20,6 +20,119 @@
...
+
+	if (count > space) {
+		fprintf(log_get_logfd(),
+		         _("%s: Not enough space for any subordinate GIDs"
+		           " (SUB_GID_MIN=%lu, SUB_GID_MAX=%lu,"
+		           " SUB_GID_COUNT=%lu)\n"),
+		         log_get_progname(),
+		         sub_gid_min, sub_gid_max, count);
+		return -1;
+	}
```
alejandro-colomar, line 110: EUSERS?

lib/find_new_sub_gids.c
```diff
@@ -20,6 +20,119 @@
...
+	id_t  product;
+	id_t  start_id;
+
+	if (__builtin_mul_overflow(uid_offset, count, &product)) {
+		fprintf(log_get_logfd(),
+		         _("%s: Overflow calculating deterministic"
+		           " subordinate GID range for UID %ju\n"),
+		         log_get_progname(), (uintmax_t)uid);
+		return -1;
+	}
```
alejandro-colomar: EOVERFLOW?

etc/login.defs
jcpunk: Because the allocated ranges are so dependent on the SUB_{U,G}ID_{MIN,MAX}, I thought it would make sense to force folks to opt in to both modes. If you tune SUB_UID_COUNT but forget SUB_GID_COUNT things will get wonky... I don't object to one big flag, but the current code most closely mirrors the existing configuration options.

These days I find explicit for opt in to be helpful in long term maintenance of under documented hosts.

lib/find_new_sub_gids.c
```diff
@@ -91,6 +105,20 @@ find_new_sub_gids_deterministic(uid_t uid,
...
+	if (allow_wrap) {
+		/*
+		* UNSAFE_SUB_GID_DETERMINISTIC_WRAP MODE
+		*
+		* Promote to uintmax_t before multiplying to avoid truncation on
+		* 32-bit platforms where unsigned long is 32 bits.
+		*/
+		uintmax_t logical_offset = (uintmax_t)uid_offset * (uintmax_t)count;
+
+		*range_start = (id_t)(sub_gid_min + (unsigned long)(logical_offset % space));
```
alejandro-colomar: The last cast to id_t is superfluous.  Integer conversions have well-defined behavior in C.

lib/find_new_sub_gids.c
jcpunk: Set

lib/find_new_sub_gids.c
jcpunk: Set

lib/find_new_sub_gids.c
jcpunk: Set

lib/find_new_sub_gids.c
jcpunk: Set

lib/find_new_sub_gids.c
jcpunk: Removed

lib/find_new_sub_gids.c
```diff
@@ -20,6 +20,119 @@
...
+{
+	unsigned long  count;
+	unsigned long  space;
+	unsigned long  uid_min;
+	unsigned long  sub_gid_max;
+	unsigned long  sub_gid_min;
+	unsigned long  uid_offset;
+
+	assert (range_start != NULL);
+	assert (range_count != NULL);
```
alejandro-colomar: These assertions should be removed.

lib/find_new_sub_gids.c
```diff
@@ -20,6 +20,119 @@
...
+
+	uid_min = getdef_ulong ("UID_MIN", 1000UL);
+	sub_gid_min = getdef_ulong ("SUB_GID_MIN", 65536UL);
+	sub_gid_max = getdef_ulong ("SUB_GID_MAX", 4294967295UL);
+	count = getdef_ulong ("SUB_GID_COUNT", 65536UL);
```
alejandro-colomar, line 72: Should these be gotten in the upper API?  That would de-duplicate this code.

lib/find_new_sub_uids.c
```diff
@@ -20,6 +20,119 @@
...
+ *   All arithmetic overflow is a hard error.  The assigned range must fit
+ *   entirely within [SUB_UID_MIN, SUB_UID_MAX].  Allocation is monotonic
+ *   and guaranteed non-overlapping.
+ *
+ * Return 0 on success, -1 if no UIDs are available.
+ */
+static int
+find_new_sub_uids_deterministic(uid_t uid,
+				id_t *range_start,
+				unsigned long *range_count)
```
alejandro-colomar, line 57: Should/can we de-duplicate with something similar to find_free_range()?

lib/find_new_sub_gids.c
jcpunk: Done

lib/find_new_sub_gids.c
jcpunk: I was thinking keeping it self-contained made things cleaner.

I have no idea how the upper API usage would be done.

alejandro-colomar: By upper API I meant find_new_sub_gids(); the one we're splitting.

lib/find_new_sub_uids.c
jcpunk: Would that go into `subordinateio.c`?  Would that also mean renaming `find_free_range` to have "`linear`" in its name to distinguish it?

lib/find_new_sub_gids.c
alejandro-colomar: This still appears in the recently-pushed code.

lib/find_new_sub_uids.c
alejandro-colomar: Yup.

lib/find_new_sub_gids.c
jcpunk: Wouldn't that make the arg list for `find_new_sub_uids_*` a bit long?

alejandro-colomar: If we make them id_t, it'd shorten significantly.

lib/find_new_sub_uids.c
alejandro-colomar: Maybe it's not a good idea.  I'm just wondering.

jcpunk: I'd be tempted to do that in a separate PR.

alejandro-colomar: Sounds reasonable; let's keep it simple for now, even if that duplicates some code.

lib/find_new_sub_gids.c
```diff
@@ -34,9 +34,6 @@ int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
 	unsigned long count;
 	id_t start;
 
-	assert (range_start != NULL);
```
hallyn, line 37: The commit message says

```
The input to these functions is always an address (&x); that's
guaranteed to be non-null.
```

Currently that may be the case, but unless there is some enforcement to ensure that will always be the case, I think removing the asserts is wrong.

alejandro-colomar: We don't assert such stuff in most places; that would clutter the code, if we were going to assert nullability in each function.

alejandro-colomar: In this case, because of how the function works, it wouldn't make much sense to pass anything other than an address.  It would be unlikely to have a bug like that, I expect.

lib/find_new_sub_gids.c
```diff
@@ -44,10 +44,9 @@ int find_new_sub_gids (id_t *range_start, unsigned long *range_count)
 	}
 
 	start = sub_gid_find_free_range(min, max, count);
-	if (start == -1) {
-		errno = EUSERS;
+	if (start == -1)
```
hallyn, line 155: The patch description says 

```
lib/: find_free_range(): Set errno on error
```
But here and in lib/find_new_sub_uids.c you are actually removing the setting of errno.  Did you intend to do that, or was that accidentally pulled into this commit?

(If it is intentional, please explain it in the commit message)

alejandro-colomar: I'm removing it there, because I'm setting it deeper, in free_find_range() (as the commit says).  Thus, this is redundant and removed.

jcpunk:
I'm out of the office this week, if you've got something actionable for me I'll be a bit slow.... You should probably @ me so it floats to the top of my inbox when I return and isn't lost in the noise..

lib/find_new_sub_gids.c
hallyn: Ah, I was thinking the the call order was reverse.

lib/find_new_sub_gids.c
hallyn: Well, if it is NULL, it'll lead to a crash, not arbitrary memory write, so I don't care all that much.

hallyn:
What is the status of this?  I'd like to see the main feature merged...

alejandro-colomar:
> What is the status of this? I'd like to see the main feature merged...

@hallyn 

I think the patches look okay.

I didn't want to OK the patch set, because I don't know these features enough.  But it looks decently reasonable.

I didn't check that the algorithms make sense.  You should check that.

Then there's also the discussion about the design of the configuration variables.  That's up to you too.

The source code looks reasonable, which is the only thing I reviewed deeply.

jcpunk:
Just double checking that nothing is waiting on me. I was out of office and kinda lost track.

alejandro-colomar:
> Just double checking that nothing is waiting on me. I was out of office and kinda lost track.

That's correct.  I'd like @hallyn and @ikerexxe to review this.

jcpunk:
Rebased off `HEAD` to clear merge conflicts.

man/login.defs.d/SUB_GID_COUNT.xml
```diff
@@ -22,3 +22,128 @@
...
+    </para>
+    <para>
+      If <option>SUB_GID_DETERMINISTIC</option> is enabled,
+      you can use
+      <command>usermod --add-subgids -S</command>
+      to produce deterministic subgids.
+    </para>
+    <para>
+      <emphasis role="bold">WARNING</emphasis>:
+      Because <option>UID_MIN</option> is used to calculate the ranges,
```
hallyn, line 60: So is SUB_GID_COUNT.  But not UID_MAX (which is listed below), iiuc.

jcpunk: I added SUB_{U,G}ID_COUNT to the documentation just now. Does that look better?

hallyn: Yes, thanks, but still why is UID_MAX listed?  It's not part of the calculation, is it?

lib/find_new_sub_uids.c
```diff
@@ -87,6 +101,20 @@ find_new_sub_uids_deterministic(uid_t uid,
...
+	if (allow_wrap) {
+		/*
+		 * UNSAFE_SUB_UID_DETERMINISTIC_WRAP MODE
+		 *
+		 * Promote to uintmax_t before multiplying to avoid truncation on
+		 * 32-bit platforms where unsigned long is 32 bits.
+		 */
+		uintmax_t logical_offset = (uintmax_t)uid_offset * (uintmax_t)count;
+
+		*range_start = (id_t)(sub_uid_min + (unsigned long)(logical_offset % space));
```
hallyn: Does this ensure that range_start + range_count will fit?  I'm not sure it does in all cases.

It seems like it would be safer to take (sub_uid_max - sub_uid_min) / count as the number of ranges which fit (call it numfit), and then using uid_offset %= numfit in the regular calculation.

I might not be thinking right, will take another look tonight if I have time.

lib/find_new_sub_gids.c
```diff
@@ -17,9 +17,118 @@
...
+	unsigned long  sub_gid_max;
+	unsigned long  sub_gid_min;
+	unsigned long  uid_offset;
+
+	uid_min = getdef_ulong ("UID_MIN", 1000UL);
+	sub_gid_min = getdef_ulong ("SUB_GID_MIN", 65536UL);
+	sub_gid_max = getdef_ulong ("SUB_GID_MAX", 4294967295UL);
+	count = getdef_ulong ("SUB_GID_COUNT", 65536UL);
+
+	if (uid < uid_min) {
```
hallyn, line 75: But so this is a pretty harsh limitation - if UID_MIN is 1000, and you create a user with uid 999, you can't give it subuids?

Note that right now useradd -f -u 999 will in fact assign an automatic subuid range.

lib/find_new_sub_uids.c
hallyn:
```
serge@sl25:~/test$ cat range.c
#include <stdio.h>
#include <unistd.h>

int main() {
	unsigned long  count;
	unsigned long  space;
	unsigned long  uid_min;
	unsigned long  uid_offset;
	unsigned long  sub_uid_max;
	unsigned long  sub_uid_min;
	uid_t uid;
	int i, spaces;
	uid_t uid_tests[] = { 1000, 1001, 65536, 100000, 0 };

	uid_min =  1000UL;
	sub_uid_min =  65536UL;
	sub_uid_max =  4294967295UL;
	count =  65536UL;

	space = sub_uid_max - sub_uid_min + 1;
	spaces = space / count;
	printf("There are %d spaces of size %lu in %lu\n", spaces, count, space);
	for (i = 0; uid_tests[i] != 0; i++) {
		uid = uid_tests[i];
		uid_offset = uid - uid_min;
		printf("uid_offset is %lu for uid %d\n", uid_offset, uid);
		printf("and so uid %d wraps to %lu\n", uid, uid_offset % spaces);
	}
}
serge@sl25:~/test$ gcc -o range range.c
serge@sl25:~/test$ ./range
There are 65535 spaces of size 65536 in 4294901760
uid_offset is 0 for uid 1000
and so uid 1000 wraps to 0
uid_offset is 1 for uid 1001
and so uid 1001 wraps to 1
uid_offset is 64536 for uid 65536
and so uid 65536 wraps to 64536
uid_offset is 99000 for uid 100000
and so uid 100000 wraps to 33465
```

man/login.defs.d/SUB_GID_COUNT.xml
jcpunk: It was my intent to use UID_MAX as a way of ensuring the ranges are bounded.  Since `uid_t` is `uint_32` right now on Linux there is a risk of wrapping. I was expecting that the upper limit on automatic allocation would be a helpful knob to have on hand.

lib/find_new_sub_gids.c
jcpunk: > if UID_MIN is 1000, and you create a user with uid 999, you can't give it subuids?

Close, with uid less than UID_MIN you can't give them subids automatically with this deterministic algorithm. You could manually assign a block of ids.

hallyn:
On Tue, Mar 31, 2026 at 08:43:29PM -0700, Pat Riehecky wrote:
> @jcpunk commented on this pull request.
> 
> 
> 
> > +				id_t *range_start,
> +				unsigned long *range_count)
> +{
> +	unsigned long  count;
> +	unsigned long  space;
> +	unsigned long  uid_min;
> +	unsigned long  sub_gid_max;
> +	unsigned long  sub_gid_min;
> +	unsigned long  uid_offset;
> +
> +	uid_min = getdef_ulong ("UID_MIN", 1000UL);
> +	sub_gid_min = getdef_ulong ("SUB_GID_MIN", 65536UL);
> +	sub_gid_max = getdef_ulong ("SUB_GID_MAX", 4294967295UL);
> +	count = getdef_ulong ("SUB_GID_COUNT", 65536UL);
> +
> +	if (uid < uid_min) {
> 
> > if UID_MIN is 1000, and you create a user with uid 999, you can't give it subuids?
> 
> Close, with uid less than UID_MIN you can't give them subids automatically with this deterministic algorithm. You could manually assign a block of ids.

Ok, please make sure that's well documented in the login.defs manpage.

hallyn:
On Tue, Mar 31, 2026 at 08:42:03PM -0700, Pat Riehecky wrote:
> @jcpunk commented on this pull request.
> 
> 
> 
> > +      This ensures the same UID always receives the same subordinate GID
> +      range on every system,
> +      making it suitable for environments with centralized user management
> +      (LDAP, NIS, etc.)
> +      or
> +      synchronized UIDs across systems.
> +    </para>
> +    <para>
> +      If <option>SUB_GID_DETERMINISTIC</option> is enabled,
> +      you can use
> +      <command>usermod --add-subgids -S</command>
> +      to produce deterministic subgids.
> +    </para>
> +    <para>
> +      <emphasis role="bold">WARNING</emphasis>:
> +      Because <option>UID_MIN</option> is used to calculate the ranges,
> 
> It was my intent to use UID_MAX as a way of ensuring the ranges are bounded.  Since `uid_t` is `uint_32` right now on Linux there is a risk of wrapping. I was expecting that the upper limit on automatic allocation would be a helpful knob to have on hand.

But you're not.  UID_MAX doesn't show up in the code.

jcpunk:
> But you're not.  UID_MAX doesn't show up in the code.

Blast, that must have gotten dropped by mistake when I was moving things around.  It probably isn't a useful control anyway.  I'll get that removed from the doc.

jcpunk:
> Ok, please make sure that's well documented in the login.defs manpage.

I've got:
```
      Users with identities less than <option>UID_MIN</option>
      are incompatible with <option>SUB_GID_DETERMINISTIC</option>,
      but can still be set manually.
```
In the man page now.  Should this be changed?

lib/find_new_sub_uids.c
jcpunk: I've just switched to the slots/spaces algorithm you've demonstrated here.

hallyn:
> > Ok, please make sure that's well documented in the login.defs manpage.
> 
> I've got:
> 
> ```
>       Users with identities less than <option>UID_MIN</option>
>       are incompatible with <option>SUB_GID_DETERMINISTIC</option>,
>       but can still be set manually.
> ```
> 
> In the man page now. Should this be changed?

Sorry, I didn't see that - that's good, thanks.

hallyn:
Thanks, I'll take one last look later today.

@alejandro-colomar since the logic has changed, probably best that you take another look and make sure we didn't just aim ourselves at another pitfall.

hallyn:
@jcpunk I think something went wrong in your rebase.  allow_wrap gets set, but never checked.  The wrapping code is missing.

jcpunk:
The wrap code somehow got into the deterministic initial code commit.  I've got it in the wrap mode commit now.  I'm not sure why the spacing looks all wonky on github.

lib/find_new_sub_gids.c
```diff
@@ -17,51 +17,163 @@
...
+			fprintf(log_get_logfd(),
+	         		_("%s: Deterministic subordinate GID range"
+				  " for UID %ju exceeds SUB_GID_MAX (%lu)\n"),
+	         		log_get_progname(),
+	         		(uintmax_t)uid, sub_gid_max);
+			return -1;
+		}
+	}
+
+	*range_start = (id_t)(sub_gid_min + slot * count);
```
alejandro-colomar: This case is superfluous.

lib/find_new_sub_uids.c
```diff
@@ -17,51 +17,163 @@
...
+			fprintf(log_get_logfd(),
+				_("%s: Deterministic subordinate UID range"
+				  " for UID %ju exceeds SUB_UID_MAX (%lu)\n"),
+				log_get_progname(),
+				(uintmax_t)uid, sub_uid_max);
+			return -1;
+		}
+	}
+
+	*range_start = (id_t)(sub_uid_min + slot * count);
```
alejandro-colomar: Superfluous cast.

alejandro-colomar:
> Thanks, looks good to me!
> 
> The only thing is, you add new printfs using strerror(), which #1381 replaces. I don't care whether we merge that first and you update to use SYSLOGEC, or this is merged first and @alejandro-colomar updates 1381.

I can update my stuff.  Don't worry.  :)

lib/find_new_sub_gids.c
alejandro-colomar: cast*

jcpunk: Fixed

lib/find_new_sub_uids.c
jcpunk: Fixed

alejandro-colomar:
@hallyn , I'm okay with it.  Feel free to merge.

PR #1387, opened 2025-11-06
Some improvements to stpecpy(), strtcpy(), stprintf(), and seprintf()

alejandro-colomar:
This is a subset of

- <https://github.com/shadow-maint/shadow/pull/1148> v15b.

I left out two commits that might be less clear.

Cc: @kees

---

Revisions:

<details>
<summary>v2</summary>

-  Improve comments

```
$ git rd 
 1:  156d72c80 =  1:  156d72c80 lib/subordinateio.c: get_owner_id(): Use snprintf_() instead of its pattern
 2:  176c04ac1 =  2:  176c04ac1 lib/string/: vsnprintf_(), strtcpy(): Set errno = E2BIG on truncation
 3:  992e21e81 =  3:  992e21e81 lib/string/sprintf/: snprintf_(): Use int instead of size_t in $2
 4:  4326914c7 =  4:  4326914c7 lib/string/: Redesign stpecpy() and stpeprintf()
 5:  0069f6492 !  5:  02145224a lib/string/: Compact documentation
    @@ Metadata
      ## Commit message ##
         lib/string/: Compact documentation
     
    +    In the case of [v]snprintf_(), anticipate the rename in the next commit.
    +
         Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
    + ## lib/string/sprintf/snprintf.h ##
    +@@
    + )
    + 
    + 
    ++// stprintf - string truncate print formatted
    + format_attr(printf, 3, 4)
    + inline int snprintf_(char *restrict s, int size, const char *restrict fmt, ...);
    ++// vstprintf - va_list string truncate print formatted
    + format_attr(printf, 3, 0)
    + inline int vsnprintf_(char *restrict s, int size, const char *restrict fmt,
    +     va_list ap);
    +
      ## lib/string/sprintf/stpeprintf.h ##
     @@
      
 6:  3edd6f0f4 !  6:  ee0002488 lib/, src/, tests/: Rename snprintf_() => stprintf()
    @@ lib/string/sprintf/snprintf.h => lib/string/sprintf/stprintf.h
      )
      
      
    + // stprintf - string truncate print formatted
      format_attr(printf, 3, 4)
     -inline int snprintf_(char *restrict s, int size, const char *restrict fmt, ...);
     +inline int stprintf(char *restrict s, int size, const char *restrict fmt, ...);
    + // vstprintf - va_list string truncate print formatted
      format_attr(printf, 3, 0)
[...]

PR #1573, opened 2026-03-10
Permit Subid storage by username or uid

jcpunk:
This PR obsoletes https://github.com/shadow-maint/shadow/pull/1570

The actual code changes are broken into smaller commits.

Closes: https://github.com/shadow-maint/shadow/issues/1554

I'm not sure the tests are well setup or have sufficient coverage.


lib/subordinateio.c
```diff
@@ -336,6 +336,49 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
 		return strcmp(range1->owner, range2->owner);
 }
 
+/*
+ * subid_owner_match: test whether two owner fields refer to the same user.
```
alejandro-colomar: I guess 'fields' means 'nsswitch.conf(5) fields'?  Please clarify.

lib/subordinateio.c
```diff
@@ -336,6 +336,49 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
+
+	if (strisdigit(a)) {
+		if (str2ui(&id_a, a) == -1)
+			return false;
+	} else {
+		const struct passwd *pw = getpwnam(a);
+		if (NULL == pw)
+			return false;
+		id_a = pw->pw_uid;
+	}
```
alejandro-colomar: We don't need to call `strisdigit()`.  If `str2ui()` fails, we can assume it's not a digit.

```c
	if (str2ui(&id_a, a) == -1) {
		struct passwd  *pw;

		pw = getpwnam(a);
		if (NULL == pw)
			return false;
		id_a = pw->pw_uid;
	}
```

lib/subordinateio.c
```diff
@@ -336,6 +336,49 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
+	uid_t id_a, id_b;
+
+	if (streq(a, b))
+		return true;
+
+	if (strisdigit(a)) {
+		if (str2ui(&id_a, a) == -1)
+			return false;
+	} else {
+		const struct passwd *pw = getpwnam(a);
```
alejandro-colomar: Please declare variables separately from their use.

See <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c#n101>.

lib/subordinateio.c
```diff
@@ -336,6 +336,49 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
+ *
+ * Return false if no match or either string cannot be resolved.
+ */
+static bool
+subid_owner_match(const char *a, const char *b)
+{
+	uid_t id_a, id_b;
+
+	if (streq(a, b))
+		return true;
```
alejandro-colomar: Should we return true without any validation?  `subid_owner_match("nonexistent", "nonexistent")`.

alejandro-colomar: I think it's better to remove the special case, and let the generic code below handle such strings.

lib/subordinateio.c
```diff
@@ -336,6 +336,49 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
+ * An exact string match has a quick return. Otherwise both strings
+ * are resolved to uid_t via passwd and compared numerically.
+ * This should account for overlaping UIDs.
+ *
+ * Return false if no match or either string cannot be resolved.
+ */
+static bool
+subid_owner_match(const char *a, const char *b)
+{
+	uid_t id_a, id_b;
```
alejandro-colomar: These would be better as uid_a and uid_b, I think.

lib/subordinateio.c
```diff
@@ -155,7 +155,7 @@ static bool range_exists(struct commonio_db *db, const char *owner)
 	const struct subordinate_range *range;
 	commonio_rewind(db);
 	while (NULL != (range = commonio_next(db))) {
-		if (streq(range->owner, owner))
+		if (subid_owner_match(range->owner, owner))
```
alejandro-colomar: Considering that this is a behavior change, I'd like the commit message to talk a bit about this.  It should justify why this expanded behavior is okay.

```
subid: Start using subid_owner_match in the easy locations

Signed-off-by: Pat Riehecky <...>
```

That is too short.

lib/subordinateio.c
```diff
@@ -914,15 +914,12 @@ static bool get_owner_id(const char *owner, enum subid_type id_type, char *id)
  */
 int list_owner_ranges(const char *owner, enum subid_type id_type, struct subid_range **in_ranges)
 {
-	// TODO - need to handle owner being either uid or username
```
alejandro-colomar, line 874: Please, add a `Fixes:` tag that refers to the commit that added this TODO.

```
Fixes: 0a7888b1 (2020-06-07; "Create a new libsubid")
```

alejandro-colomar: And also CC it's author, which is @hallyn

lib/subordinateio.c
```diff
@@ -211,60 +211,17 @@ static const struct subordinate_range *find_range(struct commonio_db *db,
...
-		range_owner_pwd = getpwnam(range->owner);
-		if (NULL == range_owner_pwd) {
+		if (!subid_owner_match(range->owner, owner))
 			continue;
-		}
 
-		if (owner_uid == range_owner_pwd->pw_uid) {
+		if ((val >= first) && (val <= last))
 			return range;
-		}
```
alejandro-colomar, line 267: I think the third commit would be better split in two.  First, move code around, and then use subid_owner_match().  I find it difficult to review as a single commit.

lib/subordinateio.c
```diff
@@ -651,7 +651,36 @@ int sub_uid_add (const char *owner, uid_t start, unsigned long count)
...
+		} else {
+			const struct passwd *pw = getpwnam(owner);
+			if (NULL == pw) {
+				errno = ENOENT;
+				return 0;
+			}
+			owner_uid = pw->pw_uid;
+		}
+
+		ret = snprintf(uid_string, ID_SIZE, "%u", owner_uid);
```
alejandro-colomar: snprintf(3) is not allowed in this project.  Please use stprintf_a().

See <lib/string/README>.

lib/subordinateio.c
```diff
@@ -651,7 +651,36 @@ int sub_uid_add (const char *owner, uid_t start, unsigned long count)
...
+		if (ret < 0 || ret >= ID_SIZE) {
+			errno = EINVAL;
+			return 0;
+		}
+
+		return add_range (&subordinate_uid_db, uid_string, start, count);
```
alejandro-colomar: Please don't use white space between function name and `(` in new code.

lib/subordinateio.c
jcpunk: I was thinking this was "similar" behavior to `find_range` 

```C
        /* 
         * Search for exact username/group specification
         *              
         * This is the original method - go fast through the db, doing only
         * exact username/group string comparison. Therefore we leave it as-is
         * for the time being, in order to keep it equally fast as it was
         * before.
         */
        commonio_rewind(db);
                unsigned long first = range->start;
                unsigned long last = first + range->count - 1;

                if (!streq(range->owner, owner))
                        continue;

                if ((val >= first) && (val <= last))
                        return range;
        }
```

Since I'm adding two `getpwnam` calls, I thought a quick return path might be helpful.

I can take out my check easily enough. Should I take out the one in `find_range`?  I don't know enough about the API/ABI guarantees to really guess at the meaning of:
```C
We only do special handling for these two files
```

lib/subordinateio.c
jcpunk: I've written a much more detailed commit message and pushed it just now.

lib/subordinateio.c
jcpunk: I have reworded this to help indicate that this is comparing what is passed in and what resolution tools are used.

Is that better?

lib/subordinateio.c
jcpunk: Fixed.

lib/subordinateio.c
jcpunk: Fixed

lib/subordinateio.c
jcpunk: Fixed

lib/subordinateio.c
alejandro-colomar: Having branches is dangerous.  The less branches one function has, the more each branch will be executed (and thus tested).

lib/subordinateio.c
jcpunk: To CC the author, do I need to put some sort of tag in the commit message or does adding the Fixes plus your ping there do the job?

I've got the Fixes line added.

lib/subordinateio.c
alejandro-colomar: You could keep it, mentioning that it's copying existing behavior, and point to the place where the existing streq(3) calls are being done.

If you remove it, you should also explain why.  In any case, whatever we do should be documented in the commit message.

lib/subordinateio.c
```diff
@@ -336,6 +336,48 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
+ * are resolved to uid_t via getpwnam and compared numerically.
+ * This should account for users with overlaping UIDs.
+ *
+ * Return false if no match or either string cannot be resolved.
+ */
+static bool
+subid_owner_match(const char *a, const char *b)
+{
+	uid_t id_a;
+	uid_t id_b;
```
alejandro-colomar: See <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c#n109>.

```

		int     i;
		char    c;
		char    *p;
		size_t  size;
```

lib/subordinateio.c
jcpunk: I'm showing the third commit as https://github.com/shadow-maint/shadow/pull/1573/changes/c1315a030d4349129e5cd3e49c17e1c104df2842 which primarily removes lines and only adds in two.

Are you asking for one commit the removes the entire secondary checking code and then another that adds in the new secondary checks?

lib/subordinateio.c
alejandro-colomar: You could add

```
```

lib/subordinateio.c
jcpunk: I'll make the change, but I'm showing a few other `snprintf` calls in this file I didn't add.  I have no idea if they are `snprintf(3)` or something else..

lib/subordinateio.c
```diff
@@ -950,13 +947,9 @@ int list_owner_ranges(const char *owner, enum subid_type id_type, struct subid_r
 		return -1;
 	}
 
-	have_owner_id = get_owner_id(owner, id_type, id);
```
alejandro-colomar, line 911: You have removed the last call to get_owner_id().  You could remove that function.

lib/subordinateio.c
alejandro-colomar: I would need to understand the second added line to be able to answer that question.  It seems like a movement of line 235, with some transformation.  Is that something that could be moved separately?  Or could you explain why the movement and transformation was done?

lib/subordinateio.c
```diff
@@ -650,7 +650,36 @@ int sub_uid_add (const char *owner, uid_t start, unsigned long count)
...
+				errno = ENOENT;
+				return 0;
+			}
+			owner_uid = pw->pw_uid;
+		}
+
+		if (stprintf_a(uid_string, ID_SIZE, owner_uid) == -1) {
+			errno = EINVAL;
+			return 0;
+		}
```
alejandro-colomar: This doesn't seem correct.  What happened to the format string?

lib/subordinateio.c
```diff
@@ -650,7 +650,36 @@ int sub_uid_add (const char *owner, uid_t start, unsigned long count)
...
+		} else {
+			const struct passwd *pw;
+
+			pw = getpwnam(owner);
+			if (NULL == pw) {
+				errno = ENOENT;
+				return 0;
+			}
+			owner_uid = pw->pw_uid;
+		}
```
alejandro-colomar: I think we should get rid of strisdigit() here too.

lib/subordinateio.c
```diff
@@ -336,6 +336,48 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
 		return strcmp(range1->owner, range2->owner);
 }
 
+/*
+ * subid_owner_match: test whether two values refer to the same user.
```
alejandro-colomar: What are `values`?  I guess you mean two `subids`?

alejandro-colomar: Sorry, two `subid owners`

lib/subordinateio.c
```diff
@@ -336,6 +336,48 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
+ *
+ * @a: first owner string (username or numeric UID)
+ * @b: second owner string (username or numeric UID)
```
alejandro-colomar: I'd remove `string` from here, as it doesn't provide much value.

lib/subordinateio.c
alejandro-colomar: Yes, I'm removing them in a different PR.

-  <https://github.com/shadow-maint/shadow/pull/1387>

lib/subordinateio.c
jcpunk: I had in my head a specific type of flow through, but if the current code isn't clear I'll try reusing the parts from the existing layout a bit more.  What I tried to do looks weird in diff...

That will come with some rebasing so my link above will probably go stale.  But hopefully the code will make more sense.

lib/subordinateio.c
jcpunk: I dropped my check, added what I think is a detailed comment above the function explaining why that check is not present.

Since the check itself was unique behavior to my commit, I'm not sure what to document.

lib/subordinateio.c
jcpunk: I didn't see a clear name in subuid(5) for what all to call the unix uid that has permission to use specific subuids/subgids.

This function is intended to take whatever "column 1 when separated by a `:` happens to be" be it a username or a uid.

lib/subordinateio.c
jcpunk: Added

lib/subordinateio.c
jcpunk: Fixed

lib/subordinateio.c
jcpunk: Sorry, that was me in a hurry not checking my local build.... I'll try to do better.

lib/subordinateio.c
jcpunk: Added commit 7 which drops the function.

lib/subordinateio.c
jcpunk: My eyes say I've added spaces and sorted things. But I may have missed some.

lib/subordinateio.c
jcpunk: Done

lib/subordinateio.c
jcpunk: Done

lib/subordinateio.c
alejandro-colomar: How about this?:

```
test whether two owner fields of subuid(5) refer to the same user.
```

or maybe 

```
test whether two owne (1st) fields of subuid(5) refer to the same user.
```

jcpunk: I'm having a hard time thinking of the right words.

This function takes as its arguments two strings which must either be valid usernames or valid UID.  Or a mix.

It returns true if they resolve to the same UID via getpwnam(3).  The function itself doesn't actually have any insight into what is passed in.

Since one of the arguments is intended to be "the user in question" and the other is sourced from "column 1 when split by `:` of /etc/sub{u,g}id" I'm genuinely at a loss for how to describe them...

The way it is currently called, argument 2 is the "owner" of the subid and argument 1 is just there to see if they match.

alejandro-colomar: Hmmm, maybe this?

```
test whether two users (either username or UID) are the same.
```

jcpunk: I broke down and asked an LLM and it offered:
```

/*
 * subid_owner_match: test whether two user identifiers refer to the same user.
 *
 * @a: user identifier (username or numeric UID string)
 * @b: user identifier (username or numeric UID string)
 *
 * Each identifier is resolved to uid_t via getpwnam(3) and compared
 * numerically. Accepts usernames, numeric UIDs, or a mix of both.
 *
 * Numeric resolution ensures stale entries for deleted users cannot
 * produce false matches, and handles systems with overlapping UIDs.
 *
 * We do not perform a streq(3) on the passed strings themselves.
 * In this way we ensure the user is resolvable on the system
 * eliminating the ability to return results from deleted users
 * with stale entries.
 *
 * Returns false if either identifier cannot be resolved, or if the
 * resolved UIDs do not match.
 */

File: lib/subordinateio.c (hunk 8 above)
Comment (jcpunk) [reply lib/subordinateio.c:267]: I think the diff looks more like what the code is supposed to achieve now.

File: lib/subordinateio.c (hunk 15 above)
Comment (alejandro-colomar) [reply lib/subordinateio.c]: > I broke down and asked an LLM and it offered:

My opinion on AI is: <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/ai>

We haven't merged yet here, but we will disallow using AI to contribute to the project:
-  <https://github.com/shadow-maint/shadow/pull/1453>

> 
> ```
> 
> /*
>  * subid_owner_match: test whether two user identifiers refer to the same user.

I don't like `user identifiers`, because that's exactly what UID means.

But we could say

```
test whether two strings identify the same user.
```

File: lib/subordinateio.c (hunk 8 above)
Comment (alejandro-colomar) [reply lib/subordinateio.c:267]: Yup, the diff makes more sense now.  Thanks!

BTW, you're free to reorder that to look like you wanted.  It's just that if you do that, it should be in a separate commit so that it's easier to review.

File: lib/subordinateio.c [hunk 17]
```diff
@@ -236,33 +222,7 @@ static const struct subordinate_range *find_range(struct commonio_db *db,
...
-		 * UID or as a literal username.
-		 *
-		 * If specified as another UID, the call to getpwnam()
-		 * will return NULL.
-		 *
-		 * If specified as literal username, we will get its
-		 * UID and compare that to UID we are looking for.
-		 */
-		const struct passwd *range_owner_pwd;
-
```
alejandro-colomar, line 259: For commit 3, which is ea150df4 ("subid: use subid_owner_match() for owner resolution in find_range"):

```
```

Thanks!

lib/subordinateio.c
```diff
@@ -958,13 +955,9 @@ int list_owner_ranges(const char *owner, enum subid_type id_type, struct subid_r
 		return -1;
 	}
 
-	have_owner_id = get_owner_id(owner, id_type, id);
-
```
alejandro-colomar, line 912: For commit 2, which is 2cf6c63b ("subid: use subid_owner_match() for owner resolution in range queries"):

```
```

lib/subordinateio.c
```diff
@@ -336,6 +336,56 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
+ * resolved UIDs do not match.
+ */
+static bool
+subid_owner_match(const char *a, const char *b)
+{
+	uid_t  id_a;
+	uid_t  id_b;
+
+	if (streq(a, b))
+		return true;
```
alejandro-colomar: Did you forget to remove the streq(3) call?  It doesn't match the comment.

lib/subordinateio.c
```diff
@@ -336,6 +336,56 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
+ * Each identifier is resolved to uid_t via getpwnam(3) and compared
+ * numerically. Accepts usernames, numeric UIDs, or a mix of both.
+ *
+ * Numeric resolution ensures stale entries for deleted users cannot
+ * produce false matches, and handles systems with overlapping UIDs.
+ *
+ * We do not perform a streq(3) on the passed strings themselves.
+ * In this way we ensure the user is resolvable on the system
+ * eliminating the ability to return results from deleted users
+ * with stale entries.
```
alejandro-colomar: I hope this is something we can do.  It sounds good, but we never know when something will be a regression for someone else.  :)

I'm okay with it, FWIW.  If anyone reports a regression, we can take care of it, and re-introduce the streq(3) call.

lib/subordinateio.c
```diff
@@ -336,6 +336,56 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
+	}
+
+	if (str2ui(&id_b, b) == -1) {
+		const struct passwd  *pw;
+
+		pw = getpwnam(b);
+		if (NULL == pw)
+			return false;
+		id_b = pw->pw_uid;
+	}
```
alejandro-colomar: If we remove the streq(3) call, it might be interesting to also call getpwuid(3) if we get a number, to be 100% sure that the user exists.

We could add

```c
const struct passwd *
getpw_uid_or_nam(const char *u)
{
	uid_t  uid;

	return str2i(uid_t, &uid, u) == 0 ? getpwuid(uid) : getpwnam(u);
}
```

in a new file pair `lib/shadow/passwd/getpw.c` (and `.h`).

And then call that here:

```c
static bool
subid_owner_match(const char *a, const char *b)
{
	const struct passwd  *pw_a, *pw_b;

	pw_a = getpw_uid_or_nam(a);
	if (pw_a == NULL)
		return false;
	pw_b = getpw_uid_or_nam(b);
	if (pw_b == NULL)
		return false;

	return pw_a->pw_uid == pw_b->pw_uid;
}
```

ikerexxe:
Just to make sure I understand the goal of this PR correctly. You'd like to store the user/group ID instead of the username/groupname in `/etc/sub*id`.  File example: `1000:165536:165536` instead of `jcpunk:165536:165536`. Is my understanding correct?

lib/subordinateio.c
jcpunk: I'll make the change to you're proposed string

lib/subordinateio.c
jcpunk: Since I've updated commit 1 with the new doc string, the hash of commit 3 changed.  It should be an identical patch, but I want to be sure I'm adding the reviewed by tag to a hash you've approved.

lib/subordinateio.c
jcpunk: Since I've updated commit 1 with the new doc string, the hash of commit 2 changed.  It should be an identical patch, but I want to be sure I'm adding the reviewed by tag to a hash you've approved.

lib/subordinateio.c
jcpunk: Is that something you'd like to see in this PR, or can it be deferred to later work?

jcpunk:
> Just to make sure I understand the goal of this PR correctly. You'd like to store the user/group ID instead of the username/groupname in `/etc/sub*id`. File example: `1000:165536:165536` instead of `jcpunk:165536:165536`. Is my understanding correct?

That is my intention.

Full disclosure, I've not be very successful at running the test suite.

lib/subordinateio.c
alejandro-colomar: As long as git-range-diff(1) says the patch is identical, yes, please add it.

The commit hashes change every time you rebase, even if the only thing that changes is the commit date.  :)

lib/subordinateio.c
alejandro-colomar: It's fine for later.

lib/subordinateio.c
jcpunk: The underlying logic is about the same, so I'll leave it as is.  Clean diffs with fewer commits is fine with me.

lib/subordinateio.c
jcpunk: Added Reviewed-by to "`subid: use subid_owner_match() for owner resolution in find_range`"

lib/subordinateio.c
jcpunk: Added Reviewed-by to "`subid: use subid_owner_match() for owner resolution in range queries`"

lib/subordinateio.c
```diff
@@ -142,6 +142,53 @@ static struct commonio_ops subordinate_ops = {
...
+ *
+ * We do not perform a streq(3) on the passed strings themselves.
+ * In this way we ensure the user is resolvable on the system
+ * eliminating the ability to return results from deleted users
+ * with stale entries.
+ *
+ * Returns false if either identifier cannot be resolved, or if the
+ * resolved UIDs do not match.
+ */
+static bool subid_owner_match(const char *a, const char *b)
```
alejandro-colomar: After tinkering with getpw_uid_or_nam(), I've been thinking that this function is really more generic than it looks.

Maybe we should call it something like `is_same_user(a,b)`?

It's not really constrained to subids at all.  This only considers users.

lib/subordinateio.c
```diff
@@ -142,6 +142,53 @@ static struct commonio_ops subordinate_ops = {
...
+{
+	uid_t  id_a;
+	uid_t  id_b;
+
+	if (str2ui(&id_a, a) == -1) {
```
alejandro-colomar: Oh, I just remembered we have an API for parsing uids specifically:

```c
$ grepc get_uid lib
lib/atoi/getnum.c:extern inline int get_uid(const char *restrict uidstr, uid_t *restrict uid);
lib/atoi/getnum.h:ATTR_STRING(1) ATTR_ACCESS(write_only, 2)
inline int get_uid(const char *restrict uidstr, uid_t *restrict uid);
lib/atoi/getnum.h:inline int
get_uid(const char *restrict uidstr, uid_t *restrict uid)
{
	return a2i(uid_t, uid, uidstr, NULL, 10, type_min(uid_t), type_max(uid_t));
}
```

We should call it here, as:

```c
if (get_uid(a, &id_a) == -1) {
```

alejandro-colomar: This will work portably, independent of what uid_t actually is.

lib/subordinateio.c
```diff
@@ -658,7 +658,29 @@ int sub_uid_add (const char *owner, uid_t start, unsigned long count)
...
+		if (str2ui(&owner_uid, owner) == -1) {
+			const struct passwd  *pw;
+
+			pw = getpwnam(owner);
+			if (NULL == pw) {
+				errno = ENOENT;
+				return 0;
+			}
+			owner_uid = pw->pw_uid;
+		}
```
alejandro-colomar: Hmmmm, this seems to be similar to the getpw_uid_or_nam() I mentioned.  Maybe we should add it already, to simplify here?

alejandro-colomar:
```c
const struct passwd  *pw;

pw = getpw_uid_or_nam(owner);
if (NULL == pw) {
	errno = ENOENT;  // I think this should actually be set within getpw_uid_or_nam().
	return 0;
}
owner_uid = pw->pw_uid;
```

lib/subordinateio.c
```diff
@@ -870,40 +870,6 @@ gid_t sub_gid_find_free_range(gid_t min, gid_t max, unsigned long count)
 	return start == ULONG_MAX ? (gid_t) -1 : start;
 }
 
-static bool get_owner_id(const char *owner, enum subid_type id_type, char *id)
```
alejandro-colomar, line 822: For "subid: Remove get_owner_id":

```
```

lib/subordinateio.c
jcpunk: Renaming to `is_same_user` makes a lot of sense.

For now I'd like to keep it in this same file since there is only the one user of the behavior set if that is OK.  It is a general utility function, but I'm not sure how wide to make exposure.

Let me know what you'd like and I'll try to do it.

If I rename the function commits 2 and 3 will change slightly.  Those have your reviewed-by on them.

alejandro-colomar: Yeah, keeping it in this file is fine.  We can move it later if/when we find more uses of it.

A simple rename would be fine.

lib/subordinateio.c
jcpunk: I had it in the wrong commit... Should be fixed now.

lib/subordinateio.c
jcpunk: Added to "`subid: Remove get_owner_id`"

lib/subordinateio.c
jcpunk: In theory this is done.

lib/subordinateio.c
jcpunk: I don't have any real grip on how the `lib` directory is sorted...

Would that follow the behavior of `getpwnam` via some sort of `xgetpw_uid_or_nam.c`?

In that I make `xgetpw_uid_or_nam.c` which would look like:
```C
/* comment */
# include headers

const struct passwd *
getpw_uid_or_nam(const char *u)
{
	uid_t  uid;

	return str2i(uid_t, &uid, u) == 0 ? getpwuid(uid) : getpwnam(u);
}
```

And add it to `prototypes.h` and `Makefile.am`?

Or since you wrote it would you rather put in the code yourself and have me rebase off that?

alejandro-colomar: I could write the code myself.  That will make it easier for you.  Thanks!

alejandro-colomar: Please pick the patch from here into your PR:

-  <https://github.com/shadow-maint/shadow/pull/1577>

ikerexxe:
> That is my intention.
> 
> Full disclosure, I've not be very successful at running the test suite.

That makes perfect sense.

It’s also a perfect example of why we generally discourage fully numeric usernames. When a file can accept either a username or a UID, a name like `1000` creates immediate parsing ambiguity, the system can't easily tell if it's looking at a name or an ID. Supporting both in this PR actually reinforces why the Linux ID Management stack struggles with numeric names.

lib/subordinateio.c
jcpunk: Rename performed, as a part of the process I've edited the commits and commit messages to use the new name.

Next up I'll start using getpw_uid_or_nam, but first I've got to put that in my branch.

lib/subordinateio.c
jcpunk: Commit put in at the top of my tree, so all my commits shifted down by 1.

I think I've got this used correctly, but I did it a bit quickly before lunch...

lib/subordinateio.c
```diff
@@ -142,6 +144,43 @@ static struct commonio_ops subordinate_ops = {
...
+ * Numeric resolution ensures stale entries for deleted users cannot
+ * produce false matches, and handles systems with overlapping UIDs.
+ *
+ * We do not perform a streq(3) on the passed strings themselves.
+ * In this way we ensure the user is resolvable on the system
+ * eliminating the ability to return results from deleted users
+ * with stale entries.
+ *
+ * Returns false if either identifier cannot be resolved, or if the
+ * resolved UIDs do not match.
```
alejandro-colomar: I'd chop all this text.  The first line plus the implementation already tells it all.

It might be interesting in the commit message, but it feels superfluous and noisy in the source code.

alejandro-colomar: I'd write it as

```c
// is_same_user - test whether two strings identify the same user
static bool
is_same_user(const char *a, const char *b)
{
        ...
}
```

lib/subordinateio.c
```diff
@@ -650,7 +650,29 @@ int sub_uid_add (const char *owner, uid_t start, unsigned long count)
...
+		if (str2ui(&owner_uid, owner) == -1) {
+			const struct passwd  *pw;
+
+			pw = getpwnam(owner);
+			if (NULL == pw) {
+				errno = ENOENT;
+				return 0;
+			}
+			owner_uid = pw->pw_uid;
+		}
```
alejandro-colomar: This still needs to be adapted to use getpw_uid_or_nam()  (in commit 6).

lib/subordinateio.c
```diff
@@ -788,7 +810,29 @@ int sub_gid_add (const char *owner, gid_t start, unsigned long count)
...
```
alejandro-colomar: And another getpw_uid_or_nam() candidate in commit 6.

lib/subordinateio.c
jcpunk: I have moved most of that into the commit message.

lib/subordinateio.c
jcpunk: Done

lib/subordinateio.c
jcpunk: Done

lib/subordinateio.c
```diff
@@ -142,6 +144,25 @@ static struct commonio_ops subordinate_ops = {
...
+		return false;
+
+	pw_b = getpw_uid_or_nam(b);
+	if (NULL == pw_b)
+		return false;
+
+	return pw_a->pw_uid == pw_b->pw_uid;
+}
+
+
```
alejandro-colomar: For commit 2, which is a60aaf40 ("subid: Add is_same_user for unified ID lookups"):

```
```

man/login.defs.d/SUB_GID_STORE_BY_UID.xml
```diff
@@ -0,0 +1,15 @@
...
+   SPDX-License-Identifier: BSD-3-Clause
+-->
+<varlistentry condition="subids">
+  <term><option>SUB_GID_STORE_BY_UID</option> (boolean)</term>
+  <listitem>
+    <para>
+      If set to <replaceable>yes</replaceable>, subordinate group ID entries
+      in <filename>/etc/subgid</filename> are stored using the numeric user
+      ID rather than the username. The default value is
+      <replaceable>no</replaceable>.
```
alejandro-colomar: Please use semantic newlines for new documentation.  See man-pages(7):

```
   Use semantic newlines

```

alejandro-colomar: That would be:

```
    <para>
      If set to <replaceable>yes</replaceable>,
      subordinate group ID entries in <filename>/etc/subgid</filename>
      are stored using the numeric user ID
      rather than the username.
      The default value is <replaceable>no</replaceable>.
    </para>
```

alejandro-colomar: Apart from this minor editorial issue,

for commit 5, which is 22654125 ("subid: Define SUB_UID_STORE_BY_UID and SUB_GID_STORE_BY_UID"):

```
```

man/login.defs.d/SUB_UID_STORE_BY_UID.xml
```diff
@@ -0,0 +1,15 @@
...
+   SPDX-License-Identifier: BSD-3-Clause
+-->
+<varlistentry condition="subids">
+  <term><option>SUB_UID_STORE_BY_UID</option> (boolean)</term>
+  <listitem>
+    <para>
+      If set to <replaceable>yes</replaceable>, subordinate user ID entries
+      in <filename>/etc/subuid</filename> are stored using the numeric user
+      ID rather than the username. The default value is
+      <replaceable>no</replaceable>.
```
alejandro-colomar: Same thing about semantic newlines.

man/Makefile.am
```diff
@@ -183,7 +183,9 @@ login_defs_v = \
 	USERGROUPS_ENAB.xml \
 	USE_TCB.xml \
 	SUB_GID_COUNT.xml \
+	SUB_GID_STORE_BY_UID.xml \
 	SUB_UID_COUNT.xml \
+	SUB_UID_STORE_BY_UID.xml \
 	SYS_GID_MAX.xml \
 	SYS_UID_MAX.xml \
```
alejandro-colomar, line 190: Should we move the `_STORE_BY_UID` after `_MAX` for consistency having `GID_` and `UID_` pairs together in this file?

man/login.defs.5.xml
```diff
@@ -204,7 +206,9 @@
       &SU_NAME;
       &SU_WHEEL_ONLY;
       &SUB_GID_COUNT; <!-- documents also SUB_GID_MIN SUB_GID_MAX -->
+      &SUB_GID_STORE_BY_UID;
       &SUB_UID_COUNT; <!-- documents also SUB_UID_MIN SUB_UID_MAX -->
+      &SUB_UID_STORE_BY_UID;
       &SYS_GID_MAX; <!-- documents also SYS_GID_MIN -->
       &SYS_UID_MAX; <!-- documents also SYS_UID_MIN -->
```
alejandro-colomar, line 213: Same issue about ordering here.

alejandro-colomar: Oh, no, I misread.  Sorry!

man/Makefile.am
alejandro-colomar: I misread; sorry!

lib/subordinateio.c
```diff
@@ -632,7 +632,24 @@ int sub_uid_add (const char *owner, uid_t start, unsigned long count)
 		errno = EOPNOTSUPP;
 		return 0;
 	}
-	return add_range (&subordinate_uid_db, owner, start, count);
+	if (getdef_bool("SUB_UID_STORE_BY_UID")) {
+		char         uid_string[ID_SIZE];
+		const struct passwd  *pw;
```
alejandro-colomar: `uid_string` should be aligned with `*pw`, as they are both the identifiers being declared.

alejandro-colomar:
```c

		char                 uid_string[ID_SIZE];
		const struct passwd  *pw;
```

lib/subordinateio.c
```diff
@@ -632,7 +632,24 @@ int sub_uid_add (const char *owner, uid_t start, unsigned long count)
...
+
+		pw = getpw_uid_or_nam(owner);
+		if (NULL == pw) {
+			errno = ENOENT;
+			return 0;
+		}
```
alejandro-colomar: getpw_uid_or_nam() already sets errno, as getpwuid(3) and getpwnam(3) from libc set errno on error.  I think it would be better to respect that errno value, which might be ENOENT, but it might also be another one such as EIO.

alejandro-colomar: I'd just

```c
if (pw == NULL)
	return 0;
```

lib/subordinateio.c
```diff
@@ -632,7 +632,24 @@ int sub_uid_add (const char *owner, uid_t start, unsigned long count)
...
+
+		if (stprintf_a(uid_string, "%u", pw->pw_uid) == -1) {
+			errno = EINVAL;
+			return 0;
+		}
```
alejandro-colomar: stprintf_a() already sets errno on error.  I'd respect that value, since EINVAL isn't much better.

I'd just

```c
if (stprintf_a(...) == -1)
	return 0;
```

man/login.defs.d/SUB_GID_STORE_BY_UID.xml
jcpunk: I've added you're reviewed by, but am a bit worried I somehow messed up the line breaks....

man/login.defs.d/SUB_UID_STORE_BY_UID.xml
jcpunk: Done

lib/subordinateio.c
jcpunk: I think I got it

lib/subordinateio.c
jcpunk: Can do

lib/subordinateio.c
jcpunk: Can do

man/login.defs.d/SUB_GID_STORE_BY_UID.xml
alejandro-colomar: No problem.  I'll have a look at it.

alejandro-colomar: I think they're fine.  What problem do you see?

lib/subordinateio.c
```diff
@@ -632,7 +632,21 @@ int sub_uid_add (const char *owner, uid_t start, unsigned long count)
...
+		if (NULL == pw)
+			return 0;
+
+		if (stprintf_a(uid_string, "%u", pw->pw_uid) == -1)
+			return 0;
+
+		return add_range(&subordinate_uid_db, uid_string, start, count);
+	} else {
+		return add_range(&subordinate_uid_db, owner, start, count);
+	}
```
alejandro-colomar, line 651: For commit 6, which is 048e9b0a1 ("subid: start using SUB_UID_STORE_BY_UID/GID"):

```
```

jcpunk: Added

lib/subordinateio.c
jcpunk: Added

man/login.defs.d/SUB_GID_STORE_BY_UID.xml
jcpunk: Mostly I don't want to put you in a position where you need to ask me to remove a Reviewed-by tag because I screwed up.

alejandro-colomar:
Thanks!  I'll merge.

alejandro-colomar:
Hmmm, there's a CI job failing.  I've tried re-running it to see if it's just a spurious failure.  Let's see.

alejandro-colomar:
Hmmm, the testsuite failure seems consistent.  Could you please have a look at the test and see why it's failing?

Cc: @hallyn

jcpunk:
I'm guess that, because the queries now require the user to resolve to a real identity that the checks for user "`foo`" are not returning anything.

jcpunk:
Digging a bit deeper, I'm showing results I don't understand...

It appears that `pw_a` is set to the content of `pw_b` at the second call to `getpw_uid_or_nam` within `is_same_user`???

Here is my diff to make print outs:
```patch
--- a/lib/shadow/passwd/getpw.h
+++ b/lib/shadow/passwd/getpw.h
@@ -23,6 +23,14 @@ getpw_uid_or_nam(const char *u)
 {
        uid_t  uid;
 
+printf("AAA: %s\n", u);
+if (get_uid(u, &uid) == 0) {
+getpwuid(uid);
+printf("BBB: %s\n", getpwuid(uid)->pw_name);
+} else {
+printf("CCC: %s\n", getpwnam(u)->pw_name);
+}
+
        return get_uid(u, &uid) == 0 ? getpwuid(uid) : getpwnam(u);
 }
 
diff --git a/lib/subordinateio.c b/lib/subordinateio.c
index 37e5100a..d41cba21 100644
--- a/lib/subordinateio.c
+++ b/lib/subordinateio.c
@@ -151,6 +151,8 @@ is_same_user(const char *a, const char *b)
        const struct passwd  *pw_a;
        const struct passwd  *pw_b;
 
+printf("XXXXX: does names %s == %s?\n", a, b);
+
        pw_a = getpw_uid_or_nam(a);
        if (NULL == pw_a)
                return false;
@@ -159,6 +161,8 @@ is_same_user(const char *a, const char *b)
        if (NULL == pw_b)
                return false;
 
+printf("YYY: does ID %s == %s?\n", pw_a->pw_name, pw_b->pw_name);
+printf("YYY: does ID %i == %i?\n", pw_a->pw_uid, pw_b->pw_uid);
        return pw_a->pw_uid == pw_b->pw_uid;
 }
 
@@ -882,6 +886,7 @@ int list_owner_ranges(const char *owner, enum subid_type id_type, struct subid_r
 
        commonio_rewind(db);
+printf("The flag is: %s\n", is_same_user(range->owner, owner) ? "true" : "false");
                if (is_same_user(range->owner, owner)) {
                        ranges = append_range(ranges, range, count++);
                        if (ranges == NULL) {
```

And I get the following output from the test with my updated ranges:
```text
----------
/tmp/subuidlist
----------------
XXXXX: does names 1000 == nobody?
AAA: 1000
BBB: ubuntu
AAA: nobody
CCC: nobody
YYY: does ID ubuntu == nobody?
YYY: does ID 1000 == 65534?
The flag is: false
XXXXX: does names 1000 == nobody?
AAA: 1000
BBB: ubuntu
AAA: nobody
CCC: nobody
YYY: does ID ubuntu == nobody?
YYY: does ID 1000 == 65534?
XXXXX: does names bin == nobody?
AAA: bin
CCC: bin
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
The flag is: true
XXXXX: does names bin == nobody?
AAA: bin
CCC: bin
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
XXXXX: does names 65534 == nobody?
AAA: 65534
BBB: nobody
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
The flag is: true
XXXXX: does names 65534 == nobody?
AAA: 65534
BBB: nobody
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
XXXXX: does names nobody == nobody?
AAA: nobody
CCC: nobody
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
The flag is: true
XXXXX: does names nobody == nobody?
AAA: nobody
CCC: nobody
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
XXXXX: does names nobody == nobody?
AAA: nobody
CCC: nobody
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
The flag is: true
XXXXX: does names nobody == nobody?
AAA: nobody
CCC: nobody
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
XXXXX: does names root == nobody?
AAA: root
CCC: root
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
The flag is: true
XXXXX: does names root == nobody?
AAA: root
CCC: root
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
XXXXX: does names 0 == nobody?
AAA: 0
BBB: root
AAA: nobody
CCC: nobody
YYY: does ID root == nobody?
YYY: does ID 0 == 65534?
The flag is: false
XXXXX: does names 0 == nobody?
AAA: 0
BBB: root
AAA: nobody
CCC: nobody
YYY: does ID root == nobody?
YYY: does ID 0 == 65534?
0: nobody 190000 10000
1: nobody 200000 10000
2: nobody 300000 10000
3: nobody 400000 10000
4: nobody 500000 1000
---------------
/tmp/subgidlist
-----------------------
XXXXX: does names bin == nobody?
AAA: bin
CCC: bin
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
The flag is: true
XXXXX: does names bin == nobody?
AAA: bin
CCC: bin
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
XXXXX: does names nobody == nobody?
AAA: nobody
CCC: nobody
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
The flag is: true
XXXXX: does names nobody == nobody?
AAA: nobody
CCC: nobody
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
XXXXX: does names 65534 == nobody?
AAA: 65534
BBB: nobody
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
The flag is: true
XXXXX: does names 65534 == nobody?
AAA: 65534
BBB: nobody
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
XXXXX: does names root == nobody?
AAA: root
CCC: root
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
The flag is: true
XXXXX: does names root == nobody?
AAA: root
CCC: root
AAA: nobody
CCC: nobody
YYY: does ID nobody == nobody?
YYY: does ID 65534 == 65534?
XXXXX: does names 0 == nobody?
AAA: 0
BBB: root
AAA: nobody
CCC: nobody
YYY: does ID root == nobody?
YYY: does ID 0 == 65534?
The flag is: false
XXXXX: does names 0 == nobody?
AAA: 0
BBB: root
AAA: nobody
CCC: nobody
YYY: does ID root == nobody?
YYY: does ID 0 == 65534?
0: nobody 190000 10000
1: nobody 200000 10000
2: nobody 210000 10000
3: nobody 500000 1000
```

Which shows at some point before `YYY` `bin` gets swapped in for `nobody`???

jcpunk:
Found the fix, the retrieved structure from `getpw_uid_or_nam`  seems to have some reused elements via the getpw* calls...

This resulted in a code change to commit 2 "`subid: Add is_same_user for unified ID lookups`" and I removed the Reviewed-by tag.

alejandro-colomar:
> Found the fix, the retrieved structure from `getpw_uid_or_nam` seems to have some reused elements via the getpw* calls...
> 
> This resulted in a code change to commit 2 "`subid: Add is_same_user for unified ID lookups`" and I removed the Reviewed-by tag.

Ahhh, yes, now it makes sense.  getpwuid(3) and getpwnam(3) reuse static memory.  We should copy the contents.

Thanks!

I'll review the changes.

lib/subordinateio.c
```diff
@@ -142,6 +144,27 @@ static struct commonio_ops subordinate_ops = {
...
+	uid_a = pw->pw_uid;
+
+	pw = getpw_uid_or_nam(b);
+	if (NULL == pw)
+		return false;
+	uid_b = pw->pw_uid;
+
+	return uid_a == uid_b;
+}
+
```
alejandro-colomar, line 167: Thanks!  The fix makes sense.

For commit 2, which is 7ac413c3 ("subid: Add is_same_user for unified ID lookups"):

```
```

tests/libsubid/01_list_ranges/list_ranges.test
```diff
@@ -17,19 +17,21 @@ trap 'log_status "$0" "FAILURE"; restore_config' 0
...
 
 echo -n "Check the subgid ranges..."
-[ $(wc -l /tmp/subgidlistout | awk '{ print $1 }') -eq 1 ]
-grep "0: foo 200000 10000" /tmp/subgidlistout
+[ $(wc -l /tmp/subgidlistout | awk '{ print $1 }') -eq 2 ]
+grep "0: nobody 200000 10000" /tmp/subgidlistout
+grep "1: nobody 210000 10000" /tmp/subgidlistout
 echo "OK"
 
 log_status "$0" "SUCCESS"
```
alejandro-colomar, line 37: I'm not too familiar with this part.  @ikerexxe , @hallyn , @stoeckmann , would you mind having a look at the last commit?

lib/subordinateio.c
jcpunk: Added

tests/libsubid/01_list_ranges/list_ranges.test
ikerexxe: I'm not really sure what the test is doing. I don't see where `nobody`'s `sub*ids` come from.

jcpunk: I thought I added the ranges in this commit https://github.com/jcpunk/shadow/commit/6a55ab4b603cdb710316d764460c2124e339eae8

alejandro-colomar: You did.  I couldn't include them in the same quote, as they're in different files.

jcpunk:
Just double checking that nothing is waiting on me. I was out of office and kinda lost track.

alejandro-colomar:
> Just double checking that nothing is waiting on me. I was out of office and kinda lost track.

That's correct.  I'm waiting on @hallyn and/or @ikerexxe OKing the changes to tests as I don't feel qualified to review them.

PR #1574, opened 2026-03-11
strchriscntrl: reject C1 control bytes (0x80-0x9F)

KhaelK138:
Shadow's chfn allows C1 control bytes (0x80-0x9F) to be written into GECOS fields in /etc/passwd. The input validation in valid_field() uses iscntrl() via strchriscntrl() to reject control characters, but glibc's iscntrl() does not classify the C1 range as control characters in any locale. The iscntrl() check was added as part of the fix for CVE-2023-29383, which blocked C0 control characters like carriage return, but C1 control bytes were never considered. The C0 range (0x00-0x1F) is correctly rejected, but the C1 range passes through.

The byte 0x9B is the C1 encoding of CSI (Control Sequence Introducer), equivalent to ESC [. On terminals that interpret C1 codes (VTE-based terminals -- GNOME Terminal, Tilix, Terminator, XFCE Terminal, etc.), an attacker can inject cursor movement and SGR sequences into a GECOS field via "chfn -r". This allows visual spoofing of /etc/passwd output -- for example, making a user's UID appear as 0 when viewed with cat. The chfn.c code also contributes to the issue, as it only warns (but does not reject) non-ASCII characters in the name and room fields, so valid_field() returning 1 for non-ASCII does not prevent the write.

This patch adds an explicit check for the 0x80-0x9F range in strchriscntrl(), so that valid_field() returns -1 (rejected) instead of 1 (non-ASCII warning) for these bytes. This causes check_fields() in chfn.c to reject the input and exit, rather than printing a warning and proceeding with the write.

Tested on Shadow 4.17.4-2ubuntu2 (Ubuntu 25.10, passwd package, aarch64) with /usr/bin/chfn as the SUID binary. Before the patch, running "chfn -r" with a payload containing 0x9B bytes exits 0 and writes the C1 bytes into /etc/passwd. After the patch, chfn prints "invalid room number" and exits 1, leaving /etc/passwd unchanged. Legitimate ASCII room numbers are still accepted normally.

### DIFF
--- a/lib/string/ctype/strchrisascii/strchriscntrl.h
+++ b/lib/string/ctype/strchrisascii/strchriscntrl.h
@@ -18,7 +18,10 @@ inline bool strchriscntrl(const char *s);
 
 
 // string character is [:cntrl:]
-// Return true if any iscntrl(3) character is found in the string.
+// Return true if any control character is found in the string.
+// Check both C0 (0x00-0x1F, 0x7F) via iscntrl(3) and C1 (0x80-0x9F)
+// explicitly, since glibc's iscntrl() does not classify C1 bytes as
+// control characters in any locale.
 inline bool
 strchriscntrl(const char *s)
 {
@@ -26,6 +29,8 @@ strchriscntrl(const char *s)
 		unsigned char  c = *s;
 
 		if (iscntrl(c))
+			return true;
+		if (c >= 0x80 && c <= 0x9F)
 			return true;
 	}
 


alejandro-colomar:
> Shadow's chfn allows C1 control bytes (0x80-0x9F) to be written into GECOS fields in /etc/passwd. The input validation in valid_field() uses iscntrl() via strchriscntrl() to reject control characters, but glibc's iscntrl() does not classify the C1 range as control characters in any locale. The iscntrl() check was added as part of the fix for [CVE-2023-29383](https://github.com/advisories/GHSA-p9w4-8hh8-crcx), which blocked C0 control characters like carriage return, but C1 control bytes were never considered. The C0 range (0x00-0x1F) is correctly rejected, but the C1 range passes through.
> 
> The byte 0x9B is the C1 encoding of CSI (Control Sequence Introducer), equivalent to ESC [. On terminals that interpret C1 codes (VTE-based terminals -- GNOME Terminal, Tilix, Terminator, XFCE Terminal, etc.), an attacker can inject cursor movement and SGR sequences into a GECOS field via "chfn -r". This allows visual spoofing of /etc/passwd output -- for example, making a user's UID appear as 0 when viewed with cat. The chfn.c code also contributes to the issue, as it only warns (but does not reject) non-ASCII characters in the name and room fields, so valid_field() returning 1 for non-ASCII does not prevent the write.
> 
> This patch adds an explicit check for the 0x80-0x9F range in strchriscntrl(), so that valid_field() returns -1 (rejected) instead of 1 (non-ASCII warning) for these bytes. This causes check_fields() in chfn.c to reject the input and exit, rather than printing a warning and proceeding with the write.
> 
> Tested on Shadow 4.17.4-2ubuntu2 (Ubuntu 25.10, passwd package, aarch64) with /usr/bin/chfn as the SUID binary. Before the patch, running "chfn -r" with a payload containing 0x9B bytes exits 0 and writes the C1 bytes into /etc/passwd. After the patch, chfn prints "invalid room number" and exits 1, leaving /etc/passwd unchanged. Legitimate ASCII room numbers are still accepted normally.

This should be included in the commit message.  It is uninformative now.

I'd also like to see some testing before and after.

KhaelK138:
Gotcha @alejandro-colomar. I've updated the commit message, and here is some of the testing I've conducted below on the updated binary (`chfn_fix`):

Verifying my `/etc/passwd` entry is normal to start:
```
ubuntu@ubuntu:~$ cat /etc/passwd | grep ubuntu
ubuntu:x:1000:1000:ubuntu,,,:/home/ubuntu:/bin/bash
```

Using the old `chfn` to inject the control characters, making my entry appear as though I'm a root user:
```
ubuntu@ubuntu:~$ chfn -r $'\xc2\x9b17D0\xc2\x9b3P\xc2\x9b1C0\xc2\x9b3P\xc2\x9b1Croot\xc2\x9b8m'
Password: 
chfn: room number with non-AS0c0croot' '
ubuntu@ubuntu:~$ cat /etc/passwd | grep ubuntu
ubuntu:x:0:0:root,,:/home/ubuntu:/bin/bash
```

Resetting:
```
ubuntu@ubuntu:~$ chfn -r ""
Password: 
ubuntu@ubuntu:~$ cat /etc/passwd | grep ubuntu
ubuntu:x:1000:1000:ubuntu,,,:/home/ubuntu:/bin/bash
```

Using the updated version, which errors out:
```
ubuntu@ubuntu:~$ chfn_fix -r $'\xc2\x9b17D0\xc2\x9b3P\xc2\x9b1C0\xc2\x9b3P\xc2\x9b1Croot\xc2\x9b8m'
Password: 
chfn: inval0o0uroot' '
ubuntu@ubuntu:~$ cat /etc/passwd | grep ubuntu
ubuntu:x:1000:1000:ubuntu,,,:/home/ubuntu:/bin/bash
```

`/etc/passwd` remains unchanged. The error message is garbled, as the C1 bytes in the output are parsed by the terminal itself.

Using the updated version for its expected purposes:
```
ubuntu@ubuntu:~$ chfn_fix -r "101A"
Password: 
ubuntu@ubuntu:~$ cat /etc/passwd | grep ubuntu
ubuntu:x:1000:1000:ubuntu,101A,,:/home/ubuntu:/bin/bash
ubuntu@ubuntu:~$ chfn_fix
Password: 
Changing the user information for ubuntu
Enter the new value, or press ENTER for the default
        Full Name: ubuntu
        Room Number [101A]: 102B
        Work Phone []: 1231231234
        Home Phone []: 1231231234
ubuntu@ubuntu:~$ cat /etc/passwd | grep ubuntu
ubuntu:x:1000:1000:ubuntu,102B,1231231234,1231231234:/home/ubuntu:/bin/bash
ubuntu@ubuntu:~$ 
```

lib/string/ctype/strchrisascii/strchriscntrl.h
```diff
@@ -18,7 +18,10 @@ inline bool strchriscntrl(const char *s);
...
 
 // string character is [:cntrl:]
-// Return true if any iscntrl(3) character is found in the string.
+// Return true if any control character is found in the string.
+// Check both C0 (0x00-0x1F, 0x7F) via iscntrl(3) and C1 (0x80-0x9F)
+// explicitly, since glibc's iscntrl() does not classify C1 bytes as
+// control characters in any locale.
 inline bool
 strchriscntrl(const char *s)
 {
```
alejandro-colomar, line 27: I think I like the idea.

It's simple, and it's obviously good, I think (unless someone shows a strong reason against).

I would like to change the name of the function, to document the class to which these characters belong.  Does this return true for all UTF-8 control characters?  If so, it might be good to call this `strchriscntrl_utf8()`.

What do you think?

KhaelK138: The function works byte-by-byte, so it rejects any byte in `0x80-0x9F` whether it's a raw C1 code or a UTF-8 continuation byte. I'd say perhaps something like `strchriscntrl_c0c1()` might work, or keeping it as is and letting the comments explain it. 

That said, I believe it would catch [all UTF-8 control characters](https://www.compart.com/en/unicode/category/Cc), so `_utf8()` could also work, though it might be a bit misleading since it just checks the bytes.

KhaelK138: One thing to note: since the check is byte-level, it would also rejects legitimate UTF-8 characters whose continuation bytes fall in `0x80-0x9F`. A room name (`-r`) of `ārnis` (`U+0101 = 0xC4 0x81`) is rejected because 0x81 falls in the range.

I'd imagine that this is acceptable today since chfn already warns on non-ASCII input, but if Shadow ever adds proper UTF-8 GECOS support (per [Debian #385085](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=385085)), this would need reworking to decode UTF-8 and only reject actual C1 codepoints (`U+0080-U+009F`).

alejandro-colomar: `strchriscntrl_c0c1` seems a reasonable name.

alejandro-colomar: > One thing to note: since the check is byte-level, it would also rejects legitimate UTF-8 characters whose continuation bytes fall in `0x80-0x9F`. A room name (`-r`) of `ārnis` (`U+0101 = 0xC4 0x81`) is rejected because 0x81 falls in the range.

This is acceptable.  I prefer a simpler solution that rejects some rare valid stuff over a complex one.

> 
> I'd imagine that this is acceptable today since chfn already warns on non-ASCII input, but if Shadow ever adds proper UTF-8 GECOS support (per [Debian #385085](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=385085)), this would need reworking to decode UTF-8 and only reject actual C1 codepoints (`U+0080-U+009F`).

ikerexxe:
IMO, the best way to test this would be to run unit-tests. That would allow us to test different possibilities, including edge cases and it's quite straightforward to implement and run.

https://github.com/shadow-maint/shadow/blob/master/doc/contributions/tests.md#unit-tests provides a brief initial explanation of how to run and implement the tests. If you need further assistance, please do not hesitate to ask here.

lib/string/ctype/strchrisascii/strchriscntrl.h
alejandro-colomar: I think I'll merge as is.  I can do the rename later, as I plan to rename these functions anyway.

alejandro-colomar:
> IMO, the best way to test this would be to run unit-tests. That would allow us to test different possibilities, including edge cases and it's quite straightforward to implement and run.
> 
> https://github.com/shadow-maint/shadow/blob/master/doc/contributions/tests.md#unit-tests provides a brief initial explanation of how to run and implement the tests. If you need further assistance, please do not hesitate to ask here.

I'll merge, as this is obviously good.  The code rejects characters in the range `0x80 .. 0x9F`, which are the C1 control characters (according to Wikipedia), and certainly not ASCII (and thus should not affect most users).

It may affect users using UTF-8, if they use multi-byte characters that have a byte that is in this range.  If so, that's something we can't easily predict (we'd need a UTF-8 expert to tell us which characters use that).  I guess it's easier to merge, and if anyone notices a regression, we can always do the more complex parsing of multi-byte strings.  But I hope we don't need to do that.

We can always merge tests later.

PR #1576, opened 2026-03-11
useradd: Cast terminating NULL in execl call

stoeckmann:
The execl function takes variadic arguments, which should be cast to a pointer. Even though shadow targets POSIX systems, which require NULL to be (void *)0 instead of just 0, the latter would lead to just 32 bit written into stack with the upper 32 bit being potential junk.

Highly hypothetical defence, after all shadow is for Linux which is POSIX compatible, but stay compatible with execl(3) requirements and other execl calls throughout the tree.

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2076,7 +2076,7 @@ static void tallylog_reset (const char *user_name)
 		break;
 	case 0: /* child */
 		pname = Basename(pam_tally2);
-		execl(pam_tally2, pname, "--user", user_name, "--reset", "--quiet", NULL);
+		execl(pam_tally2, pname, "--user", user_name, "--reset", "--quiet", (char *) NULL);
 		/* If we come here, something has gone terribly wrong */
 		perror(pam_tally2);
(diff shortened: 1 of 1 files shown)


src/useradd.c
```diff
@@ -2076,7 +2076,7 @@ static void tallylog_reset (const char *user_name)
 		break;
 	case 0: /* child */
 		pname = Basename(pam_tally2);
-		execl(pam_tally2, pname, "--user", user_name, "--reset", "--quiet", NULL);
+		execl(pam_tally2, pname, "--user", user_name, "--reset", "--quiet", (char *) NULL);
```
alejandro-colomar, line 2079: I prefer having less casts.  The ideal number of casts would be none.

We have a high dependency on POSIX already, so I think we can rely on this.

FWIW, even the C Committee is considering disallowing `0` as a null pointer constant.  That will be slow, but will eventually happen.

GCC has recently added `-Wzero-as-null-pointer-constant` --which we should enable--, to catch issues such as NULL expanding to `0`, and help an eventual transition to not allowing 0 as a null pointer constant.

PR #1582, opened 2026-03-11
useradd, usermod: Gracefully handle sysconf -1 return value

stoeckmann:
If sysconf returns -1, use predefined upper limit from `linux/limits.h`. If it's not defined, use a sane default value.

This avoids potenial out of boundary writes into a 0 byte allocated buffer.

### DIFF
--- a/lib/sysconf.c
+++ b/lib/sysconf.c
@@ -1,0 +1,43 @@
+// SPDX-FileCopyrightText: 2025, Alejandro Colomar <alx@kernel.org>
+// SPDX-FileCopyrightText: 2026, Tobias Stoeckmann <tobias@stoeckmann.org>
+// SPDX-License-Identifier: BSD-3-Clause
+
+
+#include "config.h"
+
+#include <limits.h>
+#include <unistd.h>
+
+
+#ifndef  LOGIN_NAME_MAX
+# define LOGIN_NAME_MAX  256
+#endif
+
+#ifndef  NGROUPS_MAX
+# define NGROUPS_MAX  65536
+#endif
+
+
+size_t
+login_name_max_size(void)
+{
+	long  conf;
+
+	conf = sysconf(_SC_LOGIN_NAME_MAX);
+	if (conf == -1)
+		return LOGIN_NAME_MAX;
+
+	return conf;
+}
+
+size_t
+ngroups_max_size(void)
+{
+	long  conf;
+
+	conf = sysconf(_SC_NGROUPS_MAX);
+	if (conf == -1)
+		conf = NGROUPS_MAX;
+
+	return conf;
+}
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -72,6 +72,7 @@
 #include "string/strdup/strdup.h"
 #include "string/strerrno.h"
 #include "string/strtok/stpsep.h"
+#include "sysconf.h"
 
 
 #ifndef SKEL_DIR
@@ -153,7 +154,7 @@ static bool pw_locked = false;
 static bool gr_locked = false;
 static bool spw_locked = false;
 static char **user_groups;	/* NULL-terminated list */
-static long sys_ngroups;
+static size_t sys_ngroups;
 static bool do_grp_update = false;	/* group files need to be updated */
 
 extern int allow_bad_names;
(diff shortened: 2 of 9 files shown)


src/useradd.c
```diff
@@ -2529,6 +2533,9 @@ int main (int argc, char **argv)
 #endif
 
 	sys_ngroups = sysconf (_SC_NGROUPS_MAX);
+	if (sys_ngroups == -1) {
+		sys_ngroups = NGROUPS_MAX;
+	}
```
alejandro-colomar: Should we have a library function for this?

Similar to

```c
$ grepc login_name_max_size .
./lib/chkname.c:size_t
login_name_max_size(void)
{
	long  conf;

	conf = sysconf(_SC_LOGIN_NAME_MAX);
	if (conf == -1)
		return LOGIN_NAME_MAX;

	return conf;
}
./lib/chkname.h:extern size_t login_name_max_size(void);
```

We could add `ngroups_max()` (or maybe `ngroups_max_size()`).

alejandro-colomar: Please don't use braces for a single statement.

alejandro-colomar: Coding style link:

<https://github.com/shadow-maint/shadow/blob/master/doc/contributions/coding_style.md#c-code>

<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c#n46>

stoeckmann: Adjusted

stoeckmann: We can do so. Any idea where such a function belongs?

alejandro-colomar: Maybe a new `lib/conf/sysconf.[ch]` or `lib/conf/limits.[ch]`.  But feel free to pick any other place if you have any preference.

stoeckmann: I took `lib/sysconf.[ch]` for now, since `lib/limits.[ch]` contains non-PAM specific code. Not sure if it's already worth to add a new subdirectory, but that's up to you.

lib/sysconf.c
```diff
@@ -0,0 +1,43 @@
...
+ngroups_max_size(void)
+{
+	long  conf;
+
+    conf = sysconf (_SC_NGROUPS_MAX);
+	if (conf == -1)
+		conf = NGROUPS_MAX;
+
+    return conf;
+}
```
alejandro-colomar: Missing `\n` in the source file.  This is UB.  In C23 (n3220), this is UB from J.2p1 bullet 2, which refers to 5.1.1.2p1 phase 2 (last sentence):

> A source file that is not empty shall end in a new-line character, which shall not be immediately preceded by a backslash character before any such splicing takes place.

alejandro-colomar: This is in line with POSIX, which also requires that text files (and source files are text files) are terminated by a `\n`.

alejandro-colomar: Would you mind making the white space consistent?  Maybe in a 4th commit that does white space handling.

src/useradd.c
alejandro-colomar: I'm fine with that approach.  The main concern I have with headers without a directory is that some have a name that is the same as libc ones, and thus maybe cause problems.  But since there's no sysconf.h, I'm okay.

```sh
alx@devuan:~$ apt-file find include/sysconf.h
alx@devuan:~$
```

lib/sysconf.c
stoeckmann: Even though it's nice to have an IDE to quickly jump around, such stupid settings are annoying ... Thanks for pointing out! Fixed (not in settings yet, though).

stoeckmann: Fixed. Sorry!

src/useradd.c
stoeckmann: Yeah, I would have expected a `sysconf.h` in system include directories as well. Guess it's a "berenstain vs bereinstein" moment. :)

lib/sysconf.c
```diff
@@ -0,0 +1,43 @@
...
+#include <limits.h>
+#include <unistd.h>
+
+
+#ifndef  LOGIN_NAME_MAX
+# define LOGIN_NAME_MAX	256
+#endif
+
+#ifndef  NGROUPS_MAX
+# define NGROUPS_MAX	65536
```
alejandro-colomar: I prefer two spaces instead of a tab after the macro declaration, in general.  But this is not enough of a concern, so I'll merge anyway.  Thanks!

stoeckmann:
Hang on, have to fix tests ...

alejandro-colomar:
CI is failing, which is because tests seem to depend on these APIs, and thus, we need to add the new files to some Makefile under tests/unit/.

I think the tests should link against the entire lib/, which would solve these issues once and for all...  However, I'm ignorant enough to not know how to do that in the build system.  @ikerexxe would you know how to fix this forever?

alejandro-colomar:
@stoeckmann You'll need to add it in the Makefile, not in the source of the tests.  The error comes from the linker.

stoeckmann:
> @stoeckmann You'll need to add it in the Makefile, not in the source of the tests. The error comes from the linker.

Thanks for the hint! Fixed

PR #1584, opened 2026-03-13
Tests: Add a new user with a specified non-existing gid

aborah-sudo:
This is the transformation to Python of the test located in `tests/usertools/01/05_useradd_invalid_numeric_primary_group.test`, which checks that `useradd` fails to create a new user with non-existing group

### DIFF
--- a/tests/system/tests/test_useradd.py
+++ b/tests/system/tests/test_useradd.py
@@ -328,3 +328,35 @@ def test_useradd__add_user_with_existing_uid(shadow: Shadow):
         shadow.useradd("test2 -u 4242")
 
     assert exc_info.value.rc == 4, f"Expected return code 4 (UID already in use), got {exc_info.value.rc}"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_useradd__invalid_numeric_primary_group(shadow: Shadow):
+    """
+    :title: Add a new user with a specified non-existing gid
+    :steps:
+        1. Attempt to create user with -g 4242 option
+        2. Verify command fails with appropriate error code and message
+        3. Check passwd and group entries
+        4. Check home directory
+    :expectedresults:
+        1. Useradd command fails
+        2. Return code is 6 (specified group doesn't exist) and error message indicates group doesn't exist
+        3. No entries are added to passwd and group
+        4. No home directory is created
+    :customerscenario: False
+    """
+    with pytest.raises(ProcessError) as exc_info:
+        shadow.useradd("test1 -g 4242")
+
+    actual_rc = getattr(exc_info.value, "rc", getattr(exc_info.value, "returncode", None))
+    assert actual_rc == 6, f"Expected return code 6 (specified group doesn't exist), got {actual_rc}"
+
+    error_output = exc_info.value.stderr.lower() if exc_info.value.stderr else ""
+    assert (
+        "group" in error_output and "does not exist" in error_output
+    ), f"Error message should indicate group doesn't exist. Got: {error_output}"
+
+    assert shadow.tools.getent.passwd("test1") is None, "User test1 should not be found in passwd"
+    assert shadow.tools.getent.group("test1") is None, "Group test1 should not be found"
+    assert not shadow.fs.exists("/home/test1"), "Home directory should not be created"


tests/system/tests/test_useradd.py
```diff
@@ -328,3 +328,42 @@ def test_useradd__add_user_with_existing_uid(shadow: Shadow):
...
 
     assert exc_info.value.rc == 4, f"Expected return code 4 (UID already in use), got {exc_info.value.rc}"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_useradd__invalid_numeric_primary_group(shadow: Shadow):
+    """
+    :title: Add a new user with a specified non-existing gid
+    :setup:
+        1. Verify user "test1" and group "4242" does not exist in system
```
ikerexxe: This is a clean environment, so no need for this step. Please also remove the code

tests/system/tests/test_useradd.py
```diff
@@ -328,3 +328,42 @@ def test_useradd__add_user_with_existing_uid(shadow: Shadow):
...
+    :customerscenario: False
+    """
+    assert shadow.tools.getent.passwd("test1") is None, "User test1 should not exist before test"
+    assert shadow.tools.getent.group("4242") is None, "Group 4242 should not exist before test"
+
+    with pytest.raises(ProcessError) as exc_info:
+        shadow.useradd("test1 -g 4242")
+
+    actual_rc = getattr(exc_info.value, "rc", getattr(exc_info.value, "returncode", None))
+    assert actual_rc == 6, f"Expected return code 6 (invalid group), got {actual_rc}"
```
ikerexxe: s/invalid group/specified group doesn't exist

Or at least that's what our internal documentation states

tests/system/tests/test_useradd.py
```diff
@@ -328,3 +328,42 @@ def test_useradd__add_user_with_existing_uid(shadow: Shadow):
...
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_useradd__invalid_numeric_primary_group(shadow: Shadow):
+    """
+    :title: Add a new user with a specified non-existing gid
+    :setup:
+        1. Verify user "test1" and group "4242" does not exist in system
+    :steps:
+        1. Attempt to create user with -g 4242 option
+        2. Verify command fails with appropriate error code
+        3. Check error message content
```
ikerexxe: Merge this two in one step. As an example: `Verify command fails with appropriate error code and message`

tests/system/tests/test_useradd.py
```diff
@@ -328,3 +328,42 @@ def test_useradd__add_user_with_existing_uid(shadow: Shadow):
...
+        4. Verify no changes were made to system files
```
ikerexxe: According to the wording agreed with other team members this should be `Check passwd and group entries`

tests/system/tests/test_useradd.py
```diff
@@ -328,3 +328,42 @@ def test_useradd__add_user_with_existing_uid(shadow: Shadow):
...
+        5. Verify home directory was not created
```
ikerexxe: According to the wording agreed with other team members this should be `Check home directory`

tests/system/tests/test_useradd.py
```diff
@@ -328,3 +328,42 @@ def test_useradd__add_user_with_existing_uid(shadow: Shadow):
...
+    :expectedresults:
+        1. Useradd command fails
+        2. Return code is 6 (invalid group)
+        3. Error message indicates group doesn't exist
+        4. No entries are added to passwd group
```
ikerexxe: You are missing an `and` between `passwd` and `group`

ikerexxe (commented):
Some minor comments inline

Out of curiosity, is there any test that checks the creation of a user with `-g` option and succeeds? Just that option, and no other like `-u`

aborah-sudo:
> Some minor comments inline
> 
> Out of curiosity, is there any test that checks the creation of a user with `-g` option and succeeds? Just that option, and no other like `-u`

there is test in this same file "test_useradd__two_users_same_ids" there is command:

shadow.useradd("test2 -u 4242 -g 4242 -o")  but this is a different scenario where group was added before.

ikerexxe:
> there is test in this same file "test_useradd__two_users_same_ids" there is command:
> 
> shadow.useradd("test2 -u 4242 -g 4242 -o") but this is a different scenario where group was added before.

I was looking for something more specific, but it will have to serve

Just an inline comment missing and we'll be ready to merge

aborah-sudo:
> > there is test in this same file "test_useradd__two_users_same_ids" there is command:
> > shadow.useradd("test2 -u 4242 -g 4242 -o") but this is a different scenario where group was added before.
> 
> I was looking for something more specific, but it will have to serve
> 
> Just an inline comment missing and we'll be ready to merge

I will added that test in my next PR

ikerexxe:
> I will added that test in my next PR

No need for that, but thank you

Issue #1580, opened 2026-03-11
usermod: Set `x` as password hash in /etc/passwd when /etc/shadow entry is created


stoeckmann:
The usermod utility is able to create a new /etc/shadow entry if no entry was available yet (it does not create a new /etc/shadow file though).

If this happens, the password hash is copied into /etc/shadow, but it's not guaranteed that /etc/passwd is updated as well, i.e. the `x` indicator is not written into /etc/passwd's password field.

alejandro-colomar:
Would you mind showing a reproducer?

stoeckmann:
See https://github.com/shadow-maint/shadow/pull/1578
The `!` entry is copied but not replaced with a `x`.

Alternatively, follow the reproducer and set a password before creating an empty shadow file with `passwd -P $BASE user`.

PR #1587, opened 2026-03-15
usermod: Update passwd entry when shadowing entry

stoeckmann:
If a new shadow entry is created, the passwd entry's password hash is moved and replaced with an `x`. If this happens, update the passwd file as well, otherwise the `x` is not written to disk.

Resolves: https://github.com/shadow-maint/shadow/issues/1580

Proof of Concept (run as root):
1. Create a temporary directory
```
BASE=$(mktemp -d)
mkdir -p $BASE/etc
```
2. Create a new user
```
useradd -P $BASE user
```
3. Create an empty shadow file
```
touch $BASE/etc/shadow
```
4. Disable the user account
```
usermod -P $BASE -e 0 user
```
5. Check content of passwd file
```
cat $BASE/etc/passwd
```
```
user:!:1000:1000::/home/user:/bin/bash
```

It should be `x` instead of `!`, since a new shadow file was created:
```
cat $BASE/etc/shadow
```
```
user:!:20527:::::0:
```

### DIFF
--- a/src/usermod.c
+++ b/src/usermod.c
@@ -1807,7 +1807,7 @@ static void usr_update(const struct option_flags *flags)
 	}
 
 	if (lflg || uflg || gflg || cflg || dflg || sflg || pflg
-	    || Lflg || Uflg) {
+	    || Lflg || Uflg || spwd == &spent) {
 		if (pw_update (&pwent) == 0) {
 			fprintf (stderr,
 			         _("%s: failed to prepare the new %s entry '%s'\n"),


src/usermod.c
```diff
@@ -1805,7 +1805,7 @@ static void usr_update(const struct option_flags *flags)
 	}
 
 	if (lflg || uflg || gflg || cflg || dflg || sflg || pflg
-	    || Lflg || Uflg) {
+	    || Lflg || Uflg || spwd == &spent) {
 		if (pw_update (&pwent) == 0) {
 			fprintf (stderr,
 			         _("%s: failed to prepare the new %s entry '%s'\n"),
```
alejandro-colomar, line 1813: Thanks!  LGTM.

```
```

Some minor comments:

From the commit message:

> If a new shadow entry is created, the passwd entry's password hash is moved and replaced with an "x". If this happens, update the passwd file as well, otherwise the "x" is not written to disk.

Did you mean s/moved/removed/?

Also, it might be interesting to include the reproducer in the commit message.

And, it might also be interesting to show the reproducer after the fix, to show that it works after it.  (As a sanity check.)

stoeckmann: > Did you mean s/moved/removed/?

Even though I'm great at using wrong words in commit messages, this time I really mean moved. I have clarified by stating that the hash is moved into the shadow file (and then replaced with an x in passwd file).

> Also, it might be interesting to include the reproducer in the commit message.

Added.

And last but not least: rebased.

alejandro-colomar: > > Did you mean s/moved/removed/?
> 
> Even though I'm great at using wrong words in commit messages, this time I really mean moved. I have clarified by stating that the hash is moved into the shadow file (and then replaced with an x in passwd file).

Ahhh, now I understand it.  Thanks!

PR #1591, opened 2026-03-17
Tests: Add a new user with a valid existing group as primary

aborah-sudo:
Add a new user with a valid existing group as primary with option -g

### DIFF
--- a/tests/system/tests/test_useradd.py
+++ b/tests/system/tests/test_useradd.py
@@ -360,3 +360,34 @@ def test_useradd__invalid_numeric_primary_group(shadow: Shadow):
     assert shadow.tools.getent.passwd("test1") is None, "User test1 should not be found in passwd"
     assert shadow.tools.getent.group("test1") is None, "Group test1 should not be found"
     assert not shadow.fs.exists("/home/test1"), "Home directory should not be created"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_useradd__valid_group_as_primary(shadow: Shadow):
+    """
+    :title: Add a new user with a valid existing group as primary
+    :steps:
+        1. Create a group
+        2. Check group entry
+        3. Create user with that group as primary
+        4. Check passwd entry
+    :expectedresults:
+        1. Group created successfully
+        2. Group attributes are correct
+        3. User created with the specified group as primary
+        4. User's GID matches the group's GID
+    :customerscenario: False
+    """
+    shadow.groupadd("testgroup")
+
+    group_entry = shadow.tools.getent.group("testgroup")
+    assert group_entry is not None, "Group testgroup should exist"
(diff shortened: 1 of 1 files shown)


tests/system/tests/test_useradd.py
```diff
@@ -360,3 +360,31 @@ def test_useradd__invalid_numeric_primary_group(shadow: Shadow):
...
+        2. Verify group created successfully
+    :steps:
+        1. Create user with that group as primary
+        3. Verify user created successfully
+    :expectedresults:
+        1. User is created with the specified group as primary
+        3. User's GID matches the group's GID
+    :customerscenario: False
+    """
+    shadow.groupadd("testgroup -g 5000")
```
ikerexxe: No need for `-g 5000`

tests/system/tests/test_useradd.py
```diff
@@ -360,3 +360,31 @@ def test_useradd__invalid_numeric_primary_group(shadow: Shadow):
...
     assert not shadow.fs.exists("/home/test1"), "Home directory should not be created"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_useradd__valid_group_as_primary(shadow: Shadow):
+    """
+    :title: Add a new user with a valid existing group as primary
+    :setup:
+        1. Create a group
+        2. Verify group created successfully
```
ikerexxe: This is not part of the `setup` phase, but of the `steps` and `expectedresults`

tests/system/tests/test_useradd.py
```diff
@@ -360,3 +360,32 @@ def test_useradd__invalid_numeric_primary_group(shadow: Shadow):
...
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_useradd__valid_group_as_primary(shadow: Shadow):
+    """
+    :title: Add a new user with a valid existing group as primary
+    :steps:
+        1. Create a group
+        2. Create user with that group as primary
+        3. Check user created successfully
```
ikerexxe: I'd use `Check passwd entry` instead

tests/system/tests/test_useradd.py
```diff
@@ -360,3 +360,32 @@ def test_useradd__invalid_numeric_primary_group(shadow: Shadow):
...
     assert not shadow.fs.exists("/home/test1"), "Home directory should not be created"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_useradd__valid_group_as_primary(shadow: Shadow):
+    """
+    :title: Add a new user with a valid existing group as primary
+    :steps:
+        1. Create a group
+        2. Create user with that group as primary
```
ikerexxe: You are missing a step in the middle: `Check group entry` and `Group attributes are correct` in `expectedresults`

PR #1593, opened 2026-03-18
format: clean up errant spaces.

jcpunk:
This commit should have zero functional changes, but remove the no longer expected space between a function and its arguments.

### DIFF
--- a/lib/age.c
+++ b/lib/age.c
@@ -49,18 +49,18 @@ int expire (const struct passwd *pw, /*@null@*/const struct spwd *sp)
 	 * force them to change their password.
 	 */
 
-	status = isexpired (pw, sp);
+	status = isexpired(pw, sp);
 	switch (status) {
 	case 0:
 		return 0;
 	case 1:
-		(void) fputs (_("Your password has expired."), stdout);
+		(void) fputs(_("Your password has expired."), stdout);
 		break;
 	case 2:
-		(void) fputs (_("Your password is inactive."), stdout);
+		(void) fputs(_("Your password is inactive."), stdout);
 		break;
 	case 3:
-		(void) fputs (_("Your login has expired."), stdout);
+		(void) fputs(_("Your login has expired."), stdout);
 		break;
 	}
 
@@ -72,11 +72,11 @@ int expire (const struct passwd *pw, /*@null@*/const struct spwd *sp)
 	 */
 
 	if ((status > 1) || (sp->sp_max < sp->sp_min)) {
-		(void) puts (_("  Contact the system administrator."));
-		exit (EXIT_FAILURE);
+		(void) puts(_("  Contact the system administrator."));
+		exit(EXIT_FAILURE);
 	}
-	(void) puts (_("  Choose a new password."));
-	(void) fflush (stdout);
+	(void) puts(_("  Choose a new password."));
+	(void) fflush(stdout);
 
 	/*
 	 * Close all the files so that unauthorized access won't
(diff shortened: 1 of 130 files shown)


alejandro-colomar:
The problem with such a huge patch would be that downstream is going to have problems rebasing patches.  But I'm fine with this, FWIW.  (I haven't reviewed yet, but the idea seems reasonable.)

alejandro-colomar:
BTW, if this has been scripted, please paste the scripts in the commit message, so that it can be fully reproduced and investigated.

ikerexxe:
> The problem with such a huge patch would be that downstream is going to have problems rebasing patches. But I'm fine with this, FWIW. (I haven't reviewed yet, but the idea seems reasonable.)

I definitely share those concerns, it’s going to be a headache for downstream rebases (my future self included!).

While I agree the change itself is reasonable and I won't block it, I have enough mixed feelings that I’m going to softly abstain from a formal approval. I’m happy to let this land if the rest of the team is comfortable with the tradeoff.

alejandro-colomar:
> > The problem with such a huge patch would be that downstream is going to have problems rebasing patches. But I'm fine with this, FWIW. (I haven't reviewed yet, but the idea seems reasonable.)
> 
> I definitely share those concerns, it’s going to be a headache for downstream rebases (my future self included!).
> 
> While I agree the change itself is reasonable and I won't block it, I have enough mixed feelings that I’m going to softly abstain from a formal approval. I’m happy to let this land if the rest of the team is comfortable with the tradeoff.

I think we've been surviving well with the gradual removal of spaces in new or modified code.  I'm also going to abstain.  Unless someone voices strong support for this patch, I think it might be better to not do it.

hallyn:
(lemme think about this for a bit.  I do think git apply --ignore-whitespace or whatever may make it not so horrendous for downstreams.  But so often I have cursed having to backport across such things.  And actually when I look at the patch, i find the old lines easier to read :)

lib/age.c
```diff
@@ -34,7 +34,7 @@
  *	expire() calls /bin/passwd to change the user's password
  *	if it has expired.
  */
-int expire (const struct passwd *pw, /*@null@*/const struct spwd *sp)
+int expire(const struct passwd *pw, /*@null@*/const struct spwd *sp)
```
alejandro-colomar: I'll open this comment to block merging.

Please include the sed(1) scripts you used in the commit message.

jcpunk: Updated the commit message with the exact command I ran on my system.

jcpunk:
> And actually when I look at the patch, i find the old lines easier to read :)

If it isn't worth merging, I'm not offended if you close this out unmerged.

lib/age.c
alejandro-colomar: Thanks!

hallyn:
Thanks @jcpunk.  I'm going to go ahead and close this, mainly because although some tools offer ways to ignore the whitespaces when backporting patches, I'm worried that some, like quilt, will just drive a maintainer to drink.

jcpunk:
No worries

I'll leave the find/sed here just so it isn't lost as I delete the branch later:
```shell
find lib/ libsubid/ src/ -type f -name \*.[ch] -exec sed -i -E '
/^[ \t]*$/b
/^[ \t]*#/b
/\/\*/,/\*\//b
/\/\//b
s/([[:alpha:]]) \(/\1(/g
s/\b(if|else|for|while|switch|do|return|case|default|break|continue|goto|sizeof|_Alignof|alignof|_Generic|typeof|__typeof__|__alignof__|asm|__asm|__asm__)\(/\1 (/g
' {} \;

PR #1595, opened 2026-03-22
lib/attr.h: exclude MCST LCC compiler from ATTR_MALLOC

a1batross:
This particular compiler defines GCC compatibility like Clang but doesn't support this attribute.

### DIFF
--- a/lib/attr.h
+++ b/lib/attr.h
@@ -40,7 +40,7 @@
 # define ATTR_ALLOC_SIZE(...)
 #endif
 
-#if (__GNUC__ >= 11) && !defined(__clang__)
+#if (__GNUC__ >= 11) && !defined(__clang__) && !defined(__MCST__)
 # define ATTR_MALLOC(deallocator)    [[gnu::malloc(deallocator)]]
 #else
 # define ATTR_MALLOC(deallocator)


alejandro-colomar:
> This particular compiler defines GCC compatibility like Clang but doesn't support this attribute.

Thanks!

```
```

BTW, not relevant for this PR, but I'm curious about how much compatible it is with GCC.  Does that compiler support GCC's forward declarations of function parameters?  Those are documented here: <https://gcc.gnu.org/onlinedocs/gcc/Variable-Length.html>

a1batross:
> BTW, not relevant for this PR, but I'm curious about how much compatible it is with GCC. Does that compiler support GCC's forward declarations of function parameters? Those are documented here: https://gcc.gnu.org/onlinedocs/gcc/Variable-Length.html

Doesn't seem so, unfortunately: https://ce.mentality.rip/z/TeKqjE. Even though I have quite outdated version here, on the current version it's the same. They use proprietary EDG frontend (reported as 6.0.4 in `__EDG_VERSION__`), and it honestly not that great in terms of GNU C compatibility and newer standards, even though it declares compatibility with them.

alejandro-colomar:
> > BTW, not relevant for this PR, but I'm curious about how much compatible it is with GCC. Does that compiler support GCC's forward declarations of function parameters? Those are documented here: https://gcc.gnu.org/onlinedocs/gcc/Variable-Length.html
> 
> Doesn't seem so, unfortunately: https://ce.mentality.rip/z/TeKqjE. Even though I have quite outdated version here, on the current version it's the same. They use proprietary EDG frontend (reported as 6.0.4 in `__EDG_VERSION__`), and it honestly not that great in terms of GNU C compatibility and newer standards, even though it declares compatibility with them.

Hmmm; you may want to request that they implement this.  I have a PR for using such syntax in the project.

a1batross:
> > > BTW, not relevant for this PR, but I'm curious about how much compatible it is with GCC. Does that compiler support GCC's forward declarations of function parameters? Those are documented here: https://gcc.gnu.org/onlinedocs/gcc/Variable-Length.html
> > 
> > 
> > Doesn't seem so, unfortunately: https://ce.mentality.rip/z/TeKqjE. Even though I have quite outdated version here, on the current version it's the same. They use proprietary EDG frontend (reported as 6.0.4 in `__EDG_VERSION__`), and it honestly not that great in terms of GNU C compatibility and newer standards, even though it declares compatibility with them.
> 
> Hmmm; you may want to request that they implement this. I have a PR for using such syntax in the project.

Unfortunately, Edison Design Group is winding down their operations, but at least they have plans to opensource their C/C++ frontend this year as announced previously.

I'm usually all for copyleft licensing but MCST does not want to opensource their codegen, so it would be nice if EDG published their frontend under permissive licensing.

Also, MCST LCC compilers devs do not usually add or fix anything in EDG themselves. I still could try leaving them a bugreport on this; thanks for letting me know.

PR #1597, opened 2026-03-23
Tests: Add a new user with a specified non-existing primary group

aborah-sudo:
This is the transformation to Python of the test located in `tests/usertools/01/06_useradd_invalid_named_primary_group.test`, which checks that `useradd` can not add a new user with invalid existing group as primary

### DIFF
--- a/tests/system/tests/test_useradd.py
+++ b/tests/system/tests/test_useradd.py
@@ -331,11 +331,18 @@ def test_useradd__add_user_with_existing_uid(shadow: Shadow):
 
 
 @pytest.mark.topology(KnownTopology.Shadow)
-def test_useradd__invalid_numeric_primary_group(shadow: Shadow):
+@pytest.mark.parametrize(
+    "group_identifier",
+    [
+        pytest.param(4242, id="invalid_numeric_gid"),
+        pytest.param("nekral", id="invalid_named_group"),
+    ],
+)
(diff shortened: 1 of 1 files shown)


tests/system/tests/test_useradd.py
```diff
@@ -331,11 +331,18 @@ def test_useradd__add_user_with_existing_uid(shadow: Shadow):
 
 
 @pytest.mark.topology(KnownTopology.Shadow)
-def test_useradd__invalid_numeric_primary_group(shadow: Shadow):
+@pytest.mark.parametrize(
+    "group_identifier",
+    [
+        pytest.param(4242, id="invalid_numeric_gid"),
+        pytest.param("nekral", id="invalid_named_group"),
```
ikerexxe, line 338: What's the purpose of `id`? I don't see it in the code

aborah-sudo: tests/user_add.py::test_useradd__invalid_primary_group[invalid_numeric_gid] (shadow) PASSED                                                                                                                                            
tests/user_add.py::test_useradd__invalid_primary_group[invalid_named_group] (shadow) PASSED

while test will run the id will used to distinguish the test cases

ikerexxe: aaa nice! It's good to see that it has its purpose

PR #1599, opened 2026-03-24
login: Clean up PAM resource handling

stoeckmann:
- login opens session first, then sets credentials. This follows the old [X/Open RFC 86.0](http://www.opengroup.org/rfc/mirror-rfc/rfc86.0.txt) Appendix B example but violates Linux-PAM's `pam_setcred(3)` manual page. It also violates su's cred/session order. Other examples of cred/session tools: util-linux login and su, and gdm
- Do not call `pam_close_session` if `pam_open_session` was never called
- Delete credentials if `pam_open_session` fails

See pam_setcred(3):
https://github.com/linux-pam/linux-pam/blob/cf2fc5ff7b4a8555fda2a5ebe5f6ab0e45c22996/doc/man/pam_setcred.3.xml#L33-L43

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -801,7 +801,13 @@ int main (int argc, char **argv)
  	/* Open the PAM session */
 	get_pam_user (&pam_user);
 	retcode = pam_open_session (pamh, hushed (pam_user) ? PAM_SILENT : 0);
-	PAM_FAIL_CHECK;
+	if (retcode != PAM_SUCCESS) {
+		fprintf(stderr,"\n%s\n", pam_strerror(pamh, retcode));
+		SYSLOG(LOG_ERR,"%s",pam_strerror(pamh, retcode));
+		pam_setcred(pamh, PAM_DELETE_CRED);
+		pam_end(pamh, retcode);
+		exit(1);
+	}
 
 #else				/* ! USE_PAM */
 	while (true) {	/* repeatedly get login/password pairs */


alejandro-colomar:
> * login opens session first, then sets credentials. This follows an old RFC but violates Linux-PAM's `pam_setcred(3)` manual page.

It might be good to mention the exact RFC (and possibly section thereof), and also in the commit message.

> It also violates su's cred/session order. Other examples of cred/session tools: util-linux login and su, and gdm
> 
>     * Do not call `pam_close_session` if `pam_open_session` was never called
> 
>     * Delete credentials if `pam_open_session` fails
> 
> 
> See pam_setcred(3): https://github.com/linux-pam/linux-pam/blob/cf2fc5ff7b4a8555fda2a5ebe5f6ab0e45c22996/doc/man/pam_setcred.3.xml#L33-L43

Thanks!

src/login.c
```diff
@@ -766,11 +766,6 @@ int main (int argc, char **argv)
 	}
 	PAM_FAIL_CHECK;
 
-	/* Open the PAM session */
-	get_pam_user (&pam_user);
-	retcode = pam_open_session (pamh, hushed (pam_user) ? PAM_SILENT : 0);
-	PAM_FAIL_CHECK;
-
```
alejandro-colomar, line 773: From commit message:

> According to Linux-PAM's pam_setcred manual page and shadow's su
implementation, credentials shall be set before session is opened.

Please refer to manual pages with the usual syntax: foo(3) for function foo in chapter 3 of the manual.

alejandro-colomar: In this case, it seems to be pam_setcred(3).

```sh
$ apt-file find pam_setcred
libpam-doc: /usr/share/man/man3/pam_setcred.3.gz
```

src/login.c
```diff
@@ -805,6 +800,11 @@ int main (int argc, char **argv)
 	 * into account.
 	 */
 
+ 	/* Open the PAM session */
+	get_pam_user (&pam_user);
+	retcode = pam_open_session (pamh, hushed (pam_user) ? PAM_SILENT : 0);
+	PAM_FAIL_CHECK;
+
```
alejandro-colomar, line 811: For commit 1, except for the commit message itself:

```
```

src/login.c
```diff
@@ -698,13 +696,13 @@ int main (int argc, char **argv)
 				fprintf (stderr,
 				         _("Maximum number of tries exceeded (%u)\n"),
 				         failcount);
-				PAM_END;
+				(void) pam_end(pamh, retcode);
```
alejandro-colomar: I'd like to get rid of the cast.  Since this is not an exact move, would you mind either:

-  A pre-patch that removes the cast within the macro.
or
-  Just remove the cast in the same commit.

?

alejandro-colomar: Regardless of this, for commit 2:

```
```

src/login.c
```diff
@@ -801,7 +801,13 @@ int main (int argc, char **argv)
  	/* Open the PAM session */
 	get_pam_user (&pam_user);
 	retcode = pam_open_session (pamh, hushed (pam_user) ? PAM_SILENT : 0);
-	PAM_FAIL_CHECK;
+	if (retcode != PAM_SUCCESS) {
+		fprintf(stderr,"\n%s\n", pam_strerror(pamh, retcode));
+		SYSLOG(LOG_ERR,"%s",pam_strerror(pamh, retcode));
+		(void) pam_setcred(pamh, PAM_DELETE_CRED);
```
alejandro-colomar: For commit 3:

```
```

src/login.c
stoeckmann: Adjusted

src/login.c
stoeckmann: Added

src/login.c
stoeckmann: Adjusted

src/login.c
stoeckmann: Added

alejandro-colomar (approved):
LGTM.  Thanks!

PR #1600, opened 2026-03-25
Tests: Add a new user with a specified existing numerical primary group

aborah-sudo:
This is the transformation to Python of the test located in `tests/usertools/01/07_useradd_numerical_primary_group.test`, which checks that `useradd` can add a new user with valid existing group as primary

### DIFF
--- a/tests/system/tests/test_useradd.py
+++ b/tests/system/tests/test_useradd.py
@@ -370,7 +370,14 @@ def test_useradd__invalid_primary_group(shadow: Shadow, group_identifier: int | 
 
 
 @pytest.mark.topology(KnownTopology.Shadow)
-def test_useradd__valid_group_as_primary(shadow: Shadow):
+@pytest.mark.parametrize(
+    "group_name, group_gid",
+    [
+        pytest.param("testgroup", None, id="valid_named_group"),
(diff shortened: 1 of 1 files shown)


tests/system/tests/test_useradd.py
```diff
@@ -385,13 +392,19 @@ def test_useradd__valid_group_as_primary(shadow: Shadow):
...
+    assert group_entry.name == group_name, f"Incorrect group name, expected '{group_name}', got '{group_entry.name}'"
 
-    group_entry = shadow.tools.getent.group("testgroup")
-    assert group_entry is not None, "Group testgroup should exist"
-    assert group_entry.name == "testgroup", f"Incorrect group name, expected 'testgroup', got '{group_entry.name}'"
+    if group_gid:
+        assert group_entry.gid == group_gid, f"Incorrect GID, expected {group_gid}, got {group_entry.gid}"
 
-    shadow.useradd("testuser -g testgroup")
+    shadow.useradd(f"testuser -g {user_group_arg}")
```
ikerexxe: Let's use the `group_gid` condition to create two separate paths for adding the user. Something similar to what you did to add the group. This way we don't need an additional argument (i.e. `user_group_arg`)

Issue #1598, opened 2026-03-24
recent strchriscntrl change


grawity:
Hello,

It seems to me that the recent commit 19d725da3c61d4785aec08766408ce32a81a0a0d "reject C1 control bytes" is misguided. When the terminals work in UTF-8 mode, they aren't interpreting C1 control *bytes*; they are interpreting C1 control *codepoints*, and even the example in the commit message uses UTF-8 encoded codepoints (U+009B encoded as 0xC2 0x9B) rather than single bytes. Yet the function filters raw bytes, whenever they happen to appear as part of an UTF-8 sequence.

However, while 0xC2 0x97 is indeed the UTF-8 encoding of the U+0097 control character, **0xC4** 0x97 is instead the UTF-8 encoding of the letter `ė`: not a control character, and not interpreted as one by any UTF-8 capable terminal, but will be blocked by strchriscntrl() regardless. As a result, `chfn -f` now rejects my last name as invalid.

The function needs to decode input per the locale codepage and check whether the resulting _wchar_ is in the C1 control range.

(If the user uses a UTF-8 locale but an administrator does not, I don't think that is solvable – it sounds like a self-inflicted problem – and best I can suggest is "use `cat -v` when reading untrusted files".)

alejandro-colomar:
Cc: @KhaelK138, @hallyn , @ikerexxe

alejandro-colomar:
> Hello,
> 
> It seems to me that the recent commit [19d725d](https://github.com/shadow-maint/shadow/commit/19d725da3c61d4785aec08766408ce32a81a0a0d) "reject C1 control bytes" is misguided. When the terminals work in UTF-8 mode, they aren't interpreting C1 control _bytes_; they are interpreting C1 control _codepoints_, and even the example in the commit message uses UTF-8 encoded codepoints (U+009B encoded as 0xC2 0x9B) rather than single bytes. Yet the function filters raw bytes, whenever they happen to appear as part of an UTF-8 sequence.
> 
> However, while 0xC2 0x97 is indeed the UTF-8 encoding of the U+0097 control character, **0xC4** 0x97 is instead the UTF-8 encoding of the letter `ė`: not a control character, and not interpreted as one by any UTF-8 capable terminal, but will be blocked by strchriscntrl() regardless.

Hmmm, it's funny how they made it difficult to handle those control characters.

> As a result, `chfn -f` now rejects my last name as invalid.

[...]

PR #1602, opened 2026-03-25
Revert "strchriscntrl: reject C1 control bytes (0x80-0x9F)"

alejandro-colomar:
C1 control bytes are more complicated than that.  They're represented as two bytes in UTF-8.

Commit 19d725da, has issues, rejecting otherwise valid UTF-8 multi-byte characters.

We could in theory do correct parsing of UTF, possibly parsing the multi-byte sequences, or translating to wchar_t.  However, that would complicate the source code well beyond what I'd be comfortable with.

Instead, let's revert this, and claim no intention to support UTF-8. If an admin uses a UTF-8 locale while reading /etc/passwd, that's their own fault.

Reverts: 19d725da (2026-03-13; "strchriscntrl: reject C1 control bytes (0x80-0x9F)")
Fixes: 19d725da (2026-03-13; "strchriscntrl: reject C1 control bytes (0x80-0x9F)")
Closes: <https://github.com/shadow-maint/shadow/issues/1598>
Reported-by: @grawity
Cc: @KhaelK138
Cc: @stoeckmann

---

Revisions:

<details>
<summary>v2</summary>

-  Add CAVEATS to passwd(5), documenting that the C locale should be used when reading /etc/passwd.

```
$ git rd 
1:  a2a94f2d = 1:  a2a94f2d Revert "strchriscntrl: reject C1 control bytes (0x80-0x9F)"
-:  -------- > 2:  61263b74 man/passwd.5.xml: CAVEATS: Add section
```
</details>

### DIFF
--- a/man/passwd.5.xml
+++ b/man/passwd.5.xml
@@ -163,6 +163,20 @@
     </variablelist>
   </refsect1>
 
+  <refsect1 id='caveats'>
+    <title>CAVEATS</title>
+    <para>
+      Unprivileged users may write
+      arbitrary non-ASCII characters
+      in the GECOS field,
+      notably including C1 control characters.
+      When reading this file,
+      use the C locale.
+      Otherwise,
+      the contents may look different than they are.
+    </para>
+  </refsect1>
+
   <refsect1 id='see_also'>
     <title>SEE ALSO</title>
     <para>


ikerexxe:
Is there any place where `claim no intention to support UTF-8` is stated?

alejandro-colomar:
> Is there any place where `claim no intention to support UTF-8` is stated?

I guess we could use release notes.  And/or maybe we should add something to passwd(5) in a CAVEATS section.

ikerexxe:
I prefer the second option, as people are more likely to read it.

Do you want to use this PR to update it, or shall we leave it for later?

alejandro-colomar:
> I prefer the second option, as people are more likely to read it.
> 
> Do you want to use this PR to update it, or shall we leave it for later?

I'll update this PR.

alejandro-colomar:
> > I prefer the second option, as people are more likely to read it.
> > Do you want to use this PR to update it, or shall we leave it for later?
> 
> I'll update this PR.

Done.

PR #1606, opened 2026-04-07
Tests: Tests for large UID scenarios,  supplementary groups, user wit…

aborah-sudo:
…h home directory creation

This is the transformation to Python of the test located in `tests/usertools/01/25_useradd_specified_large_UID2.test`, `tests/usertools/01/26_useradd_UID_-1.test`
which checks that `useradd` can add a new user with large UID

### DIFF
--- a/tests/system/tests/test_useradd.py
+++ b/tests/system/tests/test_useradd.py
@@ -454,26 +454,52 @@ def test_useradd__invalid_uid(shadow: Shadow, uid_value: int, expected_error: st
 
 
 @pytest.mark.topology(KnownTopology.Shadow)
-def test_useradd__specific_large_uid(shadow: Shadow):
+@pytest.mark.parametrize(
+    "uid_value, should_fail, expected_rc, expected_error",
+    [
+        pytest.param(2147483647, False, None, None, id="maximum_signed_32bit_uid"),
+        pytest.param(4294967294, False, None, None, id="large_uid_4294967294"),
+        pytest.param(4294967295, True, 3, "useradd: invalid user ID '4294967295'", id="uid_4294967295_fail"),
+    ],
+)
+def test_useradd__large_uid_scenarios(
+    shadow: Shadow, uid_value: int, should_fail: bool, expected_rc: int | None, expected_error: str | None
+):
     """
-    :title: Verify user creation at the upper boundary for UID
+    :title: Tests for large UID scenarios
+    :setup:
+        1. Create user with specified UID value
     :steps:
-        1. Create user with UID 2147483647 (2^31 - 1, maximum signed 32-bit integer)
-        2. Check passwd entry
-        3. Check group entry
+        1. Attempt to create user with UID and verify command outcome based on expected behavior
+        2. Check passwd and group entries if creation succeeds
+        3. Verify no user is created if command should fail
     :expectedresults:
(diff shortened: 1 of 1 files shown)


tests/system/tests/test_useradd.py
```diff
@@ -454,26 +454,52 @@ def test_useradd__invalid_uid(shadow: Shadow, uid_value: int, expected_error: st
...
-def test_useradd__specific_large_uid(shadow: Shadow):
+@pytest.mark.parametrize(
+    "uid_value, should_fail, expected_rc, expected_error",
+    [
+        pytest.param(2147483647, False, None, None, id="maximum_signed_32bit_uid"),
+        pytest.param(4294967294, False, None, None, id="large_uid_4294967294"),
+        pytest.param(4294967295, True, 3, "useradd: invalid user ID '4294967295'", id="uid_4294967295_fail"),
+    ],
+)
+def test_useradd__large_uid_scenarios(
```
ikerexxe: I really like the use of parametrization to keep things DRY! However, in this case, the complexity is outweighing the benefits for readability and maintenance.

Since the logic inside the test is already split between success and failure cases, I’d suggest breaking this into two separate tests:
* A happy path test for the successful scenario.
* An error path test for the failure scenarios.

This would align with the existing structure of your code and allow for more specific, descriptive docstrings for each.

tests/system/tests/test_useradd.py
```diff
@@ -503,3 +503,34 @@ def test_useradd__large_uid_scenarios(
...
+        3. Check group entry
+        4. Verify home directory exists
+    :expectedresults:
+        1. User is created successfully
+        2. passwd entry exists with correct attributes
+        3. group entry exists
+        4. Home directory is created at /home/test1
+    :customerscenario: False
+    """
+    shadow.useradd("--create-home test1")
```
ikerexxe, line 472: As it stands now `/etc/login.defs` already contains `CREATE_HOME yes`, thus you can't guarantee that the `home` directory is created by `--create-home` option. You should add a previous `setup` step, where `CREATE_HOME` is set to `no`.

By the way, I don't think the test framework can manage the option change, so you should probably implement it. You can probably copy this from SSSD.

tests/system/tests/test_useradd.py
```diff
@@ -534,3 +534,70 @@ def test_useradd__create_homedir(shadow: Shadow):
...
+            "test1 -g nogroup -G src,daemon,bin",
+            "nogroup",
+            ["src", "daemon", "bin"],
+            True,
+            ["nogroup", "src", "daemon", "bin"],
+            id="add_user_to_multiple_groups",
+        ),
+    ],
+)
+def test_useradd__supplementary_groups(
```
ikerexxe: I definitely think we should split this.

Currently, the test has to use `if/else` logic and 5 different parameters to handle the differences between default primary groups and explicit ones. This makes the test body harder to follow.

If we split this into a `test_useradd_default_primary_group` and a `test_useradd_custom_primary_and_supplementary_groups`, we can remove the `if/else` logic and the `verify_primary` toggle entirely. It would make the assertions much more direct!

aborah-sudo:
New PR: 

https://github.com/shadow-maint/shadow/pull/1607

PR #1608, opened 2026-04-09
groupadd: fix system account help text

twihno:
We found a little error in the description of `groupadd` where a line from the `useradd` help text found its way into the wrong place. This change matches the man page content (at least in english) and should fix it in the `en` and `de` locales


alejandro-colomar (approved):
Seems good.  Thanks!

PR #1609, opened 2026-04-14
Bugfix: Critical sysuser home directory creation failure with BTRFS_SUBVOLUME_HOME=yes on non-Btrfs paths

silverhadch:
This fixes a critical bug. Previously, Btrfs subvolume creation was only applied on-demand, but now it's forced for all users whenever BTRFS_SUBVOLUME_HOME=yes is set. That logic is too aggressive: it causes system users with home directories in places like /var or /run to fail because those paths aren't Btrfs-backed. KDE Linux fails to boot with Btrfs home enabled because sysuser configurations require a home directory on tmpfs, yet the option forces a Btrfs subvolume even for those state directories. Without this patch, sysusers aren't created properly, as seen in this failing pipeline: https://invent.kde.org/kde-linux/kde-linux/-/pipelines/1215866. With the patch, the system falls back to a regular directory when a Btrfs home is requested by config or flag but the target path isn't Btrfs-backed, and it throws a warning to stderr. That allows sysusers on tmpfs or other non-Btrfs state directories to be created successfully, and the pipeline passes: https://invent.kde.org/kde-linux/kde-linux/-/pipelines/1216261. This resolves the boot failure, fixes Plasma login and setup, and generally prevents breakage for any sysuser with a home on a non-Btrfs filesystem with said option set.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -1291,6 +1291,9 @@ msgstr ""
 msgid "  -r, --system                  create a system account\n"
 msgstr "  -r, --system                  Ein Systemkonto erstellen\n"
 
+msgid "  -r, --system                  create a system group\n"
+msgstr "  -r, --system                  Eine Systemgruppe erstellen\n"
+
 #, fuzzy
 #| msgid "  -l, --list                    list the members of the group\n"
 msgid ""
--- a/po/shadow.pot
+++ b/po/shadow.pot
@@ -1131,6 +1131,9 @@ msgstr ""
 msgid "  -r, --system                  create a system account\n"
 msgstr ""
 
+msgid "  -r, --system                  create a system group\n"
+msgstr ""
+
 msgid ""
 "  -U, --users USERS             comma-separated list of users to add as\n"
 "\t                         members of this group\n"
--- a/src/groupadd.c
+++ b/src/groupadd.c
@@ -116,7 +116,7 @@ usage (int status)
 	(void) fputs (_("  -o, --non-unique              allow to create groups with duplicate\n"
 	                "                                (non-unique) GID\n"), usageout);
 	(void) fputs (_("  -p, --password PASSWORD       use this encrypted password for the new group\n"), usageout);
-	(void) fputs (_("  -r, --system                  create a system account\n"), usageout);
+	(void) fputs (_("  -r, --system                  create a system group\n"), usageout);
 	(void) fputs (_("  -R, --root CHROOT_DIR         directory to chroot into\n"), usageout);
 	(void) fputs (_("  -P, --prefix PREFIX_DIR       directory prefix\n"), usageout);
 	(void) fputs (_("  -U, --users USERS             comma-separated list of users to add as\n"


silverhadch:
CC: @Conan-Kudo

Conan-Kudo:
The commit message needs to be reformatted so that it's 70 chars for the title and the rest of the information needs to be in the commit body (wrapped at 79 chars).

Conan-Kudo (approved):

src/useradd.c
```diff
@@ -2273,28 +2273,35 @@ static void create_home(const struct option_flags *flags)
 				fail_exit(E_HOMEDIR, process_selinux);
 			}
 			stpcpy(&btrfs_check[strlen(path) - strlen(cp) - 1], "");
-			if (is_btrfs(btrfs_check) <= 0) {
+			if (is_btrfs(btrfs_check) > 0) {
```
ikerexxe: I think I should have noticed this before, but I'm doing it now. is_btrfs() returns three values:

1 = BTRFS filesystem
0 = Not BTRFS
-1 = Error (permission denied, I/O error, etc.)

This is causing silent failures that mask real system problems. I'd highly recommend you to return just two values and log a message (and even failing hard) when an error happens.

src/useradd.c
```diff
@@ -2273,28 +2273,35 @@ static void create_home(const struct option_flags *flags)
...
+				/* Inform about fallback */
 				fprintf(stderr,
-					_("%s: home directory \"%s\" must be mounted on BTRFS\n"),
-					Prog, path);
-				fail_exit(E_HOMEDIR, process_selinux);
+					_("%s: warning: \"%s\" is not on BTRFS, creating regular directory instead of subvolume\n"),
+					Prog, prefix_user_home);
+				SYSLOG(LOG_WARN,
+				       "useradd: fallback to regular directory for %s (BTRFS not available at %s)",
+				       user_name, prefix_user_home);
```
ikerexxe: You should document this behaviour

silverhadch: Done.

src/useradd.c
```diff
@@ -2273,27 +2275,44 @@ static void create_home(const struct option_flags *flags)
...
+			int btrfs_ret = is_btrfs(btrfs_check);
+			if (btrfs_ret > 0) {
+				free(btrfs_check);
+				// make subvolume for user instead of directory
+				if (btrfs_create_subvolume(path)) {
+					fprintf(stderr,
+						_("%s: failed to create BTRFS subvolume: %s\n"),
+						Prog, path);
+					fail_exit(E_HOMEDIR, process_selinux);
+				}
```
alejandro-colomar: Please add relevant `Fixes:` trailer tags in the commit message.

See <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/patches/trailer#n16>
and <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/git#n46>.

man/useradd.8.xml
```diff
@@ -156,8 +157,23 @@
...
+	  </para>
+	  <para>
+	    If the filesystem type cannot be determined (e.g., because of
+	    insufficient permissions, an I/O error, or a
+	    <citerefentry><refentrytitle>statfs</refentrytitle>
+	    <manvolnum>2</manvolnum></citerefentry> failure),
+	    <command>useradd</command> treats this as a fatal error:
+	    the home directory is not created, the command exits with a
+	    non‑zero status (<literal>E_HOMEDIR</literal>, 12), and an
+	    error message is printed.
```
alejandro-colomar: Please use semantic newlines.  See man-pages(7):

```sh
   Use semantic newlines

```

src/useradd.c
```diff
@@ -2258,10 +2259,11 @@ static void create_home(const struct option_flags *flags)
...
 
-		/* Check if parent directory is BTRFS, fail if requesting
-		   subvolume but no BTRFS. The paths could be different by the
-		   trailing slash
-		 */
+		/* If requesting a subvolume and parent is BTRFS, create one.
+		   Otherwise fall back to a regular directory.
+		   However, if is_btrfs() returns an error, we treat it as a
+		   fatal error and abort. */
+		bool dir_created = false;
```
alejandro-colomar: Please separate the type from the variable name with two spaces.  Also, please declare them at the top of the block (right after the `for`).  See <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c#n100>:

```

    Variables
	Variable should be declared at the top of the block in which
	they are used.  That is, use C89 declarations.  The exception is
	loop variables; we use C99 for-loop variable declarations.

	The '*' goes with the variable name, not with the type name.
	Except if the pointer is qualified, in which case the '*' goes
	with the type name.

		int     i;
		char    c;
		char    *p;
		size_t  size;
```

src/useradd.c
```diff
@@ -2273,27 +2275,44 @@ static void create_home(const struct option_flags *flags)
...
+				free(btrfs_check);
+				// make subvolume for user instead of directory
+				if (btrfs_create_subvolume(path)) {
+					fprintf(stderr,
+						_("%s: failed to create BTRFS subvolume: %s\n"),
+						Prog, path);
+					fail_exit(E_HOMEDIR, process_selinux);
+				}
+				dir_created = true;
+			} else if (btrfs_ret == 0) {
```
alejandro-colomar: Please put the error cases first, so that we don't need to `else` (since they call exit(3)).

src/useradd.c
```diff
@@ -2273,27 +2275,44 @@ static void create_home(const struct option_flags *flags)
...
-		else
 #endif
-		if (mkdir(path, 0) != 0) {
-			fprintf(stderr, _("%s: cannot create directory %s\n"),
-				Prog, path);
-			fail_exit(E_HOMEDIR, process_selinux);
+		if (!dir_created) {
+			if (mkdir(path, 0) != 0) {
+				fprintf(stderr, _("%s: cannot create directory %s\n"),
+					Prog, path);
```
alejandro-colomar: Let's simplify the error string:

```c
fprintf(stderr, "%s: mkdir(\"%s\"): %s\n", Prog, path, strerrno());
```

This avoids having to translate the string.

src/useradd.c
```diff
@@ -2273,27 +2275,44 @@ static void create_home(const struct option_flags *flags)
...
 				fail_exit(E_HOMEDIR, process_selinux);
 			}
 		}
-		else
 #endif
-		if (mkdir(path, 0) != 0) {
-			fprintf(stderr, _("%s: cannot create directory %s\n"),
-				Prog, path);
-			fail_exit(E_HOMEDIR, process_selinux);
+		if (!dir_created) {
```
alejandro-colomar, line 2313: If I'm reading correctly, all paths that arrive here have `dir_created` set to true, because the others call fail_exit().  Am I correct?  If so, this looks like dead code.

man/useradd.8.xml
```diff
@@ -184,7 +207,7 @@
 	    user's login directory. The default is to append the
 	    <replaceable>LOGIN</replaceable> name to
 	    <replaceable>BASE_DIR</replaceable> and use that as the
-	    login directory name.  
+	    login directory name.
```
alejandro-colomar, line 210: Please fix those spaces in a separate commit.

src/useradd.c
```diff
@@ -2273,27 +2276,46 @@ static void create_home(const struct option_flags *flags)
 				fail_exit(E_HOMEDIR, process_selinux);
 			}
 			stpcpy(&btrfs_check[strlen(path) - strlen(cp) - 1], "");
-			if (is_btrfs(btrfs_check) <= 0) {
+			int btrfs_ret = is_btrfs(btrfs_check);
+
+			if (btrfs_ret < 0) {
```
alejandro-colomar, line 2281: Please check explicitly against `== -1` for the error code.  It's more readable (it very explicitly tells that it's an error code, and not just a negative mathematical value such as -42).

src/useradd.c
```diff
@@ -2273,27 +2276,46 @@ static void create_home(const struct option_flags *flags)
...
+				/* is_btrfs() returned an error – treat as fatal */
```
alejandro-colomar, line 2282: Please remove this superfluous comment.

src/useradd.c
```diff
@@ -2273,27 +2276,46 @@ static void create_home(const struct option_flags *flags)
...
+				free(btrfs_check);
 				fprintf(stderr,
-					_("%s: home directory \"%s\" must be mounted on BTRFS\n"),
-					Prog, path);
+					_("%s: cannot determine filesystem type for %s: %s\n"),
+					Prog, path, strerrno());
+				SYSLOG(LOG_ERR, "is_btrfs() failed for %s: %s",
```
alejandro-colomar, line 2287: Let's simplify the error string:

```c
"is_btrfs(\"%s\"): %s\n"
```

src/useradd.c
```diff
@@ -2273,27 +2276,46 @@ static void create_home(const struct option_flags *flags)
...
+				SYSLOG(LOG_WARN,
+				       "useradd: fallback to regular directory for %s (parent not on BTRFS)",
+				       user_name);
+			} else { /* btrfs_ret > 0 */
+				free(btrfs_check);
+				/* make subvolume for user instead of directory */
+				if (btrfs_create_subvolume(path)) {
+					fprintf(stderr,
+						_("%s: failed to create BTRFS subvolume: %s\n"),
+						Prog, path);
```
alejandro-colomar, line 2306: Let's simplify the error string:

```c
fprintf(stderr, "%s: btrfs_create_subvolume(\"%s\"): %s\n", Prog, path, strerrno())
```

src/useradd.c
```diff
@@ -2273,27 +2276,46 @@ static void create_home(const struct option_flags *flags)
...
-					Prog, path);
-				fail_exit(E_HOMEDIR, process_selinux);
+					_("%s: warning: \"%s\" is not on BTRFS, creating regular directory instead of subvolume\n"),
+					Prog, prefix_user_home);
+				SYSLOG(LOG_WARN,
+				       "useradd: fallback to regular directory for %s (parent not on BTRFS)",
+				       user_name);
+			} else { /* btrfs_ret > 0 */
+				free(btrfs_check);
+				/* make subvolume for user instead of directory */
```
alejandro-colomar, line 2302: Please remove superfluous comment.

src/useradd.c
```diff
@@ -2273,27 +2276,46 @@ static void create_home(const struct option_flags *flags)
...
 				fprintf(stderr,
-					_("%s: failed to create BTRFS subvolume: %s\n"),
-					Prog, path);
-				fail_exit(E_HOMEDIR, process_selinux);
+					_("%s: warning: \"%s\" is not on BTRFS, creating regular directory instead of subvolume\n"),
+					Prog, prefix_user_home);
+				SYSLOG(LOG_WARN,
+				       "useradd: fallback to regular directory for %s (parent not on BTRFS)",
+				       user_name);
+			} else { /* btrfs_ret > 0 */
```
alejandro-colomar, line 2300: Please remove superfluous comment.

src/useradd.c
```diff
@@ -2273,27 +2276,46 @@ static void create_home(const struct option_flags *flags)
...
+				SYSLOG(LOG_ERR, "is_btrfs() failed for %s: %s",
+				       path, strerrno());
 				fail_exit(E_HOMEDIR, process_selinux);
 			}
-			free(btrfs_check);
-			// make subvolume to mount for user instead of directory
-			if (btrfs_create_subvolume(path)) {
+			if (btrfs_ret == 0) {
+				free(btrfs_check);
+				/* Not on Btrfs – fallback to regular directory */
```
alejandro-colomar, line 2293: Please remove superfluous comment.

src/useradd.c
```diff
@@ -2273,27 +2276,46 @@ static void create_home(const struct option_flags *flags)
...
 			}
 		}
-		else
 #endif
-		if (mkdir(path, 0) != 0) {
-			fprintf(stderr, _("%s: cannot create directory %s\n"),
-				Prog, path);
-			fail_exit(E_HOMEDIR, process_selinux);
+		if (!dir_created) {
+			if (mkdir(path, 0) != 0) {
```
alejandro-colomar, line 2314: Please check `== -1` explicitly.

alejandro-colomar: Ahh, sorry; this is old code.  Forget my comment.

src/useradd.c
```diff
@@ -2273,27 +2276,46 @@ static void create_home(const struct option_flags *flags)
...
-				fail_exit(E_HOMEDIR, process_selinux);
+					_("%s: warning: \"%s\" is not on BTRFS, creating regular directory instead of subvolume\n"),
+					Prog, prefix_user_home);
+				SYSLOG(LOG_WARN,
+				       "useradd: fallback to regular directory for %s (parent not on BTRFS)",
+				       user_name);
+			} else { /* btrfs_ret > 0 */
+				free(btrfs_check);
+				/* make subvolume for user instead of directory */
+				if (btrfs_create_subvolume(path)) {
```
alejandro-colomar, line 2303: Please compare `!= 0`, as it's not a boolean.

src/useradd.c
alejandro-colomar: Also, I think we should print the same in SYSLOG() and fprintf(3).  Shouldn't we?

src/useradd.c
```diff
@@ -2273,27 +2276,46 @@ static void create_home(const struct option_flags *flags)
 				fail_exit(E_HOMEDIR, process_selinux);
 			}
 			stpcpy(&btrfs_check[strlen(path) - strlen(cp) - 1], "");
-			if (is_btrfs(btrfs_check) <= 0) {
+			int btrfs_ret = is_btrfs(btrfs_check);
```
alejandro-colomar, line 2279: Please don't initialize variables with function calls.  Declare them at the top of the block, and later assign to them with the function call.

See <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c#n100>.

That is:

```c
if (...) {
    int  btrfs_ret;

    ...

    btrfs_ret = is_btrfs(...);
    ...
}
```

ikerexxe (commented):
@alejandro-colomar I noticed something potentially confusing in this PR that I wanted your thoughts on. When I see `if (is_btrfs(path))`, I naturally expect boolean semantics where any non-zero value means "true". But here, `-1` means "error" and should be handled differently than `1` (success).

This PR correctly handles it with `if (is_btrfs(btrfs_check) > 0)`, but the function name makes this non-obvious. A future developer might write `if (is_btrfs(path))` and accidentally treat  errors as "is BTRFS".

Should we consider renaming to something like `btrfs_check_filesystem()` or `get_btrfs_status()`?

silverhadch:
> @alejandro-colomar I noticed something potentially confusing in this PR that I wanted your thoughts on. When I see `if (is_btrfs(path))`, I naturally expect boolean semantics where any non-zero value means "true". But here, `-1` means "error" and should be handled differently than `1` (success).
> 
> This PR correctly handles it with `if (is_btrfs(btrfs_check) > 0)`, but the function name makes this non-obvious. A future developer might write `if (is_btrfs(path))` and accidentally treat errors as "is BTRFS".
> 
> Should we consider renaming to something like `btrfs_check_filesystem()` or `get_btrfs_status()`?

btrfs_check_filesystem() sounds about right. I could rename it in this PR.

alejandro-colomar:
> @alejandro-colomar I noticed something potentially confusing in this PR that I wanted your thoughts on. When I see `if (is_btrfs(path))`, I naturally expect boolean semantics where any non-zero value means "true". But here, `-1` means "error" and should be handled differently than `1` (success).

Agree.  It seems confusing.

> 
> This PR correctly handles it with `if (is_btrfs(btrfs_check) > 0)`, but the function name makes this non-obvious. A future developer might write `if (is_btrfs(path))` and accidentally treat errors as "is BTRFS".
> 
> Should we consider renaming to something like `btrfs_check_filesystem()` or `get_btrfs_status()`?

I like the first suggestion.

It should return 0 if it's btrfs, and non-zero (not success) if not.

man/useradd.8.xml
```diff
@@ -207,7 +207,7 @@
...
```
alejandro-colomar, line 210: For the second commit:

```
```

lib/btrfs.c
```diff
@@ -63,9 +63,11 @@ int btrfs_is_subvolume(const char *path)
...
 	int ret;
 
-	ret = is_btrfs(path);
-	if (ret <= 0)
-		return ret;
+	ret = btrfs_check_filesystem(path);
+	if (ret == -1)
+		return -1;
+	if (ret != 0)
+		return 0;
```
alejandro-colomar, line 70: Please do the refactors separately (prior to) the actual behavior change, in a separate commit.

man/useradd.8.xml
alejandro-colomar: This was only for the second commit (now third).

Commits 1 and 2 are not reviewed.  Please remove the Reviewed-by tags from them.

lib/btrfs.c
alejandro-colomar, line 70: I would prefer the refacctor to happen before the behavior change; that is, reorder commits 1 and 2.  Would that be possible?

src/useradd.c
```diff
@@ -2258,22 +2258,21 @@ static void create_home(const struct option_flags *flags)
 			continue;
 		}
 
-		/* Check if parent directory is BTRFS, fail if requesting
+/* Check if parent directory is BTRFS, fail if requesting
 		   subvolume but no BTRFS. The paths could be different by the
 		   trailing slash
 		 */
```
alejandro-colomar: What happened here?

src/useradd.c
```diff
@@ -2258,22 +2258,21 @@ static void create_home(const struct option_flags *flags)
...
 #if WITH_BTRFS
 		if (subvolflg && (strlen(prefix_user_home) - (int)strlen(path)) <= 1) {
 			char *btrfs_check = strdup(path);
-
 			if (!btrfs_check) {
```
alejandro-colomar, line 2269: Why?

silverhadch: Merge Conflicts from the reordering I didnt fix properly. Will revert it and keep it that way.

alejandro-colomar:
Please refactor *before* changing behavior.

src/useradd.c
alejandro-colomar: Hmmmm.  Could you please try again?  It will make reviewing much easier.

src/useradd.c
```diff
@@ -2245,8 +2246,8 @@ static void create_home(const struct option_flags *flags)
 #endif
 
 	/* Check for every part of the path, if the directory
-	   exists. If not, create it with permissions 755 and
-	   owner root:root.
+		 exists. If not, create it with permissions 755 and
+		 owner root:root.
```
alejandro-colomar: Why?

src/useradd.c
```diff
@@ -2273,27 +2272,42 @@ static void create_home(const struct option_flags *flags)
...
 				fprintf(stderr,
-					_("%s: failed to create BTRFS subvolume: %s\n"),
-					Prog, path);
+					"%s: is_btrfs(\"%s\"): %s\n",
+					Prog, path, strerrno());
+				SYSLOG(LOG_ERR,
+				       "%s: is_btrfs(\"%s\"): %s",
+				       Prog, path, strerrno());
 				fail_exit(E_HOMEDIR, process_selinux);
+			} else if (btrfs_ret > 0) {
```
alejandro-colomar: Please don't else after [[noreturn]].

silverhadch: But this Code is removed in the following Commit?

alejandro-colomar: But why add it at all?

src/useradd.c
```diff
@@ -2273,27 +2272,42 @@ static void create_home(const struct option_flags *flags)
...
 				fprintf(stderr,
-					_("%s: failed to create BTRFS subvolume: %s\n"),
-					Prog, path);
+					"%s: is_btrfs(\"%s\"): %s\n",
+					Prog, path, strerrno());
+				SYSLOG(LOG_ERR,
+				       "%s: is_btrfs(\"%s\"): %s",
+				       Prog, path, strerrno());
 				fail_exit(E_HOMEDIR, process_selinux);
+			} if (btrfs_ret > 0) {
```
alejandro-colomar: `} if` seems wrong.  It should definitely be on a new line.

src/useradd.c
```diff
@@ -2273,27 +2272,42 @@ static void create_home(const struct option_flags *flags)
...
-					Prog, path);
-				fail_exit(E_HOMEDIR, process_selinux);
-			}
+
+			int btrfs_ret = is_btrfs(btrfs_check);
 			free(btrfs_check);
-			// make subvolume to mount for user instead of directory
-			if (btrfs_create_subvolume(path)) {
+
+			if (btrfs_ret == -1) {
```
alejandro-colomar: Please remove that blank line.

src/useradd.c
```diff
@@ -2264,41 +2264,42 @@ static void create_home(const struct option_flags *flags)
...
 			}
 			stpcpy(&btrfs_check[strlen(path) - strlen(cp) - 1], "");
-
-			int btrfs_ret = is_btrfs(btrfs_check);
+			int  btrfs_ret = btrfs_check_filesystem(btrfs_check);
 			free(btrfs_check);
-
-			if (btrfs_ret == -1) {
+			switch (btrfs_ret) {
+			case -1:
```
alejandro-colomar: This seems to be undoing some changes from the previous commit.

FWIW, I still think the refactoring should go first, and the behavior change later.  That will make the commit changing behavior be as clean as possible.

silverhadch: Gonna do that tomorrow.

src/useradd.c
```diff
@@ -2258,42 +2259,56 @@ static void create_home(const struct option_flags *flags)
...
-		   subvolume but no BTRFS. The paths could be different by the
-		   trailing slash
-		 */
+		bool  dir_created = false;
+
 #if WITH_BTRFS
 		if (subvolflg && (strlen(prefix_user_home) - (int)strlen(path)) <= 1) {
 			char *btrfs_check = strdup(path);
-
 			if (!btrfs_check) {
```
alejandro-colomar, line 2269: Please don't remove this blank.

src/useradd.c
```diff
@@ -2258,42 +2259,56 @@ static void create_home(const struct option_flags *flags)
...
 				fail_exit(E_HOMEDIR, process_selinux);
 			}
 			stpcpy(&btrfs_check[strlen(path) - strlen(cp) - 1], "");
-			if (btrfs_check_filesystem(btrfs_check) != 0) {
-				fprintf(stderr,
-					_("%s: home directory \"%s\" must be mounted on BTRFS\n"),
-					Prog, path);
-				fail_exit(E_HOMEDIR, process_selinux);
-			}
+			int  btrfs_ret = btrfs_check_filesystem(btrfs_check);
```
alejandro-colomar: Please declare `btrfs_ret` at the top of the block.

src/useradd.c
```diff
@@ -2258,42 +2259,56 @@ static void create_home(const struct option_flags *flags)
...
+					fprintf(stderr,
+						"%s: btrfs_create_subvolume(\"%s\"): %s\n",
+						Prog, path, strerrno());
+					fail_exit(E_HOMEDIR, process_selinux);
+				}
+				dir_created = true;
+				break;
+			default:
+				fprintf(stderr,
+					_("%s: warning: \"%s\" is not on BTRFS, creating regular directory instead of subvolume\n"),
```
alejandro-colomar: `s/,/;/`

lib/btrfs.c
alejandro-colomar, line 70: For commit 1:

```
```

Thanks!

src/useradd.c
```diff
@@ -2258,42 +2259,56 @@ static void create_home(const struct option_flags *flags)
...
+					Prog, prefix_user_home);
+				SYSLOG(LOG_WARN,
+				       "useradd: fallback to regular directory for %s (parent not on BTRFS)",
+				       user_name);
+				break;
```
alejandro-colomar, line 2304: Should these error strings be more similar?

src/useradd.c
```diff
@@ -2258,42 +2259,56 @@ static void create_home(const struct option_flags *flags)
...
+				}
+				dir_created = true;
+				break;
+			default:
+				fprintf(stderr,
+					_("%s: warning: \"%s\" is not on BTRFS, creating regular directory instead of subvolume\n"),
+					Prog, prefix_user_home);
+				SYSLOG(LOG_WARN,
+				       "useradd: fallback to regular directory for %s (parent not on BTRFS)",
+				       user_name);
```
alejandro-colomar, line 2303: SYSLOG() calls never store the program name, from what I can see.

src/useradd.c
```diff
@@ -2258,42 +2259,56 @@ static void create_home(const struct option_flags *flags)
...
+			switch (btrfs_ret) {
+			case -1:
 				fprintf(stderr,
-					_("%s: failed to create BTRFS subvolume: %s\n"),
-					Prog, path);
+					"%s: btrfs_check_filesystem(\"%s\"): %s\n",
+					Prog, path, strerrno());
+				SYSLOG(LOG_ERR,
+				       "%s: btrfs_check_filesystem(\"%s\"): %s",
+				       Prog, path, strerrno());
```
alejandro-colomar: SYSLOG() calls never store the program name.  Let's remove it, for consistency.

src/useradd.c
alejandro-colomar: Should they?

silverhadch: Done.

src/useradd.c
```diff
@@ -2273,27 +2273,44 @@ static void create_home(const struct option_flags *flags)
...
+				}
+				dir_created = true;
+				break;
+			default:
+				fprintf(stderr,
+						_("%s: warning: \"%s\" is not on BTRFS; creating regular directory instead of subvolume\n"),
+						Prog, prefix_user_home);
+				SYSLOG(LOG_WARN,
+					   _("%s: warning: \"%s\" is not on BTRFS; creating regular directory instead of subvolume"),
+					   Prog, prefix_user_home);
```
alejandro-colomar: SYSLOG() shouldn't log the program name.

src/useradd.c
```diff
@@ -2258,11 +2259,10 @@ static void create_home(const struct option_flags *flags)
...
-		/* Check if parent directory is BTRFS, fail if requesting
-		   subvolume but no BTRFS. The paths could be different by the
-		   trailing slash
-		 */
+		bool  dir_created = false;
+
 #if WITH_BTRFS
+		int     btrfs_ret;
 		if (subvolflg && (strlen(prefix_user_home) - (int)strlen(path)) <= 1) {
 			char *btrfs_check = strdup(path);
```
alejandro-colomar: That variable should be declared within the block (after the `{`).

src/useradd.c
```diff
@@ -2258,11 +2259,10 @@ static void create_home(const struct option_flags *flags)
 			continue;
 		}
 
-		/* Check if parent directory is BTRFS, fail if requesting
-		   subvolume but no BTRFS. The paths could be different by the
-		   trailing slash
-		 */
+		bool  dir_created = false;
+
```
alejandro-colomar: This variable should be declared at the top of the block.

src/useradd.c
```diff
@@ -2273,27 +2273,44 @@ static void create_home(const struct option_flags *flags)
...
 				fprintf(stderr,
-					_("%s: failed to create BTRFS subvolume: %s\n"),
-					Prog, path);
+					"%s: btrfs_check_filesystem(\"%s\"): %s\n",
+					Prog, path, strerrno());
+				SYSLOG(LOG_ERR,
+				       "btrfs_check_filesystem(\"%s\"): %s",
+				       path, strerrno());
 				fail_exit(E_HOMEDIR, process_selinux);
+				break;
```
alejandro-colomar: BTW, I see there was no SYSLOG() call before.  If we want to add SYSLOG() calls, that should be done in a separate commit.  The commit that fixes the bug should exclusively fix the bug.

alejandro-colomar:
I've pushed a few changes to your branch, which resolve some issues.  I've also significantly changed the design of the function.  Now it's much simpler.

alejandro-colomar:
Oops, I pushed by accident the contents of the master branch, and that has closed the PR.  Would you mind reopening?  You should get the commits from <https://github.com/shadow-maint/shadow/pull/1613>.

alejandro-colomar:
Oh, I can re-open myself.  (No, it doesn't re-open; github is buggy.)

PR #1611, opened 2026-04-17
Tests: Add a new user with home directory creation

aborah-sudo:
This is the transformation to Python of the test located in `tests/usertools/01/17_useradd_create_homedir.test` which checks that `useradd` can add a new user with --create-home

Implement CREATE_HOME is set to no

### DIFF
--- a/tests/system/framework/hosts/shadow.py
+++ b/tests/system/framework/hosts/shadow.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from pathlib import PurePosixPath
 from typing import Any
 
+from pytest_mh import MultihostUtility
 from pytest_mh.conn import ProcessLogLevel
 
 from .base import BaseHost, BaseLinuxHost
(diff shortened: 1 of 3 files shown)


ikerexxe (commented):
Hi Anuj,

First, let’s get those framework changes moved into their own separate commit! We need to keep the framework evolution isolated from the test logic to keep our git history clean and easy to audit.

On the technical side:
* Simplify the logic by removing `self.host.discard_file()` for `/etc/login.defs`. Changes to that specific file don’t actually impact the test results, so there's no need for the extra overhead.
* The current implementation relies on some hacks that won't scale well. We need a more robust approach that handles its own cleanup. Check out the [SSSD implementation here](https://github.com/SSSD/sssd-test-framework/blob/master/sssd_test_framework/utils/sssd.py) for inspiration. I’m expecting a similar pattern where we can call something like `shadow.login_defs("CREATE_HOME") = "no"`. This will keep the code clean.
* Right now, your configuration changes are leaking into the next tests. Please ensure the implementation includes a proper teardown mechanism (like the SSSD framework does) so that every test starts with a fresh, predictable environment.

tests/system/framework/hosts/shadow.py
```diff
@@ -201,3 +202,87 @@ def discard_file(self, origin: str) -> None:
...
+            self._backup_path = "/etc/login.defs.backup"
+            self.host.conn.run(f"cp /etc/login.defs {self._backup_path}")
+
+    def teardown(self) -> None:
+        """
+        Restore the original /etc/login.defs file.
+        """
+        if self._backup_path and self.host.fs.exists(self._backup_path):
+            self.host.conn.run(f"mv {self._backup_path} /etc/login.defs")
+            self.host.discard_file("/etc/login.defs")
```
ikerexxe: This isn't needed. The objective of this `discard_file()` method is to remove a file from a check at the end of the test. It'll be easier to understand with an example.

If you add a new group it's expected that `/etc/group` and `/etc/gshadow` change, that's why we discard those files. Files like `/etc/passwd` on the contrary didn't change, that's why we don't discard it. If you don't discard it, during the teardown phase the framework compares this file to the backup it did during the setup phase. If `/etc/passwd` file changed because the test added a new group, then the test fails because this file shouldn't change.

shadow doesn't change `/etc/login.defs`, it's the system administrator who does it, that's why it doesn't make sense to check this file. Make sure to remove all instances of `discard_file(/etc/login.defs)`.

ikerexxe (commented):
The current implementation creates `LoginDefsConfig` in a way that bypasses pytest-mh's automatic lifecycle management.

The `LoginDefsConfig` is stored as `self._login_defs` (with leading underscore) and created lazily via `@property`. This means:
  - pytest-mh's metaclass ignores it (leading underscore signals "private")
  - `setup()` is never called automatically → no backup is created
  - `teardown()` must be called manually → error-prone and inconsistent

Changes needed:
1. Remove the `@property` and `_login_defs` attribute
2. Create the utility in `__init__` with no leading underscore
3. Use `.postpone_setup()` for lazy initialization -> `self.login_defs: LoginDefsConfig = LoginDefsConfig(self.host).postpone_setup()`
4. Remove the manual `self._login_defs.teardown()` call from `Shadow.teardown()`

This gives you the same clean API (`shadow.login_defs["key"] = "value"`) but with automatic framework-managed lifecycle. Look at how other utilities like `self.fs`, `self.firewall` etc. are integrated in `BaseLinuxRole.__init__`.

The SSSD framework uses this same pattern - utilities are discovered and managed automatically by the framework, not manually by each role.

tests/system/framework/hosts/shadow.py
```diff
@@ -201,3 +202,77 @@ def discard_file(self, origin: str) -> None:
...
+    def __delitem__(self, key: str) -> None:
+        """
+        Remove an option from /etc/login.defs (comment it out).
+
+        :param key: Option name to remove
+        :type key: str
+        """
+        self.logger.info(f"Removing {key} from /etc/login.defs on {self.host.hostname}")
+
+        self.host.conn.run(f"sed -i 's/^{key}\\s.*/#{key} &/' /etc/login.defs", log_level=ProcessLogLevel.Error)
```
ikerexxe: This `sed` pattern will transform `#CREATE_HOME yes/no` into `#CREATE_HOME CREATE_HOME yes/no`, which doesn't seem very clean. Consider improving this.

tests/system/framework/hosts/shadow.py
```diff
@@ -201,3 +202,77 @@ def discard_file(self, origin: str) -> None:
...
+        :type key: str
+        :param value: Value to set
+        :type value: str
+        """
+        self.logger.info(f"Setting {key}={value} in /etc/login.defs on {self.host.hostname}")
+
+        self.host.conn.run(
+            f"""
+            if grep -q '^{key}\\s' /etc/login.defs; then
+                sed -i 's/^{key}\\s.*/{key} {value}/' /etc/login.defs
```
ikerexxe: This only looks for uncommented lines, so at some point if you have `#CREATE_HOME yes` in the file, you'll end up with:
```
#CREATE_HOME yes
CREATE_HOME no
```

This doesn't look very clean, so consider improving it.

tests/system/framework/hosts/shadow.py
```diff
@@ -201,3 +202,77 @@ def discard_file(self, origin: str) -> None:
...
+    def __init__(self, host: ShadowHost) -> None:
+        super().__init__(host)
+        self._backup_path: str | None = None
+
+    def setup(self) -> None:
+        """
+        Backup the original /etc/login.defs file.
+        """
+        if self.host.fs.exists("/etc/login.defs"):
+            self._backup_path = "/tmp/login.defs.backup"
```
ikerexxe: The hardcoded backup path `/tmp/login.defs.backup` will cause collisions if previous backup wasn't cleaned up properly.
                                                                                                    
Use `mktemp` to create unique backup paths, just like the existing [backup()](https://github.com/shadow-maint/shadow/blob/master/tests/system/framework/hosts/shadow.py#L111) method does.

tests/system/framework/hosts/shadow.py
ikerexxe: In addition, this will break if `value` contains certain characters (i.e. `/`, `&`). I'm specially concerned with the forward slashs (`/`) as this is a valid character for login.defs options

tests/system/framework/roles/shadow.py
```diff
@@ -31,6 +31,20 @@ def __init__(self, *args, **kwargs) -> None:
...
+        Usage:
+            # Set a value
+            shadow.login_defs["CREATE_HOME"] = "no"
+
+            # Get a value
+            value = shadow.login_defs["CREATE_HOME"]
+
+            # Remove an option
+            del shadow.login_defs["CREATE_HOME"]
+        """
```
ikerexxe: Move this comment to `LoginDefsConfig` class. This will help explain what the class does.

ikerexxe (commented):
Just some comments inline

tests/system/tests/test_useradd.py
```diff
@@ -566,3 +566,35 @@ def test_useradd_default_primary_group_with_supplementary(shadow: Shadow):
...
+    assert passwd_entry is not None, "User test1 should be found in passwd"
+    assert passwd_entry.name == "test1", "Incorrect username"
+    assert passwd_entry.home == "/home/test1", "Incorrect home directory"
+
+    group_entry = shadow.tools.getent.group("test1")
+    assert group_entry is not None, "Group test1 should be found"
+    assert group_entry.name == "test1", "Incorrect group name"
+
+    home_dir = "/home/test1"
+    assert shadow.fs.exists(home_dir), f"Home directory {home_dir} should exist even with CREATE_HOME=no"
```
ikerexxe: This message is a bit misleading as we are focusing on the creation with `--create-home`. What about using `Home directory {home_dir} should exist when using --create-home flag` or `Expected home directory {home_dir} to be created by --create-home`?

ikerexxe (commented):
We are almost there. Apart from the inline comment you should also improve the commit message for the `LoginDefsConfig` implementation. What about:
```
tests: implement login.defs configuration utility                                                 
                                                                                                    
Introduce LoginDefsConfig class for /etc/login.defs manipulation.
It supports getting, setting, and removing configuration options
with automatic backup and restoration.
```

In addition, let's invert the commits as the test requires the framework changes to be there before.

ikerexxe:
Ok, there’s just one last thing left to do: reorder the commits. The API for modifying the `login.defs` file is essential for the test, so it needs to come earlier in the commit history.

aborah-sudo:
> Ok, there’s just one last thing left to do: reorder the commits. The API for modifying the `login.defs` file is essential for the test, so it needs to come earlier in the commit history.

Looks like they are:

[aborah@fedora shadow]$ git log --oneline 
0272ccae (HEAD -> useradd, origin/useradd) tests: implement login.defs configuration utility
5267daf7 Tests: Add a new user with home directory creation
f6b2a43f groupadd: fix system account help text

ikerexxe:
That doesn't look like the GitHub order. You'll probably need to push.

aborah-sudo:
> That doesn't look like the GitHub order. You'll probably need to push.

Done

ikerexxe (approved):
LGTM. Thank you for the time and patience in the review process

PR #1615, opened 2026-04-23
tests: improve cloud environment compatibility

ikerexxe:
This PR enhances the system tests to work reliably in cloud VM environments by addressing two main compatibility issues:                                                                                     
1. Mount point handling: fixed `/home` directory restoration when `/home` is a separate mount point.   
2. UID and GID consistency: addressed test failures caused by different default UID allocation between containers (starting at 1000) and cloud VMs (where 1000 is typically occupied by system users like `ubuntu` or `ec2-user`).

Fixes:
1. Use selective `/home` cleanup instead of attempting to remove mounted directories.
2. Added CI user creation (UID 1000) in all distribution containers to simulate cloud VMs.
3. Updated test assertions across all affected test files to expect UID/GID 1001+ instead of 1000.

### DIFF
--- a/share/ansible/roles/ci_run/tasks/alpine.yml
+++ b/share/ansible/roles/ci_run/tasks/alpine.yml
@@ -34,6 +34,20 @@
     dest: /bin/expect
     state: link
 
+- name: Check if UID 1000 already exists
+  ansible.builtin.getent:
+    database: passwd
+    key: 1000
+  register: existing_uid_1000
+  failed_when: false
+
+- name: Create CI user to simulate cloud VM environment
+  ansible.builtin.user:
+    name: ci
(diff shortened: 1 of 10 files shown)


hallyn (approved):
LGTM, except, as I said out of band, I think it might be more robust to have the ansible jobs which create the ci user check which uids are in use.  But it's not critical, can wait for later, no objection to merging this.

ikerexxe:
> LGTM, except, as I said out of band, I think it might be more robust to have the ansible jobs which create the ci user check which uids are in use. But it's not critical, can wait for later, no objection to merging this.

Updated to take this into account. Take a look when you have a moment.

hallyn:
Thanks

Issue #1610, opened 2026-04-15
Create configure option to disable syslog


FreziDok:
Hello.

### Problem
During build of distributions with Yocto Project (and others) the `shadow-native` tools are spamming host's syslog with image-related information, which aren't relevant.

### Current workaround
 There's a [patch](https://lore.kernel.org/openembedded-core/20250228101323.1008021-1-richard.purdie@linuxfoundation.org/) that hardcodes turning the syslog off that Yocto uses. This works, but it is a Yocto-specific hack.

### Proposed solution
 Add a proper configure option, for example `./configure --disable-syslog`. When --disable-syslog is used, configure.h should define something like `USE_SYSLOG 0`, so that all SYSLOG() calls become no-ops without any #if 1 hacks.
This would allow tools like Yocto simply do something like `PACKAGECONFIG[syslog] = "--enable-syslog,--disable-syslog,,,"` and eventually get rid of the ugly patch.

If you ok with this feature, I am ready to prepare a draft patch to apply it.

hallyn:
I don't really have an objection to this.  A patch will certainly be easier to review.

PR #1618, opened 2026-04-28
Add --disable-syslog configure option

FreziDok:
Implements `--disable-syslog` configuration option, as discussed in #1610

Addresses #1610

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -145,6 +145,17 @@ AC_ARG_ENABLE([logind],
 	[enable_logind="yes"]
 )
 
+AC_ARG_ENABLE([syslog],
+  [AS_HELP_STRING([--disable-syslog], [disable logging via syslog() calls])],
+  [
+    AS_CASE([${enableval}],
+        [yes|no],[],
+        [AC_MSG_ERROR([bad parameter value for --enable-syslog=${enableval}])]
+    )
+  ],
+  [enable_syslog="yes"]
+)
+
 AC_ARG_WITH([audit],
 	[AS_HELP_STRING([--with-audit], [use auditing support @<:@default=yes if found@:>@])],
 	[with_audit=$withval], [with_audit=maybe])
@@ -200,6 +211,7 @@ AC_DEFINE_UNQUOTED([GROUP_NAME_MAX_LENGTH], [$with_group_name_max_length], [max 
 AC_SUBST([GROUP_NAME_MAX_LENGTH])
 GROUP_NAME_MAX_LENGTH="$with_group_name_max_length"
 
+AS_VAR_IF([enable_syslog],["yes"],[AC_DEFINE([USE_SYSLOG], [1], [Define to use syslog])])
 
 AM_CONDITIONAL([USE_BCRYPT], [test "x$with_bcrypt" = "xyes"])
 if test "X$with_bcrypt" = "Xyes"; then
@@ -630,5 +642,6 @@ AC_MSG_NOTICE([shadow ${PACKAGE_VERSION} has been configured with the following 
 	use file caps:			$with_fcaps
 	install su:			$with_su
 	enabled vendor dir:             $enable_vendordir
+	enable syslog: 			$enable_syslog
 
 ])
--- a/lib/io/syslog.h
+++ b/lib/io/syslog.h
@@ -24,10 +24,12 @@
 #define LOG_AUTHPRIV LOG_AUTH
 #endif
 
-#ifdef ENABLE_NLS
-#define SYSLOG(...)  SYSLOG_C(__VA_ARGS__)
+#if !USE_SYSLOG
+# define SYSLOG(...)
+#elif defined(ENABLE_NLS)
+# define SYSLOG(...)  SYSLOG_C(__VA_ARGS__)
 #else
-#define SYSLOG(...)  syslog(__VA_ARGS__)
+# define SYSLOG(...)  syslog(__VA_ARGS__)
 #endif
 
 /* The default syslog settings can now be changed here,


lib/io/syslog.h
```diff
@@ -24,11 +24,15 @@
...
 
+#if USE_SYSLOG
 #ifdef ENABLE_NLS
 #define SYSLOG(...)  SYSLOG_C(__VA_ARGS__)
 #else
 #define SYSLOG(...)  syslog(__VA_ARGS__)
 #endif
+#else 				/* !USE_SYSLOG*/
+#define SYSLOG(...)
+#endif				/* !USE_SYSLOG*/
```
alejandro-colomar: I think it would be better to invert the check, so that the `else` is closer to the `if`.

Also, indentation would be better than that comment:

```c
#if !USE_SYSLOG
# define SYSLOG(...)
#else
# ifdef ENABLE_NLS
#  define SYSLOG(...)  SYSLOG_C(__VA_ARGS__)
# else
#  define SYSLOG(...)  syslog(__VA_ARGS__)
# endif
#endif
```

configure.ac
```diff
@@ -200,6 +207,9 @@ AC_DEFINE_UNQUOTED([GROUP_NAME_MAX_LENGTH], [$with_group_name_max_length], [max
 AC_SUBST([GROUP_NAME_MAX_LENGTH])
 GROUP_NAME_MAX_LENGTH="$with_group_name_max_length"
 
+if test "X$enable_syslog" = "Xyes" ; then
+	AC_DEFINE([USE_SYSLOG], 1, [Define to use syslog])
```
alejandro-colomar: Please quote the `1` for consistency with surrounding lines, as `[1]`.

configure.ac
```diff
@@ -145,6 +145,13 @@ AC_ARG_ENABLE([logind],
...
 )
 
+AC_ARG_ENABLE([syslog],
+	[AS_HELP_STRING([--disable-syslog],
+		[disable syslog() calls @<:@default=no@:>@])],
+	[enable_syslog="${enableval}"],
+	[enable_syslog="yes"]
+)
+
 AC_ARG_WITH([audit],
```
alejandro-colomar, line 159: I don't understand why some flags are `--with`/`--without` and others are `--enable`/`--disable`.  Could you please document why you chose this?

FreziDok: Hello. As far as I know, `--with`/`--without ` are historically for external dependencies, and `--enable`/`--disable` are for internal features. Since using syslog or not is an internal choice, I decided to pick `--disable`.

configure.ac
FreziDok: Done

configure.ac
alejandro-colomar: Thanks!  Sounds reasonable.

lib/io/syslog.h
```diff
@@ -24,10 +24,14 @@
...
+#if !USE_SYSLOG
+# define SYSLOG(...)
 #else
-#define SYSLOG(...)  syslog(__VA_ARGS__)
+# ifdef ENABLE_NLS
+#  define SYSLOG(...)  SYSLOG_C(__VA_ARGS__)
+# else
+#  define SYSLOG(...)  syslog(__VA_ARGS__)
+# endif
 #endif
```
alejandro-colomar, line 33: Oh, and now that I see, the `else` and the `ifdef` can be collapsed into one directive:

```c
#if !USE_SYSLOG
# define SYSLOG(...)
#elif defined(ENABLE_NLS)
# define SYSLOG(...)  SYSLOG_C(__VA_ARGS__)
#else
# define SYSLOG(...)  syslog(__VA_ARGS__)
#endif
```

configure.ac
```diff
@@ -145,6 +145,13 @@ AC_ARG_ENABLE([logind],
 	[enable_logind="yes"]
 )
 
+AC_ARG_ENABLE([syslog],
+	[AS_HELP_STRING([--disable-syslog],
+		[disable syslog() calls @<:@default=no@:>@])],
```
lslebodn: One cause use --disable-syslog --enable-syslog as shorter versions
instead of using --enable-syslog=no

IMHO it is simpler to read "positive flags" in code and in help

WDYT about change?
```
-	[AS_HELP_STRING([--disable-syslog],
-		[disable syslog() calls @<:@default=no@:>@])],
+	[AS_HELP_STRING([--enable-syslog],
+		[enable syslog() calls @<:@default=yes@:>@])],
```

alejandro-colomar: > IMHO it is simpler to read "positive flags" in code and in help

Agree.

> WDYT about change?

Sounds reasonable.  However, I don't understand much about the build system.  Maybe others have other opinions.

Cc: @thesamesam , @Karlson2k , @jubalh

configure.ac
```diff
@@ -145,6 +145,13 @@ AC_ARG_ENABLE([logind],
 	[enable_logind="yes"]
 )
```
alejandro-colomar, line 147: From the commit message:

> Addresses https://github.com/shadow-maint/shadow/issues/1610

Please spell that as:

```
Closes: <https://github.com/shadow-maint/shadow/issues/1610>
```

configure.ac
Karlson2k: Usually parameter indicated as '--disable-something' when 'something' is enabled by default.
And the opposite: help string is '--enable-something', when 'something' is disabled by default.
Here the default value is 'enabled', it means that logically help string should be '--disable-syslog' (as '--enable-syslog' is no-op in this case).

configure.ac
```diff
@@ -145,6 +145,13 @@ AC_ARG_ENABLE([logind],
...
 
+AC_ARG_ENABLE([syslog],
+	[AS_HELP_STRING([--disable-syslog],
+		[disable syslog() calls @<:@default=no@:>@])],
+	[enable_syslog="${enableval}"],
```
Karlson2k: Useless assignment.
It is expanded by autoconf to literallly:
``` bash
enableval="$enable_syslog"

enable_syslog="${enableval}"
```

Karlson2k: Could be left empty or added value checking, like
``` autoconf
AS_CASE([${enableval}],
  [yes|no],[:],
  [AC_MSG_ERROR([wrong parameter value: --enable-syslog=${enableval}])]
)
```

configure.ac
Karlson2k: In other words: seeing '--enable-syslog' in built-in help one may think that syslog is disabled by default and must be enabled explicitly. 

Makes a little sense to document parameter '--enable-syslog' which changes nothing, while only '--disable-syslog' changes something and it is not documented explicitly.

Karlson2k: For clarity: the help string is wrong. It must be 
``` autoconf
[AS_HELP_STRING([--disable-syslog],
		[disable syslog() calls])],
```
only `--enable-something` may have optional parameter, like `--enable-something=no`.
`--disable-something` is converted automatically to `--enable-something=no`, while `--disable-something=no` is reported by `configure` as an error.

Alternatively, you may use
``` autoconf
[AS_HELP_STRING([--disable-syslog],
		[[disable syslog() calls [enable]]])],
```
or
``` autoconf
[AS_HELP_STRING([--disable-syslog],
		[[disable syslog() calls [enabled by default]]])],
```
but I don't think that it will add any clarity.

FreziDok: > only `--enable-something` may have optional parameter

It makes sense, thanks for the correction. 


I think 

> ```m4
> [[disable syslog() calls [enable]]])],
> ```

and 

> ```m4
> [[disable syslog() calls [enabled by default]]])],
> ```

are only adding obscurity, because it's hard to understand what is enabled: disable function (means syslog is disabled) or syslog function, so I think your first variant with only `[disable syslog() calls])],` is the most appropriate.

configure.ac
FreziDok: I am not an expert in autoconf, so I just looked at others options like 

```m4
AC_ARG_ENABLE([lastlog],
	[AS_HELP_STRING([--enable-lastlog],
		[enable lastlog @<:@default=no@:>@])],
	[enable_lastlog="${enableval}"],
	[enable_lastlog="no"]
)
``` 
and did the same as there. 

Could you please explain what is the difference between them?

Karlson2k: There is no difference. It was a bad style with extra redundancy and no practical effect. I think it should be cleaned-up at some moment.
In the meantime, it is better to not increase amount of bad style used in configure and the new additions should be clean.
You can use this one as an example or template:
https://github.com/shadow-maint/shadow/blob/bd1ea38423231d7735f648d80db0016ec2ba760e/configure.ac#L134-L146

For this particular situation it can be simplified to 
``` autoconf
AC_ARG_ENABLE([syslog],
  [AS_HELP_STRING([--disable-syslog], [disable logging via syslog() calls])],
  [
    AS_CASE([${enableval}],
        [yes|no],[],
        [AC_MSG_ERROR([bad parameter value for --enable-syslog=${enableval}])]
    )
  ],
  [enable_syslog="yes"]
)
```

configure.ac
Karlson2k: I agree.
See my full suggestion for this part of the configure here: https://github.com/shadow-maint/shadow/pull/1618#discussion_r3159550191

configure.ac
FreziDok: Thank you for the explanation. Applied.

configure.ac
```diff
@@ -200,6 +211,9 @@ AC_DEFINE_UNQUOTED([GROUP_NAME_MAX_LENGTH], [$with_group_name_max_length], [max
 AC_SUBST([GROUP_NAME_MAX_LENGTH])
 GROUP_NAME_MAX_LENGTH="$with_group_name_max_length"
 
+if test "X$enable_syslog" = "Xyes" ; then
+	AC_DEFINE([USE_SYSLOG], [1], [Define to use syslog])
+fi
```
Karlson2k:
```suggestion
AS_VAR_IF([enable_syslog],["yes"],[AC_DEFINE([USE_SYSLOG], [1], [Define to use syslog])])
```
Not strictly required, as autoconf will transform the suggested code to the code similar to plain shell code used by you. However, in more complex situations, autoconf should be aware about any conditionals used in the code and this information could be provided by use of autoconf macros (instead of plain shell code directly).

FreziDok: Thank you. I thought of using autoconf macros for this as you suggest, but there were a lot of plain shell code for other options, so I decided to stick to the current code style. Since (as you said) most of it should be rewritten, I suppose your variant is the best option.

hallyn:
Thanks.  As I don't regularly use autoconf, I tested by creating a minimal configure.ac:

```
serge@andre:~/autoconf-t2$ cat configure.ac
AC_INIT([test], [1.0])
# Your snippet here:
AC_ARG_ENABLE([syslog],
  [AS_HELP_STRING([--disable-syslog], [disable logging via syslog() calls])],
  [
    AS_CASE([${enableval}],
        [yes|no],[],
        [AC_MSG_ERROR([bad parameter value for --enable-syslog=${enableval}])]
    )
  ],
  [enable_syslog="yes"]
)
AS_VAR_IF([enable_syslog],["yes"],[AC_DEFINE([USE_SYSLOG], [1], [Define to use syslog])])
AC_OUTPUT
AC_MSG_NOTICE([shadow ${PACKAGE_VERSION} has been configured with the following features:
        enable syslog:                  $enable_syslog
])
serge@andre:~/autoconf-t2$ autoconf
serge@andre:~/autoconf-t2$ ./configure
configure: creating ./config.status
configure: shadow 1.0 has been configured with the following features:
        enable syslog:                  yes

serge@andre:~/autoconf-t2$ ./configure --enable-syslog=yes
configure: creating ./config.status
configure: shadow 1.0 has been configured with the following features:
        enable syslog:                  yes

serge@andre:~/autoconf-t2$ ./configure --enable-syslog=no
configure: creating ./config.status
configure: shadow 1.0 has been configured with the following features:
        enable syslog:                  no

serge@andre:~/autoconf-t2$ ./configure --enable-syslog=xxx
configure: error: bad parameter value for --enable-syslog=xxx

```

Looks good, thanks.

hallyn (approved):

hallyn:
(Please feel free to merge after tests pass.  I would enable auto-merge, but that button's not showing up... :shrug:)

alejandro-colomar:
I'll merge, from what you've said.  I haven't reviewed myself.  I assume everyone's happy with the code.  If anyone disagrees, feel free to complain.

PR #1619, opened 2026-04-28
useradd: skip btrfs subvolume creation for system users

silverhadch:
Gate subvolume creation on user_id >= UID_MIN to exclude system users regardless of their home path.

Fixes: c1d36a8acb1d (2019-05-04; "Add support for btrfs subvolumes for user homes")

### DIFF
--- a/man/useradd.8.xml
+++ b/man/useradd.8.xml
@@ -157,6 +157,19 @@
 		regardless of any configuration file settings.
 	  </para>
 	  <para>
+	    Subvolumes are only created for regular users,
+	    whose UID is within the
+	    <option>UID_MIN</option>-<option>UID_MAX</option>
+	    range defined in <filename>/etc/login.defs</filename>.
+	    System users get a regular home directory instead.
+	    Set the <option>BTRFS_SUBVOLUME_SYSTEM</option> variable
+	    in <filename>/etc/default/useradd</filename>
+	    to <replaceable>yes</replaceable>
+	    to create subvolumes for system users as well.
+	    If this variable is not set,
+	    the default value is no.
+	  </para>
+	  <para>
 	    If the parent directory of the user's home directory is
 	    <emphasis>not</emphasis> on a Btrfs filesystem,
 	    <command>useradd</command> will <emphasis>not</emphasis> create a
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -114,6 +114,7 @@ static const char *def_template = SKEL_DIR;
 static const char *def_usrtemplate = USRSKELDIR;
 static const char *def_create_mail_spool = "yes";
 static const char *def_btrfs_subvolume_home = "no";
+static const char *def_btrfs_subvolume_system = "no";
 static const char *def_log_init = "yes";
 
 static long def_inactive = -1;
@@ -223,6 +224,7 @@ static bool home_added = false;
 #define DUSRSKEL		"USRSKEL"
 #define DCREATE_MAIL_SPOOL	"CREATE_MAIL_SPOOL"
 #define DBTRFS_SUBVOLUME_HOME	"BTRFS_SUBVOLUME_HOME"
+#define DBTRFS_SUBVOLUME_SYSTEM	"BTRFS_SUBVOLUME_SYSTEM"
 #define DLOG_INIT		"LOG_INIT"
 
 /* local function prototypes */
@@ -478,6 +480,15 @@ get_defaults(const struct option_flags *flags)
 		}
 
 		/*
+		 * Create subvolume homes also for system users?
+		 */
+		else if (streq(buf, DBTRFS_SUBVOLUME_SYSTEM)) {
+			if (streq(ccp, ""))
+				ccp = "no";
+			def_btrfs_subvolume_system = xstrdup(ccp);
+		}
+
+		/*
 		 * By default do we add the user to the lastlog and faillog databases ?
 		 */
 		else if (streq(buf, DLOG_INIT)) {
@@ -512,6 +523,7 @@ static void show_defaults (void)
 	printf ("USRSKEL=%s\n", def_usrtemplate);
 	printf ("CREATE_MAIL_SPOOL=%s\n", def_create_mail_spool);
 	printf ("BTRFS_SUBVOLUME_HOME=%s\n", def_btrfs_subvolume_home);
+	printf ("BTRFS_SUBVOLUME_SYSTEM=%s\n", def_btrfs_subvolume_system);
 	printf ("LOG_INIT=%s\n", def_log_init);
 }
 
@@ -536,6 +548,7 @@ set_defaults(void)
 	bool  out_usrskel = false;
 	bool  out_create_mail_spool = false;
 	bool  out_btrfs_subvolume_home = false;
+	bool  out_btrfs_subvolume_system = false;
 	bool  out_log_init = false;
 	char  buf[1024];
 	char  *new_file = NULL;
@@ -657,6 +670,11 @@ set_defaults(void)
 			        DBTRFS_SUBVOLUME_HOME "=%s\n",
 			        def_btrfs_subvolume_home);
 			out_btrfs_subvolume_home = true;
+		} else if (!out_btrfs_subvolume_system && streq(buf, DBTRFS_SUBVOLUME_SYSTEM)) {
+			fprintf(ofp,
+			        DBTRFS_SUBVOLUME_SYSTEM "=%s\n",
+			        def_btrfs_subvolume_system);
+			out_btrfs_subvolume_system = true;
 		} else if (!out_log_init && streq(buf, DLOG_INIT)) {
 			fprintf(ofp, DLOG_INIT "=%s\n", def_log_init);
 			out_log_init = true;
@@ -693,6 +711,8 @@ set_defaults(void)
 		fprintf (ofp, DCREATE_MAIL_SPOOL "=%s\n", def_create_mail_spool);
 	if (!out_btrfs_subvolume_home)
 		fprintf (ofp, DBTRFS_SUBVOLUME_HOME "=%s\n", def_btrfs_subvolume_home);
+	if (!out_btrfs_subvolume_system)
+		fprintf (ofp, DBTRFS_SUBVOLUME_SYSTEM "=%s\n", def_btrfs_subvolume_system);
 	if (!out_log_init)
 		fprintf (ofp, DLOG_INIT "=%s\n", def_log_init);
 	/*
@@ -2211,10 +2231,20 @@ usr_update (unsigned long subuid_count, unsigned long subgid_count,
 static bool
 want_btrfs_subvolume(const char *path)
 {
+	uid_t  min, max;
+
 	if (!subvolflg)
 		return false;
+	if (strlen(prefix_user_home) - strlen(path) > 1)
+		return false;
 
-	return strlen(prefix_user_home) - strlen(path) <= 1;
+	if (strcaseeq(def_btrfs_subvolume_system, "yes"))
+		return true;
+
+	min = getdef_ulong("UID_MIN", 1000UL);
+	max = getdef_ulong("UID_MAX", 60000UL);
+
+	return user_id >= min && user_id <= max;
 }
 #endif
 
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2207,6 +2207,17 @@ usr_update (unsigned long subuid_count, unsigned long subgid_count,
 	}
 }
 
+#if WITH_BTRFS
+static bool
+want_btrfs_subvolume(const char *path)
+{
+	if (!subvolflg)
+		return false;
+
+	return strlen(prefix_user_home) - strlen(path) <= 1;
+}
+#endif
+
 /*
  * create_home - create the user's home directory
  *
@@ -2264,7 +2275,7 @@ static void create_home(const struct option_flags *flags)
 
 		dir_created = false;
 #if WITH_BTRFS
-		if (subvolflg && (strlen(prefix_user_home) - (int)strlen(path)) <= 1) {
+		if (want_btrfs_subvolume(path)) {
 			char *btrfs_check = strdup(path);
 			struct statfs  sfs;
 
--- a/tests/system/framework/hosts/shadow.py
+++ b/tests/system/framework/hosts/shadow.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from pathlib import PurePosixPath
 from typing import Any
 
-from pytest_mh import MultihostUtility
+from pytest_mh import MultihostUtility, mh_utility_used
 from pytest_mh.conn import ProcessLogLevel
 
 from .base import BaseHost, BaseLinuxHost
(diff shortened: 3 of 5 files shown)


alejandro-colomar:
The code seems good to me.

Do we need to document that BTRFS configuration is ignored for system users?

Also, I think the commit message should document why we don't want BTRFS subvolumes for system users.

silverhadch:
> The code seems good to me.
> 
> Do we need to document that BTRFS configuration is ignored for system users?
> 
> Also, I think the commit message should document why we don't want BTRFS subvolumes for system users.

Subvolumes for non intercative users are wasteful as they multiply metadata operations and generate unneccesary filesystem objects for operations, a system user who doesnt generate snapshot worthy data, wont use.

But I dont think its worth documenting as this is simply sane design. This BTRFS Backend never mattered to system users as the BTRFS Backend orginally released only supported interactive subvol creation. There is no usage for subvols (aka wasting metadata) for system users with state home dirs.

alejandro-colomar:
> > The code seems good to me.
> > Do we need to document that BTRFS configuration is ignored for system users?
> > Also, I think the commit message should document why we don't want BTRFS subvolumes for system users.
> 
> Subvolumes for non intercative users are wasteful as they multiply metadata operations and generate unneccesary filesystem objects for operations, a system user who doesnt generate snapshot worthy data, wont use.

Please copy this into the commit message.

> 
> But I dont think its worth documenting as this is simply sane design. This BTRFS Backend never mattered to system users as the BTRFS Backend orginally released only supported interactive subvol creation. There is no usage for subvols (aka wasting metadata) for system users with state home dirs.

This is still useful information for someone like me that has never used BTRFS and doesn't know anything about it other than the fact that it's a file system.  :)

I'd like to have it at least in the commit message.  I don't care too much about the manual page.

silverhadch:
> > > The code seems good to me.
> > > Do we need to document that BTRFS configuration is ignored for system users?
> > > Also, I think the commit message should document why we don't want BTRFS subvolumes for system users.
> > 
> > 
> > Subvolumes for non intercative users are wasteful as they multiply metadata operations and generate unneccesary filesystem objects for operations, a system user who doesnt generate snapshot worthy data, wont use.
> 
> Please copy this into the commit message.
> 
> > But I dont think its worth documenting as this is simply sane design. This BTRFS Backend never mattered to system users as the BTRFS Backend orginally released only supported interactive subvol creation. There is no usage for subvols (aka wasting metadata) for system users with state home dirs.
> 
> This is still useful information for someone like me that has never used BTRFS and doesn't know anything about it other than the fact that it's a file system. :)
> 
> I'd like to have it at least in the commit message. I don't care too much about the manual page.

Done. Documented in the commit message.

hallyn (changes requested):
It seems to me this should be optional, no?

I don't think everyone would agree that every user under UID_MIN does not benefit from subvolumes.

silverhadch:
> It seems to me this should be optional, no?
> 
> I don't think everyone would agree that every user under UID_MIN does not benefit from subvolumes.

How so?

Why should a System User need to waste resources?

ikerexxe:
I think this should be optional rather than hardcoded. Following the Unix philosophy of "mechanism, not policy," we should default the behavior to off (no subvolumes for system users). This addresses the performance concerns while providing a configuration toggle for edge cases or environments that still need it.

My two cents

hallyn:
Yup I'd be fine with making it the default to not create the subvolume for system users.

hallyn:
@silverhadch do you mind making the change to make this non-default?

silverhadch:
> @silverhadch do you mind making the change to make this non-default?

Sure once I finish the Security fix then I will piçk up this PR again.

hallyn:
On Wed, Jul 08, 2026 at 10:07:32PM -0700, Hadi Chokr wrote:
> silverhadch left a comment (shadow-maint/shadow#1619)
> 
> > @silverhadch do you mind making the change to make this non-default?
> 
> Sure once I finish the Security fix then I will piçk up this PR again.

Awesome, thanks.

src/useradd.c
```diff
@@ -2264,7 +2293,10 @@ static void create_home(const struct option_flags *flags)
 
 		dir_created = false;
 #if WITH_BTRFS
-		if (subvolflg && (strlen(prefix_user_home) - (int)strlen(path)) <= 1) {
+		if (subvolflg
+			&& (strcaseeq(def_btrfs_subvolume_system, "yes")
+			    || (user_id >= uid_min && user_id <= uid_max))
+			&& (strlen(prefix_user_home) - (int)strlen(path)) <= 1) {
```
alejandro-colomar: Please move the opening brace to a quasi-blank line.

See <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c#n46>.

```
    Braces
	its own.

		if (foo)
			bar();

		{
			baz();
		}
```

alejandro-colomar:
I abstain about this patch set.

I'll say that the quality of the C source code and documentation meets my expectations.

I don't have an opinion on the feature itself.

silverhadch:
> I abstain about this patch set.
> 
> I'll say that the quality of the source code meets my expectations.
> 
> I don't have an opinion on the feature itself.

Who should I ask a review from?

alejandro-colomar:
> > I abstain about this patch set.
> > I'll say that the quality of the source code meets my expectations.
> > I don't have an opinion on the feature itself.
> 
> Who should I ask a review from?

@hallyn

tests/system/framework/roles/shadow.py
```diff
@@ -190,6 +190,23 @@ def useradd(self, *args) -> ProcessResult:
 
         return cmd
 
+    def useradd_defaults(self, content: str) -> None:
```
ikerexxe: This should be implemented in a dictionary-like access (i.e. `shadow.useradd_defaults["BTRFS_SUBVOLUME_SYSTEM"] = "value"`). You can reuse the existing `LoginDefsConfig` functionality from `8408fd7e6ff5b04f75c7b96cdd35ba10cd9679a7` and make it more generic so that it can handle other files as well

man/useradd.8.xml
```diff
@@ -156,6 +156,18 @@
...
+	    Subvolumes are only created for regular users, whose UID is
+	    within the
+	    <option>UID_MIN</option>-<option>UID_MAX</option>
+	    range defined in <filename>/etc/login.defs</filename>.
+	    System users get a regular home directory instead.
+	    Set the <option>BTRFS_SUBVOLUME_SYSTEM</option> variable in
+	    <filename>/etc/default/useradd</filename> to
+	    <replaceable>yes</replaceable> to create subvolumes for
+	    system users as well.
+	    If this variable is not set, the default value is no.
```
alejandro-colomar: I'd reflow as:

```
Subvolumes are only created for regular users,
whose UID is within the
<option>UID_MIN</option>-<option>UID_MAX</option>
range defined in <filename>/etc/login.defs</filename>.
System users get a regular home directory instead.
Set the <option>BTRFS_SUBVOLUME_SYSTEM</option> variable
in <filename>/etc/default/useradd</filename>
to <replaceable>yes</replaceable>
to create subvolumes for system users as well.
If this variable is not set,
the default value is no.
```

See 'semantic newlines' in man-pages(7).

tests/system/framework/hosts/shadow.py
```diff
@@ -275,104 +275,194 @@ def discard_file(self, origin: str) -> None:
...
+        # Set a value
+        shadow.useradd_defaults["BTRFS_SUBVOLUME_SYSTEM"] = "yes"
+
+    Some minimal container images do not ship /etc/default, so the
+    directory and the file are created when missing. All changes are
+    reverted when the test is finished.
+    """
+
+    def __init__(self, host: ShadowHost) -> None:
+        super().__init__(host, "/etc/default/useradd", separator="=")
```
ikerexxe: I'd remove these subclasses, since they don't add any value

tests/system/framework/roles/shadow.py
```diff
@@ -32,6 +32,7 @@ def __init__(self, *args, **kwargs) -> None:
         """
         super().__init__(*args, **kwargs)
         self.login_defs: LoginDefsConfig = LoginDefsConfig(self.host).postpone_setup()
+        self.useradd_defaults: UseraddDefaultsConfig = UseraddDefaultsConfig(self.host).postpone_setup()
```
ikerexxe: Let's instantiate `KeyValueFileConfig` directly:
```
self.login_defs: KeyValueFileConfig = KeyValueFileConfig(self.host, "/etc/login.defs", " ").postpone_setup()
self.useradd_defaults: KeyValueFileConfig = KeyValueFileConfig(self.host, "/etc/default/useradd", "=").postpone_setup()
```

ikerexxe (commented):
@aborah-sudo, as you were the original developer of the `KeyValueFileConfig` class, I think you should take a look at the changes that have been made to it

aborah-sudo:
The refactoring from LoginDefsConfig to a generic KeyValueFileConfig looks good to me — the backup/restore, separator handling, and _ensure_exists() logic are all correct. I agree with @ikerexxe  that the LoginDefsConfig and UseraddDefaultsConfig subclasses can be dropped since they're just constructor wrappers with no added behavior. Instantiating KeyValueFileConfig directly is cleaner.

src/useradd.c
```diff
@@ -2262,7 +2293,11 @@ static void create_home(const struct option_flags *flags)
 
 		dir_created = false;
 #if WITH_BTRFS
-		if (subvolflg && (strlen(prefix_user_home) - (int)strlen(path)) <= 1) {
+		if (subvolflg
+		    && (strcaseeq(def_btrfs_subvolume_system, "yes")
+		        || (user_id >= uid_min && user_id <= uid_max))
+		    && (strlen(prefix_user_home) - (int)strlen(path)) <= 1)
```
ikerexxe: This condition is quite complex. You could make it clearer by using variables or by encapsulating it in a function

tests/system/tests/privileged/test_privileged_useradd.py
```diff
@@ -33,3 +33,55 @@ def test_useradd__btrfs_subvolume_home(shadow: Shadow) -> None:
         assert shadow.fs.exists("/home/tuser"), "Home directory should exist"
 
         assert shadow.host.btrfs.subvolume_exists("/home", "tuser")
+
+
+def test_useradd__btrfs_subvolume_home_system_user(shadow: Shadow) -> None:
+    """
+    :title: Btrfs subvolume home is skipped for system users by default
+    :steps:
+        1. Create a system user with -r -m and --btrfs-subvolume-home
```
ikerexxe: You should explain the `what` you are doing, not the `how`. As such, please avoid stating the options you are using, but what you are trying to achieve with them

tests/system/tests/privileged/test_privileged_useradd.py
```diff
@@ -33,3 +33,55 @@ def test_useradd__btrfs_subvolume_home(shadow: Shadow) -> None:
...
+        assert not shadow.host.btrfs.subvolume_exists("/home", "tsysuser"), "Home should not be a subvolume"
+
+
+def test_useradd__btrfs_subvolume_home_system_user_enabled(shadow: Shadow) -> None:
+    """
+    :title: Btrfs subvolume home is created for system users with BTRFS_SUBVOLUME_SYSTEM=yes
+    :setup:
+        1. Set BTRFS_SUBVOLUME_SYSTEM=yes in /etc/default/useradd
+    :steps:
+        1. Create a system user with -r -m and --btrfs-subvolume-home
```
ikerexxe: Same here

ikerexxe (commented):
Minor comments inline

src/useradd.c
```diff
@@ -2207,6 +2227,29 @@ usr_update (unsigned long subuid_count, unsigned long subgid_count,
...
+	if (strcaseeq(def_btrfs_subvolume_system, "yes"))
+		return true;
+
+	uid_min = getdef_ulong("UID_MIN", 1000UL);
+	uid_max = getdef_ulong("UID_MAX", 60000UL);
+
+	return user_id >= uid_min && user_id <= uid_max;
+}
+#endif
+
```
alejandro-colomar, line 2250: Please add this function in a separate (preceding) commit, separating the refactor from the actual change.

src/useradd.c
```diff
@@ -2207,6 +2227,29 @@ usr_update (unsigned long subuid_count, unsigned long subgid_count,
...
+static bool want_btrfs_subvolume(const char *path)
+{
+	uid_t  uid_min, uid_max;
+	bool   is_home_itself;
+
+	if (!subvolflg)
+		return false;
+
+	is_home_itself = (strlen(prefix_user_home) - (int)strlen(path)) <= 1;
+	if (!is_home_itself)
```
alejandro-colomar: Please don't add this single-use variable.

src/useradd.c
```diff
@@ -2207,6 +2227,29 @@ usr_update (unsigned long subuid_count, unsigned long subgid_count,
...
+#if WITH_BTRFS
+static bool want_btrfs_subvolume(const char *path)
+{
+	uid_t  uid_min, uid_max;
+	bool   is_home_itself;
+
+	if (!subvolflg)
+		return false;
+
+	is_home_itself = (strlen(prefix_user_home) - (int)strlen(path)) <= 1;
```
alejandro-colomar: The parentheses and the cast are superfluous.  If you'll refactor this, please remove them.

src/useradd.c
```diff
@@ -2207,6 +2227,29 @@ usr_update (unsigned long subuid_count, unsigned long subgid_count,
 	}
 }
 
+#if WITH_BTRFS
+static bool want_btrfs_subvolume(const char *path)
+{
```
alejandro-colomar: Please move the return type to a separate line:

```c
static bool
want_btrfs_subvolume(const char *path)
```

src/useradd.c
```diff
@@ -2207,6 +2227,29 @@ usr_update (unsigned long subuid_count, unsigned long subgid_count,
...
+	uid_t  uid_min, uid_max;
```
alejandro-colomar: Within this small function, I believe `uid_` seems too redundant.  Maybe calling them `min` and `max` would be simpler.

src/useradd.c
```diff
@@ -2207,6 +2207,17 @@ usr_update (unsigned long subuid_count, unsigned long subgid_count,
...
+#if WITH_BTRFS
+static bool
+want_btrfs_subvolume(const char *path)
+{
+	if (!subvolflg)
+		return false;
+
+	return strlen(prefix_user_home) - strlen(path) <= 1;
+}
+#endif
```
alejandro-colomar, line 2249: For commit 1:

```
```

src/useradd.c
```diff
@@ -2211,10 +2231,21 @@ usr_update (unsigned long subuid_count, unsigned long subgid_count,
...
 want_btrfs_subvolume(const char *path)
 {
+	uid_t  min, max;
+
 	if (!subvolflg)
 		return false;
 
-	return strlen(prefix_user_home) - strlen(path) <= 1;
+	if (strlen(prefix_user_home) - strlen(path) > 1)
+		return false;
```
alejandro-colomar: I think it would be interesting to remove this blank line, to pack all `return false` conditionals.  This will make the `return true` conditional stand out as more different.

Feel free to remove it in the second commit, or in the first commit, whichever you prefer.

alejandro-colomar (approved):
LGTM.  Thanks!

I still prefer that someone else approves the addition of the feature, though.

man/useradd.8.xml
```diff
@@ -156,6 +156,19 @@
 		a Btrfs subvolume is created
 		regardless of any configuration file settings.
 	  </para>
+	  <para>
```
alejandro-colomar, line 159: For commit 2:

```
```

hallyn (approved):
Lgtm, thank you very much.

Only suggestion I would have is that for the new key value setup/teardown in tests, it might be helpful if the comment mentioned login.defs as an example to look through, as otherwise a new test contributor may not know whether they can use it for some other file or not.

ikerexxe (approved):
LGTM!

@alejandro-colomar I'll let you merge since I know you requested something yesterday, but I'm not sure it was done

silverhadch:
> LGTM!
> 
> @alejandro-colomar I'll let you merge since I know you requested something yesterday, but I'm not sure it was done

It was a fixes tag, which I added.

alejandro-colomar:
Thanks!  Merging.

PR #1621, opened 2026-04-30
Add LOGIN_SHELL_FALLBACK to FOREIGNDEFS

stanislav-brabec:
util-linux-2.42 introduced new variable: LOGIN_SHELL_FALLBACK. Add it to known login.defs variables.


hallyn (approved):
Fine...  but that is a funky feature:

```
login(1), if configured with LOGIN_SHELL_FALLBACK in login.defs, can fall back
to another valid shell from /etc/shells if the user's configured shell is
inaccessible due to administrative errors.
```

hallyn:
Thanks for the patch, as indeed it's needed to avoid spurious log messages when that feature is enabled.

PR #1622, opened 2026-04-30
Tests: Usermod rename user, change password, lock password

aborah-sudo:


tests/system/tests/test_usermod.py
```diff
@@ -171,3 +171,85 @@ def test_usermod__set_expire_date_with_empty_date(shadow: Shadow, expiration_dat
...
+        1. Check passwd entry, shadow entry
+        2. Check group entry for user's primary group
+        3. Check group entry for secondary group membership
+    :expectedresults:
+        1. Passwd entry, shadow entry for renamed user exists
+        2. User's primary group still exists with old name
+        3. User is member of daemon group with new name
+    :customerscenario: False
+    """
+    shadow.useradd("-G daemon tuser1")
```
ikerexxe: Create a specific group for this test. `daemon` may not exist in some distribution and that may cause the test to fail

tests/system/tests/test_usermod.py
```diff
@@ -171,3 +171,85 @@ def test_usermod__set_expire_date_with_empty_date(shadow: Shadow, expiration_dat
...
+    shadow.usermod("-l tuser2 tuser1")
+
+    passwd_entry = shadow.tools.getent.passwd("tuser2")
+    assert passwd_entry is not None, "User should be found"
+    assert passwd_entry.name == "tuser2", "Incorrect username"
+
+    shadow_entry = shadow.tools.getent.shadow("tuser2")
+    assert shadow_entry is not None, "User should be found"
+    assert shadow_entry.name == "tuser2", "Incorrect username"
```
ikerexxe: No need for this checks

tests/system/tests/test_usermod.py
```diff
@@ -208,3 +208,49 @@ def test_usermod__rename_user_in_group(shadow: Shadow):
...
+        1. Create user
+        2. Change password using usermod -p
+    :steps:
+        1. Check shadow entry has new password
+    :expectedresults:
+        1. Password is updated in shadow file
+    :customerscenario: False
+    """
+    shadow.useradd("tuser1")
+    shadow.usermod("-p '$1$12345678$0jEt1CYOyE.5NxWFMZyZy1' tuser1")
```
ikerexxe: We have just deprecated and will remove soon MD5 passwords (the ones starting with `$1`). I highly recommend you to generate these passwords on the go, or to at least use SHA-512 (our new default)

tests/system/tests/test_usermod.py
```diff
@@ -208,3 +208,49 @@ def test_usermod__rename_user_in_group(shadow: Shadow):
...
+
+    shadow_entry = shadow.tools.getent.shadow("tuser1")
+    assert shadow_entry is not None, "User should be found"
+    assert shadow_entry.password == "$1$12345678$0jEt1CYOyE.5NxWFMZyZy1", "Password should be updated"
```
ikerexxe: Use `shadow_password_pattern()` to validate the password. This method is used in test_passwd.py, so you can follow the examples there.

tests/system/tests/test_usermod.py
```diff
@@ -208,3 +208,49 @@ def test_usermod__rename_user_in_group(shadow: Shadow):
...
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_usermod__lock_password(shadow: Shadow):
```
ikerexxe: This test should be moved to its own commit

aborah-sudo: Done

tests/system/tests/test_usermod.py
```diff
@@ -4,9 +4,13 @@
 
 from __future__ import annotations
 
+import crypt
```
ikerexxe: [crypt](https://docs.python.org/3/library/crypt.html) was removed in Python 3.13 so I'd suggest you to use [hashlib](https://docs.python.org/3/library/hashlib.html#module-hashlib) instead.

tests/system/tests/test_usermod.py
```diff
@@ -171,3 +175,87 @@ def test_usermod__set_expire_date_with_empty_date(shadow: Shadow, expiration_dat
...
+        3. Lock password
+        4. Unlock password with usermod -U
+    :steps:
+        1. Check shadow entry has unlocked password
+    :expectedresults:
+        1. Password no longer has ! prefix
+    :customerscenario: False
+    """
+    shadow.useradd("tuser1")
+    shadow.usermod("-p '$1$12345678$0jEt1CYOyE.5NxWFMZyZy1' tuser1")
```
ikerexxe: Generate the password on the go as you are doing in `test_usermod__change_password()`

tests/system/tests/test_usermod.py
```diff
@@ -171,3 +175,87 @@ def test_usermod__set_expire_date_with_empty_date(shadow: Shadow, expiration_dat
...
+    shadow.usermod("-L tuser1")
```
ikerexxe: You should add some checks to ensure the password has been locked properly

aborah-sudo: Done

tests/system/tests/test_usermod.py
```diff
@@ -171,3 +175,95 @@ def test_usermod__set_expire_date_with_empty_date(shadow: Shadow, expiration_dat
...
+        3. Lock password
+        4. Unlock password with usermod -U
+    :steps:
+        1. Check shadow entry has unlocked password
+    :expectedresults:
+        1. Password no longer has ! prefix
+    :customerscenario: False
+    """
+    shadow.useradd("tuser1")
+    password_hash = md5_crypt.hash("Secret123")
```
ikerexxe: MD5 is about to be removed so use the new default (SHA512) instead

ikerexxe (commented):
I see you've chosen `passlib` over my suggestion to use `hashlib`. Can you help me understand the technical reasoning behind this choice? I want to make sure I understand the requirements correctly.

In addition, several commit messages mention `usermode`, but that's a different binary that isn't provided by shadow. Do you mind updating the commit messages to `usermod`?

tests/system/tests/test_usermod.py
aborah-sudo: Done

ikerexxe:
I see we went back to the `crypt` module. To clarify, we cannot use `crypt` because it is removed in Python 3.13, and our code will break in the next update.

I also want to avoid `passlib` to keep our project lightweight without external dependencies.

Please implement this using `hashlib`. It is built into Python, secure, and future-proof. Does that make sense?

aborah-sudo:
> I see we went back to the `crypt` module. To clarify, we cannot use `crypt` because it is removed in Python 3.13, and our code will break in the next update.
> 
> I also want to avoid `passlib` to keep our project lightweight without external dependencies.
> 
> Please implement this using `hashlib`. It is built into Python, secure, and future-proof. Does that make sense?

Done

tests/system/tests/test_usermod.py
```diff
@@ -171,3 +176,105 @@ def test_usermod__set_expire_date_with_empty_date(shadow: Shadow, expiration_dat
...
+    password_hash = f"$6${salt}${password_digest}"
+
+    shadow.usermod(f"-p '{password_hash}' tuser1")
+    shadow.usermod("-L tuser1")
+
+    shadow_entry = shadow.tools.getent.shadow("tuser1")
+    assert shadow_entry is not None, "User should be found"
+    assert shadow_entry.password is not None, "Password should not be None"
+    assert shadow_entry.password.startswith("!"), "Password should be locked with ! prefix"
+    assert shadow_entry.password == f"!{password_hash}", "Password should have ! prefix when locked"
```
ikerexxe: This check is missing in the `steps` and `expectedresults`. You need to move some `setup` to `steps` and also add this check

ikerexxe (commented):
There's a critical issue with the current password hash generation approach that needs to be addressed. While the code that you implemented produces a string that looks like a SHA-512 crypt hash (starting with `$6$`), it's not actually a valid SHA-512 crypt hash according to the crypt(3) specification. The issue is that `hashlib.sha512()` performs a simple single-pass hash, whereas the SHA-512 crypt algorithm requires thousands of rounds of key stretching (typically 5000+ iterations) and uses a specific base64 encoding with a custom alphabet. Real SHA-512 crypt hashes have exactly 86 base64-encoded characters after the salt, not 128 hex characters.

The practical impact is that while `usermod -p` will accept and store your generated hash (since it doesn't validate the format), users would never be able to authenticate with passwords set this way because the system's authentication libraries expect real crypt-format hashes.

aborah-sudo:
> There's a critical issue with the current password hash generation approach that needs to be addressed. While the code that you implemented produces a string that looks like a SHA-512 crypt hash (starting with `$6$`), it's not actually a valid SHA-512 crypt hash according to the crypt(3) specification. The issue is that `hashlib.sha512()` performs a simple single-pass hash, whereas the SHA-512 crypt algorithm requires thousands of rounds of key stretching (typically 5000+ iterations) and uses a specific base64 encoding with a custom alphabet. Real SHA-512 crypt hashes have exactly 86 base64-encoded characters after the salt, not 128 hex characters.
> 
> The practical impact is that while `usermod -p` will accept and store your generated hash (since it doesn't validate the format), users would never be able to authenticate with passwords set this way because the system's authentication libraries expect real crypt-format hashes.

That means i have to use passlib as i used before ?

passlib is the officially recommended replacement for the crypt module. It provides cross-platform support for SHA-512 crypt hashes.

tests/system/tests/test_usermod.py
```diff
@@ -206,3 +207,28 @@ def test_usermod__rename_user_in_group(shadow: Shadow):
...
+    shadow.useradd("tuser1")
+
+    password = "Secret123"
+    password_hash = sha512_crypt.hash(password)
+    shadow.usermod(f"-p '{password_hash}' tuser1")
+
+    shadow_entry = shadow.tools.getent.shadow("tuser1")
+    assert shadow_entry is not None, "User should be found"
+    assert shadow_entry.password is not None, "Password should not be None"
+    assert shadow_entry.password.startswith("$6$"), "Password should be SHA-512 crypt hash"
```
ikerexxe, line 234: In addition to these asserts you should also assert the password validity. You can do this by checking the validity itself with `shadow_password_pattern()` or by comparing with the hash that you already generated `shadow_entry.password == password_hash`

tests/system/tests/test_usermod.py
```diff
@@ -232,3 +232,43 @@ def test_usermod__change_password(shadow: Shadow):
...
+    assert shadow_entry is not None, "User should be found"
+    assert shadow_entry.password is not None, "Password should not be None"
+    assert shadow_entry.password.startswith("!"), "Password should be locked with ! prefix"
+    assert shadow_entry.password == f"!{password_hash}", "Password should have ! prefix when locked"
+
+    shadow.usermod("-U tuser1")
+
+    shadow_entry = shadow.tools.getent.shadow("tuser1")
+    assert shadow_entry is not None, "User should be found"
+    assert shadow_entry.password == password_hash, "Password should be unlocked"
```
ikerexxe: Also check `assert shadow_entry.password is not None, "Password should not be None"` before checking the password itself

ikerexxe (commented):
I added some comments inline.

For future PRs, please provide more technical reasoning when choosing (or switching) between libraries. We went through several iterations here because the 'why' behind the implementation wasn't clear. To keep the review process efficient, it’s helpful if you argue your decisions or explain your logic upfront rather than just making silent changes.

tests/system/tests/test_usermod.py
aborah-sudo: Done

tests/system/tests/test_usermod.py
aborah-sudo: Done

ikerexxe (approved):
LGTM!

PR #1623, opened 2026-05-06
userdel: fix user busy detection for threads

haxtibal:
On Linux, userdel/usermod check all /proc/<pid> status files to ensure a to-be-modified user has no more running tasks, or abort modification otherwise.

However, the check failed to detect threads running as the user if the corresponding main thread ran as a different user. The user is deleted despite still being busy. This is due to passing a wrong value to check_status. The caller passed "<pid>/task", rather than "<pid>/task/<tid>". In consequence check_status tried to open "/proc/<pid>/task/status" - a wrong path that never exists - open fails, and check_status always returns 0. The correct status file name would have been "/proc/<pid>/task/<tid>/status" instead.

The bug can only be reproduced by rather exotic code using raw syscalls. POSIX does not allow threads to have different UIDs.

To fix it, construct the correct path to the tid status file. Also change the interface of check_status and different_namespace to take pid and tid instead of a partially constructed path string. This is simpler and makes similar bugs less likely.

Behavior without fix:

```sh
userdel testuser  # testuser has threads with tid uid != pid uid
``` 
=> no output, testuser deleted despite being busy

With the fix:

```
userdel testuser  # testuser has threads with tid uid != pid uid
userdel: user testuser is currently used by process 178863
```
=> testuser detected as busy, not deleted

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -4,8 +4,10 @@ autoreconf -v -f --install "$(dirname "$0")" || exit 1
 
 CFLAGS="-O2"
 CFLAGS="$CFLAGS -Wall"
+CFLAGS="$CFLAGS -Wformat=2"
 CFLAGS="$CFLAGS -Wextra"
 CFLAGS="$CFLAGS -Werror=discarded-qualifiers"
+CFLAGS="$CFLAGS -Werror=format"
 CFLAGS="$CFLAGS -Werror=implicit-function-declaration"
 CFLAGS="$CFLAGS -Werror=implicit-int"
 CFLAGS="$CFLAGS -Werror=incompatible-pointer-types"
--- a/lib/user_busy.c
+++ b/lib/user_busy.c
@@ -216,7 +216,7 @@ static int user_busy_processes (const char *name, uid_t uid)
 		}
 
 		/* Check if the process is in our chroot */
-		stprintf_a(root_path, "/proc/%lu/root", (unsigned long) pid);
+		stprintf_a(root_path, "/proc/%d/root", pid);
 		if (stat (root_path, &sbroot_process) != 0) {
 			continue;
 		}
@@ -236,7 +236,7 @@ static int user_busy_processes (const char *name, uid_t uid)
 			return 1;
 		}
 
-		stprintf_a(task_path, "/proc/%lu/task", (unsigned long) pid);
+		stprintf_a(task_path, "/proc/%d/task", pid);
 		task_dir = opendir (task_path);
 		if (task_dir != NULL) {
 			while (NULL != (ent = readdir(task_dir))) {
--- a/lib/user_busy.c
+++ b/lib/user_busy.c
@@ -36,7 +36,7 @@
 
 
 #ifdef __linux__
-static int check_status (const char *name, const char *sname, uid_t uid);
+static int check_status (const char *name, uid_t uid, pid_t pid, pid_t tid);
 static int user_busy_processes (const char *name, uid_t uid);
 #else				/* !__linux__ */
 static int user_busy_utmp (const char *name);
@@ -94,13 +94,13 @@ user_busy_utmp(const char *name)
 #ifdef __linux__
 #ifdef ENABLE_SUBIDS
 #define in_parentuid_range(uid) ((uid) >= parentuid && (uid) < parentuid + range)
-static int different_namespace (const char *sname)
+static int different_namespace (pid_t pid, pid_t tid)
 {
 	/* 41: /proc/xxxxxxxxxx/task/xxxxxxxxxx/ns/user + \0 */
 	char     path[41];
 	char     buf[512], buf2[512];
 
-	stprintf_a(path, "/proc/%s/ns/user", sname);
+	stprintf_a(path, "/proc/%d/task/%d/ns/user", pid, tid);
 
 	if (readlinknul_a(path, buf) == -1)
 		return 0;
@@ -116,14 +116,14 @@ static int different_namespace (const char *sname)
 #endif                          /* ENABLE_SUBIDS */
 
 
-static int check_status (const char *name, const char *sname, uid_t uid)
+static int check_status (const char *name, uid_t uid, pid_t pid, pid_t tid)
 {
 	/* 40: /proc/xxxxxxxxxx/task/xxxxxxxxxx/status + \0 */
 	char  status[40];
 	char  line[1024];
 	FILE  *sfile;
 
-	stprintf_a(status, "/proc/%s/status", sname);
+	stprintf_a(status, "/proc/%d/task/%d/status", pid, tid);
 
 	sfile = fopen (status, "r");
 	if (NULL == sfile) {
@@ -144,7 +144,7 @@ static int check_status (const char *name, const char *sname, uid_t uid)
 					return 1;
 				}
 #ifdef ENABLE_SUBIDS
-				if (    different_namespace (sname)
+				if (    different_namespace (pid, tid)
 				     && (   have_sub_uids(name, ruid, 1)
 				         || have_sub_uids(name, euid, 1)
 				         || have_sub_uids(name, suid, 1))
@@ -225,7 +225,7 @@ static int user_busy_processes (const char *name, uid_t uid)
 			continue;
 		}
 
-		if (check_status (name, tmp_d_name, uid) != 0) {
+		if (check_status (name, uid, pid, pid) != 0) {
 			(void) closedir (proc);
 #ifdef ENABLE_SUBIDS
 			sub_uid_close(true);
@@ -241,17 +241,13 @@ static int user_busy_processes (const char *name, uid_t uid)
 		if (task_dir != NULL) {
 			while (NULL != (ent = readdir(task_dir))) {
 				pid_t tid;
-				/* 27: xxxxxxxxxx/task/xxxxxxxxxx + \0 */
-				char  tid_sname[27];
-
 				if (get_pid(ent->d_name, &tid) == -1) {
 					continue;
 				}
 				if (tid == pid) {
 					continue;
 				}
-				stprintf_a(tid_sname, "%ld/task/%ld", (long) pid, (long) tid);
-				if (check_status (name, tid_sname, uid) != 0) {
+				if (check_status (name, uid, pid, tid) != 0) {
 					(void) closedir (proc);
 					(void) closedir (task_dir);
 #ifdef ENABLE_SUBIDS
--- a/lib/user_busy.c
+++ b/lib/user_busy.c
@@ -241,13 +241,17 @@ static int user_busy_processes (const char *name, uid_t uid)
 		if (task_dir != NULL) {
 			while (NULL != (ent = readdir(task_dir))) {
 				pid_t tid;
+				/* 27: xxxxxxxxxx/task/xxxxxxxxxx + \0 */
+				char  tid_sname[27];
+
 				if (get_pid(ent->d_name, &tid) == -1) {
 					continue;
 				}
 				if (tid == pid) {
 					continue;
 				}
-				if (check_status (name, task_path+6, uid) != 0) {
+				stprintf_a(tid_sname, "%ld/task/%ld", (long) pid, (long) tid);
+				if (check_status (name, tid_sname, uid) != 0) {
 					(void) closedir (proc);
 					(void) closedir (task_dir);
 #ifdef ENABLE_SUBIDS
@@ -272,4 +276,3 @@ static int user_busy_processes (const char *name, uid_t uid)
 	return 0;
 }
 #endif				/* __linux__ */
-
--- a/tests/unit/Makefile.am
+++ b/tests/unit/Makefile.am
@@ -100,6 +100,7 @@ test_snprintf_SOURCES = \
     $(NULL)
 test_snprintf_CFLAGS = \
     $(AM_CFLAGS) \
+    -Wno-format-zero-length \
     $(NULL)
 test_snprintf_LDFLAGS = \
     $(NULL)


haxtibal:
Reproducer:

*setuid.py*
```py
#!/usr/bin/env python3
import ctypes, os, pwd, sys, threading, time

SYS_setuid = 105  # setuid, x86_64
libc = ctypes.CDLL(None)
libc.syscall.restype = ctypes.c_long
libc.syscall.argtypes = [ctypes.c_long, ctypes.c_long]
uid = pwd.getpwnam(sys.argv[1]).pw_uid

def thread_func():
    libc.syscall(SYS_setuid, uid)
    while True:
        print(f"thread running as uid {uid} (pid={os.getpid()})", flush=True)
        time.sleep(5)

threading.Thread(target=thread_func, daemon=True).start()
time.sleep(60)
```

Start snippet as root. It's main thread has uid 0, spawned thread has uid of testuser. Then delete user while program is running.
```
sudo python3 setuid.py testuser &
sudo userdel testuser  # should detect thread running as testuser and abort, but doesn't detect and deletes testuser
```

lib/user_busy.c
```diff
@@ -94,13 +94,13 @@ user_busy_utmp(const char *name)
...
 #define in_parentuid_range(uid) ((uid) >= parentuid && (uid) < parentuid + range)
-static int different_namespace (const char *sname)
+static int different_namespace (pid_t pid, pid_t tid)
 {
 	/* 41: /proc/xxxxxxxxxx/task/xxxxxxxxxx/ns/user + \0 */
 	char     path[41];
 	char     buf[512], buf2[512];
 
-	stprintf_a(path, "/proc/%s/ns/user", sname);
+	stprintf_a(path, "/proc/%d/task/%d/ns/user", pid, tid);
```
alejandro-colomar, line 103: As far as I know, `pid_t` is not required by POSIX to be `int`.  If we want to use `%d`, we should at least have `-Wformat=2 -Werror=format` to make sure this doesn't produce UB.  Alternatively, you may want to use intmax_t for printing, although if we're pretty confident that `pid_t` is an `int`, `%d` should be good if combined with the compiler error.

lib/user_busy.c
```diff
@@ -94,13 +94,13 @@ user_busy_utmp(const char *name)
 #ifdef __linux__
 #ifdef ENABLE_SUBIDS
 #define in_parentuid_range(uid) ((uid) >= parentuid && (uid) < parentuid + range)
-static int different_namespace (const char *sname)
+static int different_namespace (pid_t pid, pid_t tid)
 {
```
alejandro-colomar, line 98: From the commit message:

```
To fix it, construct the correct path to the tid status file. Also
change the interface of check_status and different_namespace to take pid
and tid instead of a partially constructed path string. This is simpler
and makes similar bugs less likely.
```

I think I prefer if such refactoring occurs in a separate commit.

alejandro-colomar:
> Reproducer:
> 
> _setuid.py_
> 
> ```python
> #!/usr/bin/env python3
> import ctypes, os, pwd, sys, threading, time
> 
> SYS_setuid = 105  # setuid, x86_64
> libc = ctypes.CDLL(None)
> libc.syscall.restype = ctypes.c_long
> libc.syscall.argtypes = [ctypes.c_long, ctypes.c_long]
> uid = pwd.getpwnam(sys.argv[1]).pw_uid
> 
> def thread_func():
>     libc.syscall(SYS_setuid, uid)
>     while True:
>         print(f"thread running as uid {uid} (pid={os.getpid()})", flush=True)
>         time.sleep(5)
> 
> threading.Thread(target=thread_func, daemon=True).start()
> time.sleep(60)
> ```
> 
> Start snippet as root. It's main thread has uid 0, spawned thread has uid of testuser. Then delete user while program is running.
> 
> ```
> sudo python3 setuid.py testuser &
> sudo userdel testuser  # should detect thread running as testuser and abort, but doesn't detect and deletes testuser
> ```

Could you please use only C and sh(1) (or bash(1)) for the reproducer?  Python is hard to understand to me.

lib/user_busy.c
haxtibal: > As far as I know, pid_t is not required by POSIX to be int

This is true for POSIX, but we're now inside `#ifdef __linux__`, and for Linux we have
```
typedef int		__kernel_pid_t;
```
for all platforms (see [uapi/asm-generic/posix_types.h](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/uapi/asm-generic/posix_types.h#n28)). Further, the existing code also uses `%d` for pid_t in some places.

However the existing code does it inconsistently. If you tell what's your preference I could add a commit that aligns pid_t formatting for the whole user_busy.c file.

haxtibal:
> Could you please use only C and sh(1) (or bash(1)) for the reproducer? Python is hard to understand to me.

Sure, try this:

*setuid_thread.c*
```c
#include <pthread.h>
#include <pwd.h>
#include <stdio.h>
#include <stdlib.h>
#include <sys/syscall.h>
#include <unistd.h>

static uid_t target_uid;

static void *user_thread(void *arg)
{
        syscall(SYS_setuid, (long)target_uid);
        for (;;) {
                printf("thread running as uid %d (pid=%d)\n", (int)target_uid,
                       (int)getpid());
                sleep(5);
        }
        return NULL;
}

int main(int argc, char *argv[])
{
        if (argc < 2) {
                fprintf(stderr, "Usage: %s <username>\n", argv[0]);
                return 1;
        }

        struct passwd *pw = getpwnam(argv[1]);
        if (!pw) {
                fprintf(stderr, "user not found: %s\n", argv[1]);
                return 1;
        }
        target_uid = pw->pw_uid;

        pthread_t tid;
        pthread_create(&tid, NULL, user_thread, NULL);
        sleep(60);
        return 0;
}
```

Then
```sh
gcc setuid_thread.c -o setuid_thread
sudo useradd --no-create-home testuser
sudo ./setuid_thread testuser &
sudo userdel testuser  # should detect thread running as testuser and abort, but doesn't detect and deletes testuser
```

lib/user_busy.c
alejandro-colomar: I like using `%d`.  However, I'd like to have a separate commit adding the compiler error so that we make sure we don't have UB.

Consider the possibility that we paste that code somewhere else that is not wrapped in `#ifdef __linux__`.  It would be an easy and dangerous bug.  If the compiler prevents us from doing that thing, I'll be happier.

alejandro-colomar:
> > Could you please use only C and sh(1) (or bash(1)) for the reproducer? Python is hard to understand to me.
> 
> Sure, try this:

Thanks!

> 
> _setuid_thread.c_
> 
> ```c
> #include <pthread.h>
> #include <pwd.h>
> #include <stdio.h>
> #include <stdlib.h>
> #include <sys/syscall.h>
> #include <unistd.h>
> 
> static uid_t target_uid;
> 
> static void *user_thread(void *arg)
> {
>         syscall(SYS_setuid, (long)target_uid);
>```

Is the raw system call needed, or can we call the glibc wrapper?
```c
        setuid(target_uid);
```

From what you said, I think the raw syscall is necessary, but just to confirm.

>```c
>         for (;;) {
>                 printf("thread running as uid %d (pid=%d)\n", (int)target_uid,
>                        (int)getpid());
>                 sleep(5);
>         }
>         return NULL;
> }
> 
> int main(int argc, char *argv[])
> {
>         if (argc < 2) {
>                 fprintf(stderr, "Usage: %s <username>\n", argv[0]);
>                 return 1;
>         }
> 
>         struct passwd *pw = getpwnam(argv[1]);
>         if (!pw) {
>                 fprintf(stderr, "user not found: %s\n", argv[1]);
>                 return 1;
>         }
>         target_uid = pw->pw_uid;
> 
>         pthread_t tid;
>         pthread_create(&tid, NULL, user_thread, NULL);
>         sleep(60);
>         return 0;
> }
> ```
> 
> Then
> 
> ```shell
> gcc setuid_thread.c -o setuid_thread
> sudo useradd --no-create-home testuser
> sudo ./setuid_thread testuser &
> sudo userdel testuser  # should detect thread running as testuser and abort, but doesn't detect and deletes testuser
> ```

haxtibal:
> Is the raw system call needed,

Yes, see [man 2 setuid](https://man7.org/linux/man-pages/man2/setuid.2.html#VERSIONS):

```
At the kernel level, user IDs and group IDs are a per-thread
attribute.  However, POSIX requires that all threads in a process
share the same credentials.  The NPTL threading implementation
handles the POSIX requirements by providing wrapper functions for
the various system calls that change process UIDs and GIDs.  These
wrapper functions (including the one for setuid()) employ a
signal-based technique to ensure that when one thread changes
credentials, all of the other threads in the process also change
their credentials.
```

> or can we call the glibc wrapper?

To my knowledge there's no other way than raw syscalls to get that Linux specific feature of different thread UIDs within one process.

alejandro-colomar:
> > Is the raw system call needed,
> 
> Yes, see [man 2 setuid](https://man7.org/linux/man-pages/man2/setuid.2.html#VERSIONS):
> 
> ```
> At the kernel level, user IDs and group IDs are a per-thread
> attribute.  However, POSIX requires that all threads in a process
> share the same credentials.  The NPTL threading implementation
> handles the POSIX requirements by providing wrapper functions for
> the various system calls that change process UIDs and GIDs.  These
> wrapper functions (including the one for setuid()) employ a
> signal-based technique to ensure that when one thread changes
> credentials, all of the other threads in the process also change
> their credentials.
> ```
> 
> > or can we call the glibc wrapper?
> 
> To my knowledge there's no other way than raw syscalls to get that Linux specific feature of different thread UIDs within one process.

Thanks for conforming!

alejandro-colomar:
> > Could you please use only C and sh(1) (or bash(1)) for the reproducer? Python is hard to understand to me.
> 
> Sure, try this:
> 
> _setuid_thread.c_
> 
> ```c
> #include <pthread.h>
> #include <pwd.h>
> #include <stdio.h>
> #include <stdlib.h>
> #include <sys/syscall.h>
> #include <unistd.h>
> 
> static uid_t target_uid;
> 
> static void *user_thread(void *arg)
> {
>         syscall(SYS_setuid, (long)target_uid);
>         for (;;) {
>                 printf("thread running as uid %d (pid=%d)\n", (int)target_uid,
>                        (int)getpid());
>                 sleep(5);
>         }
>         return NULL;
> }
> 
> int main(int argc, char *argv[])
> {
>         if (argc < 2) {
>                 fprintf(stderr, "Usage: %s <username>\n", argv[0]);
>                 return 1;
>         }
> 
>         struct passwd *pw = getpwnam(argv[1]);
>         if (!pw) {
>                 fprintf(stderr, "user not found: %s\n", argv[1]);
>                 return 1;
>         }
>         target_uid = pw->pw_uid;
> 
>         pthread_t tid;
>         pthread_create(&tid, NULL, user_thread, NULL);
>         sleep(60);
>         return 0;
> }
> ```
> 
> Then
> 
> ```shell
> gcc setuid_thread.c -o setuid_thread
> sudo useradd --no-create-home testuser
> sudo ./setuid_thread testuser &
> sudo userdel testuser  # should detect thread running as testuser and abort, but doesn't detect and deletes testuser
> ```

Could you please include this program (and the corresponding shell session below it) in the commit message?  Please indent it, so that the includes are not treated as comments.

lib/user_busy.c
```diff
@@ -241,13 +241,18 @@ static int user_busy_processes (const char *name, uid_t uid)
...
 				if (get_pid(ent->d_name, &tid) == -1) {
 					continue;
 				}
 				if (tid == pid) {
 					continue;
 				}
-				if (check_status (name, task_path+6, uid) != 0) {
+				stprintf_a(tid_sname, "%lu/task/%lu",
+				           (unsigned long) pid, (unsigned long) tid);
+				if (check_status (name, tid_sname, uid) != 0) {
```
alejandro-colomar: `pid_t` is guaranteed to be a signed integer.  We should probably use `long` instead of `unsigned long`.  That would keep the lines shorter.

lib/user_busy.c
```diff
@@ -36,7 +36,7 @@
```
alejandro-colomar, line 37: In the commit message, there's:

> This is simpler and counters bugs like https://github.com/shadow-maint/shadow/commit/821db7c91334801fa5d0ff17212677da13f010db ("userdel: fix user
busy detection for threads") by design.

The commit hash is already stale, and would anyway become stale once we merge.  You could maybe instead put a link to this PR, or refer to "the previous commit".

lib/user_busy.c
alejandro-colomar: Thanks!

alejandro-colomar:
Thanks!  It looks good.  I'll test it tomorrow.

lib/user_busy.c
haxtibal: CI tests failed since I missed that `test_snprintf` intentionally tests `""` as format string, which is is now considered an error.

Added an exception for that test, it should work again. Also added a small commit that adjusts two `%d`s that were not directly related to the PR, but should be adjusted for consistency.

alejandro-colomar: Oh!  Maybe we should transform it to `"%s", ""`.  That achieves the same thing, and should appease the compiler.

alejandro-colomar: Although your approach of passing `-Wno-format-zero-length` in the tests seems better.  Thanks!

lib/user_busy.c
```diff
@@ -216,7 +216,7 @@ static int user_busy_processes (const char *name, uid_t uid)
 		}
 
 		/* Check if the process is in our chroot */
-		stprintf_a(root_path, "/proc/%lu/root", (unsigned long) pid);
+		stprintf_a(root_path, "/proc/%d/root", pid);
```
alejandro-colomar, line 219: For commit 4 ("lib: apply linux specific pid_t format consistently"):

```
```

Thanks!

autogen.sh
```diff
@@ -4,8 +4,10 @@ autoreconf -v -f --install "$(dirname "$0")" || exit 1
 
 CFLAGS="-O2"
 CFLAGS="$CFLAGS -Wall"
+CFLAGS="$CFLAGS -Wformat=2"
 CFLAGS="$CFLAGS -Wextra"
 CFLAGS="$CFLAGS -Werror=discarded-qualifiers"
+CFLAGS="$CFLAGS -Werror=format"
```
alejandro-colomar, line 10: For commit 3 ("autogen.sh: turn format issues into compiler errors "):

```
```

Thanks!

alejandro-colomar:
Sorry for the delay!  I forgot about this, due to other stuff.  I've tested it now, and I can reproduce it.

alejandro-colomar (approved):
```
```

Thanks!

haxtibal:
> Thanks!

No problem, thank you for the review :)

Issue #1629, opened 2026-05-23
shadow Security Hole Introduced / Regression in de7f1c78f70f8df4b2956f7363da3774348bc58e


MillaFleurs:
As per conversation with @alejandro-colomar logging as a public issue.

------------------------

[fix.patch](https://github.com/user-attachments/files/28183711/fix.patch)
[poc-real-useradd-prefix.sh](https://github.com/user-attachments/files/28183710/poc-real-useradd-prefix.sh)

Good Morning,

I’ve just found a security regression introduced by commit de7f1c78f70f8df4b2956f7363da3774348bc58e

In a nutshell, the commit updated lib/subordinateio.c which now allows creation of overlapping subordinate ID ranges.  This is caused because the commit changed from unsigned long to id_t.   The bug is that the internal algorithm still used an exclusive endpoint expression, max + 1, which was only safe while the intermediate type was wide enough.  So find_free_range() accepts max as an inclusive ID ceiling, but internally compares holes against max + 1.  

When max is (id_t)-1 on an unsigned-ID build, max + 1 wraps to zero. new_subid_range() passes exactly that value as the ceiling.

I’m providing a POC that shows this on current master.    The POC does NOT work in my current Debian test environments, so I suspect it’s possible it's not yet out into the wild, although I don’t know if that’s true.  I recommend either reverting the commit or using the attached patch if we want to keep the change from unsigned long to id_t.

Patch is attached for your convenience as well.

Full Disclosure: This was found with N184, my open source bug and security vulnerability scanner, although this bug report (and patch) are my own work.  You can learn more about N184 here: https://github.com/MillaFleurs/N184

Note: to use the POC you need to point to the correct useradd (which can be compiled but not installed), as in: $USERADD_BIN=/usr/sbin/useradd  ./poc.sh

Thank you,

Dan

[fix.patch](https://github.com/user-attachments/files/28183712/fix.patch)

[poc-real-useradd-prefix.sh](https://github.com/user-attachments/files/28183714/poc-real-useradd-prefix.sh)

PR #1630, opened 2026-05-23
Fix bug from de7f1c78f70f8df4b2956f7363da3774348bc58e that caused security regression

MillaFleurs:
In a nutshell, the commit updated lib/subordinateio.c which now allows creation of overlapping subordinate ID ranges.  This is caused because the commit changed from unsigned long to id_t.   The bug is that the internal algorithm still used an exclusive endpoint expression, max + 1, which was only safe while the intermediate type was wide enough.  So find_free_range() accepts max as an inclusive ID ceiling, but internally compares holes against max + 1.  

When max is (id_t)-1 on an unsigned-ID build, max + 1 wraps to zero. new_subid_range() passes exactly that value as the ceiling.

This patch fixes the above issue as detailed in #1629 

Full Disclosure: This was found with N184, my open source bug and security vulnerability scanner, although this bug report (and patch) are my own work.  You can learn more about N184 here: https://github.com/MillaFleurs/N184

### DIFF
--- a/lib/subordinateio.c
+++ b/lib/subordinateio.c
@@ -333,11 +333,12 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
 static id_t
 find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 {
-	uintmax_t                      low, max_id;
+	intmax_t                       low, max_id;
 	const struct subordinate_range *range;
 
 	if (count == 0 || max < min
-	    || (uintmax_t) count - 1 > (uintmax_t) max - min) {
+	    || (uintmax_t) count - 1 > (uintmax_t) max - min)
+	{
 		errno = ERANGE;
 		return -1;
 	}
@@ -346,38 +347,40 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 	commonio_sort (db, subordinate_range_cmp);
 	commonio_rewind(db);
 
-	low = (uintmax_t) min;
-	max_id = (uintmax_t) max;
+	low = min;
+	max_id = max;
 	while (NULL != (range = commonio_next(db))) {
-		uintmax_t  first, last;
+		intmax_t  first, last;
 
 		if (range->count == 0)
 			continue;
 
 		first = range->start;
 		if (__builtin_add_overflow(first, range->count - 1, &last))
-			last = UINTMAX_MAX;
+			last = INTMAX_MAX;
 
 		/*
-		 * Check the hole before this range as an inclusive interval.
-		 * This avoids constructing max + 1, which wraps when max is
-		 * the largest representable id_t value.
+		 * Ranges are sorted by start, and low has been advanced past
+		 * every range seen so far, so the gap [low, first - 1] (clamped
+		 * to max) contains no existing range.  A block that fits in
+		 * this gap lies below first, hence below every later range too,
+		 * so it cannot overlap any allocation.
 		 */
 		if (first > low) {
-			uintmax_t high = first - 1;
+			intmax_t high = first - 1;
 
 			/* Don't allocate IDs after max (included). */
 			if (high > max_id)
 				high = max_id;
 
 			/* Is the hole before this range large enough? */
-			if ((high >= low) && ((high - low + 1) >= count))
-				return (id_t) low;
+			if ((high >= low) && ((high - low + 1) >= (intmax_t) count))
+				return low;
 		}
 
 		/* Compute the low end of the next hole */
 		if (last >= low) {
-			if (last == UINTMAX_MAX)
+			if (last == INTMAX_MAX)
 				goto fail;
 			low = last + 1;
 		}
@@ -386,8 +389,8 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 	}
 
 	/* Is the remaining unclaimed area large enough? */
-	if (((max_id - low) + 1) >= count)
-		return (id_t) low;
+	if (((max_id - low) + 1) >= (intmax_t) count)
+		return low;
 fail:
 	errno = EUSERS;
 	return -1;
--- a/lib/subordinateio.c
+++ b/lib/subordinateio.c
@@ -17,6 +17,7 @@
 #include <pwd.h>
 #include <ctype.h>
 #include <fcntl.h>
+#include <stdint.h>
 #include <string.h>
 
 #include "alloc/malloc.h"
@@ -332,10 +333,11 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
 static id_t
 find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 {
-	id_t                           low, high;
+	uintmax_t                      low, max_id;
 	const struct subordinate_range *range;
 
-	if (count == 0 || max < min || count - 1 > max - min) {
+	if (count == 0 || max < min
+	    || (uintmax_t) count - 1 > (uintmax_t) max - min) {
 		errno = ERANGE;
 		return -1;
 	}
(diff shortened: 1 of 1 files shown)


lib/subordinateio.c
```diff
@@ -344,35 +346,48 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
...
+		}
+		if (low > max_id)
 			goto fail;
 	}
 
 	/* Is the remaining unclaimed area large enough? */
-	if (((max - low) + 1) >= count)
-		return low;
+	if (((max_id - low) + 1) >= count)
+		return (id_t) low;
```
alejandro-colomar: This cast is superfluous.  It's fine to do implicit narrowing conversions.  Please remove (unless I'm missing something).

lib/subordinateio.c
```diff
@@ -332,10 +333,11 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
 static id_t
 find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 {
-	id_t                           low, high;
+	uintmax_t                      low, max_id;
```
alejandro-colomar: To have stronger measures against this bug, maybe we should use signed integer arithmetic, so that integer overflow is a bug, and potentially produces diagnostics (including from sanitizers).  How about using `intmax_t`?

lib/subordinateio.c
```diff
@@ -332,10 +333,11 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
 	const struct subordinate_range *range;
 
-	if (count == 0 || max < min || count - 1 > max - min) {
+	if (count == 0 || max < min
+	    || (uintmax_t) count - 1 > (uintmax_t) max - min) {
 		errno = ERANGE;
 		return -1;
 	}
```
alejandro-colomar, line 344: Please move the opening brace to a separate line.  See the style for braces:

<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c#n46>

```
    Braces
	its own.

		if (foo)
			bar();

		{
			baz();
		}
```

MillaFleurs (commented):
Updates made.

lib/subordinateio.c
ikerexxe: Are you trying to avoid truncation problems here?

alejandro-colomar: @ikerexxe Are you asking me or the OP (@MillaFleurs)?

ikerexxe: It was more of a question/guess for the author

lib/subordinateio.c
```diff
@@ -332,10 +333,12 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
 find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 {
-	id_t                           low, high;
+	intmax_t                       low, max_id;
 	const struct subordinate_range *range;
 
-	if (count == 0 || max < min || count - 1 > max - min) {
+	if (count == 0 || max < min
+	    || (uintmax_t) count - 1 > (uintmax_t) max - min)
+	{
```
alejandro-colomar, line 341: I would also make these casts be signed.

I think the following should be okay:

```c
if (count == 0 || max < min
 || (intmax_t) count - 1 > (intmax_t) max - min)
{
```

Also, I think we can avoid one cast if we re-organize the comparison:

```c
if (count == 0 || max < min || (intmax_t) count > max - min + 1LL) {
```

Because both `max` and `min` are id_t, and `max >= min`, their subtraction is in the range `0 .. (id_t)-1`, and adding `1LL` to that is guaranteed to fit in long long.  Because `long long` is wider than `id_t`, this is going to be signed.  (And we also make the conditional fit in one line.)

lib/subordinateio.c
```diff
@@ -345,33 +348,48 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
...
+		if (range->count == 0)
+			continue;
 
-		/* Don't allocate IDs after max (included) */
-		if (high > max + 1) {
-			high = max + 1;
+		first = range->start;
+		if (__builtin_add_overflow(first, range->count - 1, &last))
+			last = INTMAX_MAX;
+
```
alejandro-colomar, line 361: Is this really necessary to fix the regression?

This arithmetic seems to be done on `unsigned long`, and hasn't changed to id_t in the commit that introduced the regression, AFAICS.

I'm not saying this change is not okay, but I think that's fixing a separate problem, and that should be done in a separate commit whose commit message documents clearly.

Or am I missing some reason why this would be necessary to fix the regression?

alejandro-colomar: I would like the commit that fixes the regression to be minimal.

lib/subordinateio.c
```diff
@@ -345,33 +348,48 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
...
+		/*
+		 * Ranges are sorted by start, and low has been advanced past
+		 * every range seen so far, so the gap [low, first - 1] (clamped
+		 * to max) contains no existing range.  A block that fits in
+		 * this gap lies below first, hence below every later range too,
+		 * so it cannot overlap any allocation.
+		 */
+		if (first > low) {
+			intmax_t high = first - 1;
```
alejandro-colomar, line 370: Please separate declarations from code:

```c
if (first > low) {
	intmax_t  high;

	high = first - 1;
```

lib/subordinateio.c
```diff
@@ -345,33 +348,48 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
...
 			low = last + 1;
-		if (low > max)
+		}
+		if (low > max_id)
 			goto fail;
 	}
 
 	/* Is the remaining unclaimed area large enough? */
-	if (((max - low) + 1) >= count)
+	if (((max_id - low) + 1) >= (intmax_t) count)
```
alejandro-colomar, line 392: Should we make the `count` parameter be of type `long long` to avoid these casts?

lib/subordinateio.c
```diff
@@ -345,33 +348,48 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
...
-		if (low < (last + 1))
+		if (last >= low) {
+			if (last == INTMAX_MAX)
+				goto fail;
 			low = last + 1;
-		if (low > max)
+		}
+		if (low > max_id)
 			goto fail;
 	}
```
alejandro-colomar, line 389: Considering that `INTMAX_MAX` is necessarily going to be `>= low`, we can flatten that code:

```c
if (last == INTMAX_MAX)
	goto fail;
if (last >= low)
	low = last + 1;
if (low > max_id)
	goto fail;
```

However, is it really possible for `last` to be `INTMAX_MAX`?  Or can we entirely remove that branch?

hallyn:
Hi @MillaFleurs ,

for the patch description, please a) change the first line to a concise (< 80 characters) description of the problem being fixed, such as "subids: fix security regression which can cause overlapping ranges".  Put your description from the top of the PR into the commit description.  And add a Closes tag.  So something like:

```
subids: fix security regression which can cause overlapping ranges

Fix bug from de7f1c78f70f8df4b2956f7363da3774348bc58e that changed from
using unsigned long to id_t.  new_subid_range() returns the same value
as the ceiling allowing potentially overlapping subordinate ID ranges.

In a nutshell, the commit updated lib/subordinateio.c which now allows
creation of overlapping subordinate ID ranges. This is caused because
the commit changed from unsigned long to id_t. The bug is that the
internal algorithm still used an exclusive endpoint expression, max + 1,
which was only safe while the intermediate type was wide enough. So
find_free_range() accepts max as an inclusive ID ceiling, but internally
compares holes against max + 1.

When max is (id_t)-1 on an unsigned-ID build, max + 1 wraps to zero.
new_subid_range() passes exactly that value as the ceiling.

This was found with N184, my open source bug and security vulnerability
scanner, although this bug report (and patch) are my own work.  You can
learn more about N184 here: https://github.com/MillaFleurs/N184

Closes: <https://github.com/shadow-maint/shadow/issues/1629>
```

[Note: edited by @alejandro-colomar  :]

lib/subordinateio.c
MillaFleurs: Note you're looking at an outdated commit @ikerexxe and the explicit conversion was removed.  

The thought process is:

id_t (what we're returning and what we don't want wrap / overflow on) is 32 bits wide so intermediate values like max + 1 or first + count - 1 could exceed the max of id_t, so we need more headroom.  The current commit uses intmax_t which is 64 bits wide which is more than sufficient.

Once the comparison logic is done, there is no need to keep it in intmax_t.  Presumably there won't be much overhead because everything's 64 bit these days but we're also not in a critical code branch where we would care about conversion overhead.

By using signed intmax_t for this (per @alejandro-colomar), an overflow would be UB a sanitizer can catch rather than a silent wrap, and it returns low directly. 

Does that make sense?

lib/subordinateio.c
MillaFleurs: Done by most recent commit

lib/subordinateio.c
```diff
@@ -332,10 +333,12 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
 static id_t
 find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 {
```
alejandro-colomar, line 335: Since we're relying on the fact that `intmax_t` is strictly wider than `id_t`, we could declare that in a static assertion (just to preclude any hypothetical future systems where `id_t` might be 64-bit).

How about adding this at the top of this function?

```c
static_assert(sizeof(intmax_t) > sizeof(id_t), "");
```

lib/subordinateio.c
ikerexxe: Ok, thank you for the explanation

silverhadch:
@MillaFleurs 

Are you still interested on working on this? If not, you mind me taking this over?

alejandro-colomar:
> @MillaFleurs
> 
> Are you still interested on working on this? If not, you mind me taking this over?

@hallyn and I were talking about this last week.  We'd like to release soon-ish, so if you want to work on this regression now, it'd be very helpful.  Please keep @MillaFleurs as

```
Co-authored-by: Dan Anderson <https://github.com/MillaFleurs>
```

in the commit messages if you continue his work.

silverhadch:
@alejandro-colomar 

This can be closed then.

alejandro-colomar:
Yup.  @MillaFleurs , if you think there's anything else that needs to be done, please open a new PR.

Closing as superseded by #1668.

PR #1631, opened 2026-06-01
Expand the use of --with-fcaps to other binaries

anthonyryan1:
This is defense in depth attack surface reduction.

During recent kernel bugs, PoCs overwrote the kernel page cache to
inject attacker payloads into setuid binaries for execution.

The kernel forms the trusted computing base for shadow, so it's not
technically shadow's responsibility to mitigate.

But it's relatively simple change that gives us protection from both
future or undiscovered bugs in shadow binaries, and limits the number
of useful setuid targets to attackers weaponizing future kernel bugs.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -550,7 +550,7 @@ fi
 AM_CONDITIONAL([USE_PAM], [test "X$with_libpam" = "Xyes"])
 
 AC_ARG_WITH([fcaps],
-	[AS_HELP_STRING([--with-fcaps], [use file capabilities instead of suid binaries for newuidmap/newgidmap @<:@default=no@:>@])],
+	[AS_HELP_STRING([--with-fcaps], [use file capabilities instead of suid binaries where possible @<:@default=no@:>@])],
 	[with_fcaps=$withval], [with_fcaps=no])
 AM_CONDITIONAL([FCAPS], [test "x$with_fcaps" = "xyes"])
 
--- a/lib/find_new_sub_gids.c
+++ b/lib/find_new_sub_gids.c
@@ -10,6 +10,7 @@
 
 #include <stdio.h>
 #include <errno.h>
+#include <stdint.h>
 
 #include "prototypes.h"
 #include "subordinateio.h"
--- a/lib/pwd_init.c
+++ b/lib/pwd_init.c
@@ -51,6 +51,7 @@ void pwd_init (void)
 	signal (SIGTERM, SIG_IGN);
 	signal (SIGTSTP, SIG_IGN);
 	signal (SIGTTOU, SIG_IGN);
+	signal (SIGXFSZ, SIG_IGN);
 
 	umask (077);
 }
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -61,17 +61,22 @@ noinst_PROGRAMS = sulogin
 
 suidusbins     =
 suidbins       =
-suidubins      = chage chfn chsh gpasswd newgrp
 if WITH_SU
 suidbins      += su
 endif
+
+privubins      = chage chfn chsh gpasswd newgrp
 if !WITH_TCB
-suidubins += passwd
+privubins     += passwd
 endif
 if ENABLE_SUBIDS
-if !FCAPS
-suidubins += newgidmap newuidmap
+privubins     += newgidmap newuidmap
 endif
+
+if FCAPS
+suidubins      =
+else
+suidubins      = $(privubins)
 endif
 
 if WITH_TCB
@@ -142,6 +147,16 @@ if WITH_TCB
 		chown root:shadow $(DESTDIR)$(ubindir)/$$i; \
 		chmod $(sgidperms) $(DESTDIR)$(ubindir)/$$i; \
 	done
+endif
+if FCAPS
+	setcap cap_dac_read_search+ep $(DESTDIR)$(ubindir)/chage
+	setcap cap_setgid,cap_dac_read_search+ep $(DESTDIR)$(ubindir)/newgrp
+	setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/chfn
+	setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/chsh
+	setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/gpasswd
+if !WITH_TCB
+	setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/passwd
+endif
 endif
 if ENABLE_SUBIDS
 if FCAPS
--- a/src/chfn.c
+++ b/src/chfn.c
@@ -395,13 +395,7 @@ static void update_gecos(const char *user, char *gecos, const struct option_flag
 
 	process_selinux = !flags->chroot;
 
-	/*
-	 * Before going any further, raise the ulimit to prevent colliding
-	 * into a lowered ulimit, and set the real UID to root to protect
-	 * against unexpected signals. Any keyboard signals are set to be
-	 * ignored.
-	 */
-	if (setuid (0) != 0) {
+	if (geteuid () == 0 && setuid (0) != 0) {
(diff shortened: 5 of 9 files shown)


lib/find_new_sub_gids.c
```diff
@@ -10,6 +10,7 @@
 
 #include <stdio.h>
 #include <errno.h>
+#include <stdint.h>
```
alejandro-colomar, line 13: For commit 1:

```
```

configure.ac
```diff
@@ -550,7 +550,7 @@ fi
 AM_CONDITIONAL([USE_PAM], [test "X$with_libpam" = "Xyes"])
 
 AC_ARG_WITH([fcaps],
-	[AS_HELP_STRING([--with-fcaps], [use file capabilities instead of suid binaries for newuidmap/newgidmap @<:@default=no@:>@])],
+	[AS_HELP_STRING([--with-fcaps], [use file capabilities instead of suid binaries where possible @<:@default=no@:>@])],
```
alejandro-colomar, line 553: Let's CC the author of that code.

Cc: @giuseppe

src/Makefile.am
```diff
@@ -143,6 +150,24 @@ if WITH_TCB
...
+	setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/chfn
+	setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/chsh
+
+# rewrite group + gshadow
+	setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/gpasswd
+if !WITH_TCB
+# rewrite shadow
+	setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/passwd
+endif
+endif
```
alejandro-colomar, line 160: I don't understand these comments.  Would you mind explaining what you meant?

anthonyryan1: The code comments there were my observations of the privileged operations each tool needed to function. I wrote the comments before any capabilities, to determine the minimal privileges each tool needed.

I thought that might be helpful to explain the purpose of each capability for anyone reviewing or auditing the specific arguments into setcap.

I have no objection to removing the comments if the internal team finds the capability requirements entirely obvious.

For me as someone working on shadow for the first time, I had used about half of the tools, but had to look at the source for a few I had never used as an end user (eg. chfn, chage and newgrp) to understand what they did.

anthonyryan1: It does look like a few of my code comments aren't as polished as the code and may be a hint misleading, so I'm going to proceed with removing them, and add a comment in commonio.c on the function that ultimately uses all the non-obvious capabilities.

alejandro-colomar: You could put the comments in the commit message.  There I find them quite useful, and you can take more space to explain it carefully.

giuseppe (approved):
LGTM

src/chsh.c
```diff
@@ -376,16 +376,23 @@ static void update_shell (const char *user, char *newshell, const struct option_
...
+	 * handling with pwd_init() so that updating the password file is
+	 * not interrupted in a way that leaves a stale lock behind.
+	 *
+	 * Without file capabilities the real UID is also switched to root.
+	 *
+	 * With file capabilities (USE_FCAPS) we deliberately do not become
+	 * root: the process keeps the invoking user's UID and relies on the
+	 * cap_dac_override capability to update /etc/passwd
 	 */
+#ifndef USE_FCAPS
```
floppym: Instead of checking a compile-time constant, it would probably make more sense to check if the current effective user id is 0 at runtime.

This would allow the file capabilities to be applied to the binary independently of the configure option.

lib/pwd_init.c
```diff
@@ -43,6 +43,16 @@ void pwd_init (void)
...
+	 * temporary files behind.
+	 *
+	 * SIGKILL & SIGSTOP can't be caught, so we rely on commonio_close()'s
+	 * atomic rename to protect the shadow file from corruption.
+	 *
+	 * The stale lockfile is harmless.
+	 */
 	signal (SIGALRM, SIG_IGN);
 	signal (SIGHUP, SIG_IGN);
 	signal (SIGINT, SIG_IGN);
```
alejandro-colomar: I'm averse to adding comments.  Please move that to the commit message.

src/chfn.c
```diff
@@ -396,12 +396,17 @@ static void update_gecos(const char *user, char *gecos, const struct option_flag
...
+	 * lock the real and saved-set-user-ID to root as well, preventing
+	 * any later drop back to the invoking user's UID.  When running
+	 * with file capabilities instead (euid != 0), the necessary
+	 * capabilities are already in effect and no UID change is needed.
 	 */
-	if (setuid (0) != 0) {
+	if (geteuid () == 0 && setuid (0) != 0) {
 		fputs (_("Cannot change ID to root.\n"), stderr);
 		SYSLOG(LOG_ERR, "can't setuid(0)");
 		fail_exit (E_NOPERM, process_selinux);
```
alejandro-colomar: I would also prefer moving this comment to the commit message (and if the existing comment becomes stale, just remove it and document reality in the commit message.

The commit message stays connected to the diff, and thus we can always relate that documentation with what it changed.  The source-code comment is likely to become stale, and confuse.

anthonyryan1:
Amended based on @floppym's feedback.

`--with-fcaps` now decides what our Makefile sets by default, but no longer changes runtime behaviour. If a distro or user installs it with setuid, and later changes it to filecaps, the binary will now recognize that and "just work" in the other mode, and vice versa.

This makes it simpler for downstream distro packagers to implement filecaps based on their own install-time criteria, rather than needing to determine that at configure time.

src/chfn.c
anthonyryan1: Ack. Will remove the comments and update the commit message.

anthonyryan1:
I've moved all the explanations and trade-offs into the commit message and removed them from the code. Makes for a much smaller diff.

Happy to keep iterating on this if there's other changes you would like to see. Or if you're not happy with any of the trade-offs chosen (like assuming a sane rlimit/ulimit rather than adding cap_sys_resource).

floppym (approved):

lib/pwd_init.c
```diff
@@ -51,6 +51,7 @@ void pwd_init (void)
 	signal (SIGTERM, SIG_IGN);
 	signal (SIGTSTP, SIG_IGN);
 	signal (SIGTTOU, SIG_IGN);
+	signal (SIGXFSZ, SIG_IGN);
```
alejandro-colomar, line 55: From the commit message:

```
The --with-fcaps argument now changes what our Makefile does, but
runtime behaviour doesn't depend on the configuration argument.
```

Should this go in a separate commit?  It doesn't seem to be the same change as `Expand the use of --with-fcaps to other binaries`.

hallyn (approved):
It might be slightly neater to have the setcap for newuidmap and newgidmap be inside the same `if FSCAPS`, so something like

```
if FCAPS
       setcap cap_dac_read_search+ep $(DESTDIR)$(ubindir)/chage
       setcap cap_setgid,cap_dac_read_search+ep $(DESTDIR)$(ubindir)/newgrp
       setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/chfn
       setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/chsh
       setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/gpasswd
if !WITH_TCB
       setcap cap_chown,cap_dac_override,cap_fowner+ep $(DESTDIR)$(ubindir)/passwd
endif
if ENABLE_SUBIDS
       setcap cap_setuid+ep $(DESTDIR)$(ubindir)/newuidmap
       setcap cap_setgid+ep $(DESTDIR)$(ubindir)/newgidmap
endif
endif  // FCAPS
```

But no big deal either way.

lib/pwd_init.c
alejandro-colomar: Ping.

PR #1632, opened 2026-06-03
Tests: Add two users with same UID using -o flag

aborah-sudo:
This is the transformation to Python of the test located in `tests/usertools/01/04_useradd_add_user_with_existing_UID_with_-o.test` which checks that `useradd` add two users with same UID using -o flag

### DIFF
--- a/tests/system/tests/test_useradd.py
+++ b/tests/system/tests/test_useradd.py
@@ -305,6 +305,47 @@ def test_useradd__two_users_same_ids(shadow: Shadow):
 
 
 @pytest.mark.topology(KnownTopology.Shadow)
+def test_useradd__two_users_same_uid(shadow: Shadow):
+    """
+    :title: Add two users with same UID using -o flag (allow non-unique UID)
+    :setup:
+        1. Create first user with UID 4242
+        2. Create second user with same UID using -o flag (allows non-unique UIDs)
+    :steps:
+        1. Check passwd entries for both users
+        2. Check group entries for both users
+    :expectedresults:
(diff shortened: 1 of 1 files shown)


tests/system/tests/test_useradd.py
```diff
@@ -304,6 +304,47 @@ def test_useradd__two_users_same_ids(shadow: Shadow):
     assert group_entry2 is None, "Group test2 should not be found"
 
 
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_useradd__add_user_with_existing_uid_and_o_flag(shadow: Shadow):
```
ikerexxe: Following the nomenclature from the previous test (`test_useradd__two_users_same_ids`), I'd name this `test_useradd__two_users_same_uid`

tests/system/tests/test_useradd.py
```diff
@@ -304,6 +304,47 @@ def test_useradd__two_users_same_ids(shadow: Shadow):
...
+    """
+    :title: Add two users with same UID using -o flag
+    :setup:
+        1. Create first user with UID 4242
+        2. Create second user with same UID using -o flag
```
ikerexxe: You could add context explaining what `-o` option does, but this is just a suggestion

tests/system/tests/test_useradd.py
aborah-sudo: Done

tests/system/tests/test_useradd.py
aborah-sudo: Done

ikerexxe (approved):
LGTM

PR #1559, opened 2026-03-01
Re-design strr[c]spn(), to make it simpler and more useful

alejandro-colomar:
Cc: @kees

---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase

```
$ git rd 
1:  a9b6d5b2 = 1:  8378517b lib/string/strspn/: strr[c]spn(): Return the length counting from the end
2:  f7aa9729 = 2:  adb25eda src/login_nopam.c: from_match(): Use strrspn_() instead of its pattern
```
</details>

<details>
<summary>v1c</summary>

-  Rebase

```
$ git rd 
1:  8378517b39bf = 1:  13befe5ae03a lib/string/strspn/: strr[c]spn(): Return the length counting from the end
2:  adb25eda7a2a = 2:  8f1850df8c38 src/login_nopam.c: from_match(): Use strrspn_() instead of its pattern
```
</details>

<details>
<summary>v1d</summary>

-  Rebase

```
$ git rd 
1:  13befe5ae03a = 1:  d318a743e56d lib/string/strspn/: strr[c]spn(): Return the length counting from the end
2:  8f1850df8c38 = 2:  dd2d8c59a6b7 src/login_nopam.c: from_match(): Use strrspn_() instead of its pattern
```
</details>

<details>
<summary>v2</summary>

-  Document the change in the lib/string/README.

```
$ git rd 
1:  d318a743e56d ! 1:  913a991b0185 lib/string/strspn/: strr[c]spn(): Return the length counting from the end
    @@ Commit message
     
         Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
    + ## lib/string/README ##
    +@@ lib/string/README: sprintf/ - Formatted string creation
    + strspn/ - String span searching
    + 
    +   Naming conventions:
    +-  -  'r': reverse (search from the end).
    ++  -  'r': rear (search from the end).
    +   -  'c': complement (negate the second argument).
    +   -  'stp': return a pointer instead of a length.
    + 
    +@@ lib/string/README: strspn/ - String span searching
    +   Like strspn(3), but return a pointer instead of an offset.
    +   It is useful for skipping leading white space.
    + 
    +-    strrspn()
    +-  Use stprspn() instead.
    +-  Return the offset of a suffix segment consisting only of
    ++    strrspn_()
    ++  Return the length of a trailing span consisting only of
    +   characters in 'accept'.
    ++  It is useful for determining if a string ends with characters
    ++  in "accept".
    +     stprspn()
[...]

PR #1633, opened 2026-06-03
passwd: add UPN validation support

### LINKED ISSUE #1626: `passwd` does not accept fully qualified usernames `(user@domain)`
As reported in https://github.com/canonical/authd/issues/1487

Attempting to use `passwd` command for UPN formatted usernames fails

```bash
$ sudo passwd <user>@<domain>
passwd: Provided user name is not a valid name
```

It is failing because this function call was added [`is_valid_user_name()`](https://github.com/shadow-maint/shadow/blob/f5422fbc679592bbdda9b3642f7a3f9c0ffaccd7/src/passwd.c#L996-L997). The `@` character isn't in the [allowed set](https://github.com/shadow-maint/shadow/blob/f5422fbc679592bbdda9b3642f7a3f9c0ffaccd7/lib/chkname.c#L66-L67), causing `passwd user@domain.com` to fail.

ikerexxe:
Add User Principal Name (UPN) validation to allow `passwd` command to accept usernames in `user@domain.com` format. This addresses rejection of valid UPN usernames by implementing proper RFC-compliant domain validation.

Changes include:
* Add `is_valid_upn(), is_valid_domain_name()`, and `is_valid_domain_label()` functions to `lib/chkname.c` with RFC 1035/1123 compliant validation
* Update `passwd.c` to accept both traditional usernames and UPN format
* Add comprehensive unit tests covering UPN validation edge cases
* Domain names limited to 253 chars total and 63 chars per label
* Maintains backward compatibility with existing username validation

Fixes: 326889ca (2024-10-22; "Fix coverity unbound buffer issues")
Closes: https://github.com/shadow-maint/shadow/issues/1626
Reported-by: @nooreldeenmansour
Supersedes: https://github.com/shadow-maint/shadow/pull/1627

### DIFF
--- a/lib/chkname.c
+++ b/lib/chkname.c
@@ -7,8 +7,7 @@
 
 
 /*
- * is_valid_user_name(), is_valid_group_name() - check the new user/group
- * name for validity;
+ * check user/group names for valid syntax
  * return values:
  *   true  - OK
  *   false - bad name
@@ -33,10 +32,17 @@
 
 #include "defines.h"
 #include "chkname.h"
+#include "sizeof.h"
 #include "string/ctype/isascii.h"
 #include "string/strcmp/streq.h"
 #include "string/strcmp/strcaseeq.h"
+#include "string/strspn/strrcspn.h"
+#include "string/strtok/stpsep.h"
 #include "sysconf.h"
+
+
+#define DOMAIN_MAXLEN	255
+#define LABEL_MAXLEN	63
 
 
 int allow_bad_names = false;
@@ -125,4 +131,86 @@ is_valid_group_name(const char *name)
 	}
 
 	return is_valid_name (name);
+}
+
+
+// Validate a single DNS domain label according to RFC 1035 2.3.1.
+static bool
+is_valid_domain_label(const char *label)
+{
+	if (strlen(label) > LABEL_MAXLEN) {
+		errno = EOVERFLOW;
+		return false;
+	}
+	if (!strspn(label, CTYPE_ALPHA_C) || !strrcspn(label, "-")) {
+		errno = EINVAL;
+		return false;
+	}
+	if (!streq(stpspn(label, CTYPE_ALNUM_C "-"), "")) {
+		errno = EINVAL;
+		return false;
+	}
+
+	return true;
+}
+
+
+/*
+ * Validate a domain name according to RFC 1035 by splitting
+ * it into labels and validating each label individually.
+ */
+static bool
+is_valid_domain_name(const char *domain)
+{
+	char        *d;
+	const char  *l;
+
+	if (!strcspn(domain, ".")) {
+		errno = EINVAL;
+		return false;
+	}
+
+	if (strlen(domain) + !!strrcspn(domain, ".") > DOMAIN_MAXLEN) {
+		errno = EOVERFLOW;
+		return false;
+	}
+	d = strdupa(domain);
+
+	while (NULL != (l = strsep(&d, "."))) {
+		if (d == NULL && streq(l, ""))  // trailing root label
+			break;
+		if (!is_valid_domain_label(l))
+			return false;
+	}
+
+	return true;
+}
+
+
+/*
+ * is_valid_upn - is valid User Principal Name
+ *
+ * Check UPN format (user@domain) for validity.
+ *
+ * This function only validates syntax, not whether the UPN exists
+ * in any authentication system.
+ */
+bool
+is_valid_upn(const char *upn)
+{
+	char  *u, *d;
+
+	if (strlen(upn) >= LOGIN_NAME_MAX + STRLEN("@") + DOMAIN_MAXLEN) {
+		errno = EOVERFLOW;
+		return false;
+	}
+	u = strdupa(upn);
+
+	d = stpsep(u, "@");
+	if (d == NULL) {
+		errno = EINVAL;
+		return false;
+	}
+
+	return is_valid_user_name(u) && is_valid_domain_name(d);
 }
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -993,7 +993,7 @@ main(int argc, char **argv)
 	}
 	myname = xstrdup (pw->pw_name);
 	if (optind < argc) {
-		if (!is_valid_user_name (argv[optind])) {
+		if (!is_valid_user_name (argv[optind]) && !is_valid_upn (argv[optind])) {
 			fprintf (stderr, _("%s: Provided user name is not a valid name\n"), Prog);
 			fail_exit (E_NOPERM, process_selinux);
 		}
--- a/tests/unit/test_chkname.c
+++ b/tests/unit/test_chkname.c
@@ -30,6 +30,12 @@ static void test_is_valid_user_name_nok_empty(MAYBE_UNUSED void ** _1);
 static void test_is_valid_user_name_nok_numeric(MAYBE_UNUSED void ** _1);
 static void test_is_valid_user_name_nok_otherchars(MAYBE_UNUSED void ** _1);
 static void test_is_valid_user_name_long(MAYBE_UNUSED void ** _1);
+static void test_is_valid_upn_ok(MAYBE_UNUSED void ** _1);
+static void test_is_valid_upn_nok_not_upn(MAYBE_UNUSED void ** _1);
+static void test_is_valid_upn_nok_structure(MAYBE_UNUSED void ** _1);
+static void test_is_valid_upn_nok_domain(MAYBE_UNUSED void ** _1);
+static void test_is_valid_upn_ok_limits(MAYBE_UNUSED void ** _1);
+static void test_is_valid_upn_nok_limits(MAYBE_UNUSED void ** _1);
 
 
 int
@@ -45,6 +51,12 @@ main(void)
         cmocka_unit_test(test_is_valid_user_name_nok_numeric),
         cmocka_unit_test(test_is_valid_user_name_nok_otherchars),
         cmocka_unit_test(test_is_valid_user_name_long),
+        cmocka_unit_test(test_is_valid_upn_ok),
+        cmocka_unit_test(test_is_valid_upn_nok_not_upn),
+        cmocka_unit_test(test_is_valid_upn_nok_structure),
+        cmocka_unit_test(test_is_valid_upn_nok_domain),
+        cmocka_unit_test(test_is_valid_upn_ok_limits),
+        cmocka_unit_test(test_is_valid_upn_nok_limits),
     };
 
     return cmocka_run_group_tests(tests, NULL, NULL);
@@ -145,4 +157,123 @@ test_is_valid_user_name_long(MAYBE_UNUSED void ** _1)
 	assert_true(is_valid_user_name(name));
 
 	free(name);
+}
+
+
+static void
+test_is_valid_upn_ok(MAYBE_UNUSED void ** _1)
+{
+	assert_true(is_valid_upn("user@example.com"));
+	assert_true(is_valid_upn("john.doe@corp.example.org"));
+	assert_true(is_valid_upn("test@sub.domain.net"));
+	assert_true(is_valid_upn("a@b.c"));
+	assert_true(is_valid_upn("user123@test123.example"));
+	assert_true(is_valid_upn("user_name@example-domain.com"));
+	assert_true(is_valid_upn("test.user@example.domain.org"));
+	assert_true(is_valid_upn("user@domain"));
+	assert_true(is_valid_upn("user@domain.com."));
+	assert_true(is_valid_upn("user@sub."));
+}
+
+
+static void
+test_is_valid_upn_nok_not_upn(MAYBE_UNUSED void ** _1)
+{
+	assert_true(false == is_valid_upn("regularuser"));
+	assert_true(false == is_valid_upn("user.name"));
+	assert_true(false == is_valid_upn("user_name"));
+	assert_true(false == is_valid_upn("123user"));
+	assert_true(false == is_valid_upn("USER"));
+}
+
+
+static void
+test_is_valid_upn_nok_structure(MAYBE_UNUSED void ** _1)
+{
+	// Empty parts
+	assert_true(false == is_valid_upn("@domain.com"));
+	assert_true(false == is_valid_upn("user@"));
+	assert_true(false == is_valid_upn("@"));
+
+	// Multiple @ symbols
+	assert_true(false == is_valid_upn("user@domain@com"));
+	assert_true(false == is_valid_upn("@@domain.com"));
+	assert_true(false == is_valid_upn("user@@domain.com"));
+
+	// Empty string
+	assert_true(false == is_valid_upn(""));
+
+	// Invalid prefix
+	assert_true(false == is_valid_upn("-user@domain.com"));
+	assert_true(false == is_valid_upn("123@domain.com"));
+	assert_true(false == is_valid_upn("user space@domain.com"));
+}
+
+
+static void
+test_is_valid_upn_nok_domain(MAYBE_UNUSED void ** _1)
+{
+	// Invalid domain formats
+	assert_true(false == is_valid_upn("user@.domain.com"));
+	assert_true(false == is_valid_upn("user@domain..com"));
+
+	// Invalid domain characters
+	assert_true(false == is_valid_upn("user@domain_name.com"));
+	assert_true(false == is_valid_upn("user@domain name.com"));
+}
+
+
+static void
+test_is_valid_upn_ok_limits(MAYBE_UNUSED void ** _1)
+{
+	char    *upn;
+	char    *domain;
+
+	// Test domain at maximum allowed length (254 chars + implicit root = 255)
+	domain = malloc_T(254 + 1, char);
+	assert_true(domain != NULL);
+	strcpy(domain,
+		"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa." // 63 a's + dot
+		"bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb." // 63 b's + dot
+		"ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc." // 63 c's + dot
+		"ddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd."      // 59 d's + dot
+		"co");                                                             // 2
+	// Total: 63 + 1 + 63 + 1 + 63 + 1 + 59 + 1 + 2 (+ 1 implicit root) = 255 chars
+
+	upn = malloc_T(5 + 254 + 1, char);  // "user@" + domain + '\0'
+	assert_true(upn != NULL);
+	strcpy(upn, "user@");
+	strcat(upn, domain);
+	assert_true(is_valid_upn(upn));
+
+	free(upn);
+	free(domain);
+}
+
+
+static void
+test_is_valid_upn_nok_limits(MAYBE_UNUSED void ** _1)
+{
+	char    *upn;
+	char    *domain;
+
+	// Test domain exceeding maximum length (255 chars + implicit root = 256 > 255 limit)
+	domain = malloc_T(255 + 1, char);
+	assert_true(domain != NULL);
+	memset(domain, 'a', 252);
+	strcpy(&domain[252], ".co");
+
+	upn = malloc_T(5 + 255 + 1, char);  // "user@" + domain + '\0'
+	assert_true(upn != NULL);
+	strcpy(upn, "user@");
+	strcat(upn, domain);
+	assert_true(false == is_valid_upn(upn));
+
+	free(upn);
+	free(domain);
+
+	// Domain label too long (>63 chars)
+	assert_true(false == is_valid_upn("user@"
+		"verylongdomainlabelthatexceedssixtythreecharacterslimitsetbyRFC1035"
+		".com"));
 }
--- a/lib/chkname.h
+++ b/lib/chkname.h
@@ -28,5 +28,6 @@
 
 extern bool is_valid_user_name (const char *name);
 extern bool is_valid_group_name (const char *name);
+extern bool is_valid_upn (const char *name);
 
 #endif


ikerexxe:
This is what I’ve managed to put in place with the time I had available...

ikerexxe:
@MarcinDigitic do you mind testing?

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+		return false;
+	}
+
+	if (len > 1) {
+		if (!((label[len-1] >= 'a' && label[len-1] <= 'z') ||
+		      (label[len-1] >= 'A' && label[len-1] <= 'Z') ||
+		      (label[len-1] >= '0' && label[len-1] <= '9'))) {
+			return false;
+		}
+	}
```
alejandro-colomar: This could be nicely written as

```c
if (!strspn(label, CTYPE_ALNUM_C))
	return false;
if (!strrspn_(label, CTYPE_ALNUM_C))
	return false;
```

For that, we need to merge #1471 and #1559.  Please review those PRs.  I'd like to block this PR until we have those.  I'm not happy writing all this code, which is error-prone.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+	for (i = 1; i < len - 1; i++) {
+		if (!((label[i] >= 'a' && label[i] <= 'z') ||
+		      (label[i] >= 'A' && label[i] <= 'Z') ||
+		      (label[i] >= '0' && label[i] <= '9') ||
+		      label[i] == '-')) {
+			return false;
+		}
+	}
+
+	return true;
```
alejandro-colomar: This can be written as:

```c
return streq(stpspn(s, CTYPE_ALNUM_C "-"), "");
```

I might want to add a macro for that, but for now that should work.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+ *
+ * RETURN VALUE
+ *	true	Valid domain label.
+ *	false	Invalid domain label.
+ *
+ * ERRORS
+ *	This function does not set errno.
+ */
+static bool
+is_valid_domain_label(const char *label, size_t len)
```
alejandro-colomar: Please don't pass the length.  It is not necessary.  (You could use strlen(3) inside the function, if needed.)

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+	if (local_len == 0) {
+		errno = EINVAL;
+		return false;
+	}
+
+	local_part = malloc(local_len + 1);
+	if (local_part == NULL) {
+		errno = ENOMEM;
+		return false;
+	}
```
alejandro-colomar: malloc(3) already sets errno.

Also, we should call malloc_T() for safety reasons.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+
+	strncpy(local_part, name, local_len);
+	local_part[local_len] = '\0';
```
alejandro-colomar: Never ever use strncpy(3).  See the string README.

alejandro-colomar: This very much looks like

```c
local_part = strndup(name, local_len);
```

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+	if (domain_part == NULL) {
+		errno = ENOMEM;
+		free(local_part);
+		return false;
+	}
+
+	if (streq(domain_part, "")) {
+		errno = EINVAL;
+		goto cleanup;
+	}
```
alejandro-colomar: is_valid_domain_name() already checks for an empty string, AFAICS.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+ *	bool is_valid_upn(const char *name);
+ *
+ * ARGUMENTS
+ *	name	UPN (User Principal Name) string to validate.
+ *
+ * DESCRIPTION
+ *	Check UPN format (user@domain) for validity. A UPN consists of a
+ *	prefix (before @) and suffix (after @). The prefix must follow the
+ *	same rules as traditional usernames. The suffix part must be a
+ *	valid domain name.
```
alejandro-colomar: Which specific section of the RFCs specifies this?

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+	at_pos = strchr(name, '@');
+	if (at_pos == NULL) {
+		return false; /* Not a UPN */
+	}
+
+	/* Check for multiple @ symbols */
+	if (strchr(at_pos + 1, '@') != NULL) {
+		errno = EINVAL;
+		return false;
+	}
```
alejandro-colomar: This is already validated by is_valid_domain_name(), by validating that all characters are alnum or `-`.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+ *
+ * DESCRIPTION
+ *	Validate a single DNS domain label according to RFC 1035/1123.
+ *	Domain labels are components separated by dots in a domain name.
+ *	For example, in "example.com", both "example" and "com" are labels.
+ *
+ * RETURN VALUE
+ *	true	Valid domain label.
+ *	false	Invalid domain label.
+ *
```
alejandro-colomar: This is obvious.  Please remove.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+
+
+/*
+ * SYNOPSIS
+ *	static bool is_valid_domain_label(const char *label, size_t len);
+ *
+ * ARGUMENTS
+ *	label	Domain label string to validate.
+ *	len	Length of the label string.
+ *
```
alejandro-colomar: This is too obvious; please remove.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+ *	Domain labels are components separated by dots in a domain name.
+ *	For example, in "example.com", both "example" and "com" are labels.
+ *
+ * RETURN VALUE
+ *	true	Valid domain label.
+ *	false	Invalid domain label.
+ *
+ * ERRORS
+ *	This function does not set errno.
+ */
```
alejandro-colomar: This is unimportant; please remove.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+	start = domain;
+	while ((end = strchr(start, '.')) != NULL) {
+		has_dot = true;
+		label_len = end - start;
+
+		/* Check for consecutive dots */
+		if (label_len == 0) {
+			errno = EINVAL;
+			return false;
+		}
```
alejandro-colomar: This is already validated by the empty label check.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+	if (label_len == 0 || !is_valid_domain_label(start, label_len)) {
+		errno = EILSEQ;
+		return false;
+	}
+
+	/* Require FQDN (at least one dot) */
+	if (!has_dot) {
+		errno = EINVAL;
+		return false;
+	}
```
alejandro-colomar: Do we really want this?  Is it possible that someone might want to use non-fully qualified domain names?  I don't know.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+	if (domain_len > DOMAIN_MAXLEN) {
+		errno = EOVERFLOW;
+		return false;
+	}
+
+	/* Check for leading or trailing dots */
+	if (domain[0] == '.' || domain[domain_len-1] == '.') {
+		errno = EINVAL;
+		return false;
+	}
```
alejandro-colomar: Isn't a trailing dot valid?  The root label is represented by a trailing dot, and it is always valid, AFAIK, and implied if not present.

I'm not an expert in this, though.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+ *	domain	Domain name string to validate.
+ *
+ * DESCRIPTION
+ *	Validate a domain name according to RFC 1035/1123 by splitting
+ *	it into labels and validating each label individually.
+ *	Requires FQDN format (at least one dot).
+ *
+ * RETURN VALUE
+ *	true	Valid domain name.
+ *	false	Invalid domain name.
```
alejandro-colomar: This is obvious; please remove.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+}
+
+
+/*
+ * SYNOPSIS
+ *	static bool is_valid_domain_name(const char *domain);
+ *
+ * ARGUMENTS
+ *	domain	Domain name string to validate.
+ *
```
alejandro-colomar: This is obvious; please remove.

lib/chkname.c
```diff
@@ -127,3 +131,238 @@ is_valid_group_name(const char *name)
...
+	}
+
+	domain_len = strlen(domain);
+	if (domain_len > DOMAIN_MAXLEN) {
+		errno = EOVERFLOW;
+		return false;
+	}
+
+	/* Check for leading or trailing dots */
+	if (domain[0] == '.' || domain[domain_len-1] == '.') {
```
alejandro-colomar: This could be written as

```c
if (!strcspn(domain, ".") || !strrcspn(domain, "."))
```

This needs #1559 .

lib/chkname.c
alejandro-colomar: Or maybe invert the check, to be more explicit about the fact that we're rejecting leading or trailing hyphens:

```c
if (!strcspn(label, "-") || !strrcspn(label, "-"))
	return false;
```

alejandro-colomar: BTW, please clarify where exactly this is specified in the relevant RFCs.

alejandro-colomar: Please consider reviewing those 2 PRs soon, so that we can use them here.

alejandro-colomar:
I've fixed several issues for you.

v2:

-  Use strspn(3) and strcspn(3) instead of manual byte operations.
-  Use strndup(3) instead of its pattern.
-  Allow trailing dot.
-  Allow non-FQDN.
-  Remove dead code.
-  Simplify documentation.
-  Other minor improvements.

<details>
<summary>range-diff</summary>

```
$ git range-diff --creation-factor=99 HEAD^^^ origin/passwd-upn passwd-upn -w
1:  a3d59004 ! 1:  8943055e lib: add UPN validation support
    @@ Metadata
     Author: Iker Pedrosa <ipedrosa@redhat.com>
     
      ## Commit message ##
    -    lib: add UPN validation support
    +    lib/chkname.*: Add UPN validation support
     
         Add is_valid_upn() function to validate User Principal Name format. UPN
         validation splits on @ and validates the prefix using existing username
         rules and suffix part using RFC 1035/1123 compliant domain name
         validation.
     
    +    Co-authored-by: Iker Pedrosa <ipedrosa@redhat.com>
    +    Co-authored-by: Alejandro Colomar <alx@kernel.org>
    +    Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
      ## lib/chkname.c ##
    +@@
    + 
    + 
    + /*
    +- * is_valid_user_name(), is_valid_group_name() - check the new user/group
    +- * name for validity;
    ++ * check user/group names for valid syntax
    +  * return values:
    +  *   true  - OK
    +  *   false - bad name
    +@@
    +  *   EINVAL       Invalid name
    +  *   EILSEQ       Invalid name character sequence (acceptable with --badname)
    +  *   EOVERFLOW    Name longer than maximum size
    ++ *   ENOMEM     Cannot allocate memory
    +  */
    + 
    + 
     @@
      #include "sysconf.h"
      
    @@ lib/chkname.c: is_valid_group_name(const char *name)
     +
     +
     +/*
    -+ * SYNOPSIS
    -+ *        static bool is_valid_domain_label(const char *label, size_t len);
    -+ *
    -+ * ARGUMENTS
    -+ *        label   Domain label string to validate.
    -+ *        len     Length of the label string.
    -+ *
    -+ * DESCRIPTION
     + * Validate a single DNS domain label according to RFC 1035/1123.
     + * Domain labels are components separated by dots in a domain name.
     + * For example, in "example.com", both "example" and "com" are labels.
    -+ *
    -+ * RETURN VALUE
    -+ *        true    Valid domain label.
    -+ *        false   Invalid domain label.
    -+ *
    -+ * ERRORS
    -+ *        This function does not set errno.
     + */
     +static bool
    -+is_valid_domain_label(const char *label, size_t len)
    ++is_valid_domain_label(const char *label)
     +{
    -+  size_t i;
    ++  size_t  len;
     +
    -+  /* Empty label is invalid */
    -+  if (len == 0 || len > LABEL_MAXLEN) {
    ++  if (strlen(label) > LABEL_MAXLEN) {
    ++          errno = EOVERFLOW;
     +          return false;
     +  }
    -+
    -+  /* Must start and end with alphanumeric */
    -+  if (!((label[0] >= 'a' && label[0] <= 'z') ||
    -+        (label[0] >= 'A' && label[0] <= 'Z') ||
    -+        (label[0] >= '0' && label[0] <= '9'))) {
    ++  if (!strcspn(label, "-")) {
    ++          errno = EINVAL;
     +          return false;
     +  }
    -+
    -+  if (len > 1) {
    ++  len = strlen(label);
     +  if (!((label[len-1] >= 'a' && label[len-1] <= 'z') ||
     +        (label[len-1] >= 'A' && label[len-1] <= 'Z') ||
    -+                (label[len-1] >= '0' && label[len-1] <= '9'))) {
    ++        (label[len-1] >= '0' && label[len-1] <= '9')))
    ++  {
    ++          errno = EINVAL;
     +          return false;
     +  }
    -+  }
     +
    -+  /* Check middle characters (can include hyphen) */
    -+  for (i = 1; i < len - 1; i++) {
    ++  for (size_t i = 0; i < len; i++) {
     +          if (!((label[i] >= 'a' && label[i] <= 'z') ||
     +                (label[i] >= 'A' && label[i] <= 'Z') ||
     +                (label[i] >= '0' && label[i] <= '9') ||
    -+                label[i] == '-')) {
    ++                label[i] == '-'))
    ++          {
    ++                  errno = EINVAL;
     +                  return false;
     +          }
     +  }
    @@ lib/chkname.c: is_valid_group_name(const char *name)
     +
     +
     +/*
    -+ * SYNOPSIS
    -+ *        static bool is_valid_domain_name(const char *domain);
    -+ *
    -+ * ARGUMENTS
    -+ *        domain  Domain name string to validate.
    -+ *
    -+ * DESCRIPTION
     + * Validate a domain name according to RFC 1035/1123 by splitting
     + * it into labels and validating each label individually.
    -+ *        Requires FQDN format (at least one dot).
    -+ *
    -+ * RETURN VALUE
    -+ *        true    Valid domain name.
    -+ *        false   Invalid domain name.
    -+ *
    -+ * ERRORS
    -+ *        EINVAL          Invalid domain structure (empty, leading/trailing dots, consecutive dots).
    -+ *        EILSEQ          Invalid domain label.
    -+ *        EOVERFLOW       Domain too long.
     + */
     +static bool
     +is_valid_domain_name(const char *domain)
     +{
    -+  const char *start, *end;
    -+  size_t domain_len, label_len;
    -+  bool has_dot = false;
    ++  char        *d;
    ++  const char  *l;
     +
    -+  if (streq(domain, "")) {
    -+          errno = EINVAL;
    -+          return false;
    -+  }
    -+
    -+  domain_len = strlen(domain);
    -+  if (domain_len > DOMAIN_MAXLEN) {
    ++  if (strlen(domain) > DOMAIN_MAXLEN) {
     +          errno = EOVERFLOW;
     +          return false;
     +  }
    -+
    -+  /* Check for leading or trailing dots */
    -+  if (domain[0] == '.' || domain[domain_len-1] == '.') {
    ++  if (!strcspn(domain, ".")) {
     +          errno = EINVAL;
     +          return false;
     +  }
     +
    -+  /* Parse and validate each label */
    -+  start = domain;
    -+  while ((end = strchr(start, '.')) != NULL) {
    -+          has_dot = true;
    -+          label_len = end - start;
    ++  d = strdupa(domain);
     +
    -+          /* Check for consecutive dots */
    -+          if (label_len == 0) {
    -+                  errno = EINVAL;
    -+                  return false;
    -+          }
    -+
    -+          if (!is_valid_domain_label(start, label_len)) {
    -+                  errno = EILSEQ;
    -+                  return false;
    -+          }
    -+
    -+          start = end + 1;
    -+  }
    -+
    -+  /* Validate final label */
    -+  label_len = strlen(start);
    -+  if (label_len == 0 || !is_valid_domain_label(start, label_len)) {
    -+          errno = EILSEQ;
    -+          return false;
    -+  }
    -+
    -+  /* Require FQDN (at least one dot) */
    -+  if (!has_dot) {
    -+          errno = EINVAL;
    ++  while (NULL != (l = strsep(&d, "."))) {
    ++          if (d == NULL && streq(l, ""))  // trailing root label
    ++                  break;
    ++          if (!is_valid_domain_label(l))
     +                  return false;
     +  }
     +
    @@ lib/chkname.c: is_valid_group_name(const char *name)
     +
     +
     +/*
    -+ * SYNOPSIS
    -+ *        bool is_valid_upn(const char *name);
    ++ * is_valid_upn - is valid User Principal Name
     + *
    -+ * ARGUMENTS
    -+ *        name    UPN (User Principal Name) string to validate.
    -+ *
    -+ * DESCRIPTION
    -+ *        Check UPN format (user@domain) for validity. A UPN consists of a
    -+ *        prefix (before @) and suffix (after @). The prefix must follow the
    -+ *        same rules as traditional usernames. The suffix part must be a
    ++ * Check UPN format (user@domain) for validity.  The user must follow
    ++ * the same rules as traditional usernames.  The domain part must be a
     + * valid domain name.
     + *
     + * This function only validates syntax, not whether the UPN exists
     + * in any authentication system.
    -+ *
    -+ * RETURN VALUE
    -+ *        true    Valid UPN format.
    -+ *        false   Invalid UPN format or not a UPN.
    -+ *
    -+ * ERRORS
    -+ *        EINVAL          Invalid UPN structure (empty parts, multiple @).
    -+ *        EILSEQ          Invalid character sequence.
    -+ *        EOVERFLOW       Name longer than maximum size.
    -+ *        ENOMEM          Memory allocation failure.
     + */
     +bool
    -+is_valid_upn(const char *name)
    ++is_valid_upn(const char *upn)
     +{
    ++  bool        ret;
    ++  char        *u;
     +  const char  *at_pos;
    -+  char *local_part, *domain_part;
    -+  size_t local_len;
    -+  bool result = false;
     +
    -+  /* Find exactly one @ symbol */
    -+  at_pos = strchr(name, '@');
    -+  if (at_pos == NULL) {
    -+          return false; /* Not a UPN */
    -+  }
    -+
    -+  /* Check for multiple @ symbols */
    -+  if (strchr(at_pos + 1, '@') != NULL) {
    -+          errno = EINVAL;
    ++  at_pos = strchr(upn, '@');
    ++  if (at_pos == NULL)
     +          return false;
    -+  }
     +
    -+  /* Extract prefix */
    -+  local_len = at_pos - name;
    -+  if (local_len == 0) {
    -+          errno = EINVAL;
    ++  u = strndup(upn, at_pos - upn);
    ++  if (u == NULL)
     +          return false;
    -+  }
     +
    -+  local_part = malloc(local_len + 1);
    -+  if (local_part == NULL) {
    -+          errno = ENOMEM;
    -+          return false;
    -+  }
    ++  ret = false;
     +
    -+  strncpy(local_part, name, local_len);
    -+  local_part[local_len] = '\0';
    -+
    -+  /* Extract suffix */
    -+  domain_part = strdup(at_pos + 1);
    -+  if (domain_part == NULL) {
    -+          errno = ENOMEM;
    -+          free(local_part);
    -+          return false;
    -+  }
    -+
    -+  if (streq(domain_part, "")) {
    -+          errno = EINVAL;
    ++  if (!is_valid_user_name(u))
     +          goto cleanup;
    -+  }
    -+
    -+  /* Validate local part using existing username rules */
    -+  if (!is_valid_user_name(local_part)) {
    ++  if (!is_valid_domain_name(at_pos + 1))
     +          goto cleanup;
    -+  }
    -+
    -+  /* Validate domain part using RFC-compliant domain name rules */
    -+  if (!is_valid_domain_name(domain_part)) {
    -+          goto cleanup;
    -+  }
    -+
    -+  result = true;
     +
    ++  ret = true;
     +cleanup:
    -+  free(local_part);
    -+  free(domain_part);
    -+  return result;
    ++  free(u);
    ++  return ret;
     +}
     
      ## lib/chkname.h ##
2:  5aa78162 = 2:  d9f3bc29 tests: add unit tests for UPN validation
3:  4ea29c1b = 3:  97b1a5c0 passwd: add UPN validation support
```
</details>

lib/chkname.c
alejandro-colomar: Partially resolved in v2.  I've changed it to use strspn(3), which is ISO C.  We will be able to use strrspn_() once we merge the other PRs.

lib/chkname.c
alejandro-colomar: Solved in v2.

lib/chkname.c
alejandro-colomar: Solved in v2.

lib/chkname.c
alejandro-colomar: Solved in v2.

alejandro-colomar: Solved in v2.

lib/chkname.c
alejandro-colomar: Solved in v2.

lib/chkname.c
alejandro-colomar: Solved in v2.

lib/chkname.c
alejandro-colomar: Removed in v2.

lib/chkname.c
alejandro-colomar: Removed in v2.

lib/chkname.c
alejandro-colomar: Removed in v2.

lib/chkname.c
alejandro-colomar: Solved in v2.

lib/chkname.c
alejandro-colomar: I've removed this in v2, since it was simpler to implement without it.  But feel free to discuss if you think we want it.

lib/chkname.c
alejandro-colomar: I've allowed a trailing dot in v2.

lib/chkname.c
alejandro-colomar: Removed in v2.

lib/chkname.c
alejandro-colomar: Removed in v2.

lib/chkname.c
alejandro-colomar: Solved in v2.  Because the former is ISO C and the latter is not needed if we allow a trailing dot, this can be done already.

lib/chkname.c
```diff
@@ -127,3 +131,116 @@ is_valid_group_name(const char *name)
...
+		return false;
+	}
+	len = strlen(label);
+	if (!((label[len-1] >= 'a' && label[len-1] <= 'z') ||
+	      (label[len-1] >= 'A' && label[len-1] <= 'Z') ||
+	      (label[len-1] >= '0' && label[len-1] <= '9')))
+	{
+		errno = EINVAL;
+		return false;
+	}
```
alejandro-colomar: We still need strrcspn_() to replace all this.

lib/chkname.c
```diff
@@ -127,3 +131,116 @@ is_valid_group_name(const char *name)
...
+	for (size_t i = 0; i < len; i++) {
+		if (!((label[i] >= 'a' && label[i] <= 'z') ||
+		      (label[i] >= 'A' && label[i] <= 'Z') ||
+		      (label[i] >= '0' && label[i] <= '9') ||
+		      label[i] == '-'))
+		{
+			errno = EINVAL;
+			return false;
+		}
+	}
```
alejandro-colomar: And we still need CTYPE_ALNUM_C to replace all this.

lib/chkname.c
```diff
@@ -127,3 +131,116 @@ is_valid_group_name(const char *name)
...
+
+	return true;
+}
+
+
+/*
+ * Validate a domain name according to RFC 1035/1123 by splitting
+ * it into labels and validating each label individually.
+ */
```
alejandro-colomar, line 161: We should clarify where in the standards this is exactly specified.

lib/chkname.c
```diff
@@ -40,6 +40,10 @@
 #include "sysconf.h"
 
 
+#define DOMAIN_MAXLEN	253
```
alejandro-colomar: This seems to already account for the trailing root label, but if explicitly present, the limit is 254, AFAIK.

lib/chkname.c
```diff
@@ -127,3 +131,116 @@ is_valid_group_name(const char *name)
...
+static bool
+is_valid_domain_name(const char *domain)
+{
+	char        *d;
+	const char  *l;
+
+	if (strlen(domain) > DOMAIN_MAXLEN) {
+		errno = EOVERFLOW;
+		return false;
+	}
```
alejandro-colomar: For accounting for the trailing root label correctly, regardless of whether it's implicit or explicit, this could be:

```c
#define DOMAIN_MAXLEN  254

if (strlen(domain) + !!strrcspn(domain, ".") > DOMAIN_MAXLEN)
```

lib/chkname.c
```diff
@@ -127,3 +131,116 @@ is_valid_group_name(const char *name)
...
+		return false;
+
+	u = strndup(upn, at_pos - upn);
+	if (u == NULL)
+		return false;
+
+	ret = false;
+
+	if (!is_valid_user_name(u))
+		goto cleanup;
```
alejandro-colomar: I'm a bit worried about this.  We're allocating a string before we've validated the input length.  The user is in control of this string, so it could crash the program.  I think we should validate the string length before allocating.

alejandro-colomar:
I've fixed a few more issues for you.

v3:

-  The first check is now a length check.  I test against compile-time constants, which provides guarantees that the maximum string is quite short, and thus fits in the stack.
-  The above allows me to drop the malloc(3) call (called through strndup(3)).  Now that the string is really short, we can strdupa(3).  This avoids ENOMEM entirely, and a vulnerability with it: the user can't force a crash with a huge UPN.
-  Use stpsep() to simplify.
-  Set errno in all false cases.
-  Further reduce documentation.

<details>
<summary>range-diff</summary>

```
$ git range-diff HEAD^^^ origin/passwd-upn passwd-upn 
1:  8943055e ! 1:  6020aa6c lib/chkname.*: Add UPN validation support
    @@ lib/chkname.c
       *   true  - OK
       *   false - bad name
     @@
    -  *   EINVAL       Invalid name
    -  *   EILSEQ       Invalid name character sequence (acceptable with --badname)
    -  *   EOVERFLOW    Name longer than maximum size
    -+ *   ENOMEM     Cannot allocate memory
    -  */
      
    - 
    -@@
    + #include "defines.h"
    + #include "chkname.h"
    ++#include "sizeof.h"
    + #include "string/ctype/strchrisascii/strchriscntrl.h"
    + #include "string/ctype/strisascii/strisdigit.h"
    + #include "string/strcmp/streq.h"
    + #include "string/strcmp/strcaseeq.h"
    ++#include "string/strtok/stpsep.h"
      #include "sysconf.h"
      
      
    @@ lib/chkname.c: is_valid_group_name(const char *name)
     +/*
     + * is_valid_upn - is valid User Principal Name
     + *
    -+ * Check UPN format (user@domain) for validity.  The user must follow
    -+ * the same rules as traditional usernames.  The domain part must be a
    -+ * valid domain name.
    ++ * Check UPN format (user@domain) for validity.
     + *
     + * This function only validates syntax, not whether the UPN exists
     + * in any authentication system.
    @@ lib/chkname.c: is_valid_group_name(const char *name)
     +bool
     +is_valid_upn(const char *upn)
     +{
    -+  bool        ret;
    -+  char        *u;
    -+  const char  *at_pos;
    ++  char  *u, *d;
     +
    -+  at_pos = strchr(upn, '@');
    -+  if (at_pos == NULL)
    ++  if (strlen(upn) >= LOGIN_NAME_MAX + STRLEN("@") + DOMAIN_MAXLEN) {
    ++          errno = EOVERFLOW;
     +          return false;
    ++  }
     +
    -+  u = strndup(upn, at_pos - upn);
    -+  if (u == NULL)
    ++  u = strdupa(upn);
    ++  d = stpsep(u, "@");
    ++  if (d == NULL) {
    ++          errno = EINVAL;
     +          return false;
    ++  }
     +
    -+  ret = false;
    -+
    -+  if (!is_valid_user_name(u))
    -+          goto cleanup;
    -+  if (!is_valid_domain_name(at_pos + 1))
    -+          goto cleanup;
    -+
    -+  ret = true;
    -+cleanup:
    -+  free(u);
    -+  return ret;
    ++  return is_valid_user_name(u) && is_valid_domain_name(d);
     +}
     
      ## lib/chkname.h ##
2:  d9f3bc29 = 2:  01d04883 tests: add unit tests for UPN validation
3:  97b1a5c0 = 3:  0428deec passwd: add UPN validation support
```
</details>

------

v3b:

-  Move the length checks closer to the corresponding strdupa(3) calls, to make it easier to verify that strdupa(3) calls are safe.

<details>
<summary>range-diff</summary>

```
$ git range-diff HEAD^^^ origin/passwd-upn passwd-upn 
1:  6020aa6c ! 1:  cf710e51 lib/chkname.*: Add UPN validation support
    @@ lib/chkname.c: is_valid_group_name(const char *name)
     +  char        *d;
     +  const char  *l;
     +
    -+  if (strlen(domain) > DOMAIN_MAXLEN) {
    -+          errno = EOVERFLOW;
    -+          return false;
    -+  }
     +  if (!strcspn(domain, ".")) {
     +          errno = EINVAL;
     +          return false;
     +  }
     +
    ++  if (strlen(domain) > DOMAIN_MAXLEN) {
    ++          errno = EOVERFLOW;
    ++          return false;
    ++  }
     +  d = strdupa(domain);
     +
     +  while (NULL != (l = strsep(&d, "."))) {
    @@ lib/chkname.c: is_valid_group_name(const char *name)
     +          errno = EOVERFLOW;
     +          return false;
     +  }
    -+
     +  u = strdupa(upn);
    ++
     +  d = stpsep(u, "@");
     +  if (d == NULL) {
     +          errno = EINVAL;
2:  01d04883 = 2:  9635aa1b tests: add unit tests for UPN validation
3:  0428deec = 3:  d2f556b7 passwd: add UPN validation support
```
</details>

lib/chkname.c
alejandro-colomar: Solved in v3.

src/passwd.c
```diff
@@ -993,7 +993,7 @@ main(int argc, char **argv)
 	}
 	myname = xstrdup (pw->pw_name);
 	if (optind < argc) {
-		if (!is_valid_user_name (argv[optind])) {
+		if (!is_valid_user_name (argv[optind]) && !is_valid_upn (argv[optind])) {
 			fprintf (stderr, _("%s: Provided user name is not a valid name\n"), Prog);
 			fail_exit (E_NOPERM, process_selinux);
 		}
```
alejandro-colomar, line 999: Hmmmm, now I've realized another issue that was introduced when we added syntax checking of the user names.  There's no way to relax the syntax checking, since passwd(1) has no --badname option.  Thus, we've effectively removed --badname support.

We should probably also fix that.

alejandro-colomar: Maybe we should set badname to true in these programs that didn't check syntax before.

tests/unit/test_chkname.c
```diff
@@ -147,3 +159,123 @@ test_is_valid_user_name_long(MAYBE_UNUSED void ** _1)
...
+{
+	char    *upn;
+	char    *domain;
+
+	// Test domain exceeding maximum length (254 chars)
+	domain = malloc_T(254 + 1, char);
+	assert_true(domain != NULL);
+	memset(domain, 'a', 251);
+	strcpy(&domain[251], ".co");
+	domain[254] = '\0';
```
alejandro-colomar: This is dead code.  strcpy(3) already writes a NUL byte.

lib/chkname.c
ikerexxe: > BTW, please clarify where exactly this is specified in the relevant RFCs.

Done

> Please consider reviewing those 2 PRs soon, so that we can use them here.

I'll try to, but it'll take me some time since I'm traveling these days

tests/unit/test_chkname.c
ikerexxe: Updated

ikerexxe:
I realize that one of the unit tests is failing, I'll fix it when related PRs are merged and I can base my work on stable APIs

alejandro-colomar:
> I realize that one of the unit tests is failing, I'll fix it when related PRs are merged and I can base my work on stable APIs

I think the reason for the failing tests is that I changed the code to accept `user@domain.com.` with a trailing root label, and also `user@domain` with a non-fully-qualified domain.

alejandro-colomar:
I've fixed a few more things for you.

v4:

-  Rebase

<details>
<summary>range-diff</summary>

```
$ git range-diff origin/passwd-upn^^^..origin/passwd-upn upstream/master..passwd-upn 
1:  adc23074fe38 ! 1:  052644053b3d lib/chkname.*: Add UPN validation support
    @@ lib/chkname.c
      #include "defines.h"
      #include "chkname.h"
     +#include "sizeof.h"
    - #include "string/ctype/strchrisascii/strchriscntrl.h"
    - #include "string/ctype/strisascii/strisdigit.h"
    + #include "string/ctype/isascii.h"
      #include "string/strcmp/strcaseeq.h"
     +#include "string/strtok/stpsep.h"
2:  f1f9f623dadd = 2:  522faa7a5861 tests: add unit tests for UPN validation
3:  d71db363028d = 3:  c3625ee0efcb passwd: add UPN validation support
```
</details>

v5:

-  Test the last character against '-' to simplify.
-  Use `streq(stpspn(s, CTYPE_ALNUM_C "-"), "")` to simplify.

<details>
<summary>range-diff</summary>

```
$ git range-diff origin/passwd-upn^^^..origin/passwd-upn upstream/master..passwd-upn 
1:  052644053b3d ! 1:  5e50c4a34637 lib/chkname.*: Add UPN validation support
    @@ lib/chkname.c: is_valid_group_name(const char *name)
     +static bool
     +is_valid_domain_label(const char *label)
     +{
    -+  size_t  len;
    -+
     +  if (strlen(label) > LABEL_MAXLEN) {
     +          errno = EOVERFLOW;
     +          return false;
     +  }
    -+  if (!strcspn(label, "-")) {
    ++  if (!strcspn(label, "-") || label[strlen(label)-1] == '-') {
     +          errno = EINVAL;
     +          return false;
     +  }
    -+  len = strlen(label);
    -+  if (!((label[len-1] >= 'a' && label[len-1] <= 'z') ||
    -+        (label[len-1] >= 'A' && label[len-1] <= 'Z') ||
    -+        (label[len-1] >= '0' && label[len-1] <= '9')))
    -+  {
    ++  if (!streq(stpspn(s, CTYPE_ALNUM_C "-"), "")) {
     +          errno = EINVAL;
     +          return false;
     +  }
     +
    -+  for (size_t i = 0; i < len; i++) {
    -+          if (!((label[i] >= 'a' && label[i] <= 'z') ||
    -+                (label[i] >= 'A' && label[i] <= 'Z') ||
    -+                (label[i] >= '0' && label[i] <= '9') ||
    -+                label[i] == '-'))
    -+          {
    -+                  errno = EINVAL;
    -+                  return false;
    -+          }
    -+  }
    -+
     +  return true;
     +}
     +
2:  522faa7a5861 = 2:  f0b23874316b tests: add unit tests for UPN validation
3:  c3625ee0efcb = 3:  48d9c38e8f0e passwd: add UPN validation support
```
<details>

lib/chkname.c
```diff
@@ -126,3 +131,93 @@ is_valid_group_name(const char *name)
...
+ * and hyphen. Labels must be 63 characters or less.
+ */
+static bool
+is_valid_domain_label(const char *label)
+{
+	if (strlen(label) > LABEL_MAXLEN) {
+		errno = EOVERFLOW;
+		return false;
+	}
+	if (!strcspn(label, "-") || label[strlen(label)-1] == '-') {
```
alejandro-colomar: We still need strrcspn_() to replace the second test.

lib/chkname.c
```diff
@@ -126,3 +131,93 @@ is_valid_group_name(const char *name)
...
+
+
+/*
+ * Validate a single DNS domain label according to RFC 1035/1123.
+ * Domain labels are components separated by dots in a domain name.
+ * For example, in "example.com", both "example" and "com" are labels.
+ *
+ * According to RFC 1035 2.3.1 labels must start with a letter, end with
+ * a letter or digit, and have as interior characters only letters, digits,
+ * and hyphen. Labels must be 63 characters or less.
```
alejandro-colomar: According to my reading of that RFC in the section you referenced, the first character in a label must be a letter; a digit is not allowed.

Would you mind revising?

alejandro-colomar: Oh, actually, this comment is correct, but the source code doesn't match it.  I';ll remove most of it, and move it to the commit message.

alejandro-colomar:
I've fixed a bug for you.

v6:

-  The first character in a label must be a letter.  The comment correctly said this, but the code didn't do it.  Fix the code, and remove the comment.  Add a link to the specification in the commit message.
-  Use 2 spaces as inter-sentence spacing in commit message.

<details>
<summary>range-diff</summary>

```
$ git range-diff origin/passwd-upn^^^..origin/passwd-upn upstream/master..passwd-upn 
1:  5e50c4a34637 ! 1:  b1822d4782a6 lib/chkname.*: Add UPN validation support
    @@ Metadata
      ## Commit message ##
         lib/chkname.*: Add UPN validation support
     
    -    Add is_valid_upn() function to validate User Principal Name format. UPN
    +    Add is_valid_upn() function to validate User Principal Name format.  UPN
         validation splits on @ and validates the prefix using existing username
         rules and suffix part using RFC 1035/1123 compliant domain name
         validation.
     
    +    Link: <https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.1>
         Co-authored-by: Iker Pedrosa <ipedrosa@redhat.com>
         Co-authored-by: Alejandro Colomar <alx@kernel.org>
    @@ lib/chkname.c: is_valid_group_name(const char *name)
      }
     +
     +
    -+/*
    -+ * Validate a single DNS domain label according to RFC 1035/1123.
    -+ * Domain labels are components separated by dots in a domain name.
    -+ * For example, in "example.com", both "example" and "com" are labels.
    -+ *
    -+ * According to RFC 1035 2.3.1 labels must start with a letter, end with
    -+ * a letter or digit, and have as interior characters only letters, digits,
    -+ * and hyphen. Labels must be 63 characters or less.
    -+ */
    ++// Validate a single DNS domain label according to RFC 1035 2.3.1.
     +static bool
     +is_valid_domain_label(const char *label)
     +{
    @@ lib/chkname.c: is_valid_group_name(const char *name)
     +          errno = EOVERFLOW;
     +          return false;
     +  }
    -+  if (!strcspn(label, "-") || label[strlen(label)-1] == '-') {
    ++  if (!strspn(label, CTYPE_ALPHA_C) || label[strlen(label)-1] == '-') {
     +          errno = EINVAL;
     +          return false;
     +  }
2:  f0b23874316b = 2:  339dde667336 tests: add unit tests for UPN validation
3:  48d9c38e8f0e = 3:  23c7a5328623 passwd: add UPN validation support
```
</details>

ikerexxe:
I'm actively looking into this PR now that the dependencies are merged. Please do not push

lib/chkname.c
ikerexxe: This should be fixed by now

lib/chkname.c
ikerexxe: Done

src/passwd.c
ikerexxe: Yeah, probably, but that should be fixed in a different PR

lib/chkname.c
ikerexxe: Done

ikerexxe:
This is ready for review again

lib/chkname.c
```diff
@@ -126,3 +132,85 @@ is_valid_group_name(const char *name)
...
+		errno = EINVAL;
+		return false;
+	}
+	if (!streq(stpspn(label, CTYPE_ALNUM_C "-"), "")) {
+		errno = EINVAL;
+		return false;
+	}
+
+	return true;
+}
```
alejandro-colomar, line 155: LGTM.

lib/chkname.c
```diff
@@ -33,12 +32,19 @@
...
 #include "string/ctype/isascii.h"
 #include "string/strcmp/streq.h"
 #include "string/strcmp/strcaseeq.h"
+#include "string/strspn/strrcspn.h"
+#include "string/strtok/stpsep.h"
 #include "sysconf.h"
 
 
+#define DOMAIN_MAXLEN	255
+#define LABEL_MAXLEN	63
```
alejandro-colomar, line 45: LGTM.

lib/chkname.c
```diff
@@ -126,3 +132,85 @@ is_valid_group_name(const char *name)
...
+
+	while (NULL != (l = strsep(&d, "."))) {
+		if (d == NULL && streq(l, ""))  // trailing root label
+			break;
+		if (!is_valid_domain_label(l))
+			return false;
+	}
+
+	return true;
+}
```
alejandro-colomar, line 187: LGTM.

lib/chkname.c
```diff
@@ -126,3 +132,85 @@ is_valid_group_name(const char *name)
...
+
+
+/*
+ * Validate a domain name according to RFC 1035/1123 by splitting
+ * it into labels and validating each label individually.
+ */
+static bool
+is_valid_domain_name(const char *domain)
+{
```
alejandro-colomar, line 164: Is RFC 1123 needed here?  I think RFC 1035 (section 2.3) is the most relevant part.  Do we need anything from the other one?

lib/chkname.c
```diff
@@ -126,3 +132,85 @@ is_valid_group_name(const char *name)
...
+	u = strdupa(upn);
+
+	d = stpsep(u, "@");
+	if (d == NULL) {
+		errno = EINVAL;
+		return false;
+	}
+
+	return is_valid_user_name(u) && is_valid_domain_name(d);
+}
```
alejandro-colomar, line 216: LGTM.

alejandro-colomar (commented):
Here's a review of the first commit.  Except for one comment, it looks good to me.

tests/unit/test_chkname.c
```diff
@@ -146,3 +158,122 @@ test_is_valid_user_name_long(MAYBE_UNUSED void ** _1)
...
+	assert_true(false == is_valid_upn("@@domain.com"));
+	assert_true(false == is_valid_upn("user@@domain.com"));
+
+	// Empty string
+	assert_true(false == is_valid_upn(""));
+
+	// Invalid suffix
+	assert_true(false == is_valid_upn("-user@domain.com"));
+	assert_true(false == is_valid_upn("123@domain.com"));
+	assert_true(false == is_valid_upn("user space@domain.com"));
```
alejandro-colomar: You meant invalid "prefix", right?

tests/unit/test_chkname.c
```diff
@@ -146,3 +158,122 @@ test_is_valid_user_name_long(MAYBE_UNUSED void ** _1)
...
+}
+
+
+static void
+test_is_valid_upn_nok_domain(MAYBE_UNUSED void ** _1)
+{
+	// Invalid domain formats
+	assert_true(false == is_valid_upn("user@.domain.com"));
+	assert_true(false == is_valid_upn("user@domain..com"));
```
alejandro-colomar, line 218: You could say "Empty non-trailing label" to be more precise, but it's fine as it.

src/passwd.c
alejandro-colomar: Sounds reasonable.

alejandro-colomar (approved):
Thanks!  LGTM.

```
```

I only have some comments about the source-code comments, but the source code is good.

lib/chkname.c
ikerexxe: I read it again and not really

tests/unit/test_chkname.c
ikerexxe: Yes, I probably copied this from somewhere and forgot to edit

tests/unit/test_chkname.c
ikerexxe: Let's leave it as it is

ikerexxe:
@alejandro-colomar feel free to merge if you agree with the changes

alejandro-colomar:
Merged.

BTW, I forgot...  I had this in mind a few weeks ago, but forgot about it...  Is passwd the only program that needs this?  Do we need to do the same thing for other programs?

ikerexxe:
Yes, I think so. `passwd` is the only tool that can change data to external users

alejandro-colomar:
Thanks!  Let's keep it as is, and if someone finds and reports issues with any other tool, we'll address them.  :)

alejandro-colomar:
> Yes, I think so. `passwd` is the only tool that can change data to external users

@ikerexxe  How about gpasswd(1)?  Is there a Group Principal Name?  Or are groups always local?

ikerexxe:
Hmmmm it depends. `gpasswd` does not use PAM, thus it can't change the password for a centrally managed group. 

But, this isn't the only use case for `gpasswd` as we can also manage other things like user/group membership from this tool. Do we want to support that a centrally managed user is a member of a local group? I've always have mixed feelings about it

alejandro-colomar:
> Hmmmm it depends. `gpasswd` does not use PAM, thus it can't change the password for a centrally managed group.
> 
> But, this isn't the only use case for `gpasswd` as we can also manage other things like user/group membership from this tool. Do we want to support that a centrally managed user is a member of a local group? I've always have mixed feelings about it

No idea.  :)

I guess we can keep it as is, and if someone shows a good case for it, we can consider allowing it.

PR #1634, opened 2026-06-05
Fix memory leak: free path/rel in shadowtcb_remove()

savoptik:
In function shadowtcb_remove() (lib/tcbfuncs.c), memory allocated for 'path' and 'rel' pointers was not released, causing a memory leak. Added explicit free(path) and free(rel) calls before returning.

### DIFF
--- a/lib/tcbfuncs.c
+++ b/lib/tcbfuncs.c
@@ -409,14 +409,25 @@ shadowtcb_status shadowtcb_remove (const char *name)
 	shadowtcb_status ret = SHADOWTCB_SUCCESS;
 	char *path = shadowtcb_path_existing (name);
 	char *rel = shadowtcb_path_rel_existing (name);
-	if ((NULL == path) || (NULL == rel) || (rmdir (path) != 0)) {
+	if (NULL == path) {
+		free(rel);
 		return SHADOWTCB_FAILURE;
 	}
+	if (NULL == rel) {
+		free(path);
+		return SHADOWTCB_FAILURE;
+	}
+	if (rmdir (path) != 0) {
+		free(path);
+		free(rel);
+		return SHADOWTCB_FAILURE;
+	}
+	free(path);
 	if (rmdir_leading (rel) == SHADOWTCB_FAILURE) {
+		free(rel);
 		return SHADOWTCB_FAILURE;
 	}
-	free (path);
-	free (rel);
+	free(rel);
 	path = aprintf(TCB_DIR "/%s", name);
 	if (path == NULL) {
 		OUT_OF_MEMORY;


alejandro-colomar:
Please add a Fixes tag in the commit message if possible, to identify when this memory leak was introduced and how/why.

See:
<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/git#n46>

lib/tcbfuncs.c
```diff
@@ -410,9 +410,13 @@ shadowtcb_status shadowtcb_remove (const char *name)
...
+		free (path);
+		free (rel);
 		return SHADOWTCB_FAILURE;
 	}
 	if (rmdir_leading (rel) == SHADOWTCB_FAILURE) {
+		free (path);
+		free (rel);
 		return SHADOWTCB_FAILURE;
 	}
 	free (path);
```
alejandro-colomar: If you move the existing `free(path);` before this conditional, you'll save one line of code.

lib/tcbfuncs.c
```diff
@@ -410,9 +410,13 @@ shadowtcb_status shadowtcb_remove (const char *name)
 	char *path = shadowtcb_path_existing (name);
 	char *rel = shadowtcb_path_rel_existing (name);
 	if ((NULL == path) || (NULL == rel) || (rmdir (path) != 0)) {
+		free (path);
+		free (rel);
 		return SHADOWTCB_FAILURE;
 	}
```
alejandro-colomar, line 419: Interestingly, if `NULL == rel` we don't `rmdir(path)`.  Is that desirable?

savoptik: Done

lib/tcbfuncs.c
savoptik: Done

savoptik:
> Please add a Fixes tag in the commit message if possible, to identify when this memory leak was introduced and how/why.
> 
> See: https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/patches/trailer#n16 https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/git#n46

Should I perform a squash and force push?

alejandro-colomar:
> > Please add a Fixes tag in the commit message if possible, to identify when this memory leak was introduced and how/why.
> > See: https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/patches/trailer#n16 https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/git#n46
> 
> Should I perform a squash and force push?

Yup; please.

lib/tcbfuncs.c
```diff
@@ -408,14 +408,22 @@ shadowtcb_status shadowtcb_remove (const char *name)
 {
 	shadowtcb_status ret = SHADOWTCB_SUCCESS;
 	char *path = shadowtcb_path_existing (name);
+	if (NULL == path) {
+		return SHADOWTCB_FAILURE;
+	}
```
alejandro-colomar, line 415: Please don't use braces here.

savoptik: Done

savoptik:
> > > Please add a Fixes tag in the commit message if possible, to identify when this memory leak was introduced and how/why.
> > > See: https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/patches/trailer#n16 https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/git#n46
> > 
> > 
> > Should I perform a squash and force push?
> 
> Yup; please.

Done

lib/tcbfuncs.c
```diff
@@ -408,14 +408,20 @@ shadowtcb_status shadowtcb_remove (const char *name)
...
+	if (NULL == path)
+		return SHADOWTCB_FAILURE;
+	if (rmdir (path) != 0) {
+		free (path);
 		return SHADOWTCB_FAILURE;
 	}
+	free (path);
+	char *rel = shadowtcb_path_rel_existing (name);
+	if (NULL == rel)
+		return SHADOWTCB_FAILURE;
```
alejandro-colomar, line 423: Are you sure it's fine to move the rmdir(path) call from after to before the shadowtcb_path_rel_existing(name) call?

I haven't checked what shadowtcb_path_rel_existing() does.  It would be nice to see some analysis in the commit message clarifying why this is possible.

lib/tcbfuncs.c
alejandro-colomar: It would be interesting to also mention this in the commit message (or even better, split that to a separate commit that documents that fix separately).

lib/tcbfuncs.c
savoptik: I’m sorry, I’ve restored the original order.

lib/tcbfuncs.c
```diff
@@ -408,14 +408,23 @@ shadowtcb_status shadowtcb_remove (const char *name)
...
 	shadowtcb_status ret = SHADOWTCB_SUCCESS;
 	char *path = shadowtcb_path_existing (name);
+	if (NULL == path)
+		return SHADOWTCB_FAILURE;
 	char *rel = shadowtcb_path_rel_existing (name);
-	if ((NULL == path) || (NULL == rel) || (rmdir (path) != 0)) {
+	if (NULL == rel) {
+		free (path);
 		return SHADOWTCB_FAILURE;
 	}
```
alejandro-colomar, line 415: This also changes behavior:

Previously, even if shadowtcb_path_existing(name) failed, shadowtcb_path_rel_existing(name) was executed.

After this patch, if the former fails, we don't execute the latter.

I'm not sure which of the two behaviours should be preferred, but any commit that changes the behavior should document the change.  And that should probably be done in a separate commit from the one that fixes the memory leak.

savoptik: Fixed.

lib/tcbfuncs.c
```diff
@@ -409,13 +409,25 @@ shadowtcb_status shadowtcb_remove (const char *name)
 	shadowtcb_status ret = SHADOWTCB_SUCCESS;
 	char *path = shadowtcb_path_existing (name);
 	char *rel = shadowtcb_path_rel_existing (name);
-	if ((NULL == path) || (NULL == rel) || (rmdir (path) != 0)) {
+	if (NULL == path) {
+		if (NULL != rel)
+			free (rel);
```
alejandro-colomar: free(NULL) is valid C.  See free(3).  This conditional is dead code.

savoptik: Done.

lib/tcbfuncs.c
```diff
@@ -413,13 +413,13 @@ shadowtcb_status shadowtcb_remove (const char *name)
 		free (rel);
 		return SHADOWTCB_FAILURE;
 	}
-	if (NULL == rel) {
+	if (rmdir (path) != 0) {
 		free (path);
+		free (rel);
 		return SHADOWTCB_FAILURE;
```
alejandro-colomar, line 418: For the second commit, I'd like to see in the commit message some analysis of what shadowtcb_path_rel_existing() does, because it's complex enough that I don't know if we can remove the directory or not if it failed.

lib/tcbfuncs.c
```diff
@@ -409,13 +409,24 @@ shadowtcb_status shadowtcb_remove (const char *name)
 	shadowtcb_status ret = SHADOWTCB_SUCCESS;
 	char *path = shadowtcb_path_existing (name);
 	char *rel = shadowtcb_path_rel_existing (name);
-	if ((NULL == path) || (NULL == rel) || (rmdir (path) != 0)) {
```
alejandro-colomar, line 412: For the first commit:

```
```

Thanks!

lib/tcbfuncs.c
savoptik: I suggest removing the second commit and keeping the current behaviour.

alejandro-colomar: That would be fine.

lib/tcbfuncs.c
```diff
@@ -409,13 +409,24 @@ shadowtcb_status shadowtcb_remove (const char *name)
...
+		free (rel);
 		return SHADOWTCB_FAILURE;
 	}
-	if (rmdir_leading (rel) == SHADOWTCB_FAILURE) {
+	if (rmdir (path) != 0) {
+		free (path);
+		free (rel);
+		return SHADOWTCB_FAILURE;
+	}
+	if (NULL == rel) {
```
alejandro-colomar: Please don't add spaces between the function name and the opening `(` in new code.

savoptik: Fixed.

alejandro-colomar (approved):
Thanks!

PR #1423, opened 2025-12-14
lib/list.c: realloc(3) and memmove(3) to prevent memory leaks

alejandro-colomar:
Cc: @uecker 

---

Revisions:

<details>
<summary>v2</summary>

-  Improve commit message.
-  Also avoid memory leaks in del_list().

```
$ git rd 
1:  378deaa4b ! 1:  7ba6b1f82 lib/list.c: realloc(3) to prevent memory leaks
    @@ Metadata
     Author: Alejandro Colomar <alx@kernel.org>
     
      ## Commit message ##
    -    lib/list.c: realloc(3) to prevent memory leaks
    +    lib/list.c: add_list(): realloc(3) to prevent memory leaks
     
         Each call to add_list() that resulted in adding a group to the list was
         leaking the old list.  There's no need to allocate a fresh list.
-:  --------- > 2:  55b3d70d9 lib/list.c: del_list(): Reimplement with memmove(3) and realloc(3) to avoid memory leaks
```
</details>

<details>
<summary>v2b</summary>

-  Clarify assumption.

```
$ git rd 
1:  7ba6b1f82 = 1:  7ba6b1f82 lib/list.c: add_list(): realloc(3) to prevent memory leaks
2:  55b3d70d9 ! 2:  c4c02cc30 lib/list.c: del_list(): Reimplement with memmove(3) and realloc(3) to avoid memory leaks
    @@ Commit message
         pointer.  We already return the original pointer in some cases, so this
         must be okay.
     
    +    This change assumes that list members are unique.  As far as I can see,
    +    they are.  add_list() makes sure this is true.  If a list was manually
    +    constructed without add_list(), and uniqueness wouldn't be preserved,
    +    that would be a problem.
    +
         Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
      ## lib/list.c ##
```
</details>

<details>
<summary>v3</summary>

-  del_list(): Return early, to avoid a no-op realloc(3) call.  It also reduces indentation.

```
$ git rd 
1:  7ba6b1f82 = 1:  7ba6b1f82 lib/list.c: add_list(): realloc(3) to prevent memory leaks
2:  c4c02cc30 ! 2:  ba4c3dcb8 lib/list.c: del_list(): Reimplement with memmove(3) and realloc(3) to avoid memory leaks
    @@ lib/list.c: add_list(/*@returned@*/ /*@only@*/char **list, const char *member)
     -          if (!streq(list[i], member)) {
     -                  j++;
     -          }
[...]

PR #1638, opened 2026-06-08
fix: memory leak of ngrp in updates_groups()

savoptik:
In src/userdel.c:update_groups(), the 'ngrp' pointer was not freed
    after use, leading to a memory leak. Added free(ngrp) call.

    In src/userdel.c:update_groups(), the 'nsgrp' pointer was not freed
    between loop iterations, leading to cumulative memory leaks. Added
    free(nsgrp) call before proceeding to next iteration.

### DIFF
--- a/lib/list.c
+++ b/lib/list.c
@@ -67,6 +67,8 @@ add_list(/*@returned@*/ /*@only@*/char **list, const char *member)
 	tmp[i] = xstrdup (member);
 	tmp[i+1] = NULL;
 
+	free (list);
+
 	return tmp;
 }
 
--- a/lib/list.c
+++ b/lib/list.c
@@ -121,6 +121,8 @@ del_list(/*@returned@*/ /*@only@*/char **list, const char *member)
 		if (!streq(list[i], member)) {
 			tmp[j] = list[i];
 			j++;
+		} else {
+			free (list[i]);
 		}
 	}
 
--- a/lib/list.c
+++ b/lib/list.c
@@ -128,6 +128,8 @@ del_list(/*@returned@*/ /*@only@*/char **list, const char *member)
 
 	tmp[j] = NULL;
 
+	free (list);
+
 	return tmp;
 }
 
--- a/src/groupadd.c
+++ b/src/groupadd.c
@@ -67,8 +67,6 @@ static const char Prog[] = "groupadd";
 static /*@null@*/char *group_name;
 static gid_t group_id;
 static /*@null@*/char *group_passwd;
-static /*@null@*/char *empty_list = NULL;
-
 static const char *prefix = "";
 static char *user_list;
 
@@ -151,7 +149,7 @@ static void new_grent (struct group *grent)
 		grent->gr_passwd = SHADOW_PASSWD_STRING;	/* XXX warning: const */
 	}
 	grent->gr_gid = group_id;
-	grent->gr_mem = &empty_list;
+	grent->gr_mem = comma_to_list("");
 }
 
 #ifdef	SHADOWGRP
@@ -170,8 +168,8 @@ static void new_sgent (struct sgrp *sgent)
 	} else {
 		sgent->sg_passwd = "!";	/* XXX warning: const */
 	}
-	sgent->sg_adm = &empty_list;
-	sgent->sg_mem = &empty_list;
+	sgent->sg_adm = comma_to_list("");
+	sgent->sg_mem = comma_to_list("");
 }
 #endif				/* SHADOWGRP */
 
@@ -247,6 +245,14 @@ grp_update(void)
 		         Prog, sgr_dbname (), sgrp.sg_namp);
 		fail_exit (E_GRP_UPDATE);
 	}
+#endif				/* SHADOWGRP */
+	free_list(grp.gr_mem);
+	free(grp.gr_mem);
+#ifdef	SHADOWGRP
+	free_list(sgrp.sg_mem);
+	free(sgrp.sg_mem);
+	free_list(sgrp.sg_adm);
+	free(sgrp.sg_adm);
 #endif				/* SHADOWGRP */
 }
 
--- a/src/groupmod.c
+++ b/src/groupmod.c
@@ -241,7 +241,7 @@ grp_update(void)
 			sgrp.sg_namp   = xstrdup (grp.gr_name);
 			sgrp.sg_passwd = xstrdup (grp.gr_passwd);
 			sgrp.sg_adm    = &empty;
-			sgrp.sg_mem    = dup_list (grp.gr_mem);
+			sgrp.sg_mem    = grp.gr_mem;
 			new_sgent (&sgrp);
 			osgrp = &sgrp; /* entry needs to be committed */
 		}
--- a/src/groupmod.c
+++ b/src/groupmod.c
@@ -259,8 +259,7 @@ grp_update(void)
 			grp.gr_mem[0] = NULL;
 		} else {
 			// append to existing groups
-			if (NULL != grp.gr_mem[0])
-				grp.gr_mem = dup_list (grp.gr_mem);
+			grp.gr_mem = dup_list (grp.gr_mem);
 		}
 #ifdef	SHADOWGRP
 		if (NULL != osgrp) {
@@ -268,8 +267,7 @@ grp_update(void)
 				sgrp.sg_mem = xmalloc_T(1, char *);
 				sgrp.sg_mem[0] = NULL;
 			} else {
-				if (NULL != sgrp.sg_mem[0])
-					sgrp.sg_mem = dup_list(sgrp.sg_mem);
+				sgrp.sg_mem = dup_list(sgrp.sg_mem);
 			}
 		}
 #endif				/* SHADOWGRP */
--- a/src/groupmod.c
+++ b/src/groupmod.c
@@ -325,6 +325,16 @@ grp_update(void)
 		}
 	}
 #endif				/* SHADOWGRP */
+	if (NULL != user_list) {
+		free_list(grp.gr_mem);
+		free(grp.gr_mem);
+#ifdef	SHADOWGRP
+		if (NULL != osgrp) {
+			free_list(sgrp.sg_mem);
+			free(sgrp.sg_mem);
+		}
+#endif				/* SHADOWGRP */
+	}
 }
 
 /*


src/userdel.c
```diff
@@ -282,6 +282,7 @@ static void update_groups (bool process_selinux)
 #endif				/* WITH_AUDIT */
 		SYSLOG(LOG_INFO, "delete '%s' from shadow group '%s'\n",
 		       user_name, nsgrp->sg_namp);
+		free (nsgrp);
```
alejandro-colomar: __sgr_grp() is not malloc()-like.  It's quite more complex, and needs a specialized cleanup function.

savoptik: Done

src/userdel.c
```diff
@@ -282,6 +282,7 @@ static void update_groups (bool process_selinux)
 #endif				/* WITH_AUDIT */
 		SYSLOG(LOG_INFO, "delete '%s' from shadow group '%s'\n",
 		       user_name, nsgrp->sg_namp);
+		sgr_free (nsgrp);
```
alejandro-colomar, line 283: This code is quite delicate, so it would be good to see some valgrind session that shows the memory leaks, and another session with the patched programs that show the leak is fixed.

It would also be good to see some sessions that show that this doesn't regress the behavior.

alejandro-colomar: @savoptik Were you able to reproduce the bug (and fix) with valgrind?  How about the behavior?

savoptik: Sorry, not yet. I’m on vacation right now. After it, I’ll tackle the issue more closely.

savoptik: > @savoptik Were you able to reproduce the bug (and fix) with valgrind? How about the behavior?

done.
[savoptik-patches.valgrind.txt](https://github.com/user-attachments/files/30188693/savoptik-patches.valgrind.txt)
[del_list-fix.valgrind.txt](https://github.com/user-attachments/files/30188699/del_list-fix.valgrind.txt)

lib/list.c
```diff
@@ -119,11 +119,19 @@ del_list(/*@returned@*/ /*@only@*/char **list, const char *member)
...
+			free (list[i]);
 		}
 	}
 
 	tmp[j] = NULL;
 
+	/* The surviving strings are now owned by 'tmp'; only the old
+	 * container array is stale. */
+	free (list);
+
```
alejandro-colomar, line 132: We already have a PR for addressing this memory leak.  But thanks!

-  <https://github.com/shadow-maint/shadow/pull/1423>

alejandro-colomar: Hmmm, actually, that PR addresses the leak at the bottom, but not the one above.

It would be interesting to separate this fix into two commits, since each fixes a different leak.

src/userdel.c
alejandro-colomar: Thanks!!  Please include these in the commit messages.

savoptik: The commit message mentions what was added. Are you asking to make this a separate paragraph, or do you want to add something else?

alejandro-colomar: I mean pasting the valgrind results in the commit messages.

savoptik: Done

src/userdel.c
```diff
@@ -220,6 +220,7 @@ static void update_groups (bool process_selinux)
 #endif				/* WITH_AUDIT */
 		SYSLOG(LOG_INFO, "delete '%s' from group '%s'\n",
 			user_name, ngrp->gr_name);
+		gr_free (ngrp);
 	}
```
alejandro-colomar, line 223: For commit 1:

```
```

Thanks!

I've checked that gr_update() doesn't take ownership of this, so we can indeed free.

src/useradd.c
```diff
@@ -1067,6 +1067,7 @@ static void grp_update (bool process_selinux)
 		              SHADOW_AUDIT_SUCCESS);
 #endif
 		SYSLOG(LOG_INFO, "add '%s' to group '%s'", user_name, ngrp->gr_name);
+		gr_free (ngrp);
 	}
```
alejandro-colomar, line 1067: For commit 2:

```
```

src/useradd.c
```diff
@@ -1131,6 +1131,7 @@ static void grp_update (bool process_selinux)
 #endif
 		SYSLOG(LOG_INFO, "add '%s' to shadow group '%s'",
 		       user_name, nsgrp->sg_namp);
+		sgr_free (nsgrp);
```
alejandro-colomar, line 1128: For commit 3:

```
```

Thanks!

src/userdel.c
alejandro-colomar, line 283: For commit 4:

```
```

Thanks!

lib/list.c
```diff
@@ -67,6 +67,8 @@ add_list(/*@returned@*/ /*@only@*/char **list, const char *member)
 	tmp[i] = xstrdup (member);
 	tmp[i+1] = NULL;
 
+	free (list);
+
 	return tmp;
 }
```
alejandro-colomar, line 74: About commit 5:

From the commit message:

```

  - groupmod: in the append path (-a), grp.gr_mem / sgrp.sg_mem still
    pointed into the gr_locate()/sgr_locate() database entry whenever the
    existing member list was empty, because dup_list() was skipped in
    that case.  add_list() then free()d that borrowed database array,
    producing a use-after-free and an invalid free() in gr_update() /
    sgr_update().  Duplicate the list unconditionally so add_list()
    always owns it, and free the owned member lists once the entry has
    been stored.  This fixes the memory corruption and the pre-existing
    leak of the duplicated array:

        valgrind ... src/groupmod -a -U alice emptygrp
        before this commit:  invalid free() + use-after-free,
                             definitely lost: 32 bytes in 2 blocks
        after this commit:   0 errors, definitely lost: 0 bytes in 0 blocks
```

I think this use-after-free bug would deserve its own commit, prior to commit 5.

savoptik: Done

src/groupmod.c
```diff
@@ -259,17 +262,18 @@ grp_update(void)
...
 		if (NULL != osgrp) {
 			if (!aflg) {
 				sgrp.sg_mem = xmalloc_T(1, char *);
 				sgrp.sg_mem[0] = NULL;
 			} else {
-				if (NULL != sgrp.sg_mem[0])
-					sgrp.sg_mem = dup_list(sgrp.sg_mem);
+				sgrp.sg_mem = dup_list(sgrp.sg_mem);
 			}
 		}
```
alejandro-colomar, line 272: From the commit message:

> While here: drop the eager dup_list() in the new-sgent branch;
sgrp.sg_mem borrows grp.gr_mem until the append path takes its
own copy, avoiding a redundant allocation.

Could we do this in a separate commit?  I think combining delicate stuff in a single commit complicates review unnecessarily.

alejandro-colomar, line 272: From the commit message:

> , producing a
use-after-free and an invalid free in the subsequent gr_update() /
sgr_update() call.

Does gr_update() free the input?  If so, I'd appreciate some more detailed analysis, as this code is difficult to follow.

savoptik: Both of these calls lead to `lib/commonio.c` ```int commonio_update (struct commonio_db *db, const void *eptr)```.
In different execution paths, the cleanup function ```db->ops->cio_free()```` is called.

alejandro-colomar: > Both of these calls lead to `lib/commonio.c` `int commonio_update (struct commonio_db *db, const void *eptr)`.

Yup.

```c
alx@devuan:~/src/shadow/shadow/master$ grepc -h -tfd gr_update .
int gr_update (const struct group *gr)
{
	return commonio_update (&group_db, gr);
}
alx@devuan:~/src/shadow/shadow/master$ grepc -h -tfd sgr_update .
int sgr_update (const struct sgrp *sg)
{
	return commonio_update (&gshadow_db, sg);
}
```

> In different execution paths, the cleanup function ```db->ops->cio_free()```` is called.

Yup, `...->cio_free()` ends up being called, but it seems to be called on memory allocated within the function itself, AFAICS.


```c
alx@devuan:~/src/shadow/shadow/master$ grepc -h -tfd commonio_update . | grep -e '^\w' -e cio_free
int commonio_update (struct commonio_db *db, const void *eptr)
			db->ops->cio_free(nentry);
		db->ops->cio_free(p->eptr);
		db->ops->cio_free(nentry);
```

If that memory points to memory owned by the caller, at least it's not obvious to me (but this function is hard to read, so it could be).

savoptik: Done.

src/groupmod.c
```diff
@@ -335,6 +335,22 @@ grp_update(void)
...
 	}
 #endif				/* SHADOWGRP */
+	/*
+	 * gr_update()/sgr_update() store a duplicate of the entry, so the
+	 * member lists built above (owned heap arrays) can be released now.
+	 * They are owned only when a member list was (re)built, i.e. when
+	 * user_list was given.
+	 */
```
alejandro-colomar: Please remove the comments.  Let's rely on the commit message for that (feel free to extend the commit message if you feel it's appropriate).

src/groupmod.c
```diff
@@ -335,6 +335,22 @@ grp_update(void)
...
+	if (NULL != user_list) {
+		free_list (grp.gr_mem);
+		free (grp.gr_mem);
+#ifdef	SHADOWGRP
+		if (NULL != osgrp) {
+			free_list (sgrp.sg_mem);
+			free (sgrp.sg_mem);
+		}
+#endif				/* SHADOWGRP */
+	}
```
alejandro-colomar, line 337: Please don't add white space between a function name and the opening parenthesis in new code.  We're slowly removing those spaces in old code.

src/groupadd.c
```diff
@@ -169,8 +167,8 @@ static void new_sgent (struct sgrp *sgent)
 	} else {
 		sgent->sg_passwd = "!";	/* XXX warning: const */
 	}
-	sgent->sg_adm = &empty_list;
-	sgent->sg_mem = &empty_list;
+	sgent->sg_adm = comma_to_list("");
+	sgent->sg_mem = comma_to_list("");
```
alejandro-colomar, line 172: Apart from minor comments, for commit 7:

```
```

lib/list.c
```diff
@@ -121,11 +121,19 @@ del_list(/*@returned@*/ /*@only@*/char **list, const char *member)
 		if (!streq(list[i], member)) {
 			tmp[j] = list[i];
 			j++;
+		} else {
+			/* The removed entry is no longer referenced by the
+			 * new list, so free its string here. */
+			free (list[i]);
```
alejandro-colomar, line 125: Please separate the last commit message in two.  Also, please move the comments into the commit message and remove them from the source code.

src/groupadd.c
savoptik: done

src/groupmod.c
savoptik: done

src/groupmod.c
savoptik: Done

lib/list.c
savoptik: Could you please clarify what you mean by “splitting the commit message”?  

Also, I’ve removed the comments you pointed out — do you mean I should remove all comments from the code, or just the ones we discussed?

alejandro-colomar: > Could you please clarify what you mean by “splitting the commit message”?

Oops, I meant the commit, not commit message.  :)

> Also, I’ve removed the comments you pointed out — do you mean I should remove all comments from the code, or just the ones we discussed?

The existing comments don't need to be removed.  I remove them when modifying the related code enough, and otherwise leave them.

But I'm quite averse to adding new source-code comments.

savoptik: I removed all comments from the source code in all my commits.

Are you sure the last (8th) commit needs to be split?

alejandro-colomar: Yes, please; those are two different leaks.  Thanks!

savoptik: done

savoptik:
ping

alejandro-colomar:
> ping

Thanks for the ping!  I either missed this or forgot about it (I'm not sure which, since I had unusually high email traffic these days).

src/groupmod.c
```diff
@@ -241,7 +241,7 @@ grp_update(void)
 			sgrp.sg_namp   = xstrdup (grp.gr_name);
 			sgrp.sg_passwd = xstrdup (grp.gr_passwd);
 			sgrp.sg_adm    = &empty;
-			sgrp.sg_mem    = dup_list (grp.gr_mem);
+			sgrp.sg_mem    = grp.gr_mem;
```
alejandro-colomar, line 244: Is it possible to move this commit ("fix: src/groupmod.c: borrow gr_mem in new gshadow-entry branch") before the previous one, or does it depend on it?

If it can be moved, it might be better to move it, as it would make it easier to review.

lib/list.c
```diff
@@ -121,6 +121,8 @@ del_list(/*@returned@*/ /*@only@*/char **list, const char *member)
 		if (!streq(list[i], member)) {
 			tmp[j] = list[i];
 			j++;
+		} else {
+			free (list[i]);
```
alejandro-colomar, line 125: For commit 8:

```
```

lib/list.c
```diff
@@ -128,6 +128,8 @@ del_list(/*@returned@*/ /*@only@*/char **list, const char *member)
 
 	tmp[j] = NULL;
 
+	free (list);
+
```
alejandro-colomar, line 132: For commit 9:

```
```

src/groupmod.c
savoptik: done

lib/list.c
savoptik: done

lib/list.c
savoptik: done

src/groupmod.c
alejandro-colomar: I believe this hasn't been changed.  Those commits are still in the same order (and in fact, their commit date remains of 5 days ago).

savoptik: done.

alejandro-colomar: So, it doesn't depend on the other change?

alejandro-colomar:
Patches 1..4 are pretty trivial, and independent of the rest.  I think it would be good if you opened a separate PR with those patches only, and then I'd merge it already.  Then we can continue to discuss the remaining patches here.

src/groupmod.c
savoptik: This is getting ridiculous. You asked me to split these commits, and I did. You asked me to reorder them, and I did. From a Git perspective, they’re independent. Initially, they were in a single patch because they’re related.

So far, the whole discussion has boiled down to me moving the same commits back and forth just because you’re asking me to do it for the sake of your review convenience.

alejandro-colomar: > This is getting ridiculous. You asked me to split these commits, and I did. You asked me to reorder them, and I did.

I asked you whether that could be done (and yes, if it could be done, to do it).

In general, doing things saying 'done' without answering doubts is only partially useful.  I need to understand the patches.

> From a Git perspective, they’re independent.

My question is whether we could merge one without the other without introducing a bug.

If you had opened many small PRs, then it would be clear that they don't have any dependency among them.  By having all patches in a single PR, we have to do more work to understand which patches are dependent (and determine in which order they must be applied).

> Initially, they were in a single patch because they’re related.
> 
> So far, the whole discussion has boiled down to me moving the same commits back and forth just because you’re asking me to do it for the sake of your review convenience.

savoptik:
> Patches 1..4 are pretty trivial, and independent of the rest. I think it would be good if you opened a separate PR with those patches only, and then I'd merge it already. Then we can continue to discuss the remaining patches here.

https://github.com/shadow-maint/shadow/pull/1711

src/groupmod.c
savoptik: These two commits are completely independent of each other. Commits 1 and 2 remove duplication of lists from different parts of the code.

alejandro-colomar: Thanks for clarifying!

savoptik:
I took the liberty of synchronizing the branch with master to avoid having too many differences. Could you please let me know if you have any further questions regarding these changes?

alejandro-colomar:
> I took the liberty of synchronizing the branch with master to avoid having too many differences.

Thanks!

> Could you please let me know if you have any further questions regarding these changes?

I'll review tomorrow (I've been traveling today).

savoptik:
> I'll review tomorrow (I've been traveling today).

Have a great rest and an interesting journey!

src/groupmod.c
alejandro-colomar, line 244: For commit 1:

```
```

src/groupmod.c
```diff
@@ -259,17 +259,15 @@ grp_update(void)
 			grp.gr_mem[0] = NULL;
 		} else {
 			// append to existing groups
-			if (NULL != grp.gr_mem[0])
-				grp.gr_mem = dup_list (grp.gr_mem);
+			grp.gr_mem = dup_list (grp.gr_mem);
 		}
```
alejandro-colomar, line 263: For commit 2:

```
```

(I wonder how this hasn't been found earlier, given that a use-after-free and double-free should crash the program hard.)

src/groupmod.c
alejandro-colomar: This is still unresolved.

alejandro-colomar: Maybe this is the reason why it hasn't been found earlier.  If it's only a use-after-free but not a double-free, then it need not crash hard, and thus would remain hidden.

src/groupmod.c
savoptik: done

src/groupmod.c
savoptik: done

alejandro-colomar:
Thanks!  The only unresolved issue is <https://github.com/shadow-maint/shadow/pull/1638#discussion_r3673893431>.

alejandro-colomar:
> > I'll review tomorrow (I've been traveling today).
> 
> Have a great rest and an interesting journey!

Thanks!  I'm back from vacation.  :)

> Thanks! The only unresolved issue is [#1638 (comment)](https://github.com/shadow-maint/shadow/pull/1638#discussion_r3673893431).

Please check this.  Once we clarify that from the commit message, I'll merge.

savoptik:
> > > I'll review tomorrow (I've been traveling today).
> > 
> > 
> > Have a great rest and an interesting journey!
> 
> Thanks! I'm back from vacation. :)
> 
> > Thanks! The only unresolved issue is [#1638 (comment)](https://github.com/shadow-maint/shadow/pull/1638#discussion_r3673893431).
> 
> Please check this. Once we clarify that from the commit message, I'll merge.

I updated the commit description.

alejandro-colomar (approved):
Thanks!

PR #1645, opened 2026-06-14
po/shadow.pot: Update pot file

alejandro-colomar:
I've updated it by running 'make dist'.

Cc: @mariobl, @mlh85386

### DIFF
--- a/po/shadow.pot
+++ b/po/shadow.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: shadow 4.19.0\n"
 "Report-Msgid-Bugs-To: pkg-shadow-devel@lists.alioth.debian.org\n"
-"POT-Creation-Date: 2025-12-30 14:17+0100\n"
+"POT-Creation-Date: 2026-06-14 21:27+0200\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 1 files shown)


hallyn:
I wonder whether CI should regenerate shadow.pot whenever any of the files listed in po/POTFILES.in are update.

hallyn:
Anyway, lgtm if you want to merge this.

alejandro-colomar:
> I wonder whether CI should regenerate shadow.pot whenever any of the files listed in po/POTFILES.in are update.

I don't like automated commits.  I'd like to oppose that.

Maybe we can "regularly" (i.e., whenever we remember) regenerate it.  Like maybe once every one or two months, when one remembers to do it.

alejandro-colomar:
> Anyway, lgtm if you want to merge this.

I'll regenerate the other file, then open for review.  Thanks!

mariobl:
> Maybe we can "regularly" (i.e., whenever we remember) regenerate it. Like maybe once every one or two months, when one remembers to do it.

Assuming that shadow will one day be available on the Translation Project, it would be sufficient to update both .pot files when a pre-release tarball is sent to the robot. In the meantime, translators can update the files on demand (maybe the instuctions for this could be reviewed and clarified).

alejandro-colomar:
> > Maybe we can "regularly" (i.e., whenever we remember) regenerate it. Like maybe once every one or two months, when one remembers to do it.
> 
> Assuming that shadow will one day be available on the Translation Project, it would be sufficient to update both .pot files when a pre-release tarball is sent to the robot. In the meantime, translators can update the files on demand (maybe the instuctions for this could be reviewed and clarified).

Maybe we can release an -rc already.  After all, the plan was to release on June.

We won't be able to release on June because we have some regressions that need to be fixed, but we could do an -rc already.

@hallyn , would you do an -rc?

hallyn:
Lemme see if I hit any roadblocks in making an rc right now.

PR #1648, opened 2026-06-17
4.20.0 rc1

hallyn:
Line up a pre-release.

Update the libsubid and shadow version numbers.

Update translations.

Fix an issue that made re-calculating translations far more volatile than
it should be.  There's more to be done there, but this was a huge improvement.


hallyn:
I will go ahead and upload the tarballs as a pre-release.  If there are problems with it, we'll just have to jump to rc2.

PR #1453, opened 2025-12-26
doc/contributions/: Add guidelines banning AI for contributing

alejandro-colomar:
This policy has been copied verbatim from the Linux man-pages project, which itself derived it from (and is more restrictive than) Gentoo.

Cc: @ikerexxe 
Cc: @hallyn 

Link: <https://arstechnica.com/ai/2025/07/study-finds-ai-tools-made-open-source-software-developers-19-percent-slower/>
Link: <https://petri.com/ai-coding-tools-rising-software-defects/>
Link: <https://ia.acs.org.au/article/2024/ai-coding-tools-may-produce-worse-software-.html>
Link: <https://carbonate.dev/blog/posts/the-ai-code-quality-crisis>

---

Revisions:

<details>
<summary>v2</summary>

-  Allow minimal uses of AI.

```
$ git rd 
1:  7d360348a ! 1:  dafd0c440 doc/contributions/: Add guidelines banning AI for contributing
    @@ Metadata
     Author: Alejandro Colomar <alx@kernel.org>
     
      ## Commit message ##
    -    doc/contributions/: Add guidelines banning AI for contributing
    +    doc/contributions/: Add guidelines severely restricting use of AI for contributing
     
    -    This policy has been copied verbatim from the Linux man-pages project,
    -    which itself derived it from (and is more restrictive than) Gentoo.
    +    This policy has been derived from Gentoo.  I added a requirement that
    +    use of AI is disclosed.  And changes resulting from said use should be
    +    disclosed in detail.  Also, I left a note saying we'll reject
    +    non-negligible use of AI, which is a bit of an escape allowing us to
    +    just say "too much".
     
         Cc: Iker Pedrosa <ipedrosa@redhat.com>
         Cc: Serge Hallyn <serge@hallyn.com>
    @@ doc/contributions/ai.txt (new)
     +
     +Description
     +  It is expressly forbidden to contribute to this project any
    -+  content that has been created or derived with the assistance of
    ++  content that has been created with the assistance of
     +  AI tools.
     +
    -+  This includes AI assistive tools used in the contributing
    -+  process, even if such tools do not directly generate the
    -+  contributed code but are used to derive the contribution.  For
    -+  example, AI linters, AI static analyzers, and AI tools that
    -+  summarize input are forbidden.
    ++  The use of AI tools should be minimal, limited to reviewing
    ++  a contribution once it's almost finished.  Use of AI tools in
    ++  the process of a contribution should be disclosed explicitly and
[...]

PR #1657, opened 2026-06-23
tests: link unit tests with libshadow.la

### LINKED ISSUE #1654: Missing symbols in unit tests with CFLAGS=-Os
Downstream bug: https://bugs.gentoo.org/972915

If I configure shadow with `CFLAGS=-Os`, gcc avoids inlining most function calls.

This results in several missing symbols in the unit tests which do not include the necessary `extern inline` declarations for functions declared `inline` in headers.

Example:

```
bin/sh ../../libtool  --tag=CC   --mode=link gcc   -Os   -o test_atoi_strtoi ../../lib/atoi/strtoi/test_atoi_strtoi-strtoi.o ../../lib/atoi/strtoi/test_atoi_strtoi-strtou_noneg.o test_atoi_strtoi-test_atoi_strtoi.o -lcmocka  -lbsd 
libtool: link: gcc -Os -o test_atoi_strtoi ../../lib/atoi/strtoi/test_atoi_strtoi-strtoi.o ../../lib/atoi/strtoi/test_atoi_strtoi-strtou_noneg.o test_atoi_strtoi-test_atoi_strtoi.o  -lcmocka -lbsd
/usr/x86_64-pc-linux-gnu/binutils-bin/2.46.1/ld: test_atoi_strtoi-test_atoi_strtoi.o: in function `test_strtou':
test_atoi_strtoi.c:(.text+0x9ac): undefined reference to `strtou_'
/usr/x86_64-pc-linux-gnu/binutils-bin/2.46.1/ld: test_atoi_strtoi.c:(.text+0xa13): undefined reference to `strtou_'
/usr/x86_64-pc-linux-gnu/binutils-bin/2.46.1/ld: test_atoi_strtoi.c:(.text+0xa7b): undefined reference to `strtou_'
/usr/x86_64-pc-linux-gnu/binutils-bin/2.46.1/ld: test_atoi_strtoi.c:(.text+0xb07): undefined reference to `strtou_'
/usr/x86_64-pc-linux-gnu/binutils-bin/2.46.1/ld: test_atoi_strtoi.c:(.text+0xb93): undefined reference to `strtou_'
/usr/x86_64-pc-linux-gnu/binutils-bin/2.46.1/ld: test_atoi_strtoi-test_atoi_strtoi.o:test_atoi_strtoi.c:(.text+0xc2b): more undefined references to `strtou_' follow
collect2: error: ld returned 1 exit status
```

I see three possible solutions:

1. Add the necessary source files which contain the `extern inline` declarations to test_xxx_SOURCES.
2. Add libshadow.la to test_xxx_LDADD.
3. Change the inline functions in headers to `static inline` and drop the `extern inline` declarations.

floppym:
This fixes errors when the compiler chooses not to inline function calls.

Fixes: https://github.com/shadow-maint/shadow/issues/1654

### DIFF
--- a/tests/unit/Makefile.am
+++ b/tests/unit/Makefile.am
@@ -1,5 +1,7 @@
 AM_CPPFLAGS = -I$(top_srcdir)/lib -I$(top_srcdir)
 
+LIBSHADOW = $(top_builddir)/lib/libshadow.la
+
 if HAVE_CMOCKA
 TESTS = $(check_PROGRAMS)
 
@@ -23,8 +25,6 @@ check_PROGRAMS += \
     $(NULL)
 
 test_adds_SOURCES = \
-    ../../lib/adds.c \
-    ../../lib/search/cmp/cmp.c \
     test_adds.c \
     $(NULL)
 test_adds_CFLAGS = \
@@ -33,12 +33,11 @@ test_adds_CFLAGS = \
 test_adds_LDFLAGS = \
     $(NULL)
 test_adds_LDADD = \
+    $(LIBSHADOW) \
     $(CMOCKA_LIBS) \
     $(NULL)
 
 test_atoi_strtoi_SOURCES = \
-    ../../lib/atoi/strtoi/strtoi.c \
-    ../../lib/atoi/strtoi/strtou_noneg.c \
     test_atoi_strtoi.c \
     $(NULL)
 test_atoi_strtoi_CFLAGS = \
@@ -47,11 +46,11 @@ test_atoi_strtoi_CFLAGS = \
 test_atoi_strtoi_LDFLAGS = \
     $(NULL)
 test_atoi_strtoi_LDADD = \
+    $(LIBSHADOW) \
     $(CMOCKA_LIBS) \
     $(NULL)
 
 test_chkhash_SOURCES = \
-    ../../lib/chkhash.c \
     test_chkhash.c \
     $(NULL)
 test_chkhash_CFLAGS = \
@@ -60,12 +59,11 @@ test_chkhash_CFLAGS = \
 test_chkhash_LDFLAGS = \
     $(NULL)
 test_chkhash_LDADD = \
+    $(LIBSHADOW) \
     $(CMOCKA_LIBS) \
     $(NULL)
 
 test_chkname_SOURCES = \
-    ../../lib/chkname.c \
-    ../../lib/sysconf.c \
     test_chkname.c \
     $(NULL)
 test_chkname_CFLAGS = \
@@ -74,11 +72,11 @@ test_chkname_CFLAGS = \
 test_chkname_LDFLAGS = \
     $(NULL)
 test_chkname_LDADD = \
+    $(LIBSHADOW) \
     $(CMOCKA_LIBS) \
     $(NULL)
 
 test_logind_SOURCES = \
-    ../../lib/logind.c \
     test_logind.c \
     $(NULL)
 test_logind_CFLAGS = \
@@ -90,12 +88,12 @@ test_logind_LDFLAGS = \
     -Wl,-wrap,sd_uid_get_sessions \
     $(NULL)
 test_logind_LDADD = \
+    $(LIBSHADOW) \
     $(CMOCKA_LIBS) \
     $(LIBSYSTEMD) \
     $(NULL)
 
 test_stprintf_SOURCES = \
-    ../../lib/string/sprintf/stprintf.c \
     test_stprintf.c \
     $(NULL)
 test_stprintf_CFLAGS = \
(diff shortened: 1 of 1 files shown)


tests/unit/Makefile.am
```diff
@@ -1,5 +1,7 @@
 AM_CPPFLAGS = -I$(top_srcdir)/lib -I$(top_srcdir)
 
+LIBSHADOW = $(top_builddir)/lib/libshadow.la
+
```
alejandro-colomar, line 4: In the commit message:

s/Fixes:/Closes:/

<https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-your-changes>

We use the Fixes tag is for referring to commits that introduce bad code which is fixed.
We use the Closes tag for referring to issue tickets that are closed (resolved) by the change.

tests/unit/Makefile.am
```diff
@@ -90,12 +88,12 @@ test_logind_LDFLAGS = \
     -Wl,-wrap,sd_uid_get_sessions \
     $(NULL)
 test_logind_LDADD = \
+    $(LIBSHADOW) \
     $(CMOCKA_LIBS) \
     $(LIBSYSTEMD) \
     $(NULL)
```
alejandro-colomar, line 94: Does the order matter?  Since cmocka is an interposer, it might matter.  Just wondering.

alejandro-colomar (approved):

tests/unit/Makefile.am
floppym: I am not familiar enough with cmocka to answer this.

The tests appear to work with my changes in effect.

alejandro-colomar: Okay; if it works, that's convincing enough.  :)

alejandro-colomar:
Thanks!

tests/unit/Makefile.am
floppym: On a whim, I asked Claude.ai:

> I have a project which uses automake for a build system.  I am adding unit tests which use the cmocka library. The unit tests utilize code in a helper library libshadow.la. How should I define the LDADD variable in Makefile.am?

It responded with a long explanation, but the gist was this:
```
check_PROGRAMS = test_shadow

test_shadow_SOURCES = test_shadow.c

# Compiler flags - point at headers for cmocka and your library
test_shadow_CFLAGS = $(CMOCKA_CFLAGS) -I$(top_srcdir)/src

# Linker flags/libs
test_shadow_LDADD = libshadow.la $(CMOCKA_LIBS)

TESTS = test_shadow
```
> Order matters (sometimes). LDADD is passed to the linker in order, and on platforms with strict one-pass linkers (some older GNU ld configurations, or --as-needed setups), symbols need to be resolvable left-to-right. Since your test code calls into libshadow.la, and libshadow.la itself might depend on system libs, the typical convention is:
> test_shadow_LDADD = libshadow.la $(CMOCKA_LIBS)
> i.e., your own code/libs first, then third-party libs, then system libs last. If libshadow.la itself needs cmocka symbols (unusual, but possible if you have test helpers baked into it), you may need cmocka listed twice or reordered — but normally cmocka is only needed by the test file itself.

Take from that whatever you will.

alejandro-colomar: > On a whim, I asked Claude.ai:

-  We recently merged <https://github.com/shadow-maint/shadow/pull/1453>.

[...]

> > Order matters (sometimes). LDADD is passed to the linker in order, and on platforms with strict one-pass linkers (some older GNU ld configurations, or --as-needed setups), symbols need to be resolvable left-to-right. Since your test code calls into libshadow.la, and libshadow.la itself might depend on system libs, the typical convention is:
> > test_shadow_LDADD = libshadow.la $(CMOCKA_LIBS)
> > i.e., your own code/libs first, then third-party libs, then system libs last. If libshadow.la itself needs cmocka symbols (unusual, but possible if you have test helpers baked into it), you may need cmocka listed twice or reordered — but normally cmocka is only needed by the test file itself.

The AI seems to have got it completely wrong.  Because CMOCKA is an interposer, it should probabbly go first in the linker flags.  Because it interposes sometimes, it probably needs to call libshadow functions in the cases where it doesn't interpose; that means that libshadow should go after cmocka in the linker.

Of course, our library can't have dependencies on cmocka symbols, so cmocka going after libshadow wouldn't be useful.

But since it worked, and we already merged it, let's keep it as is.  If someone ever reports an issue, it might be that cmocka needs to go first, so we'll check if/when that happens.

Thanks!

floppym: > > On a whim, I asked Claude.ai:
> 
> * We recently merged [doc/contributions/: Add guidelines banning AI for contributing #1453](https://github.com/shadow-maint/shadow/pull/1453).

Gotcha. I do not make regular use of AI when coding, so this wont be a problem.
I only asked it as a curiosity to see if it could answer the question you had posed.

alejandro-colomar: Okay.  :)

PR #1660, opened 2026-06-24
Avoid falling prey to tiocsti

hallyn:
We've long known (see
https://www.halfdog.net/Security/2012/TtyPushbackPrivilegeEscalation/ and https://jdebp.uk/FGA/dont-abuse-su-for-dropping-privileges.html) that su should be used to gain but not drop privilege.  Much more recently, linux added the ability to prevent TIOCSTI through a configurable /proc/sys/dev/tty/legacy_tiocsti setting.

If /proc/sys/dev/tty/legacy_tiocsti is set to 0, then we are protected from the callee injecting commands on caller's tty through TIOCSTI. If it's 1, or doesn't exist, then we are not.  That can be dangerous if caller is root.  We currently give up the controlling terminal for non-interactive uses of su (-c).  Let's do that for interactive calls as well, only in the dangerous case.

### DIFF
--- a/src/su.c
+++ b/src/su.c
@@ -56,6 +56,7 @@
 #ifdef USE_PAM
 #include "pam_defs.h"
 #endif				/* USE_PAM */
+#include "io/fgets/fgets.h"
 #include "pwauth.h"
 #include "prototypes.h"
 #include "shadowlog.h"
@@ -996,6 +997,24 @@ static void set_environment (struct passwd *pw)
 
 }
 
+// See linux.git 83efeeeb3d04 (2022-10-22; "tty: Allow TIOCSTI to be disabled")
+static bool legacy_tiocsti_is_disabled(void)
+{
+	char buf[3];
+	FILE *fp;
+	void *ret;
+
+	fp = fopen("/proc/sys/dev/tty/legacy_tiocsti", "r");
+	if (NULL == fp)
+		return false;
+	ret = fgets_a(buf, fp);
+	fclose(fp);
+	if (ret == NULL)
+		return false;
+
+	return streq(buf, "0\n");
+}
+
 /*
  * su - switch user id
  *
@@ -1010,6 +1029,7 @@ int main (int argc, char **argv)
 {
 	const char *cp;
 	struct passwd *pw = NULL;
+	bool need_pty_prot;
 
 #ifdef USE_PAM
 	int ret;
@@ -1022,6 +1042,8 @@ int main (int argc, char **argv)
 	(void) textdomain (PACKAGE);
 
 	save_caller_context();
+
+	need_pty_prot = caller_is_root && !legacy_tiocsti_is_disabled();
 
 	OPENLOG (Prog);
 
@@ -1152,7 +1174,7 @@ int main (int argc, char **argv)
 
 	set_environment (pw);
 
-	if (!doshell) {
+	if (!doshell || need_pty_prot) {
 		/* There is no need for a controlling terminal.
 		 * This avoids the callee to inject commands on
 		 * the caller's tty. */


src/su.c
```diff
@@ -107,6 +107,19 @@ extern size_t newenvc; /* lib/env.c */
...
+	char c;
+	int fd = open("/proc/sys/dev/tty/legacy_tiocsti", O_RDONLY);
+	if (fd == -1)
+		return false;
+	ret = read(fd, &c, 1);
+	close(fd);
+	if (ret != 1)
+		return false;
+	return c == '0';
+}
```
alejandro-colomar: Some style fixes (as mentioned in the email).
Also, I'd rename it by adding an `_is_`.
Also, I think it'd be simpler to use fgets(3) instead of read(2), since it's a text file.

```c
#include <limits.h>
#include <stdbool.h>
#include <stddef.h>
#include <stdio.h>

#include "io/fgets/fgets.h"

static bool legacy_tiocsti_is_disabled(void);

static bool
legacy_tiocsti_is_disabled(void)
{
	char  ln[LINE_MAX + 1];
	char  *p;
	FILE  *f;

	f = fopen("/proc/sys/dev/tty/legacy_tiocsti", "r");
	if (f == NULL)
		return false;

	p = fgets_a(ln, f);
	fclose(f);
	if (p == NULL)
		return false;

	return streq(ln, "0\n");
}
```

src/su.c
```diff
@@ -1023,6 +1037,9 @@ int main (int argc, char **argv)
 
 	save_caller_context();
 
+	if (caller_is_root && !legacy_tiocsti_disabled())
+		need_pty_prot = true;
+
```
alejandro-colomar: By refactoring this, we can remove the initialization above.

```c
need_pty_prot = (caller_is_root && !legacy_tiocsti_is_disabled());
```

src/su.c
```diff
@@ -107,6 +107,19 @@ extern size_t newenvc; /* lib/env.c */
 
 /* local function prototypes */
 
+static bool legacy_tiocsti_disabled() {
```
alejandro-colomar: Please move this function to where the function definitions are.

src/su.c
ikerexxe: I was adding comments to improve the code when I saw this comment and it looks better

src/su.c
hallyn: Yeah, but that's a lot harder to read.

hallyn: Well, maybe it's not so bad.

src/su.c
```diff
@@ -996,6 +997,29 @@ static void set_environment (struct passwd *pw)
 
 }
 
+/*
+ * See the following kernel commit:
+ * commit 83efeeeb3d04b22aaed1df99bc70a48fe9d22c4d
+ * Author: Kees Cook <kees@kernel.org>
+ * Date:   Sat Oct 22 11:29:49 2022 -0700
+ * Subject: tty: Allow TIOCSTI to be disabled
+ */
```
alejandro-colomar: How about compacting that?:

```c
// See linux.git 83efeeeb3d04 (2022-10-22; "tty: Allow TIOCSTI to be disabled")
```

src/su.c
```diff
@@ -996,6 +997,29 @@ static void set_environment (struct passwd *pw)
...
+static bool legacy_tiocsti_is_disabled(void) {
```
alejandro-colomar: Please move the brace to a new line (and optionally the type, but at least the brace).

src/su.c
```diff
@@ -996,6 +997,29 @@ static void set_environment (struct passwd *pw)
...
+
+	fp = fopen("/proc/sys/dev/tty/legacy_tiocsti", "r");
+	if (NULL == fp)
+		return false;
+	ret = fgets_a(buf, fp);
+	fclose(fp);
+	if (ret == NULL)
+		return false;
+
+	return buf[0] == '0';
```
alejandro-colomar: I'd like to avoid all direct operations on bytes of strings.  The good thing about str*() functions is that you can fortify them with _FORTIFY_SOURCE, and other niceties.  They're also easier to grep(1).

I suggest using one of the following alternatives:

```c
return streq(buf, "0\n");

return !!strspn(buf, "0");
```

alejandro-colomar: I prefer the former, because it also implicitly validates the rest of the string.

Maybe it would be interesting to validate and remove the newline explicitly separately:

```c
if (stpspn(buf, "\n") == NULL)
	return false;

return streq(buf, "0");
```

This makes the streq() call more readable, while still validating the entire line.

src/su.c
```diff
@@ -996,6 +997,29 @@ static void set_environment (struct passwd *pw)
...
 
+/*
+ * See the following kernel commit:
+ * commit 83efeeeb3d04b22aaed1df99bc70a48fe9d22c4d
+ * Author: Kees Cook <kees@kernel.org>
+ * Date:   Sat Oct 22 11:29:49 2022 -0700
+ * Subject: tty: Allow TIOCSTI to be disabled
+ */
+static bool legacy_tiocsti_is_disabled(void) {
+	char buf[3];
```
alejandro-colomar, line 1003: The array size LGTM.

src/su.c
hallyn: I'm not understanding the last one.  stpspn doesn't change the string, right?  You're not clearing out the \n, so why would streq(buf, "0") return true?

alejandro-colomar: Ouch!  You're right; I meant to use `stpsep()`.

alejandro-colomar: (I think the `strspn(3)` family should have been `strspan()`, which would be less prone to confusion with similar function names.  I often confuse strspn(3) and strsep(3).)

alejandro-colomar (approved):
```
```

src/su.c
alejandro-colomar: Feel free to merge already, or use stpsep(), as you wish.

PR #1661, opened 2026-06-24
Translations

alejandro-colomar:
I'm forwarding these from the TP, received via email.  I've attributed the commit authorship to the authors of the translations, and the author date to the Date of the email in which we received them.  The commit messages also contain a From: trailer field and Message-ID: trailer field, which identify the emails in which we received them.

Cc: @ramses-70 , @yurchor


ikerexxe (approved):
From a technical point of view, it’s fine. I don’t know the languages, so I can’t check whether the translations are correct, but I understand that they’ve already taken care of that

PR #1664, opened 2026-06-30
Tests: Change multiple user attributes at once

aborah-sudo:
This is the transformation to Python of the test located in `tests/usertools/01/09_usermod_change_user_info.test` which checks that `usermod` can change multiple user attributes at once

### DIFF
--- a/tests/system/tests/test_usermod.py
+++ b/tests/system/tests/test_usermod.py
@@ -274,3 +274,36 @@ def test_usermod__unlock_password(shadow: Shadow):
     assert shadow_entry is not None, "User should be found"
     assert shadow_entry.password is not None, "Password should not be None"
     assert shadow_entry.password == password_hash, "Password should be unlocked"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_usermod__modify_multiple_attributes(shadow: Shadow):
+    """
+    :title: Change multiple user attributes at once
+    :setup:
+        1. Create test group
+        2. Create user
+        3. Change GID, comment, expiration, inactive, shell, and home
+    :steps:
+        1. Check passwd entry
+        2. Check shadow entry
+    :expectedresults:
+        1. Passwd entry for the user exists and the attributes are correct
+        2. Shadow entry for the user exists and the attributes are correct
+    :customerscenario: False
+    """
+    shadow.groupadd("tgroup -g 9999")
+    shadow.useradd("tuser1")
+    shadow.usermod("-g 9999 --comment 'Test Comment' -e 2000-09-01 -f 17 -s /bin/bash -d /tmp tuser1")
+
+    passwd_entry = shadow.tools.getent.passwd("tuser1")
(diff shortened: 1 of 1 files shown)


tests/system/tests/test_usermod.py
```diff
@@ -274,3 +274,35 @@ def test_usermod__unlock_password(shadow: Shadow):
     assert shadow_entry is not None, "User should be found"
     assert shadow_entry.password is not None, "Password should not be None"
     assert shadow_entry.password == password_hash, "Password should be unlocked"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_usermod__change_user_info(shadow: Shadow):
+    """
+    :title: Change multiple user attributes at once
+    :description: Converted from usertools/01/09_usermod_change_user_info.test
```
ikerexxe: This field isn't used by any other test in the codebase. Can you remove it?

tests/system/tests/test_usermod.py
```diff
@@ -274,3 +274,35 @@ def test_usermod__unlock_password(shadow: Shadow):
...
+    :setup:
+        1. Create user
+        2. Change GID, comment, expiration, inactive, shell, and home
+    :steps:
+        1. Check passwd entry for GID, comment, shell, home
+        2. Check shadow entry for expiration and inactive
```
ikerexxe: The established pattern throughout the repository is to describe what component is being checked (`Check passwd entry` and `Check shadow entry`) without entering into details. Let's follow that pattern

tests/system/tests/test_usermod.py
```diff
@@ -274,3 +274,35 @@ def test_usermod__unlock_password(shadow: Shadow):
...
+    :expectedresults:
+        1. All passwd attributes are updated correctly
+        2. All shadow attributes are updated correctly
```
ikerexxe: Same here. This should be:
```
1. passwd entry for the user exists and the attributes are correct
2. shadow entry for the user exists and the attributes are correct
```

tests/system/tests/test_usermod.py
```diff
@@ -274,3 +274,35 @@ def test_usermod__unlock_password(shadow: Shadow):
     assert shadow_entry is not None, "User should be found"
     assert shadow_entry.password is not None, "Password should not be None"
     assert shadow_entry.password == password_hash, "Password should be unlocked"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_usermod__change_user_info(shadow: Shadow):
```
ikerexxe: This name is too generic. What about renaming it to `test_usermod__modify_multiple_attributes` or `test_usermod__change_multiple_fields_simultaneously`?

ikerexxe (commented):
Some minor comments inline.

In addition, the test is failing in openSUSE and I suspect it's because there isn't a group with ID 1, but I'm not 100% sure so I recommend you to investigate it

ikerexxe (approved):
LGTM!

PR #1668, opened 2026-07-06
Fix find free range 1629

### LINKED ISSUE #1629: shadow Security Hole Introduced / Regression in de7f1c78f70f8df4b2956f7363da3774348bc58e

silverhadch:
Fixes the security regression from de7f1c78f70f8df4b2956f7363da3774348bc58e (overlapping subordinate ID ranges, #1629), continuing @MillaFleurs' work from #1630 with all review comments applied. Kept Co-authored-by: Dan Anderson <https://github.com/MillaFleurs> in both commits as requested.

Closes #1629
Replaces #1630

Full disclosure carried over from the original PR: this was found with N184, @MillaFleurs' open source bug and security vulnerability scanner (https://github.com/MillaFleurs/N184).

### DIFF
--- a/lib/subordinateio.c
+++ b/lib/subordinateio.c
@@ -360,6 +360,9 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 	while (NULL != (range = commonio_next(db))) {
 		intmax_t  first, last;
 
+		if (range->count == 0)
+			continue;
+
 		first = range->start;
 		last = first - 1LL + range->count;
 
--- a/lib/subordinateio.c
+++ b/lib/subordinateio.c
@@ -14,6 +14,7 @@
 #include "getdef.h"
 #include "../libsubid/subid.h"
 #include <sys/types.h>
+#include <sys/param.h>
 #include <pwd.h>
 #include <ctype.h>
 #include <fcntl.h>
@@ -113,10 +114,13 @@ subordinate_parse(const char *line)
 	if (streq(fields[2], ""))
 		return NULL;
 	range.owner = fields[0];
-	if (str2ul(&range.start, fields[1]) == -1)
+	if (a2ul(&range.start, fields[1], NULL, 0, 0, maxof(id_t)) == -1)
 		return NULL;
-	if (str2ul(&range.count, fields[2]) == -1)
+	if (a2ul(&range.count, fields[2], NULL, 0, 0,
+	         MIN(maxof(id_t) + 1LL - range.start, maxof(id_t))) == -1)
+	{
 		return NULL;
+	}
 
 	return &range;
 }
--- a/lib/subordinateio.c
+++ b/lib/subordinateio.c
@@ -17,6 +17,7 @@
 #include <pwd.h>
 #include <ctype.h>
 #include <fcntl.h>
+#include <stdint.h>
 #include <string.h>
 
 #include "alloc/malloc.h"
@@ -29,6 +30,9 @@
 #include "string/strcmp/streq.h"
 #include "string/strtok/strsep2arr.h"
 #include "typetraits.h"
+
+#undef NDEBUG
+#include <assert.h>
 
 
 #define ID_SIZE 31
@@ -333,10 +337,13 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
 static id_t
 find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 {
-	id_t                           low, high;
+	static_assert(sizeof(long long) > sizeof(id_t), "");
+
+	intmax_t                       n, low, high;
 	const struct subordinate_range *range;
 
-	if (count == 0 || max < min || count - 1 > max - min) {
+	n = count;
+	if (n == 0 || max < min || n > max - min + 1LL) {
 		errno = ERANGE;
 		return -1;
 	}
@@ -347,21 +354,21 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 
 	low = min;
 	while (NULL != (range = commonio_next(db))) {
-		id_t  first, last;
+		intmax_t  first, last;
 
 		first = range->start;
-		last = first + range->count - 1;
+		last = first - 1LL + range->count;
 
 		/* Find the top end of the hole before this range */
 		high = first;
 
 		/* Don't allocate IDs after max (included) */
-		if (high > max + 1) {
-			high = max + 1;
+		if (high > max + 1LL) {
+			high = max + 1LL;
 		}
 
 		/* Is the hole before this range large enough? */
-		if ((high > low) && ((high - low) >= count))
+		if ((high > low) && ((high - low) >= n))
 			return low;
 
 		/* Compute the low end of the next hole */
@@ -372,7 +379,7 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 	}
 
 	/* Is the remaining unclaimed area large enough? */
-	if (((max - low) + 1) >= count)
+	if (max - low >= n - 1)
 		return low;
 fail:
 	errno = EUSERS;


silverhadch:
CC: @alejandro-colomar

lib/subordinateio.c
```diff
@@ -332,10 +334,12 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
 find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 {
-	id_t                           low, high;
+	static_assert(sizeof(intmax_t) > sizeof(id_t), "");
+
+	intmax_t                       low, high;
 	const struct subordinate_range *range;
 
-	if (count == 0 || max < min || count - 1 > max - min) {
+	if (count == 0 || max < min || count > (uintmax_t) max - min + 1) {
```
alejandro-colomar: This should probably be using signed arithmetic, as said here:
<https://github.com/shadow-maint/shadow/pull/1630#discussion_r3298262198>

Especially, since the static assertion above is using intmax_t and not uintmax_t.

lib/subordinateio.c
```diff
@@ -346,21 +350,21 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
...
+		last = first + (intmax_t) range->count - 1;
 
 		/* Find the top end of the hole before this range */
 		high = first;
 
 		/* Don't allocate IDs after max (included) */
-		if (high > max + 1) {
-			high = max + 1;
+		if (high > (intmax_t) max + 1) {
+			high = (intmax_t) max + 1;
```
alejandro-colomar: I'd use 1LL to avoid the casts here.

lib/subordinateio.c
```diff
@@ -346,21 +350,21 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
...
 		}
 
 		/* Is the hole before this range large enough? */
-		if ((high > low) && ((high - low) >= count))
+		if ((high > low) && ((high - low) >= (intmax_t) count))
 			return low;
```
alejandro-colomar, line 379: Is this cast necessary?

lib/subordinateio.c
```diff
@@ -371,7 +375,7 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 	}
 
 	/* Is the remaining unclaimed area large enough? */
-	if (((max - low) + 1) >= count)
+	if (((max - low) + 1) >= (intmax_t) count)
 		return low;
```
alejandro-colomar, line 390: This cast is bogus; it doesn't prevent the overflow, which can happen in `(max - low) + 1`

lib/subordinateio.c
```diff
@@ -332,46 +334,70 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
+
+	intmax_t                       n, low, high;
 	const struct subordinate_range *range;
 
-	if (count == 0 || max < min || count - 1 > max - min) {
+	if (count == 0 || max < min || (intmax_t) count > max - min + 1LL) {
 		errno = ERANGE;
 		return -1;
 	}
+	n = count;
```
alejandro-colomar: If you set `n` before this, you'll avoid that cast too.

```c
n = count;
if (n == 0 || max < min || n > max - min + 1LL) {
...
}
```

lib/subordinateio.c
```diff
@@ -332,35 +334,38 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
...
 	commonio_rewind(db);
 
 	low = min;
 	while (NULL != (range = commonio_next(db))) {
-		id_t  first, last;
+		intmax_t  first, last;
 
 		first = range->start;
-		last = first + range->count - 1;
+		last = first + (intmax_t) range->count - 1;
```
alejandro-colomar: How about this to avoid the cast?

```c
last = first - 1LL + range->count;
```

The order isn't nice, but avoiding the cast compensates for that.

lib/subordinateio.c
```diff
@@ -353,8 +353,23 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
...
+
+		/*
+		 * start and count are parsed from the subid file as
+		 * unsigned long, so they are not bounded by id_t.  A range
+		 * starting above INTMAX_MAX also starts above max, because
+		 * intmax_t is strictly wider than id_t, so it cannot
+		 * constrain an allocation, which ends at or below max.
+		 */
+		if (range->start > INTMAX_MAX)
+			continue;
```
alejandro-colomar: How about adding validation code where we parse it, to make sure it fits a id_t?

alejandro-colomar: Cc: @hallyn

lib/subordinateio.c
```diff
@@ -353,8 +353,23 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
...
+
 		first = range->start;
-		last = first + (intmax_t) range->count - 1;
+		if (__builtin_add_overflow(first, range->count - 1, &last))
+			last = INTMAX_MAX;
```
alejandro-colomar: Should we report an ERANGE here too?  Or would that be problematic?

lib/subordinateio.c
```diff
@@ -110,9 +111,19 @@ subordinate_parse(const char *line)
...
+	 * entries are skipped when iterating the database, but the raw
+	 * line is preserved if the file is rewritten.
+	 */
+	if (a2ul(&range.start, fields[1], NULL, 0, 0, type_max(id_t)) == -1)
 		return NULL;
-	if (str2ul(&range.count, fields[2]) == -1)
+	if (a2ul(&range.count, fields[2], NULL, 0, 0, type_max(id_t)) == -1)
+		return NULL;
+	if (range.count != 0 && range.count - 1 > type_max(id_t) - range.start)
 		return NULL;
```
alejandro-colomar: Hmmm, `range.count - 1 > type_max(id_t) - range.start` is weird, as it relies on unsigned wrap-around.  It is also strange, because we usually do the range checks within the a2ul() call.  How about this?

```c
if (a2ul(&range.count, fields[2], NULL, 0, 0, MIN(type_max(id_t) + 1LL - range.start, type_max(id_t))) == -1)
	return NULL;
```

lib/subordinateio.c
```diff
@@ -110,9 +111,19 @@ subordinate_parse(const char *line)
...
 	range.owner = fields[0];
-	if (str2ul(&range.start, fields[1]) == -1)
+
+	/*
+	 * start and count denote IDs, so accept them only if the whole
+	 * range, including its last ID, start + count - 1, fits in id_t.
+	 * a2ul() reports ERANGE for out-of-range values.  Rejected
+	 * entries are skipped when iterating the database, but the raw
+	 * line is preserved if the file is rewritten.
+	 */
```
alejandro-colomar: I'd remove this comment.  If anything from here is not already in the commit message, I'd move it to the commit message.

lib/subordinateio.c
```diff
@@ -353,6 +364,10 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 	while (NULL != (range = commonio_next(db))) {
 		intmax_t  first, last;
 
+		/* Skip empty ranges; they constrain nothing. */
+		if (range->count == 0)
+			continue;
+
```
alejandro-colomar: I think this belongs in a third commit, since it's not inherently related to the id_t vs intmax_t difference.

lib/subordinateio.c
```diff
@@ -332,10 +334,13 @@ static int subordinate_range_cmp (const void *p1, const void *p2)
 static id_t
 find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 {
-	id_t                           low, high;
+	static_assert(sizeof(long long) > sizeof(id_t), "");
+
+	intmax_t                       n, low, high;
```
alejandro-colomar, line 346: For the _first_ commit:

```
```

Thanks!

lib/subordinateio.c
```diff
@@ -14,9 +14,11 @@
 #include "getdef.h"
 #include "../libsubid/subid.h"
 #include <sys/types.h>
+#include <assert.h>
 #include <pwd.h>
```
alejandro-colomar, line 18: `<assert.h>` should be the last include, and must be preceded by `#undef NDEBUG`.  See all other existing inclusions of this header.

lib/subordinateio.c
```diff
@@ -112,9 +114,10 @@ subordinate_parse(const char *line)
 	if (streq(fields[2], ""))
 		return NULL;
 	range.owner = fields[0];
-	if (str2ul(&range.start, fields[1]) == -1)
+	if (a2ul(&range.start, fields[1], NULL, 0, 0, type_max(id_t)) == -1)
 		return NULL;
-	if (str2ul(&range.count, fields[2]) == -1)
+	if (a2ul(&range.count, fields[2], NULL, 0, 0,
+	         MIN(type_max(id_t) + 1LL - range.start, type_max(id_t))) == -1)
 		return NULL;
```
alejandro-colomar, line 122: For the second commit:

```
```

Thanks!

alejandro-colomar, line 122: I would add braces here:

```c
if (a2ul(&range.count, fields[2], NULL, 0, 0,
         MIN(type_max(id_t) + 1LL - range.start, type_max(id_t))) == -1)
{
	return NULL;
}
```

See <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c#n54>.

lib/subordinateio.c
```diff
@@ -358,6 +358,10 @@ find_free_range(struct commonio_db *db, id_t min, id_t max, unsigned long count)
 	while (NULL != (range = commonio_next(db))) {
 		intmax_t  first, last;
 
+		/* Skip empty ranges; they constrain nothing. */
+		if (range->count == 0)
+			continue;
```
alejandro-colomar: I'd remove the comment.  The commit message should be enough.

Anyway:

```
```

(All three commits are reviewed now.)

alejandro-colomar (approved):
Thanks!  I'll leave it open for a day or two, in case anyone has comments.  It looks good to me (with minor comments).

lib/subordinateio.c
```diff
@@ -108,9 +114,10 @@ subordinate_parse(const char *line)
 	if (streq(fields[2], ""))
 		return NULL;
 	range.owner = fields[0];
-	if (str2ul(&range.start, fields[1]) == -1)
+	if (a2ul(&range.start, fields[1], NULL, 0, 0, type_max(id_t)) == -1)
```
alejandro-colomar: We've merged today a PR that renamed `type_max()` to `maxof()` (for consistency with GCC 16 `_Maxof()`).

alejandro-colomar (approved):
Thanks!  I'll merge.

PR #1398, opened 2025-12-04
lib/chkname.c: Don't allow '$' in user/group names

alejandro-colomar:
The dollar is dangerous.  Require users to use --badname to allow it.

Cc: @zeha , @thesamesam , @jubalh , @Zugschlus , @stoeckmann 

---

Revisions:

<details>
<summary>v1b</summary>

-  Expand commit message.

```
$ git rd 
1:  91f74f912 ! 1:  895b1b420 lib/chkname.c: Don't allow '$' in user/group names
    @@ Metadata
      ## Commit message ##
         lib/chkname.c: Don't allow '$' in user/group names
     
    -    The dollar is dangerous.  Require users to use --badname to allow it.
    +    The dollar is dangerous.  Most languages use it as a special character,
    +    such as the shell --which uses it with many different meanings,
    +    depending on the context--.
    +
    +    Require users to use --badname to allow it.
     
         Signed-off-by: Alejandro Colomar <alx@kernel.org>
     
```
</details>

<details>
<summary>v1c</summary>

-  Rebase

```
$ git rd 
1:  895b1b420 ! 1:  0d058ad4e lib/chkname.c: Don't allow '$' in user/group names
    @@ Commit message
     
      ## lib/chkname.c ##
     @@ lib/chkname.c: is_valid_name(const char *name)
    +    * User/group names must match BRE regex:
    +    *    [a-zA-Z0-9_.][a-zA-Z0-9_.-]*$\?
    +    *
    +-   * as a non-POSIX, extension, allow "$" as the last char for
    +-   * sake of Samba 3.x "add machine script"
    +-   *
    +    * Also do not allow fully numeric names or just "." or "..".
    +    */
      
    -   /*
    -          * User/group names must match BRE regex:
    --         *    [a-zA-Z0-9_.][a-zA-Z0-9_.-]*$\?
    --         *
    --         * as a non-POSIX, extension, allow "$" as the last char for
    --         * sake of Samba 3.x "add machine script"
    -+         *    [a-zA-Z0-9_.][a-zA-Z0-9_.-]*
    -          *
    -          * Also do not allow fully numeric names or just "." or "..".
    -          */
     @@ lib/chkname.c: is_valid_name(const char *name)
                      (*name >= '0' && *name <= '9') ||
                      *name == '_' ||
```
</details>

<details>
<summary>v1d</summary>

-  Rebase

```
$ git rd 
1:  0d058ad4e ! 1:  54148256a lib/chkname.c: Don't allow '$' in user/group names
[...]

PR #1671, opened 2026-07-12
src/{chfn,chsh,passwd}.c: Allow bad names (regression fix)

alejandro-colomar:
```
    These programs don't add new users.  Because they handle existing users,
    they should handle them with whatever names they already have, including
    bad names.
    
    A recent commit that added some username syntax validation blindly
    removed support for bad names in these programs.  Fix that regression.
```

Also, pass the badname flag as a function parameter instead of an extern variable.  This makes it easier to analyze which programs allow badnames and which don't.

Also, remove an inconsistent (and seemingly accidental) allowance of badnames in group names in newusers(8).

Fixes: 326889ca (2024-10-22; "Fix coverity unbound buffer issues")
Fixes: a2cd3e9ef03a (2019-10-04; "chkname.c, pwck.c, useradd.c, usermod.c, newusers.c: Allow names that do not conform to standards")
Cc: @MarcinDigitic , @edneville

### DIFF
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -946,7 +946,7 @@ main(int argc, char **argv)
 	}
 	myname = xstrdup (pw->pw_name);
 	if (optind < argc) {
-		if (!is_valid_user_name(argv[optind], false) && !is_valid_upn(argv[optind], false)) {
+		if (!is_valid_user_name(argv[optind], true) && !is_valid_upn(argv[optind], true)) {
 			fprintf (stderr, _("%s: Provided user name is not a valid name\n"), Prog);
 			fail_exit (E_NOPERM, process_selinux);
 		}
--- a/src/passwd.c
+++ b/src/passwd.c
@@ -946,7 +946,7 @@ main(int argc, char **argv)
 	}
 	myname = xstrdup (pw->pw_name);
 	if (optind < argc) {
-		if (!is_valid_user_name (argv[optind]) && !is_valid_upn (argv[optind])) {
+		if (!is_valid_user_name(argv[optind], false) && !is_valid_upn(argv[optind], false)) {
 			fprintf (stderr, _("%s: Provided user name is not a valid name\n"), Prog);
 			fail_exit (E_NOPERM, process_selinux);
 		}
--- a/lib/chkname.c
+++ b/lib/chkname.c
@@ -111,7 +111,7 @@ is_valid_user_name(const char *name, bool badnames)
 
 
 bool
-is_valid_group_name(const char *name, bool badnames)
+is_valid_group_name(const char *name)
 {
 	/*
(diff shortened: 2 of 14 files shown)


src/passwd.c
```diff
@@ -946,7 +946,7 @@ main(int argc, char **argv)
 	}
 	myname = xstrdup (pw->pw_name);
 	if (optind < argc) {
-		if (!is_valid_user_name (argv[optind]) && !is_valid_upn (argv[optind])) {
+		if (!is_valid_user_name(argv[optind], true) && !is_valid_upn(argv[optind], true)) {
```
ikerexxe, line 949: I’m not 100 per cent sure we should allow UPNs with `badname` format, but I’m not convinced by the opposite option either...

ikerexxe (commented):
The changes look great, but I think we can make a couple of quick changes to improve the review. If you could squash commits 1 and 4, it would really help tidy up the commit history and make the final diff much easier to read.
Also, to keep our pattern consistent and avoid introducing a new global variable, I'd highly suggest expanding the existing `option_flags` struct to include the `badnames` boolean instead. That way, we can just pass it down wherever it's needed and keep our state centralized.

src/passwd.c
alejandro-colomar: There's the Samba names which contain `$`, which I want to disallow (and thus hide under --badname).

See <https://github.com/shadow-maint/shadow/pull/1398>.

alejandro-colomar:
> The changes look great, but I think we can make a couple of quick changes to improve the review.


> If you could squash commits 1 and 4, it would really help tidy up the commit history and make the final diff much easier to read.

One problem with that is that it would squash a breaking change with a refactor.  I prefer having it separate, so that the diff for the breaking change is minimal.

> Also, to keep our pattern consistent and avoid introducing a new global variable, I'd highly suggest expanding the existing `option_flags` struct to include the `badnames` boolean instead. That way, we can just pass it down wherever it's needed and keep our state centralized.

I'd like to refactor that separately.  I also want to unify the `struct option_flags` and put it in a header file.  I prefer doing that in a separate PR, though.

ikerexxe:
> One problem with that is that it would squash a breaking change with a refactor. I prefer having it separate, so that the diff for the breaking change is minimal.

I see your point here. This was just a preference, so let's keep it as it is

> I'd like to refactor that separately. I also want to unify the `struct option_flags` and put it in a header file. I prefer doing that in a separate PR, though.

I’m really looking forward to seeing this change. At first, I considered doing it this way, but I felt that there were utilities that wouldn’t use all the options, so I ruled it out. I also considered using getters and setters for the values, but that’s very OOP

ikerexxe (approved):
LGTM! So, let's merge it

PR #1684, opened 2026-07-17
useradd: fix subuid allocation

### LINKED ISSUE #1683: useradd: automatic subuid creation is broken
The should_assign_subuid() fn checks for user_id == 0.  But user_id is not assigned until after this fn is called.  We can't just move the fn, because open_files() will only open the subuid db if is_sub_uid is true, which gets set to should_assign_subuid().

hallyn:
closes #1683

The decision on whether to allocate a subuid is dependent on user_id != 0, but user_id is not filled in until after the call to this.  I spent a few minutes trying to track down where this ordering changed, and haven't found it yet. We can add a fixes tag if we track it down.

Move the decision on whether to calculate it to after we calculate user_id.  Then, because that decision is used to decide whether to open the subuid_db, and pull the subuid db opening out of open_files and run it later.

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -1769,8 +1769,11 @@ static void open_files (bool process_selinux)
 	/* shadow file will be opened by open_shadow(); */
 
 	open_group_files (process_selinux);
+}
 
 #ifdef ENABLE_SUBIDS
+static void open_subid_files (bool process_selinux)
+{
 	if (is_sub_uid) {
 		if (sub_uid_lock () == 0) {
 			fprintf (stderr,
@@ -1801,8 +1804,8 @@ static void open_files (bool process_selinux)
 			fail_exit (E_SUB_GID_UPDATE, process_selinux);
 		}
 	}
-#endif				/* ENABLE_SUBIDS */
 }
+#endif				/* ENABLE_SUBIDS */
 
 static void open_group_files (bool process_selinux)
 {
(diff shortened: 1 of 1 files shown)


src/useradd.c
```diff
@@ -1801,8 +1804,8 @@ static void open_files (bool process_selinux)
 			fail_exit (E_SUB_GID_UPDATE, process_selinux);
 		}
 	}
-#endif				/* ENABLE_SUBIDS */
 }
```
alejandro-colomar, line 1807: I've also been trying to figure out when this was broken, and can't find it.  I'm happy to fix it without a fixes tag, if we don't find it.

alejandro-colomar (approved):
I'm happy with this.

```
```

Thanks!

Feel free to add a fixes tag if you find out when it was broken.

hallyn:
For posterity: this was apparently always broken, but the opposite way.  The user_id part of the check was always true if !user_id, and user_id was always 0, so that was ignored.  Alex fixed that by switching to if (user_id == 0) return false, in 8508d61607996556cab976659dcb0ded80c73567.  Yeah, I should have added this to the commit message.  Too late now.

alejandro-colomar:
> For posterity: this was apparently always broken, but the opposite way. The user_id part of the check was always true if !user_id, and user_id was always 0, so that was ignored. Alex fixed that by switching to if (user_id == 0) return false, in [8508d61](https://github.com/shadow-maint/shadow/commit/8508d61607996556cab976659dcb0ded80c73567). Yeah, I should have added this to the commit message. Too late now.

Maybe you could add some patch that fixes some nearby white space and comment this in the commit message?

PR #1689, opened 2026-07-19
find_range: avoid count==0 bug

hallyn:
If someone has a subuid range with count==0, the variable last is calculated as first + count - 1, meaning last == first - 1.  Since the calculated range wraps around, this can invalidate later checks.

Check for this case explicitly.

Because of other checks, this can only lead to erroneous ranges being created if the first loweruid is 0.  This means that the subid range has to be user:0:0, which is doubly wrong, and should never be seen in the real world.

A note on formatting: the calculations of first and last are done in their declarations.  If range->count == 0, these go unused.  The alternatives would be to declare the variables later in the scope - which we do not do - or to assign them later, adding 3 more lines in each scope.

Since this case (again) should never happen, it's not worth avoiding the extra assignments in this error case.

Reported-by: Mohammad Hossein Abedini <mhmd.abedinii@gmail.com>

### DIFF
--- a/lib/subordinateio.c
+++ b/lib/subordinateio.c
@@ -222,6 +222,9 @@ static const struct subordinate_range *find_range(struct commonio_db *db,
 		unsigned long first = range->start;
 		unsigned long last = first + range->count - 1;
 
+		if (range->count == 0)
+			continue;
+
 		if (!streq(range->owner, owner))
 			continue;
 
@@ -248,6 +251,9 @@ static const struct subordinate_range *find_range(struct commonio_db *db,
 	while (NULL != (range = commonio_next(db))) {
 		unsigned long first = range->start;
 		unsigned long last = first + range->count - 1;
+
+		if (range->count == 0)
+			continue;
 
 		/* For performance reasons check range before using getpwnam() */
 		if ((val < first) || (val > last)) {


alejandro-colomar (approved):
```
```

I personally prefer separating declarations from code, but that can be addressed later.
If we separate declarations from code, as a side effect, we wouldn't calculate this unnecessarily.

alejandro-colomar:
FWIW, if the vertical size of the block were a concern (and in general, I think it is a concern; blocks should be small), the best gains would come from chopping braces and unimportant/obsolete comments:

```c
		unsigned long first = range->start;
		unsigned long last = first + range->count - 1;

		if (range->count == 0)
			continue;

		/* For performance reasons check range before using getpwnam() */
		if ((val < first) || (val > last)) {
			continue;
		}
		if (is_same_user(range->owner, owner)) {
			return range;
		}
	}
```
vs
```c
		unsigned long  first, last;

		if (range->count == 0)
			continue;

		first = range->start;
		last = first + range->count - 1;

		if (val < first || val > last)
			continue;
		if (is_same_user(range->owner, owner))
			return range;
	}
```

In this specific case, since we use the variables exactly once, maybe we could even remove the variables.

```c
		if (range->count == 0)
			continue;
		if (val < range->start || val > range->start + range->count - 1)
			continue;
		if (is_same_user(range->owner, owner))
			return range;
	}
```
It packs the conditional too much, but we could split it (which also improves readability by aligning similar stuff):

```c
		if (range->count == 0)
			continue;
		if (val < range->start)
			continue;
		if (val > range->start + range->count - 1)
			continue;
		if (is_same_user(range->owner, owner))
			return range;
	}
```

PR #1454, opened 2025-12-27
lib/salt.c: gensalt(): Default to SHA512 instead of DES

alejandro-colomar:
DES is insecure; use it only if explicitly requested.
    
-  Closes: <https://github.com/shadow-maint/shadow/issues/1278>
-  Reported-by:@andreboscatto
-  Cc: @ikerexxe 

---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase

```
$ git rd 
1:  f4dd13ccb108 < -:  ------------ */: Remove support for MD5_CRYPT_ENAB
2:  85771ad90a6c < -:  ------------ lib/salt.c: Compact conditionals
3:  fdc32380c44a < -:  ------------ */: chpasswd(8): -m,--md5: Remove option
4:  43b5bef9f464 < -:  ------------ */: chgpasswd(8): -m,--md5: Remove option
5:  5bb41add8c8a = 1:  0c12025560a2 lib/salt.c: gensalt(): Default to SHA512 instead of DES
```
</details>

<details>
<summary>v1c</summary>

-  Rebase

```
$ git rd 
1:  0c120255 = 1:  887f62eb lib/salt.c: gensalt(): Default to SHA512 instead of DES
```
</details>

<details>
<summary>v2</summary>

-  Update login.defs(5).  [@hallyn ]

```
$ git rd 
1:  887f62eb ! 1:  f7c5a949 lib/salt.c: gensalt(): Default to SHA512 instead of DES
    @@ Metadata
     Author: Alejandro Colomar <alx@kernel.org>
     
      ## Commit message ##
    -    lib/salt.c: gensalt(): Default to SHA512 instead of DES
    +    lib/, man/: Default to SHA512 instead of DES
     
         DES is insecure; use it only if explicitly requested.
     
    @@ lib/salt.c: static /*@observer@*/const char *gensalt (size_t salt_size)
        }
      
      #if USE_XCRYPT_GENSALT
    +
    + ## man/login.defs.d/ENCRYPT_METHOD.xml ##
    +@@
    +     <para>
    +       It can take one of these values: <phrase condition="bcrypt">
    +       <replaceable>BCRYPT</replaceable>,</phrase>
    +-      <replaceable>DES</replaceable> (default),
    ++      <replaceable>DES</replaceable>,
    +       <replaceable>MD5</replaceable>,
    +       <replaceable>SHA256</replaceable>,
    +-      <replaceable>SHA512</replaceable>,
    ++      <replaceable>SHA512</replaceable> (default),
[...]

PR #1691, opened 2026-07-20
tests/system/etc/login.defs: set ENCRYPT_METHOD SHA512

ikerexxe:
Commit 609bd9194174d ("Remove support for DES") commented out ENCRYPT_METHOD, which caused tests to use the container's default password hashing. On OpenSUSE this defaults to DES, but since the shadow code no longer supports DES, tests fail.

Explicitly setting ENCRYPT_METHOD to SHA512 ensures tests use a supported modern hashing algorithm regardless of container defaults

Fixes: 609bd9194174 (2026-07-17; "*/: Remove support for DES")

### DIFF
--- a/tests/system/etc/login.defs
+++ b/tests/system/etc/login.defs
@@ -301,7 +301,7 @@ CHFN_RESTRICT		rwh
 # Note: If you use PAM, it is recommended to use a value consistent with
 # the PAM modules configuration.
 #
-#ENCRYPT_METHOD SHA512
+ENCRYPT_METHOD SHA512
 
 #
 # Only works if ENCRYPT_METHOD is set to SHA256 or SHA512.


ikerexxe:
@alejandro-colomar please review before 4.20 release

alejandro-colomar:
What do you mean by "container defaults"?  You mean that OpenSUSE has an explicit `ENCRYPT_METHOD=DES` in /etc/login.defs?

How does a container set its default?

ikerexxe:
> What do you mean by "container defaults"? You mean that OpenSUSE has an explicit `ENCRYPT_METHOD=DES` in /etc/login.defs?
> 
> How does a container set its default?

By container defaults I meant OpenSUSE defaults in general.

I don't know exactly how this distribution manages the configuration, but I assume there must be some place (i.e. /etc/login.defs) where `ENCRYPT_METHOD=DES`.

alejandro-colomar:
> > What do you mean by "container defaults"? You mean that OpenSUSE has an explicit `ENCRYPT_METHOD=DES` in /etc/login.defs?
> > How does a container set its default?
> 
> By container defaults I meant OpenSUSE defaults in general.
> 
> I don't know exactly how this distribution manages the configuration, but I assume there must be some place (i.e. /etc/login.defs) where `ENCRYPT_METHOD=DES`.

Hmmmmm, if OpenSUSE defaults to `DES` explicitly, maybe we should let it fail.  That will make it more evident that OpenSUSE needs to fix their defaults?

Cc: @jubalh

jubalh:
Hi guys,
We use `ENCRYPT_METHOD SHA512` in openSUSE by default.
See https://build.opensuse.org/projects/Base:System/packages/shadow/files/shadow-login_defs-suse.patch?expand=1

alejandro-colomar:
@jubalh would you mind looking at CI and see why it's consistently failing for OpenSUSE in master?  It's weird that this PR fixes the issue given OpenSUSE defaults to SHA512 as you say.

ikerexxe:
From a quick look, the `opensuse/tumbleweed` container lacks both `/etc/login.defs` and `/etc/login.defs.d/` by default, which seems to trigger the fallback to `DES`

alejandro-colomar:
> From a quick look, the `opensuse/tumbleweed` container lacks both `/etc/login.defs` and `/etc/login.defs.d/` by default, which seems to trigger the fallback to `DES`

But we now default to SHA512, right.  Do the tests run the system shadow utils, or do they run the shadow utils we build?  Or a combination of both?  The ones we build don't fall back to DES anymore.

<https://github.com/shadow-maint/shadow/pull/1454/commits/a9cd6347d71abc2bbafdff57fdc57fc16c6611fe>

alejandro-colomar:
> From a quick look, the `opensuse/tumbleweed` container lacks both `/etc/login.defs` and `/etc/login.defs.d/` by default, which seems to trigger the fallback to `DES`

Hmmm, this sounds like something OpenSUSE should fix.  (But we could workaround it, in case they have good reasons for that.)

ikerexxe:
> But we now default to SHA512, right. Do the tests run the system shadow utils, or do they run the shadow utils we build? Or a combination of both? The ones we build don't fall back to DES anymore.

The CI builds the latest shadow, installs it and then tests it. In this case that means that your changes to remove DES are included in the build and test for this PR.

alejandro-colomar:
> > But we now default to SHA512, right. Do the tests run the system shadow utils, or do they run the shadow utils we build? Or a combination of both? The ones we build don't fall back to DES anymore.
> 
> The CI builds the latest shadow, installs it and then tests it. In this case that means that your changes to remove DES are included in the build and test for this PR.

If CI uses the latest shadow, why does it need `ENCRYPT_METHOD` to use `SHA512`?  We already default to `SHA512` if there's no configuration:

<https://github.com/shadow-maint/shadow/blob/da3d7afa63375604a74cfd5af9062e7e4010e674/lib/salt.c#L356>

ikerexxe:
> If CI uses the latest shadow, why does it need `ENCRYPT_METHOD` to use `SHA512`? We already default to `SHA512` if there's no configuration:
> 
> https://github.com/shadow-maint/shadow/blob/da3d7afa63375604a74cfd5af9062e7e4010e674/lib/salt.c#L356

I don't know, but from the tests I can tell you it's defaulting to `DES`. Let's see if I have some time to investigate this problem in the following days.

alejandro-colomar:
> > If CI uses the latest shadow, why does it need `ENCRYPT_METHOD` to use `SHA512`? We already default to `SHA512` if there's no configuration:
> > https://github.com/shadow-maint/shadow/blob/da3d7afa63375604a74cfd5af9062e7e4010e674/lib/salt.c#L356
> 
> I don't know, but from the tests I can tell you it's defaulting to `DES`. Let's see if I have some time to investigate this problem in the following days.

I suspect we're somehow calling the binary that comes with the distro instead of the one we build.  That's the only binary that is supposed to default to DES.

ikerexxe:
> I suspect we're somehow calling the binary that comes with the distro instead of the one we build. That's the only binary that is supposed to default to DES.

No, I just checked this and we are calling the built one.

From what I can see OpenSUSE is built with `--with-libpam` option set, but the container doesn't contain any PAM service file (`/etc/pam.d` is empty). If I create a very basic file there for `chpasswd` and `passwd`, then the generated password uses the proper algorithm.

@jubalh do you mind helping us fix this? I'm not sure which is the best way to solve this. Do you have an additional package installing the PAM service files? Should shadow package provide the PAM service files? I also wonder whether this is a problem only happening in containers or something that could happen in all image types.

Vogtinator:
> From a quick look, the `opensuse/tumbleweed` container lacks both `/etc/login.defs` and `/etc/login.defs.d/` by default, which seems to trigger the fallback to `DES`

In openSUSE, the distro supplied config file is `/usr/etc/login.defs`. shadow is built with libeconf and `--enable-vendordir=/usr/etc`.

alejandro-colomar:
> > I suspect we're somehow calling the binary that comes with the distro instead of the one we build. That's the only binary that is supposed to default to DES.
> 
> No, I just checked this and we are calling the built one.
> 
> From what I can see OpenSUSE is built with `--with-libpam` option set, but the container doesn't contain any PAM service file (`/etc/pam.d` is empty). If I create a very basic file there for `chpasswd` and `passwd`, then the generated password uses the proper algorithm.

Hmmm, I still don't understand the cause of the bug.

> 
> @jubalh do you mind helping us fix this? I'm not sure which is the best way to solve this. Do you have an additional package installing the PAM service files? Should shadow package provide the PAM service files? I also wonder whether this is a problem only happening in containers or something that could happen in all image types.

Yes, please.

I'd like to fully understand the problem.

Vogtinator:
> From what I can see OpenSUSE is built with --with-libpam option set, but the container doesn't contain any PAM service file (/etc/pam.d is empty).

Like with login.defs (https://github.com/shadow-maint/shadow/pull/1691#issuecomment-5069755091), that's not in `/etc/pam.d/` but in `/usr/lib/pam.d/`.

FWICT the CI fails because `./src/chpasswd` is built with PAM support, calls into the distro provided `pam_unix`, but that reads the CI provided `/etc/login.defs` which no longer sets `ENCRYPT_METHOD`, so `pam_unix` defaults to DES.

alejandro-colomar:
> > From what I can see OpenSUSE is built with --with-libpam option set, but the container doesn't contain any PAM service file (/etc/pam.d is empty).
> 
> Like with login.defs ([#1691 (comment)](https://github.com/shadow-maint/shadow/pull/1691#issuecomment-5069755091)), that's not in `/etc/pam.d/` but in `/usr/lib/pam.d/`.
> 
> FWICT the CI fails because `./src/chpasswd` is built with PAM support, calls into the distro provided `pam_unix`, but that reads the CI provided `/etc/login.defs` which no longer sets `ENCRYPT_METHOD`, so `pam_unix` defaults to DES.

Ok, so libpam still defaults to DES?  This makes more sense.  @ldv-alt, @thkukuk , do you think it's time for PAM to change the default to SHA512?  That will make it consistent with shadow, since we've changed the default for 4.20.0 (to be released this week).

thkukuk:
`ENCRYPT_METHOD` from login.defs is the standard to set the default hash algorithm for a lot of utilities since I think over 20 years now, but I have no overview who else does read this variable and what the fallbacks are.
So I expect quite some fallout's from this change, not only pam_unix.so.

alejandro-colomar:
> `ENCRYPT_METHOD` from login.defs is the standard to set the default hash algorithm for a lot of utilities since I think over 20 years now, but I have no overview who else does read this variable and what the fallbacks are. So I expect quite some fallout's from this change, not only pam_unix.so.

Well, the only problem we have is in tests, because some runs still work with DES where we expect it would use SHA512.

We could fix that by changing the tests to specifically use SHA512, and I wouldn't worry much here.

The main problem I see here (and it's not really a big problem) is that people might think our reasonably safe default is the same default as with other programs, and then rely on not needing to specify it --where they need to specify it, to avoid old programs still using DES--.  But nothing should break in the short term, other than maybe tests.

However, this is something important to document in the release notes: other programs that also parse login.defs(5) may still default to DES.  Maybe also in a HISTORY section in login.defs(5) too.

ikerexxe:
> FWICT the CI fails because `./src/chpasswd` is built with PAM support, calls into the distro provided `pam_unix`, but that reads the CI provided `/etc/login.defs` which no longer sets `ENCRYPT_METHOD`, so `pam_unix` defaults to DES.

Ok, this explains the problem perfectly. Thank you!

ikerexxe:
@alejandro-colomar I just merged https://github.com/shadow-maint/shadow/pull/1697

Even though shadow-utils doesn't directly need to declare this variable in `login.defs`, dependencies like Linux-PAM rely on it to enable modern hashing algorithms. This PR fixes those specific use cases, so let's go ahead and merge it too.

thkukuk:
> Ok, so libpam still defaults to DES? This makes more sense. @ldv-alt, @thkukuk , do you think it's time for PAM to change the default to SHA512? That will make it consistent with shadow, since we've changed the default for 4.20.0 (to be released this week).

I created a PR: https://github.com/linux-pam/linux-pam/pull/1006

alejandro-colomar:
> @alejandro-colomar I just merged #1697
> 
> Even though shadow-utils doesn't directly need to declare this variable in `login.defs`, dependencies like Linux-PAM rely on it to enable modern hashing algorithms. This PR fixes those specific use cases, so let's go ahead and merge it too.

Sounds reasonable; I'll review later today.

tests/system/etc/login.defs
```diff
@@ -301,7 +301,7 @@ CHFN_RESTRICT		rwh
 # Note: If you use PAM, it is recommended to use a value consistent with
 # the PAM modules configuration.
 #
-#ENCRYPT_METHOD SHA512
+ENCRYPT_METHOD SHA512
 
 #
```
alejandro-colomar, line 306: > Commit https://github.com/shadow-maint/shadow/commit/609bd9194174d631a5fbc9224f5e0dac4dfd4a8c ("Remove support for DES") commented out
ENCRYPT_METHOD, which caused tests to use the container's default
password hashing. On OpenSUSE this defaults to DES, but since the
shadow code no longer supports DES, tests fail.

The commit message should be updated to say that it's PAM that still defaults to DES.

ikerexxe: Done

alejandro-colomar (approved):
Thanks @ikerexxe for the patch!  And thank you all for helping figure out what was the actual problem!

```
```

alejandro-colomar:
Ironically, OpenSUSE CI is still failing, but this time it seems to be due to unrelated reasons.  I've run it again, in case it works in the second time.  Even if it fails again, I think we should probably merge when it finishes.

Vogtinator:
> Ironically, OpenSUSE CI is still failing, but this time it seems to be due to unrelated reasons.

Yup, that looks unrelated. Most likely the container image was a bit newer than what the online repos had, which should fix itself rather soonish.

PR #1693, opened 2026-07-21

asakure:
existing bash tests with numbering 22, 23, 24.

### DIFF
--- a/tests/system/framework/roles/shadow.py
+++ b/tests/system/framework/roles/shadow.py
@@ -46,9 +46,12 @@ class Shadow(BaseLinuxRole[ShadowHost]):
         """
         super().teardown()
 
-    def _parse_args(self, *args) -> Dict[str, str]:
-        args_list = shlex.split(*args[0])
-        name = args_list[-1]
+    def _parse_args(self, args) -> Dict[str, str]:
+        if not args or not args[0]:
+            return {"name": ""}
+
+        args_list = shlex.split(args[0])
+        name = args_list[-1] if args_list else ""
 
         return {"name": name}
 
@@ -230,7 +233,7 @@ class Shadow(BaseLinuxRole[ShadowHost]):
         """
         args_dict = self._parse_args(args)
         self.logger.info(f'Creating group "{args_dict["name"]}" on {self.host.hostname}')
-        cmd = self.host.conn.run("groupadd " + args[0], log_level=ProcessLogLevel.Error)
+        cmd = self.host.conn.run("groupadd " + (args[0] if args else ""), log_level=ProcessLogLevel.Error)
 
         self.host.discard_file("/etc/group")
         self.host.discard_file("/etc/gshadow")
--- a/tests/system/tests/test_groupadd.py
+++ b/tests/system/tests/test_groupadd.py
@@ -8,6 +8,7 @@ import re
 
 import pytest
 from passlib.hash import sha512_crypt
+from pytest_mh.conn import ProcessError
 
 from framework.misc import shadow_password_pattern
 from framework.roles.shadow import Shadow
@@ -150,3 +151,21 @@ def test_groupadd__force_group_creation(shadow: Shadow):
         gshadow_entry = shadow.tools.getent.gshadow("tgroup")
         assert gshadow_entry is not None, "Group should be found"
         assert gshadow_entry.name == "tgroup", "Incorrect groupname"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_groupadd__usage(shadow: Shadow):
+    """
+    :title: Groupadd command displays usage with -h option and exits successfully
+    :setup:
+        1. None required
+    :steps:
+        1. Run groupadd command with -h option
+        2. Verify that groupadd command exits successfully
+    :expectedresults:
+        1. Command shows usage help
+        2. groupadd command completes successfully
+    :customerscenario: False
+    """
+    result = shadow.groupadd("-h")
+    assert result.rc == 0, f"Expected return code 0(success), got {result.rc}"
--- a/tests/system/tests/test_groupadd.py
+++ b/tests/system/tests/test_groupadd.py
@@ -188,3 +188,22 @@ def test_groupadd__no_group(shadow: Shadow):
     with pytest.raises(ProcessError) as exc_info:
         shadow.host.conn.run("groupadd")
     assert exc_info.value.rc == 2, f"Expected return code 2(invalid usage), got {exc_info.value.rc}"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_groupadd__two_groups(shadow: Shadow):
+    """
+    :title: Groupadd command fails when two groups are mentioned
+    :setup:
+        1. None required
+    :steps:
+        1. Run groupadd command with two groupnames
+        2. Verify that groupadd command fails
+    :expectedresults:
+        1. Command with two groupnames fails
+        2. groupadd command fails with error (invalid usage)
+    :customerscenario: False
+    """
+    with pytest.raises(ProcessError) as exc_info:
+        shadow.groupadd("tgroup1 tgroup2")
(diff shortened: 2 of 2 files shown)


tests/system/tests/test_groupadd.py
```diff
@@ -464,3 +464,21 @@ def test_groupadd__existing_group(shadow: Shadow):
...
+    :steps:
+        1. Run groupadd command with -h option
+        2. Verify that groupadd command exits successfully
+    :expectedresults:
+        1. Command shows usage help
+        2. groupadd command completes successfully and displays usage information
+    :customerscenario: False
+    """
+    result = shadow.groupadd("-h")
+    assert result.rc == 0, f"Expected return code 0(success), got {result.rc}"
```
ikerexxe, line 486: You could add additional asserts checking specific parts of the output

tests/system/framework/roles/shadow.py
```diff
@@ -230,7 +230,7 @@ def groupadd(self, *args) -> ProcessResult:
         """
         args_dict = self._parse_args(args)
         self.logger.info(f'Creating group "{args_dict["name"]}" on {self.host.hostname}')
-        cmd = self.host.conn.run("groupadd " + args[0], log_level=ProcessLogLevel.Error)
+        cmd = self.host.conn.run("groupadd " + (args[0] if args else ""), log_level=ProcessLogLevel.Error)
```
ikerexxe: Move the changes in this file to a separate commit

tests/system/tests/test_groupadd.py
```diff
@@ -485,19 +485,26 @@ def test_groupadd__usage(shadow: Shadow):
...
         2. Verify that groupadd command fails
     :expectedresults:
-        1. Command without any argument fails
+        1. Command with invalid argument fails
         2. groupadd command fails with error (invalid usage)
     :customerscenario: False
     """
     with pytest.raises(ProcessError) as exc_info:
-        shadow.groupadd()
+        shadow.groupadd(f"{args}")
```
ikerexxe: You don't need to prepend the `f`

tests/system/tests/test_groupadd.py
```diff
@@ -483,3 +483,22 @@ def test_groupadd__usage(shadow: Shadow):
...
+        1. Run groupadd command without any argument
+        2. Verify that groupadd command fails
+    :expectedresults:
+        1. Command without any argument fails
+        2. groupadd command fails with error (invalid usage)
+    :customerscenario: False
+    """
+    with pytest.raises(ProcessError) as exc_info:
+        shadow.groupadd()
+    assert exc_info.value.rc == 2, f"Expected return code 2(invalid usage), got {exc_info.value.rc}"
```
ikerexxe: Add an empty line between these two

tests/system/framework/roles/shadow.py
```diff
@@ -47,8 +47,8 @@ def teardown(self) -> None:
         super().teardown()
 
     def _parse_args(self, *args) -> Dict[str, str]:
-        args_list = shlex.split(*args[0])
-        name = args_list[-1]
+        args_list = shlex.split(args[0][0] if args[0] else "")
+        name = args_list[-1] if args_list else ""
```
ikerexxe: `args[0][0]` takes the first character of the argument, not the whole argument string.

I think a good approach would be to have the defensive programming first and then handle the parsing:
```
if not args or not args[0]:
    return {"name": ""}

args_list = shlex.split(args[0])
name = args_list[-1] if args_list else ""
```

tests/system/tests/test_groupadd.py
```diff
@@ -464,3 +464,48 @@ def test_groupadd__existing_group(shadow: Shadow):
         existing_gshadow_entry = shadow.tools.getent.gshadow("tgroup")
         assert existing_gshadow_entry is not None, "Group should be found"
         assert existing_gshadow_entry.name == "tgroup", "Incorrect groupname"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_groupadd__usage(shadow: Shadow):
+    """
+    :title: Groupadd command displays usage with -h option and exits successfully
```
ikerexxe: Let's avoid using `-h option` and only state what the test does `display usage`

tests/system/tests/test_groupadd.py
```diff
@@ -464,3 +464,48 @@ def test_groupadd__existing_group(shadow: Shadow):
...
+    :setup:
+        1. None required
+    :steps:
+        1. Run groupadd command with -h option
```
ikerexxe: No need to mention `with -h option`, as that's the how

tests/system/tests/test_groupadd.py
```diff
@@ -464,3 +464,48 @@ def test_groupadd__existing_group(shadow: Shadow):
...
+        2. Verify that groupadd command exits successfully
```
ikerexxe, line 477: You are missing an additional step here: `Check usage information`

tests/system/tests/test_groupadd.py
```diff
@@ -464,3 +464,48 @@ def test_groupadd__existing_group(shadow: Shadow):
...
+    :expectedresults:
+        1. Command shows usage help
```
ikerexxe: I'd say this should be `Command runs successfully`

tests/system/tests/test_groupadd.py
```diff
@@ -464,3 +464,48 @@ def test_groupadd__existing_group(shadow: Shadow):
...
+        2. groupadd command completes successfully and displays usage information
```
ikerexxe: In this step you are checking that the command run successfully, thus update the message accordingly

tests/system/tests/test_groupadd.py
```diff
@@ -464,3 +464,48 @@ def test_groupadd__existing_group(shadow: Shadow):
...
+    :customerscenario: False
```
ikerexxe, line 483: Missing another step here to `Check usage information`

tests/system/tests/test_groupadd.py
```diff
@@ -464,3 +464,48 @@ def test_groupadd__existing_group(shadow: Shadow):
...
+@pytest.mark.parametrize(
+    "args",
+    [
+        pytest.param("", id="no_group"),
+        pytest.param("tgroup1 tgroup2", id="two_groups"),
+    ],
+)
+def test_groupadd__invalid_usage(shadow: Shadow, args: str):
+    """
+    :title: Groupadd command fails with invalid usage
```
ikerexxe: s/usage/arguments

tests/system/tests/test_groupadd.py
```diff
@@ -464,3 +464,48 @@ def test_groupadd__existing_group(shadow: Shadow):
...
+    :setup:
+        1. None required
+    :steps:
+        1. Run groupadd command with invalid arguments
+        2. Verify that groupadd command fails
+    :expectedresults:
+        1. Command with invalid argument fails
```
ikerexxe: I'd say this should be `Attempt to create groups` or something similar

tests/system/framework/roles/shadow.py
asakure: Hi @ikerexxe with this change, test fails with error:
```
FAILED tests/test_groupadd_working.py::test_groupadd__no_group (shadow) - IndexError: tuple index out of range
```
I think an additional change would be required here: https://github.com/shadow-maint/shadow/blob/5e7ae07fa785af8bb614c0f9e24e06a82b3915a8/tests/system/framework/roles/shadow.py#L233 ? Kindly suggest.

ikerexxe: If I'm not mistake you can use `(args[0] if args else "")`. Didn't you have something similar before?

asakure: Right, I added `(args[0] if args else "")`, now all test PASSED successfully.

ikerexxe (approved):
LGTM

PR #1695, opened 2026-07-22

asakure:
existing bash tests with numbering 25, 26, 27.

### DIFF
--- a/tests/system/tests/test_groupadd.py
+++ b/tests/system/tests/test_groupadd.py
@@ -209,3 +209,32 @@ def test_groupadd__o_without_g(shadow: Shadow):
     if shadow.host.features["gshadow"]:
         gshadow_entry = shadow.tools.getent.gshadow("tgroup")
         assert gshadow_entry is None, "Group should not be found"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_groupadd__invalid_option(shadow: Shadow):
+    """
+    :title: Group creation fails with invalid option
+    :setup:
+        1. None required
+    :steps:
+        1. Attempt to create group
+        2. Verify that groupadd command fails
+        3. Check group and gshadow entries
+    :expectedresults:
+        1. Group entry is not created
+        2. groupadd command fails with error (invalid usage)
+        3. No group or gshadow entries are found
+    :customerscenario: False
+    """
+    with pytest.raises(ProcessError) as exc_info:
+        shadow.groupadd("-invalid tgroup")
+
+    assert exc_info.value.rc == 2, f"Expected return code 2 (invalid usage), got {exc_info.value.rc}"
+
+    group_entry = shadow.tools.getent.group("tgroup")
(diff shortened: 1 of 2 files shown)


tests/system/tests/test_groupadd.py
```diff
@@ -512,3 +512,34 @@ def test_groupadd__invalid_arguments(shadow: Shadow, args: str):
...
+        3. Check that /etc/gshadow file is not present
+    :expectedresults:
+        1. Group is created
+        2. Group entry is found
+        3. /etc/gshadow file is not found
+    :customerscenario: False
+    """
+    shadow.fs.rm("/etc/gshadow")
+
+    shadow.login_defs["FORCE_SHADOW"] = "no"
```
ikerexxe, line 536: I can't find this change in the original bash test. Are you sure this is how the test should behave?

tests/system/tests/test_groupadd.py
```diff
@@ -543,3 +543,32 @@ def test_groupadd__no_gshadow(shadow: Shadow):
 
     gshadow_file = shadow.fs.exists("/etc/gshadow")
     assert not gshadow_file, "/etc/gshadow file should not be found"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_groupadd__o_without_g(shadow: Shadow):
```
ikerexxe: Think of a better name that explains what the test is doing

tests/system/tests/test_groupadd.py
```diff
@@ -543,3 +543,32 @@ def test_groupadd__no_gshadow(shadow: Shadow):
...
+    """
+    :title: Groupadd command fails when -o is mentioned without -g
```
ikerexxe: Same here. What does `-o` do? And `-g`?

tests/system/tests/test_groupadd.py
asakure: Yes because by default `FORCE_SHADOW   yes` is set in `login.defs` which leads to forced creation of `/etc/shadow` file and this test would fail with AssertionError.

ikerexxe: Ok, I see

ikerexxe (approved):
LGTM!

ikerexxe:
I just discovered that `FORCE_SHADOW` isn't documented anywhere.

@asakure would you like to document it? If you are interested you'd need to do something similar to the changes in https://github.com/shadow-maint/shadow/commit/203693c0edb17ea42105d172967d97aa0cdbfe15 for `man/login.defs.5.xml` and `man/login.defs.d/SUB_GID_STORE_BY_UID.xml`

asakure:
Hi @ikerexxe, I have raised https://github.com/shadow-maint/shadow/pull/1730 to document `FORCE_SHADOW` option in man-page. Do I need to add it in `CROSS REFERENCES` section as well?

PR #1696, opened 2026-07-23
Tests: Add system tests for groupdel in test_groupdel.py

asakure:
Add system tests for groupdel in test_groupdel.py for
existing bash tests with numbering 2,3.

### DIFF
--- a/tests/system/tests/test_groupdel.py
+++ b/tests/system/tests/test_groupdel.py
@@ -64,3 +64,31 @@ def test_groupdel__delete_group_no_gshadow_entry(shadow: Shadow):
     if shadow.host.features["gshadow"]:
         gshadow_entry = shadow.tools.getent.gshadow("tgroup")
         assert gshadow_entry is None, "Group should not be found"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_groupdel__delete_group_no_gshadow_file(shadow: Shadow):
+    """
+    :title: Group deletion with no gshadow file
+    :setup:
+        1. Create group
+        2. Remove gshadow file
+    :steps:
+        1. Delete group
+        2. Check group entry
+        3. Verify that gshadow file doesn't exist
+    :expectedresults:
+        1. Group entry is deleted successfully
+        2. No group entry is found
+        3. No gshadow file is found
+    :customerscenario: False
+    """
+    shadow.groupadd("tgroup")
+
+    shadow.fs.rm("/etc/gshadow")
+
+    shadow.groupdel("tgroup")
+
+    group_entry = shadow.tools.getent.group("tgroup")
+    assert group_entry is None, "Group should not be found"
+    assert not shadow.fs.exists("/etc/gshadow"), "/etc/gshadow file should not be found"
--- a/tests/system/tests/test_groupdel.py
+++ b/tests/system/tests/test_groupdel.py
@@ -33,3 +33,34 @@ def test_groupdel__delete_group(shadow: Shadow):
 
     gshadow_entry = shadow.tools.getent.gshadow("tgroup")
     assert gshadow_entry is None, "Group should not be found"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_groupdel__delete_group_no_gshadow_entry(shadow: Shadow):
+    """
+    :title: Group deletion with no gshadow entry
+    :setup:
+        1. Create group
+        2. Remove gshadow entry manually
+    :steps:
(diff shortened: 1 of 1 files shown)


tests/system/tests/test_groupdel.py
```diff
@@ -33,3 +33,34 @@ def test_groupdel__delete_group(shadow: Shadow):
...
+def test_groupdel__delete_group_no_gshadow_entry(shadow: Shadow):
+    """
+    :title: Group deletion with no corresponding gshadow entry
+    :setup:
+        1. Create group
+        2. Remove gshadow entry manually
+    :steps:
+        1. Check group entry
+        2. Check gshadow entry
+        3. Delete group
```
ikerexxe: The order is different: first you delete the group, and then you check the entries

tests/system/tests/test_groupdel.py
```diff
@@ -33,3 +33,34 @@ def test_groupdel__delete_group(shadow: Shadow):
...
+    :expectedresults:
+        1. Group entry is found
+        2. No gshadow entry is found
+        3. Group is deleted
```
ikerexxe: Same here

tests/system/tests/test_groupdel.py
```diff
@@ -64,3 +64,31 @@ def test_groupdel__delete_group_no_gshadow_entry(shadow: Shadow):
...
+def test_groupdel__delete_group_no_gshadow_file(shadow: Shadow):
+    """
+    :title: Group deletion with no gshadow file
+    :setup:
+        1. Create group
+        2. Remove gshadow file
+    :steps:
+        1. Delete group
+        2. Check group entry
+        3. Verify that gshadow file doesn't exist
```
ikerexxe, line 79: You aren't checking this step in the code. Do we need the step? Or are you missing the code?

ikerexxe (commented):
Minor comments inline

tests/system/tests/test_groupdel.py
asakure: I thought `assert` would be sufficient but I have removed it as it doesn't seem to be necessary.

asakure: @ikerexxe as I checked further, in existing bash test we have `test ! -f /etc/gshadow` to confirm that `gshadow` file doesn't exist.

ikerexxe: Then, let's add the check. I think you can use `shadow.fs.exists()`

tests/system/tests/test_groupdel.py
```diff
@@ -33,3 +33,59 @@ def test_groupdel__delete_group(shadow: Shadow):
...
+def test_groupdel__delete_group_no_gshadow_entry(shadow: Shadow):
+    """
+    :title: Group deletion with no corresponding gshadow entry
+    :setup:
+        1. Create group
+        2. Remove gshadow entry manually
+    :steps:
+        1. Delete group
+        1. Check group entry
+        2. Check gshadow entry
```
ikerexxe: Step numbering is wrong, please update it accordingly

tests/system/tests/test_groupdel.py
```diff
@@ -33,3 +33,59 @@ def test_groupdel__delete_group(shadow: Shadow):
...
+    :customerscenario: False
+    """
+    shadow.groupadd("tgroup")
+
+    shadow.fs.sed("/^tgroup:/d", "/etc/gshadow")
+
+    shadow.groupdel("tgroup")
+
+    group_entry = shadow.tools.getent.group("tgroup")
+    assert group_entry is None, "Group should not be found"
```
ikerexxe, line 62: Update the assertion message as the group entry should be found

tests/system/tests/test_groupdel.py
```diff
@@ -33,3 +33,59 @@ def test_groupdel__delete_group(shadow: Shadow):
...
+    """
+    :title: Group deletion with no gshadow file
+    :setup:
+        1. Create group
+        2. Remove gshadow file
+    :steps:
+        1. Delete group
+        2. Check group entry
+    :expectedresults:
+        1. Group entry is deleted successfully
```
ikerexxe: I'd remove `entry`

tests/system/tests/test_groupdel.py
asakure: My bad, indeed group entry should not be found as we delete it in previous step. I corrected `:expectedresults:`

ikerexxe (approved):
LGTM!

PR #1714, opened 2026-08-07
man: Document UAPI merging in login.defs

### LINKED ISSUE #1712: Document UAPI in login.defs
https://github.com/shadow-maint/shadow/commit/b52ce71c added support for a vendor directory.

Right now, `man 5 login.defs` and various utility manpages (useradd, passwd, ..) state that /etc/login.defs is the configuration file. 

AFAIK openSUSE and [Fedora](https://src.fedoraproject.org/rpms/shadow-utils/blob/rawhide/f/shadow-utils.spec#_45) both use libeconf so for those just stating `/etc/login.defs` wouldn't be true.

We could change the man pages to be compile flag dependant but I would suggest to just document the situation in plain text.

In `man/login.defs.5.xml` we would have to remove:

https://github.com/shadow-maint/shadow/blob/8c1e4b96c0cc9deea7c735985237834744dd40dc/man/login.defs.5.xml#L118-L121

with something like:

```
The configuration for the shadow password suite is defined by login.defs. Depending on configuration the settings are either read from /etc/login.defs or, if it is compiled with libeconf, merged hierarchically following the UAPI configuration file specification.
```

And then maybe go on and list the hierarchical directories.

In the individual man pages we could either just leave https://github.com/shadow-maint/shadow/blob/8c1e4b96c0cc9deea7c735985237834744dd40dc/man/chfn.1.xml#L190 or or replace it with `login.defs` without the `/etc`. Maybe expanding the https://github.com/shadow-maint/shadow/blob/8c1e4b96c0cc9deea7c735985237834744dd40dc/man/chfn.1.xml#L192 with `See login.defs for details on the configuration format and search paths`.

Before opening a PR I thought its best to create this issue to hear what others think.

jubalh:
b52ce71c added support for UAPI/libeconf.
The first release to contain this change was 4.8.

Update our man pages to properly describe the various locations of login.defs if shadow is compiled with libeconf and without.

Ref: b52ce71c
Resolves: https://bugzilla.suse.com/show_bug.cgi?id=1264297
Fixes: https://github.com/shadow-maint/shadow/issues/1712

### DIFF
--- a/man/chfn.1.xml
+++ b/man/chfn.1.xml
@@ -187,7 +187,11 @@
     <title>FILES</title>
     <variablelist>
       <varlistentry>
-	<term><filename>/etc/login.defs</filename></term>
+	<term>
+	  <citerefentry>
+	    <refentrytitle>login.defs</refentrytitle><manvolnum>5</manvolnum>
+	  </citerefentry>
+	</term>
 	<listitem>
 	  <para>Shadow password suite configuration.</para>
 	</listitem>
--- a/man/login.defs.5.xml
+++ b/man/login.defs.5.xml
@@ -115,10 +115,23 @@
   <refsect1 id='description'>
     <title>DESCRIPTION</title>
     <para>
-      The <filename>/etc/login.defs</filename> file defines the
-      site-specific configuration for the shadow password suite. This file
-      is required. Absence of this file will not prevent system operation,
-      but will probably result in undesirable operation.
+      The configuration for the shadow password suite is defined by
+      <filename>login.defs</filename>.
+    </para>
+    <para>
+      Depending on compile-time configuration,
+      the settings are either read from a single file
+      (<filename>/etc/login.defs</filename>),
+      or merged hierarchically
+      (if compiled with <literal>libeconf</literal> support):
+      from <filename>/usr/etc/login.defs</filename>,
+      <filename>/usr/lib/login.defs</filename>,
+      <filename>/etc/login.defs</filename>,
+      and drop-ins under <filename>/etc/login.defs.d/</filename>.
+    </para>
+    <para>
+      The merging follows the UAPI configuration file specification:
+      https://uapi-group.org/specifications/specs/configuration_files_specification/
     </para>
 
     <para>
--- a/man/login.defs.5.xml
+++ b/man/login.defs.5.xml
@@ -115,10 +115,49 @@
   <refsect1 id='description'>
     <title>DESCRIPTION</title>
     <para>
-      The <filename>/etc/login.defs</filename> file defines the
-      site-specific configuration for the shadow password suite. This file
-      is required. Absence of this file will not prevent system operation,
-      but will probably result in undesirable operation.
+      The configuration for the shadow password suite is defined by
+      <filename>login.defs</filename>.
+    </para>
+    <para>
+      Depending on compile-time configuration, the settings are either read
+      from a single file (<filename>/etc/login.defs</filename>), or merged
+      hierarchically following the UAPI configuration file specification
+      (if compiled with <literal>libeconf</literal> support):
+      https://uapi-group.org/specifications/specs/configuration_files_specification/
+    </para>
+    <para>
+      When the stateless merging model is used, the configurations are parsed
+      and merged in the following order of precedence (later files override
+      earlier ones):
+    </para>
+    <itemizedlist>
+      <listitem>
+        <para>
+          <emphasis>Vendor defaults:</emphasis> typically located in
+          <filename>/usr/etc/login.defs</filename> or <filename>/usr/lib/login.defs</filename>.
+          These files are owned by the distribution packages and should not be
+          manually modified.
+        </para>
+      </listitem>
+      <listitem>
+        <para>
+          <emphasis>Local admin overrides:</emphasis> located in
+          <filename>/etc/login.defs</filename>. Administrators should use this file
+          to override vendor presets.
+        </para>
+      </listitem>
+      <listitem>
+        <para>
+          <emphasis>Admin drop-ins:</emphasis> located in
+          <filename>/etc/login.defs.d/</filename>. This directory allows for modular,
+          package-specific overrides without editing the main file.
+        </para>
+      </listitem>
+    </itemizedlist>
+    <para>
+      Neither <filename>/etc/login.defs</filename> nor the drop-in directories are
+      strictly required. If they are absent, the shadow utilities will seamlessly fall
+      back to the vendor-provided defaults.
     </para>
 
     <para>
--- a/man/chgpasswd.8.xml
+++ b/man/chgpasswd.8.xml
@@ -218,7 +218,11 @@
 	</listitem>
       </varlistentry>
       <varlistentry>
-	<term><filename>/etc/login.defs</filename></term>
+	<term>
+	  <citerefentry>
+	    <refentrytitle>login.defs</refentrytitle><manvolnum>5</manvolnum>
+	  </citerefentry>
+	</term>
 	<listitem>
 	  <para>Shadow password suite configuration.</para>
 	</listitem>
--- a/man/chpasswd.8.xml
+++ b/man/chpasswd.8.xml
@@ -272,7 +272,11 @@
 	</listitem>
       </varlistentry>
       <varlistentry>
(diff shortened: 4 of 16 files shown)


man/chfn.1.xml
```diff
@@ -187,7 +187,11 @@
     <title>FILES</title>
     <variablelist>
       <varlistentry>
-	<term><filename>/etc/login.defs</filename></term>
+	<term>
+	  <citerefentry>
+	    <refentrytitle>login.defs</refentrytitle><manvolnum>5</manvolnum>
+	  </citerefentry>
+	</term>
 	<listitem>
```
alejandro-colomar, line 195: From the commit message:

```
Ref: b52ce71c
Resolves: https://bugzilla.suse.com/show_bug.cgi?id=1264297
Fixes: https://github.com/shadow-maint/shadow/issues/1712
```

'Ref' should be 'Fixes'.
'Fixes' should be 'Closes'.
'Resolves' should also be 'Closes'.

alejandro-colomar: And the URIs should be wrapped in `<>`

alejandro-colomar: The Fixes (currently 'Ref') field should contain the commit date and subject too.

See <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/git#n46> and <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/patches/trailer#n16>.

man/login.defs.5.xml
```diff
@@ -115,10 +115,49 @@
...
+      The configuration for the shadow password suite is defined by
+      <filename>login.defs</filename>.
+    </para>
+    <para>
+      Depending on compile-time configuration, the settings are either read
+      from a single file (<filename>/etc/login.defs</filename>), or merged
+      hierarchically following the UAPI configuration file specification
+      (if compiled with <literal>libeconf</literal> support):
+      https://uapi-group.org/specifications/specs/configuration_files_specification/
+    </para>
```
alejandro-colomar: Please use semantic newlines.  See man-pages(7):

```
   Use semantic newlines
```

man/login.defs.5.xml
```diff
@@ -115,10 +115,49 @@
...
+      <listitem>
+        <para>
+          <emphasis>Vendor defaults:</emphasis> typically located in
+          <filename>/usr/etc/login.defs</filename> or <filename>/usr/lib/login.defs</filename>.
+          These files are owned by the distribution packages and should not be
+          manually modified.
+        </para>
+      </listitem>
+      <listitem>
+        <para>
```
alejandro-colomar: This seems like documenting the usual behavior of libeconf, right?  Nothing specific of this project.  If so, would it be possible to just refer to some libeconf manual page (or other documentation)?

(I don't know much of libeconf, so I might be wrong.)

man/chfn.1.xml
jubalh: > 'Ref' should be 'Fixes'.

Should it really? I think "fix" is a bit hard word to just add documentation for what was done in that commit. In my POV Ref is more appropriate.

> 'Resolves' should also be 'Closes'.
In 903593249630054ab5df327481f7386f718088cc the Resolves is used for an bugzilla link.

jubalh: > And the URIs should be wrapped in `<>`

In 9035932, 99c4f4 and various other commits I see http used without `<>`.

alejandro-colomar: > > And the URIs should be wrapped in `<>`
> 
> In [9035932](https://github.com/shadow-maint/shadow/commit/903593249630054ab5df327481f7386f718088cc), 99c4f4 and various other commits I see http used without `<>`.

Yup, consistency isn't perfect.

See uri(7) (first sentence):

```
$ man uri | awk '/Writing a URI/,/^$/'
   Writing a URI
     When  written,  URIs  should  be  placed  inside  double   quotes   (e.g.,
     "http://www.kernel.org"),    enclosed    in    angle    brackets    (e.g.,
     <http://lwn.net>), or placed on a line by themselves.  A warning for those
     who use double‐quotes: never move extraneous punctuation (such as the  pe‐
     riod  ending  a  sentence or the comma in a list) inside a URI, since this
     will change the value of the URI.  Instead, use angle brackets instead, or
     switch to a quoting system that never includes extraneous  characters  in‐
     side  quotation  marks.  This latter system, called the ’new’ or ’logical’
     quoting system by "Hart’s Rules" and the "Oxford  Dictionary  for  Writers
     and  Editors", is preferred practice in Great Britain and in various Euro‐
     pean languages.  Older documents suggested  inserting  the  prefix  "URL:"
     just before the URI, but this form has never caught on.

```

alejandro-colomar: > > 'Ref' should be 'Fixes'.
> 
> Should it really? I think "fix" is a bit hard word to just add documentation for what was done in that commit. In my POV Ref is more appropriate.

If you prefer `Ref` I guess it's okay.  I don't see `Fixes` as to hard; it just addresses something that should have been done in that commit.  It doesn't necessarily mean that it's a vulnerability or something too bad, IMO.  But I don't mind too much.

man/login.defs.5.xml
jubalh: AFAIK there is no man but I changed it to `The merging follows the UAPI configuration file specification` with a link to the file now.

jubalh:
@alejandro-colomar please give it another go :) I hope I changed everything as requested.

man/chfn.1.xml
```diff
@@ -187,7 +187,11 @@
     <title>FILES</title>
     <variablelist>
       <varlistentry>
-	<term><filename>/etc/login.defs</filename></term>
+	<term>
+	  <citerefentry>
```
alejandro-colomar, line 191: From the commit message:

```
Ref: b52ce71c (2019-10-05; \"Add support for a vendor directory and libeconf\")
```

Those `\` should not be there.

man/login.defs.5.xml
```diff
@@ -115,10 +115,23 @@
...
+      or merged hierarchically
+      from <filename>/usr/etc/login.defs</filename>,
+      <filename>/usr/lib/login.defs</filename>,
+      <filename>/etc/login.defs</filename>,
+      and drop-ins under <filename>/etc/login.defs.d/</filename>.
+    </para>
+    <para>
+      The merging follows the UAPI configuration file specification
+      (if compiled with <literal>libeconf</literal> support):
+      https://uapi-group.org/specifications/specs/configuration_files_specification/
```
alejandro-colomar, line 134: The parenthetical reads as if merging follows the UAPI specification if configured with libeconf support, and doesn't follow it but still is possible if configured without libeconf support.

I think that parenthetical belongs  in the first paragraph, right after 'merged hierarchically'.

man/chfn.1.xml
jubalh: done

man/login.defs.5.xml
jubalh: Not sure if I understood but I updated the commit, let me know if its better.

alejandro-colomar: Yup, that's right.

man/login.defs.5.xml
```diff
@@ -115,10 +115,23 @@
...
+    <para>
+      Depending on compile-time configuration,
+      the settings are either read from a single file
+      (<filename>/etc/login.defs</filename>),
+      or merged hierarchically
+      (if compiled with <literal>libeconf</literal> support):
+      from <filename>/usr/etc/login.defs</filename>,
+      <filename>/usr/lib/login.defs</filename>,
+      <filename>/etc/login.defs</filename>,
+      and drop-ins under <filename>/etc/login.defs.d/</filename>.
```
alejandro-colomar, line 130: The `:` should have been kept before the link in the second paragraph.

jubalh: Not sure what you mean. "The merging follows the UAPI configuration file specification" should have an ":"?

alejandro-colomar: Yes.

jubalh: done

alejandro-colomar (approved):
Thanks!

```
```

PR #1716, opened 2026-08-10
useradd: Only apply USRSKEL if the default SKEL is used

jubalh:
Historically useradd(8) copied skeleton files from the directory specified by SKEL.
Since 74c17c7 (release 4.14.0) it also copies files from USRSKEL.

If an admin explicitly specifies an alternative skeleton directory using -k (for instance, to create an empty home directory with `-k /var/lib/empty`), the files from USRSKEL are still unconditionally merged in.

If the user has explicitly requested an alternative skeleton directory we shouldn't silently merge in files from USRSKEL behind their back.

However if they specify `-k /etc/skel` (explicitly requesting the default skeleton), they most likely still expect USRSKEL to be merged.

With this commit we only copy the USRSKEL files when the selected skeleton directory matches the default.

Resolves: https://bugzilla.suse.com/show_bug.cgi?id=1191490
Fixes: 74c17c7 (2022-11-09; "Add support for skeleton files from /usr/etc/skel")

### DIFF
--- a/src/useradd.c
+++ b/src/useradd.c
@@ -2469,6 +2469,7 @@ int main (int argc, char **argv)
 	unsigned long subgid_count = 0;
 	struct option_flags  flags = {.chroot = false, .prefix = false};
 	bool process_selinux;
+	const char *default_skel;
 
 	log_set_progname(Prog);
 	log_set_logfd(stderr);
@@ -2500,6 +2501,8 @@ int main (int argc, char **argv)
 #endif
 
 	get_defaults (&flags);
+
+	default_skel = def_template;
 
 	process_flags (argc, argv, &flags);
 	process_selinux = !flags.chroot && !flags.prefix;
@@ -2660,8 +2663,10 @@ int main (int argc, char **argv)
 		if (home_added) {
 			copy_tree (def_template, prefix_user_home, false,
 			           (uid_t)-1, user_id, (gid_t)-1, user_gid);
-			copy_tree (def_usrtemplate, prefix_user_home, false,
-			           (uid_t)-1, user_id, (gid_t)-1, user_gid);
+			if (streq(def_template, default_skel)) {
+				copy_tree(def_usrtemplate, prefix_user_home, false,
+				           (uid_t)-1, user_id, (gid_t)-1, user_gid);
+			}
 		} else {
 			eprintf(_("%s: warning: the home directory %s already exists.\n"
 			          "%s: Not copying any file from skel directory into it.\n"),


src/useradd.c
```diff
@@ -2469,6 +2469,7 @@ int main (int argc, char **argv)
 	unsigned long subgid_count = 0;
 	struct option_flags  flags = {.chroot = false, .prefix = false};
 	bool process_selinux;
+	const char *default_skel = NULL;
```
alejandro-colomar: This `NULL` seems to be dead code, isn't it?

jubalh: I followed the initialization like done for `subgid_count`.

jubalh: changed.

alejandro-colomar: > I followed the initialization like done for `subgid_count`.

Initialization is bad, and hides bugs.  Yes, much existing code does this, but it's something that we shouldn't follow in new code.

jubalh: Usually people say "follow the existing coding style". And I don't see how this hides bugs. It will be optimized by the compiler anyways.

alejandro-colomar: > Usually people say "follow the existing coding style".

I don't categorize initialization as a style (since it's not just cosmetic), but rather as a bad programming practice.

> And I don't see how this hides bugs.

Static analyzers are able to detect paths that don't handle something that they should.  This is often done by detecting variables used but not set, and variables set but not used.

By initializing a variable unconditionally, static analyzers can't detect whether it is used but not set, because it would be always set.  They must assume that you want to use that value, even if you don't.

src/useradd.c
```diff
@@ -2660,8 +2663,10 @@ int main (int argc, char **argv)
 		if (home_added) {
 			copy_tree (def_template, prefix_user_home, false,
 			           (uid_t)-1, user_id, (gid_t)-1, user_gid);
-			copy_tree (def_usrtemplate, prefix_user_home, false,
-			           (uid_t)-1, user_id, (gid_t)-1, user_gid);
+			if (streq (def_template, default_skel)) {
+				copy_tree (def_usrtemplate, prefix_user_home, false,
+				           (uid_t)-1, user_id, (gid_t)-1, user_gid);
+			}
```
alejandro-colomar: Please remove the spaces after the function name.  We're avoiding those spaces in new code.

alejandro-colomar: Also, please put a space after the casts, while at it.

See the style guide we're using: <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/style/c#n60>.

```
    Spaces
	Treat sizeof() and similar operators as functions, not keywords.
	Use a space after keywords, but not after functions.

	Use a space to separate binary and ternary operators (except
	`.` and `->`), but not to separate unary operators.

	Use a space between a cast and the expression it converts.
```

jubalh: done

alejandro-colomar:
Thanks!  The source code LGTM.

I prefer if @hallyn could have a look at it too, since he was involved in the original patch (the one we're fixing).

hallyn:
On Tue, Aug 18, 2026 at 03:11:09AM -0700, Alejandro Colomar wrote:
> alejandro-colomar left a comment (shadow-maint/shadow#1716)
> 
> Thanks!  The source code LGTM.
> 
> I prefer if @hallyn could have a look at it too, since he was involved in the original patch (the one we're fixing).

How about just checking kflg?  I know the user could enter
the default as kflg option, but then who knows what they
really meant in that case anyway.

I mainly think checking kflg will be easier to understand
to future readers.

alejandro-colomar:
> On Tue, Aug 18, 2026 at 03:11:09AM -0700, Alejandro Colomar wrote: alejandro-colomar left a comment ([shadow-maint/shadow#1716](https://github.com/shadow-maint/shadow/pull/1716)) Thanks! The source code LGTM. I prefer if @hallyn could have a look at it too, since he was involved in the original patch (the one we're fixing).
> How about just checking kflg? I know the user could enter the default as kflg option, but then who knows what they really meant in that case anyway. I mainly think checking kflg will be easier to understand to future readers.

I'm not sure I understand.  Would you mind showing some diff?  (Or maybe using more/different words to clarify it?)

hallyn:
> > On Tue, Aug 18, 2026 at 03:11:09AM -0700, Alejandro Colomar wrote: alejandro-colomar left a comment ([shadow-maint/shadow#1716](https://github.com/shadow-maint/shadow/pull/1716)) Thanks! The source code LGTM. I prefer if @hallyn could have a look at it too, since he was involved in the original patch (the one we're fixing).
> > How about just checking kflg? I know the user could enter the default as kflg option, but then who knows what they really meant in that case anyway. I mainly think checking kflg will be easier to understand to future readers.
> 
> I'm not sure I understand. Would you mind showing some diff? (Or maybe using more/different words to clarify it?)

Well, I see that he explicitly addressed it in the commit message.

I'm really not sure what the right behavior is here.  But since USRSKEL was added by him for openSUSE explicitly, I suppose he knows best :) Let's just merge it.

hallyn (approved):

PR #1491, opened 2026-01-10
lib/: Use the comma operator to perform lvalue conversion

alejandro-colomar:
-  Queued after <https://github.com/shadow-maint/shadow/pull/1717>.

---

Compound literals are lvalues, and thus somewhat dangerous.  Their address can be taken, and they can be assigned to.

We were using statement expressions to perform lvalue conversion on compound literals, transforming them to rvalues, and thus removing their dangers.  However, statement expressions are non-standard, and quite complex within the compiler, so it would be interesting to use simpler compiler features to achieve the same.

The comma operator also performs lvalue conversion, and we can use a dummy (void)0 expression to introduce it.  This is significantly simpler, and is more portable than the statement expression (it is valid all the way back to ANSI C89).

By using a simpler feature, we have a smaller risk of running into a compiler bug.

Suggested-by: @uecker
Cc: @chrisbazley
Cc: @kees
Cc: @flatcap

---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase

```
$ git rd 
1:  9558736a5b42 = 1:  5732c14f453e lib/: Use the comma operator to perform lvalue conversion
```
</details>

<details>
<summary>v1c</summary>

-  Rebase

```
$ git rd 
1:  5732c14f = 1:  8956e27a lib/: Use the comma operator to perform lvalue conversion
```
</details>

<details>
<summary>v1d</summary>

-  Rebase

```
$ git rd 
1:  8956e27a = 1:  5a03c618 lib/: Use the comma operator to perform lvalue conversion
```
</details>

<details>
<summary>v1e</summary>

-  Clarify portability.

```
$ git rd
1:  5a03c618 ! 1:  b7ab3f97 lib/: Use the comma operator to perform lvalue conversion
    @@ Commit message
     
         The comma operator also performs lvalue conversion, and we can use
         a dummy (void)0 expression to introduce it.  This is significantly
    -    simpler, and is more portable than the statement expression (it is valid
    -    all the way back to ANSI C89).
    +    simpler, and is more portable than the statement expression: it is valid
    +    all the way back to C99 (the comma operator and the (void)0 expression
    +    are portable to C89, but the compound literal is from C99).
     
         By using a simpler feature, we have a smaller risk of running into
         a compiler bug.
```
</details>

<details>
<summary>v1f</summary>

-  Rebase

```
$ git rd 
1:  b7ab3f97a0c8 = 1:  617ded6a961c lib/: Use the comma operator to perform lvalue conversion
```
</details>

<details>
[...]

PR #1717, opened 2026-08-20
lib/: Use the comma operator to perform lvalue conversion

alejandro-colomar:
Compound literals are lvalues, and thus somewhat dangerous.  Their address can be taken, and they can be assigned to.

We were using statement expressions to perform lvalue conversion on compound literals, transforming them to rvalues, and thus removing their dangers.  However, statement expressions are non-standard, and quite complex within the compiler, so it would be interesting to use simpler compiler features to achieve the same.

The comma operator also performs lvalue conversion, and we can use a dummy (void)0 expression to introduce it.  This is significantly simpler, and is more portable than the statement expression: it is valid all the way back to C99 (the comma operator and the (void)0 expression are portable to C89, but the compound literal is from C99).

By using a simpler feature, we have a smaller risk of running into a compiler bug.

Suggested-by: @uecker 
Cc: @chrisbazley
Cc: @kees
Cc: @flatcap

---

-  This is a subset of <https://github.com/shadow-maint/shadow/pull/1491> at v2g, to make it easier to review.

---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase

```
$ git rd 
1:  a5953e443b47 = 1:  b24b07bf2a36 lib/: Use the comma operator to perform lvalue conversion
```
</details>

### DIFF
--- a/lib/alloc/calloc.h
+++ b/lib/alloc/calloc.h
@@ -17,9 +17,10 @@
 // calloc_T - calloc type-safe
 #define calloc_T(n, T)   calloc_T_(n, typeas(T))
 #define calloc_T_(n, T)                                               \
-({                                                                    \
-	(T *){calloc(n, sizeof(T))};                                  \
-})
+(                                                                     \
+	(void)0,                                                      \
+	(T *){calloc(n, sizeof(T))}                                   \
+)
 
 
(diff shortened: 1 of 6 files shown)


alejandro-colomar:
-  Superseded by #1721

PR #1718, opened 2026-08-21
Tets: Fix indentation in test_groupadd.py

asakure:
This patch fixes the indentation issue introduced by python tranformation of bash test `06_groupadd_-f_add_existing_group` to keep code aligned with other python tests in test_groupadd.py


ikerexxe (approved):
LGTM

PR #1719, opened 2026-08-21
Fix Debian CI.

silverhadch:
make install doesn't want to override /sbin/nologin so we shall force it.

### DIFF
--- a/share/ansible/roles/ci_run/tasks/debian.yml
+++ b/share/ansible/roles/ci_run/tasks/debian.yml
@@ -72,11 +72,34 @@
     chdir: /usr/local/src/shadow/
   ignore_errors: true
 
-- name: Install
-  ansible.builtin.command:
-    make install
+# make install fails with "File exists" on /sbin/nologin, which the base image
+# already provides; install into a staging tree and apply it by hand instead.
+- name: Install into a staging tree
+  ansible.builtin.shell: |
+    rm -rf /tmp/rootfs
+    make install DESTDIR=/tmp/rootfs > install.log 2>&1
   args:
     chdir: /usr/local/src/shadow/
+  ignore_errors: true
+
+# The targets are removed before extracting: tar(1) fails with ETXTBSY when it
+# overwrites a running binary in place.  -U (--unlink-first) would do that for
+# us, but it also unlinks the merged-/usr symlinks (/bin, /sbin, /lib) that
+# --keep-directory-symlink is here to preserve, so do it explicitly.
+- name: Apply the staging tree to the system
+  ansible.builtin.shell: |
+    exec > apply.log 2>&1
+    set -euxo pipefail
+    find /tmp/rootfs -not -type d -printf '/%P\n' \
+    | sort > install-manifest.txt
+    xargs -d '\n' -r rm -f -- < install-manifest.txt
+    tar -C /tmp/rootfs -cf - . \
+    | tar -C / -xpf - --keep-directory-symlink --no-overwrite-dir
+    ldconfig
+    test -x /bin/sh
+  args:
+    chdir: /usr/local/src/shadow/
+    executable: /bin/bash
   ignore_errors: true
 
 - name: Copy logs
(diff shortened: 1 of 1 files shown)


share/ansible/roles/ci_run/tasks/debian.yml
```diff
@@ -49,49 +49,85 @@
...
+  failed_when: false
 
-- name: Install
-  ansible.builtin.command:
-    make install
+# --keep-directory-symlink preserves merged-/usr symlinks (/bin, /sbin, /lib)
+- name: Apply the staging tree to the system
+  ansible.builtin.shell: |
+    set -euo pipefail
+    find /tmp/rootfs \( -type f -o -type l \) \
```
alejandro-colomar: Can we simplify this (drop those escaped parentheses) with `-not -type d`?  Or do we really need this?

share/ansible/roles/ci_run/tasks/debian.yml
```diff
@@ -49,49 +49,88 @@
...
-- name: Install
-  ansible.builtin.command:
-    make install
+# --keep-directory-symlink preserves merged-/usr symlinks (/bin, /sbin, /lib);
+# do NOT add tar -U, it defeats --keep-directory-symlink
+- name: Apply the staging tree to the system
+  ansible.builtin.shell: |
+    exec > apply.log 2>&1
+    set -euxo pipefail
+    find /tmp/rootfs ! -type d \
```
alejandro-colomar: `-not` is better than `!`, as it's POSIX-compliant, and it also avoids a character that may have issues with the shell (think of pasting the command in an interactive shell).

ikerexxe (commented):
I think this PR is mixing two separate improvements that should be handled differently:

1. The main fix (addresses the immediate issue):
- DESTDIR staging to avoid `/sbin/nologin` collision
- The new tar-based installation logic
- Adding the missing configuration files

2. Error handling modernization (general improvement):
- Switching from `ignore_errors: true` to `register + failed_when: false`
- Adding `stderr` redirection to log captures
- The deferred failure checking at the end

Suggested approach:
- Separate commit for the error handling improvements, since they're valuable but unrelated to the collision fix
- Port the error handling changes to other distributions: fedora.yml, alpine.yml and opensuse.yml. They would benefit from the same improved logging and failure detection

This would make the changes easier to review and test.

share/ansible/roles/ci_run/tasks/debian.yml
silverhadch: Actually the other way around.
https://pubs.opengroup.org/onlinepubs/9799919799/utilities/find.html

```
! expression
    Negation of a primary; the unary NOT operator.
```
is POSIX.

alejandro-colomar: Oh, then the page I read was wrong, or I misread it.  I can't find it now.  :|

So, yeah, if POSIX blessed `!`, let's go for it.

share/ansible/roles/ci_run/tasks/debian.yml
```diff
@@ -72,11 +72,34 @@
     chdir: /usr/local/src/shadow/
   ignore_errors: true
 
-- name: Install
-  ansible.builtin.command:
-    make install
```
alejandro-colomar, line 77: > make install fails with "File exists" on /sbin/nologin

Why does it fail?  Doesn't `make install` overwrite whatever is on the system?  That's what I'd have expected.

ikerexxe: I second these questions. If `make install` is failing we should fix the root cause in the Makefile itself rather than adding workarounds in the Ansible role. Fixing it at the build level keeps the toolchain clean and independent

silverhadch: Seems like it just fixed itself lol.

silverhadch:
No longer required.

PR #1721, opened 2026-08-22
rvalue()

alejandro-colomar:
Compound literals are lvalues, and thus somewhat dangerous.  Their address can be taken, and they can be assigned to.

We were using statement expressions to perform lvalue conversion on compound literals, transforming them to rvalues, and thus removing their dangers.  However, statement expressions are non-standard, and quite complex within the compiler, so it would be interesting to use simpler compiler features to achieve the same.

The comma operator also performs lvalue conversion, and we can use a dummy (void)0 expression to introduce it.  This is significantly simpler, and is more portable than the statement expression: it is valid all the way back to C99 (the comma operator and the (void)0 expression are portable to C89, but the compound literal is from C99).

By using a simpler feature, we have a smaller risk of running into a compiler bug.

Suggested-by: @uecker 
Cc: @chrisbazley
Cc: @kees

---

-  This PR supersedes #1717.

---

Revisions:

<details>
<summary>v1b</summary>

-  Rebase.
-  Add link in commit message.

```
$ git rd 
1:  ce2379cc29a1 = 1:  98d3200f5c25 lib/cast.h: rvalue(): Add macro for performing lvalue conversion
2:  711e15b8c147 = 2:  e120ea45e39f lib/: Use rvalue() instead of its pattern
3:  86b395f375f1 ! 3:  3ec82c67cd03 lib/cast.h: rvalue(): Use the comma operator to perform lvalue conversion
    @@ Commit message
         By using a simpler feature, we have a smaller risk of running into
         a compiler bug.
     
    +    Link: <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88773>
         Suggested-by: Martin Uecker <uecker@tugraz.at>
         Cc: Christopher Bazley <chris.bazley@arm.com>
         Cc: Kees Cook <kees@kernel.org>
```
</details>


alejandro-colomar:
As predicted, using the comma operator to get an rvalue seems better.  Here's a GCC bug with statement expressions:

<https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88773>

PR #1732, opened 2026-09-07

asakure:
existing bash tests with numbering 3.

### DIFF
--- a/tests/system/tests/test_groupadd.py
+++ b/tests/system/tests/test_groupadd.py
@@ -700,3 +700,32 @@ def test_groupadd__no_system_gids_available(shadow: Shadow):
     if shadow.host.features["gshadow"]:
         gshadow_entry = shadow.tools.getent.gshadow("sgroup3")
         assert gshadow_entry is None, "System group should not be found"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_groupadd__override_GID_MIN(shadow: Shadow):
+    """
+    :title: Group creation respects GID_MIN override
+    :setup:
+        1. None required
+    :steps:
+        1. Create group with GID_MIN override
+        2. Check group and gshadow entries
+        3. Check GID for group
+    :expectedresults:
+        1. Group is created
+        2. Group and gshadow entries are found
+        3. Group entry is found with GID = 2000
+    :customerscenario: False
+    """
+    shadow.groupadd("-K GID_MIN=2000 tgroup")
+
+    group_entry = shadow.tools.getent.group("tgroup")
+    assert group_entry is not None, "Group should be found"
+    assert group_entry.name == "tgroup", "Incorrect groupname"
+    assert group_entry.gid == 2000, "Incorrect GID"
+
+    if shadow.host.features["gshadow"]:
+        gshadow_entry = shadow.tools.getent.gshadow("tgroup")
+        assert gshadow_entry is not None, "Group should be found in gshadow"
+        assert gshadow_entry.name == "tgroup", "Incorrect groupname"


tests/system/tests/test_groupadd.py
```diff
@@ -700,3 +700,33 @@ def test_groupadd__no_system_gids_available(shadow: Shadow):
     if shadow.host.features["gshadow"]:
         gshadow_entry = shadow.tools.getent.gshadow("sgroup3")
         assert gshadow_entry is None, "System group should not be found"
+
+
+@pytest.mark.topology(KnownTopology.Shadow)
+def test_groupadd__override_GID_MIN(shadow: Shadow):
+    """
+    :title: Group creation respects -K GID_MIN override
```
ikerexxe: No need to mention `-K` in the title

tests/system/tests/test_groupadd.py
```diff
@@ -700,3 +700,33 @@ def test_groupadd__no_system_gids_available(shadow: Shadow):
...
+        3. Group entry is found with GID >= 2000
+    :customerscenario: False
+    """
+    shadow.groupadd("-K GID_MIN=2000 tgroup")
+
+    group_entry = shadow.tools.getent.group("tgroup")
+    assert group_entry is not None, "Group should be found"
+    assert group_entry.name == "tgroup", "Incorrect groupname"
+    assert group_entry.gid is not None, "Group should have a GID"
+    assert group_entry.gid >= 2000, "Group GID should be >= GID_MIN (2000)"
```
ikerexxe: Let's update the assert and the message to mimic the original test case with `==` instead of `>=`

tests/system/tests/test_groupadd.py
```diff
@@ -700,3 +700,33 @@ def test_groupadd__no_system_gids_available(shadow: Shadow):
...
+    :setup:
+        1. None required
+    :steps:
+        1. Create group with GID_MIN override
+        2. Check group and gshadow entries
+        3. Check GID for group
+    :expectedresults:
+        1. Group is created
+        2. Group and gshadow entries are found
+        3. Group entry is found with GID >= 2000
```
ikerexxe: You should also update this comment

tests/system/tests/test_groupadd.py
```diff
@@ -700,3 +700,33 @@ def test_groupadd__no_system_gids_available(shadow: Shadow):
...
+    group_entry = shadow.tools.getent.group("tgroup")
+    assert group_entry is not None, "Group should be found"
+    assert group_entry.name == "tgroup", "Incorrect groupname"
+    assert group_entry.gid == 2000, "Incorrect GID"
+
+    if shadow.host.features["gshadow"]:
+        gshadow_entry = shadow.tools.getent.gshadow("tgroup")
+        assert gshadow_entry is not None, "Group should be found in gshadow"
+        assert gshadow_entry.name == "tgroup", "Incorrect groupname"
+        assert gshadow_entry.password == "!", "Incorrect password"
```
ikerexxe: Why did you add this line? I don't remember seeing it before

asakure: It's there in first `groupadd` test https://github.com/shadow-maint/shadow/blob/3d4f43c70a98ed2eea80e6fe94785a23d1c37d30/tests/system/tests/test_groupadd.py#L43  
so added but we haven't used it in any other tests, I will remove it.

asakure: @ikerexxe I've removed that line.

ikerexxe: The first test in `groupadd` tries to be as thorough as possible checking all `group` and `gshadow` fields. Other test cases don't need to do that

ikerexxe (approved):
LGTM

PR #1744, opened 2026-09-17
share/ansible/roles/ci_run/tasks/: Run ldconfig after make install

thc1006:
The programs that link libsubid can't run in the Fedora and openSUSE CI containers: make install puts the library under /usr/local/lib, which ld.so doesn't search on Fedora, and whose cache openSUSE never refreshes. Nothing had run such a program there before the getsubids tests in #1739 did. This adds the directory on Fedora and runs ldconfig on both after make install. Split out of #1739, which keeps the commit until this merges, since its Python tests need it.

- Link: <https://github.com/shadow-maint/shadow/pull/1739>


alejandro-colomar (approved):
LGTM.  Thanks!

I'll merge this evening.  I'll let a few hours of margin in case there are any comments from others.

Cc: @ikerexxe , @jubalh

ikerexxe (approved):
LGTM! Thank you for updating this

PR #1745, opened 2026-09-18
introduction.md: fix "use of AI" link

vincentdephily:
The file is actually `ai.txt` not `ai.md`.

But maybe that file should be converted to md instead, to match the others in that folder ? I can send that PR instead if you prefer.


alejandro-colomar (approved):
Thanks!

I dislike .md, so let's keep is as .txt for now.

PR #1748, opened 2026-09-22
Use [[deprecated]] to prohibit using functions we already disallow

alejandro-colomar:
These functions were already disallowed, although this was only enforced during review, and there wasn't anything enforced through the compiler.  Let's enforce this, which is actually easy thanks to the `[[deprecated]]` attribute and some GCC pragmas to white-list the few use cases which we allow (essentially, the known-safe wrappers that are meant to be used instead).

The pragmas are used in a `DEPRECATED()` macro, which is used to mark all the very-few known-safe use cases.  That macro is quite similar to Rust's `unsafe`, except we exclusively disable errors due to using the disallowed functions, but we don't disable any other errors, which makes this much safer than Rust's `unsafe`.


alejandro-colomar:
Cc: @uecker , @kees

@kees, this is the strong prohibition you were asking for about strncpy(3), except I've went further and prohibited all kinds of stuff, such as memcpy(3) and many others where we provide a better wrapper.  :)

